PR #16, opened 2019-06-27
Move header auth to filter

kahowell:


kahowell:
FYI, I am looking into moving authorization to an annotation via [Spring Security - Method Security](https://spring.io/guides/topicals/spring-security-architecture#_method_security).

awood:
> FYI, I am looking into moving authorization to an annotation via [Spring Security - Method Security](https://spring.io/guides/topicals/spring-security-architecture#_method_security).

:+1: Spring Security is good stuff.  It's been 9 years since I worked with it, but at the time I remember being really impressed with the expressiveness and completeness of the API.

src/main/java/org/candlepin/subscriptions/resteasy/IdentityHeaderAuthFilter.java
```diff
@@ -0,0 +1,62 @@
...
+public class IdentityHeaderAuthFilter implements ContainerRequestFilter {
+
+    private final ObjectMapper mapper;
+
+    public IdentityHeaderAuthFilter(ObjectMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
```
awood: This breaks serving the openapi spec, right?  What if we inject

```java
@Context
private ResourceInfo resourceInfo;
```

And then waive the security requirement if the resource class is OpenApiJsonResource or OpenApiYamlResource?

awood: This breaks the serving of the openapi json right?  What if we did

```java
@Context
private ResourceInfo resinfo;
```

and then looked at the `resinfo.getResourceClass()` method and waived authentication for the OpenApi*Resource classes?

src/main/java/org/candlepin/subscriptions/resteasy/IdentityHeaderAuthFilter.java
```diff
@@ -0,0 +1,62 @@
...
+        String identityHeader = requestContext.getHeaderString("x-rh-identity");
+        if (identityHeader == null) {
+            throw new SubscriptionsException(ErrorCode.VALIDATION_FAILED_ERROR,
+                Response.Status.UNAUTHORIZED, "Unauthorized", "No identity header supplied");
+        }
+        requestContext.setSecurityContext(new IdentityHeaderSecurityContext(mapper, identityHeader));
```
awood: Can we drop a TODO here for Spring to manage this via a factory or something?  If we plan on moving to Spring Security, all this will change.  If not, we can at least go back and address this later.

kahowell: I don't want to confuse people by mentioning Spring Security here. I did a little investigation and this whole class goes away if we move to Spring Security, so I'd rather just keep the details in a card.

src/main/java/org/candlepin/subscriptions/resteasy/IdentityHeaderAuthFilter.java
kahowell: DONE, I used `requestContext.getUriInfo().getMatchedResources()` instead.

PR #20, opened 2019-07-08
Add pagination information; remove account number from API.

awood:
Insights has a standard for responses that we are supposed to follow.
That standard includes allowing for pagination in requests and returning
a block of navigation links in responses.

This commit also removes the account number as a path parameter from the
API.  The account number that is delivered via the x-rh-identity header
is the only account the API will service.

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -12,21 +12,25 @@ servers:
   - url: https://cloud.redhat.com/api/rhsm-subscriptions/v1
 
 paths:
-  /tally/accounts/{account_number}/products/{product_id}:
+  /tally/products/{product_id}:
     description: "Operations for a tally report of a specific account and product."
     parameters:
-      - name: account_number
-        in: path
-        required: true
-        schema:
-          type: string
-        description: "Account number to fetch data for. Must match the authenticated account number."
       - name: product_id
         in: path
         required: true
         schema:
           type: string
         description: "The ID for the product we wish to query"
+      - name: offset
+        in: query
+        schema:
+          type: integer
+        description: "The number of items to skip before starting to collect the result set"
+      - name: limit
+        in: query
+        schema:
+          type: integer
+        description: "The numbers of items to return"
     get:
       summary: "Fetch a tally report for an account and product."
       operationId: getTallyReport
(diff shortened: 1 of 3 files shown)


api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -138,8 +142,39 @@ components:
         application/vnd.api+json:
           schema:
             $ref: "#/components/schemas/Errors"
-
   schemas:
+    InsightsResult:
```
kahowell: So I think a couple of small tweaks are necessary here. First, I think that the items in data should be an array of `TallySnapshot`. Also, the other properties currently in `TallyReport` (`product`, `granularity`) should go into `meta`. I think it would be clearer to go ahead and rename `InsightsResult` to `TallyReport`, and then adjust accordingly:

```yaml
    TallyReport:
      properties:
        data:
          type: array
          items:
            $ref: "#/components/schemas/TallySnapshot"
        links:
          type: object
          properties:
            first:
              type: string
            last:
              type: string
            previous:
              type: string
            next:
              type: string
          required:
            - first
            - previous
            - last
            - next
        meta:
          type: object
          properties:
            resultSetSize:
              type: integer
            product:
              type: string
            granularity:
              description: "Describes the significance of each date in the TallySnapshot list. For example if the
                granularity is set to 'weekly', the dates in the TallySnapshot list will represent the start of a
                seven day period."
              type: string
          required:
            - resultSetSize
```
We could go ahead and set up a generic InsightsResult and use the composition pattern outlined in the link if you'd like, I could go either way.

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -59,7 +63,7 @@ paths:
           content:
             application/vnd.api+json:
               schema:
-                $ref: "#/components/schemas/TallyReport"
+                $ref: "#/components/schemas/InsightsResult"
```
kahowell: I think this should stay as `TallyReport`; see following comment.

awood: I was thinking we'd name it InsightsResult because we could drop other things into it using Openapi [polymorphism](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/).  I.e. the `data` section of an InsightsResult can contain a TallyReport or a SomethingElseReport.  If you don't think we need that level of extensibility, I'll flip it back.

api/rhsm-subscriptions-api-spec.yaml
awood: Looks like this now:

```json
{
  "data": [
    {
      "date": "2019-07-12T17:48:44.275Z",
      "instance_count": 0,
      "cores": 0
    }
  ],
  "links": {
    "first": "string",
    "last": "string",
    "previous": "string",
    "next": "string"
  },
  "meta": {
    "resultSetSize": 0,
    "product": "string",
    "granularity": "string"
  }
}
```

PR #25, opened 2019-07-16
Count sockets and add to snapshots and rollups

mstead:
Sockets are counted from inventory hosts and stored in the
daily snapshots. When the daily snapshots are rolled, the
max value of the sockets for the time period is used.


src/main/java/org/candlepin/subscriptions/tally/roller/DailySnapshotRoller.java
```diff
@@ -67,6 +67,7 @@ public void rollSnapshots(List<String> accounts) {
         produceDailySnapshots(accounts, calculateMaxValuesFromInventory(accounts));
     }
 
+    @SuppressWarnings("squid:S3776")
```
awood: What is this?  I have never seen this before.

PR #28, opened 2019-07-18
Roll snapshots based on calculations from inventory

mstead:
Snapshots are now created and rolled based on the lastest
data from the inventory DB as apposed to querying for max
values from the relevant previously created snapshots.

Notes:
- Created an InventoryAccountUsageCollector that will pull
  the latest hosts for the configured accounts and tally
  up the sockets/cores/instances on a per account/product
  basis.
- The snapshot rollers now use the account calculations to
  create/update the relevant snapshots for their respective
  granularity.
- The daily snapshot roller does not apply the max value
  restriction when creating/updating snapshots. Each time
  the UsageSnapshotProducer is run, the value reflected by
  the calculation will be applied for the current day. The
  snapshots produced by the other rollers will still reflect
  the largest value that was seen for the day.


src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @SuppressWarnings("squid:S3776")
+    @Transactional(value = "inventoryTransactionManager", readOnly = true)
+    public Collection<AccountUsageCalculation> collect(Collection<String> products,
+        Collection<String> accounts) {
+        Map<String, AccountUsageCalculation> calcsByAccount = new HashMap<>();
+        try (Stream<InventoryHost> hostStream = inventoryRepository.findByAccountIn(accounts)) {
+            hostStream.forEach(host -> {
+                String account = host.getAccount();
+                if (!calcsByAccount.containsKey(account)) {
```
awood: I think this `if` and the line below could be replaced with [putIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putIfAbsent-K-V-)?

mstead: DONE

PR #29, opened 2019-07-18
Create a job to purge snapshots periodically.

awood:


src/main/java/org/candlepin/subscriptions/controller/TallyRetentionController.java
```diff
@@ -37,12 +41,26 @@
...
+        this.accountListSource = accountListSource;
+    }
+
+    public void purgeSnapshots() {
+        try {
+            List<String> accountList = accountListSource.list();
+            accountList.forEach(this::cleanStaleSnapshotsForAccount);
+        }
+        catch (IOException e) {
+            throw new JobFailureException("Could not purge snapshots", e);
```
mstead: I'm not sure that this should be throwing a JobFailureException. I'd like to see the job itself trap any exceptions from this method, then have it throw the JobFailureException.

src/main/java/org/candlepin/subscriptions/controller/TallyRetentionController.java
```diff
@@ -37,12 +41,26 @@
...
+        }
     }
 
-    public void cleanStaleSnapshotsForAccount(String accountNumber) {
+    protected void cleanStaleSnapshotsForAccount(String accountNumber) {
```
mstead: Should this have an @Transactional annotation? I couldn't find one elsewhere.

src/main/java/org/candlepin/subscriptions/jobs/PurgeSnapshotsJob.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Autowired
+    public PurgeSnapshotsJob(TallyRetentionController retentionController) {
+        this.retentionController = retentionController;
+    }
+
+    @Override
+    protected void executeInternal(JobExecutionContext context) throws JobExecutionException {
+        log.info("Starting snapshot purge.");
+        // FIXME
```
mstead: REMOVEME

src/main/java/org/candlepin/subscriptions/jobs/PurgeSnapshotsJob.java
```diff
@@ -0,0 +1,52 @@
...
+        retentionController.purgeSnapshots();
+    }
```
mstead: Perhaps add a log message stating that the purge has completed.

src/main/java/org/candlepin/subscriptions/jobs/PurgeSnapshotsJob.java
awood: :laughing:

mstead:
Just need to get the test failures fixed up and this should be good to go.

awood:
> Just need to get the test failures fixed up and this should be good to go.

Should be fixed now.  I forgot a `@TestProperties` annotation.

mstead:
One small item left to make sonarqube happy. Need to make the method you added the transactional annotation to public.

PR #30, opened 2019-07-19
Add sonarcloud.io PR decoration

kahowell:


kahowell:
Actually this may not be necessary, as I set the equivalent parameters in project settings on sonarcloud.io.

PR #33, opened 2019-07-23
Disable jobs via config property

mstead:
A configuration property (rhsm-subscriptions.enableJobProcessing=true|false)
was added to disable job scheduling when the application is started.
This allows dedicating an application instance to handling requests
while not having the burden of running jobs.

**NOTE:**
When jobs are disabled, we are unfortunately still spinning up a quartz scheduler. To avoid this, we would have to find a way to remove the quartz config from the application.yaml and move it to our properties file. I've left this out of the PR given our time constraints. I'd prefer to make this change at a later date, but let me know if we should invest the time to do it now.


mstead:
Need to also disable the API endpoint. Update coming.

PR #35, opened 2019-07-25
Snapshot Generation Improvements

mstead:
This PR drastically improves performance on snapshot generation runs. It is very near completion but requires removing some very nasty code that was put in just to get this POC working.


src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHostFacts.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private Set<String> asProducts(String productJson) {
+        if (!StringUtils.hasText(productJson)) {
+            return new HashSet<>();
+        }
+
+        // FIXME Should be able to do this via the query. This code is a GROSS quick
+        //       hack to see if the whole query concept would make a difference. This
+        //       has to go.
```
mstead: This has to go and is the remaining bit of work left for this PR. There should be a way to produce a comma separated list via the query.

src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
```diff
@@ -20,30 +20,53 @@
...
+                @ColumnResult(name = "products"),
+                @ColumnResult(name = "sync_timestamp")
+            }
+        )
+    }
+)
+@NamedNativeQuery(name = "InventoryHost.getFacts",
+    query =
+        "select " +
+            "account, display_name, facts->'rhsm'->>'orgId' as org_id, facts->'rhsm'->>'CPU_CORES' as cores, " +
```
awood: I have never seen this arrow notation before.  Is this a Spring Data thing?  Do you have a link explaining it handy?

awood: Ah, I see.  JSON dereferencing.  Cool.

src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHostFacts.java
```diff
@@ -0,0 +1,140 @@
...
+    private String account;
+    private String displayName;
+    private String orgId;
+    private int cores;
+    private int sockets;
+    private boolean isRhel;
+    private String syncTimestamp;
+    private Set<String> products;
+
+    public InventoryHostFacts(String account, String displayName, String orgId, String cores, String sockets,
```
kahowell: SonarQube flagged this constructor for having too many arguments. I have to agree, especially given that they are all `String`. Let's remove this constructor, and use the setters instead. We can add a few overloaded setters that take `String` where we want to use `asInt`, `asBoolean`, etc.

src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHostFacts.java
```diff
@@ -0,0 +1,140 @@
...
+import org.springframework.util.StringUtils;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Represents an inventory host's facts.
+ */
+public class InventoryHostFacts {
+    private static final Logger log = LoggerFactory.getLogger(InventoryHostFacts.class);
```
kahowell: Unused logger, caught by SonarQube.

src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
```diff
@@ -45,32 +42,81 @@
 
     private static final Logger log = LoggerFactory.getLogger(FactNormalizer.class);
```
kahowell: Unused logger, caught by SonarQube.

src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHostFacts.java
awood: I don't know that we can do that.  The object is created in `InventoryHost` via a `@SqlResultSetMapping` annotation.  That annotation also allows for mapping to an entity via field injection, but `InventoryHostFacts` isn't an entity.  It's just a convenience class we use as a grab bag.

kahowell: Ack.

PR #38, opened 2019-07-31
Parse rh_products_installed from QPC to detect RHEL

kahowell:


src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
```diff
@@ -67,13 +68,21 @@
...
+        "qpc_certs.product_ids as qpc_product_ids " +
         "from hosts h " +
         "cross join lateral ( " +
-            "select string_agg(cj.elem::text, ',') as products " +
-            "from json_array_elements_text(h.facts::json->'rhsm'->'RH_PROD') as cj(elem)) cj " +
+            "select string_agg(cj.elem, ',') as products " +
+            "from jsonb_array_elements_text(h.facts->'rhsm'->'RH_PROD') as cj(elem)) cj " +
+        "cross join lateral ( " +
+            "select string_agg(cj.elem, ',') as products " +
+            "from jsonb_array_elements_text(h.facts->'qpc'->'rh_products_installed') as cj(elem)) " +
```
awood: Is this right?  The clause is referenced as `qpc_prods` but inside the clause, you're using `cj`.  Shouldn't it be

```
cross join lateral (
  select string_agg(qpc_prods.elem, ',') as products
  from jsonb_array_elements_text(h.facts->'qpc'->'rh_products_installed') as qpc_prods(elem)) qpc_prods
```

src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
```diff
@@ -67,13 +68,21 @@
...
+            "qpc_prods " +
+        "cross join lateral ( " +
+            "select string_agg(cj.elem, ',') as product_ids " +
+            "from jsonb_array_elements_text(h.facts->'qpc'->'rh_product_certs') as cj(elem)) qpc_certs " +
```
awood: Same issue here as above.

kahowell:
@awood I rebased and rewrote the query (the join part especially) to make it a little less confusing. Please take a look and let me know your thoughts.

awood:
> @awood I rebased and rewrote the query (the join part especially) to make it a little less confusing. Please take a look and let me know your thoughts.

Yes.  What you have now is very nice and superior to what I wrote.  To be honest the `cj.elem` and `as cj(elem)` bit was black magic to me that I just took from an SO answer without any true understanding.  It's interesting to see that it's not necessary at all.

awood:
@kahowell 
Something I stumbled on after reading up more on [lateral queries](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-LATERAL)

```
It is often particularly handy to LEFT JOIN to a LATERAL subquery, so that source rows
will appear in the result even if the LATERAL subquery produces no rows for them. For
example, if get_product_names() returns the names of products made by a manufacturer,
but some manufacturers in our table currently produce no products, we could find out
which ones those are like this: ...
```

So maybe instead of a CROSS JOIN we want a LEFT JOIN ... ON true?  The `string_agg` command returns null if you send it an empty array.

```
select string_agg(d, ',') from json_array_elements_text('[]') as d;
```

I'm not sure if an empty RH_PROD is possible, but it's something we may want to consider.

awood:
> I'm not sure if an empty RH_PROD is possible, but it's something we may want to consider.

I did some testing with an empty JSON array and the result in the column was null with both a CROSS JOIN and a LEFT JOIN ... ON true, so I guess there's no point in upsetting the apple cart.

PR #43, opened 2019-08-20
Fixed missing transaction issue when purging snapshots

mstead:
Spring does not honor @Transactional when the method it is on
is called from within the same class. In this particular case,
we would have to mark purgeSnapshots to make it work. Since we
want a transaction around the purge for each account, the
annotation was placed on the repository method itself.

See Pitfall#2:
https://codete.com/blog/5-common-spring-transactional-pitfalls/


awood:
TheRock

Looks good to me!

PR #46, opened 2019-09-10
Add capacity API

kahowell:
It is designed to be symmetric with the tally API for consistency.


kahowell:
Looking for feedback from at least one backend dev; UI input welcome.

@priley86, your input welcome too!

cdcabrera:
Much appreciated on the `has_infinite_quantity` description!

PR #49, opened 2019-09-12
Add TallyJmxBean

kahowell:
To test, run jconsole against a running instance and observe that the
operation is available under org.candlepin.subscriptions.jmx ->
TallyJmxBean -> tallyJmxBean -> Operations -> tallyAccount.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/UsageSnapshotProducer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/UsageSnapshotProducer.java
@@ -43,6 +43,7 @@ import org.springframework.transaction.annotation.Transactional;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


barnabycourt:
Would it be possible to add an option to run tally for all accounts in addition to the method here that runs it just for a single account?

kahowell:
> Would it be possible to add an option to run tally for all accounts in addition to the method here that runs it just for a single account?

Will do!

src/main/java/org/candlepin/subscriptions/tally/UsageSnapshotProducer.java
```diff
@@ -84,28 +85,36 @@ public UsageSnapshotProducer(FactNormalizer factNormalizer, AccountListSource ac
...
-                dailyRoller.rollSnapshots(accounts, accountCalcs);
-                weeklyRoller.rollSnapshots(accounts, accountCalcs);
-                monthlyRoller.rollSnapshots(accounts, accountCalcs);
-                yearlyRoller.rollSnapshots(accounts, accountCalcs);
-                quarterlyRoller.rollSnapshots(accounts, accountCalcs);
-                count += accounts.size();
-                log.info("{}/{} accounts processed.", count, accountList.size());
-            }
-            log.info("Finished producing snapshots for all accounts.");
+            produceSnapshotsForAccounts(accountList);
```
awood: See [pitfall 2](https://codete.com/blog/5-common-spring-transactional-pitfalls/)

```

* The method visibility can’t be any other than public.
* The invocation must come from outside of the bean.
```

So the internal invocation here will not be transactional; we need to refactor out the call to another class. :slightly_frowning_face:

kahowell: @awood, thanks for the info; I believe it's okay in this case because `produceSnapshotsForAccounts(...)` is a private method, and the two public methods:
 - `produceSnapshotsForAccount`
 - `produceSnapshots`

are both annotated with `@Transactional`.

kahowell: i.e. all entrypoints to public methods are transactional, and there are no call paths that involve trying to invoke a transactional method from inside the bean.

awood: Ah, the diff cut off that the method you changed was already transactional.

PR #50, opened 2019-09-13
Generate mock capacity responses (for short term)

kahowell:


src/main/java/org/candlepin/subscriptions/resource/CapacityResource.java
```diff
@@ -0,0 +1,113 @@
...
+            .data(createMockData(granularity, beginning, ending))
+            .links(createMockLinks())
+            .meta(createMockMeta(productId, granularity, beginning, ending));
+    }
+
+    private CapacityReportMeta createMockMeta(String productId, @NotNull String granularity,
+        @NotNull OffsetDateTime beginning, @NotNull OffsetDateTime ending) {
+
+        return new CapacityReportMeta()
+            .count(createMockData(granularity, beginning, ending).size())
```
mstead: Any reason to create the data again just to get the size? Can we just pass the snapshot count, productid and granularity?

kahowell: @mstead, addressed!

PR #55, opened 2019-09-18
Add logic to get effective products for a subscription

kahowell:
This is decoupled from the actual pool/subscription representation, and
is intended to plugged into later work.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/CapacityProductExtractor.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/CapacityProductExtractor.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2009 - 2019 Red Hat, Inc.
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/subscriptions/capacity/CapacityProductExtractor.java
```diff
@@ -0,0 +1,56 @@
...
+        throws IOException {
+
+        this.productIdToProductsMap = productIdToProductsMapSource.getValue();
+    }
+
+    public Set<String> getProducts(Collection<Integer> productIds) {
+        // NOTE: this logic can grow over time (e.g. to blacklist certain combinations, etc).
+        return productIds.stream()
+            .map(productIdToProductsMap::get).filter(Objects::nonNull).flatMap(List::stream)
+            .collect(Collectors.toSet());
```
mstead: Neat-o!

PR #62, opened 2019-09-25
Implement capacity API using repository

kahowell:
Involved some refactoring:

 - Renamed `TallyGranularity` to `Granularity` as it is used for
   capacity too.
 - Pulled `SnapshotTimeAdjuster` interface out of `ReportFiller` (and
   changed to compose rather than inherit, making it easy to share
   time adjustment between the filler and capacity snapshot building.
 - Refactored some common functionality for both resources into
   `AbstractReportResource`.


src/main/java/org/candlepin/subscriptions/resource/AbstractReportResource.java
```diff
@@ -0,0 +1,66 @@
...
+
+    protected String getAccountNumber() {
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        return auth.getName();
+    }
+
+    @NotNull
+    Pageable getPageable(Integer offset, Integer limit) {
+        if (limit == null) {
+            limit = DEFAULT_LIMIT;
```
awood: Looking later on in the code, it appears that this method is meant to cover the bases if offset has been specified alone, if limit has been specified alone, and if they have both been specified.

I'd like to suggest another way.  We consider it a bad request if only one of the offset and limit have been given (and handle this via the validation level).  Optionally, for offset we can put a [default value](https://swagger.io/docs/specification/describing-parameters/#default) of zero in the API definition which results in a JAX-RS `@DefaultValue` being inserted into the method signature.  In that case, only the presence of a limit would be diagnostic of a request for paging.

We can then do the second part (where we check that the offset is a multiple of the limit) as a custom validation.  This approach removes the need for both resource classes to inherit from the same class and moves the concerns (setting a default and validation) out of the method itself (where someone might remove it or it gets clobbered in a merge) and into a higher level.

The `getAccountNumber` method can either be moved to a static utility class or just removed since it's not all that complicated.

Also, we need to validate that limit is non-zero (probably we want it to be positive) if we're going to be dividing by it at any point.

kahowell: > Looking later on in the code, it appears that this method is meant to cover the bases if offset has been specified alone, if limit has been specified alone, and if they have both been specified.

> I'd like to suggest another way. We consider it a bad request if only one of the offset and limit have been given (and handle this via the validation level). Optionally, for offset we can put a default value of zero in the API definition which results in a JAX-RS @DefaultValue being inserted into the method signature. In that case, only the presence of a limit would be diagnostic of a request for paging.

Unfortunately we have to perform paging if either is specified (i.e. is valid to specify offset=10 without specifying limit), so we still have to check for the presence of either, so I'm inclined to not change it. However, I'm happy to revisit whether we decouple the offset/limit validation into another layer at some point.

> The getAccountNumber method can either be moved to a static utility class or just removed since it's not all that complicated.

I changed AsbtractReportResource to ResourceUtils by making methods static.

> Also, we need to validate that limit is non-zero (probably we want it to be positive) if we're going to be dividing by it at any point.

I added `minimum: 1` to the spec for this.

PR #64, opened 2019-10-01
Skip Satellite 6 offerings in capacity calculations.

awood:
Since the RHEL provided with a Satellite 6 instance is meant to be tied
to the Satellite 6 usage, we don't want to consider it in the general
capacity calculations.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/CapacityProductExtractor.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/CapacityProductExtractor.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/subscriptions/capacity/CapacityProductExtractor.java
```diff
@@ -48,9 +50,32 @@ public CapacityProductExtractor(ProductIdToProductsMapSource productIdToProducts
...
+    public boolean setIsInvalid(Set<String> products) {
+        return products.stream().anyMatch(matchesRhel()) && products.stream().anyMatch((matchesSatellite()));
+    }
+
+    private Predicate<String> matchesRhel() {
+        return x -> x.matches("^RHEL.*");
+    }
+
+    private Predicate<String> matchesSatellite() {
+        return x -> x.matches("^Satellite.*");
```
kahowell: For this to work, we need to add `Satellite 6` and `Satellite 6 Capsule` to the product ID mapping. The Eng IDs we should use for this are 250 (Satellite) and 269 (Capsule).

kahowell: I also just realized we can avoid regex here; this can just be `x.startsWith('Satellite')`

src/main/java/org/candlepin/subscriptions/capacity/CapacityProductExtractor.java
```diff
@@ -48,9 +50,32 @@ public CapacityProductExtractor(ProductIdToProductsMapSource productIdToProducts
...
+     * Return whether this set of products should be considered for capacity calculations.
+     * @param products a set of product names
+     * @return true if the set is invalid for capacity calculations
+     */
+    public boolean setIsInvalid(Set<String> products) {
+        return products.stream().anyMatch(matchesRhel()) && products.stream().anyMatch((matchesSatellite()));
+    }
+
+    private Predicate<String> matchesRhel() {
+        return x -> x.matches("^RHEL.*");
```
kahowell: Can just be x.startsWith("RHEL")

awood:
The review comments are resolved.  I just want to make sure that this PR is redundant with #66

awood:
> The review comments are resolved. I just want to make sure that this PR is redundant with #66

This PR should be in line with #6 now in terms of behavior: kick out the RHEL product ID instead of returning an empty set.

PR #66, opened 2019-10-02
Skip RHEL when Satellite product is present

mstead:
Since RHEL is provided with Satellite, we do not want
to include a snapshot for RHEL as well. Therefore, we filter
the RHEL product when normalizing the product facts.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
@@ -35,6 +35,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
```diff
@@ -76,9 +77,24 @@ public NormalizedFacts normalize(InventoryHostFacts hostFacts) {
...
         normalizeConflictingOrMissingRhelVariants(normalizedFacts);
+        pruneProducts(normalizedFacts);
         return normalizedFacts;
     }
 
+    @SuppressWarnings("indentation")
+    private void pruneProducts(NormalizedFacts normalizedFacts) {
+        // If a Satellite product was found, do not include RHEL or its variants.
+        boolean hasSatellite = normalizedFacts.getProducts().stream()
+            .anyMatch(s -> s.matches("^.*[Ss]atellite.*$"));
```
kahowell: Since we control the names of the extracted products, I don't think we need to be quite so flexible. Specifically, I'd like to suggest that we simply go ahead and pick names for the Satellite variants. In #64 I suggested `Satellite 6` and `Satellite 6 Capsule`. That'd eliminate the need for regex, and you could just go ahead and use `s.startsWith('Satellite')`

mstead: DONE

PR #68, opened 2019-10-07
Added physical data counts to snapshots

mstead:


mstead:
Missed some test cases. Please do not review until they have been added.

mstead:
DO NOT MERGE - Need some clarification on whether this is actually what we want.

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -87,9 +87,19 @@ public InventoryAccountUsageCollector(FactNormalizer factNormalizer,
...
+                        // Calculate the physical usage data
+                        if (!hostFacts.isHypervisor() && !hostFacts.isVirtual()) {
+                            prodCalc.addPhysical(cores, sockets, 1);
+                        }
+                        // Assume uncategorized classification and just add to the general totals.
+                        else {
+                            prodCalc.addUncategorized(cores, sockets, 1);
+                        }
+
+                        // Accumulate hypervisor and guest data here
```
awood: Is this a TODO?

mstead: Ya. This will be done as part of another story.

mstead: Already removed in the code I'm working on.

PR #72, opened 2019-10-29
Use orgId over accountNumber.

awood:
In certain cases, a pool may be missing an account number.  For example,
if a customer acquires a developer subscription and then buys from Red
Hat, the developer subscription won't be updated with their account
number.  Consequently, relying on the account number as a component
of a primary key in a table is undesirable.

This patch changes the capacity table to use the orgId (which should
always be present) as a part of the composite primary key.
Additionally, it ties the principal to the orgId from the authentication
header and performs necessary queries with it.


mstead:
This looks good to me. :+1: 

Need to address the sonar complaints, but once done I'll get this merged.

awood:
Okay, I rewrote that test as another unit test that is closer to the actual issue.

PR #74, opened 2019-11-07
Rework tally to use account number, drop account not null

kahowell:
Testing the tally change
------------------------
Try the tally API via deployment with `./gradlew bootRun` followed by invocation of APIs through swagger-ui. Be sure to try both tally and capacity APIs.

Testing the DB change
---------------------
Try:

```
liquibase --username rhsm-subscriptions --classpath=/usr/share/java/postgresql-jdbc.jar --url jdbc:postgresql://localhost:5432/rhsm-subscriptions --changeLogFile liquibase/changelog.xml update
```

Observe that the not null constraint is dropped via:

```
psql -U rhsm-subscriptions -h localhost -c '\d subscription_capacity'
```

Note: I added a new annotation: `@WithMockRedHatPrincipal` to replace `@WithMockUser`.

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -299,10 +299,25 @@ components:
       type: apiKey
       in: header
       name: x-rh-identity
-      description: >-
+      description: |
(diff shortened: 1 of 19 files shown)


mstead:
Looks good to me, ACK.

PR #80, opened 2019-11-19
Restrict Tally/Capacity APIs to admin + account whitelist

mstead:
Both APIs are now restricted to users having the admin role
as well as their account must be in the configured white list
file.

APIs can now be restricted in this mannor by annotating the API
method with the new annotation: @AdminWithWhitelistedAccountOnly

The white list can be configured by adding a new property to the
config file as follows:

rhsm-subscriptions.reporting_account_whitelist_resource_location=file:<PATH_TO_YOUR_ACCOUNT_LIST_FILE>

This file simply contains a list of accounts that are allowed to access the reporting APIs.

NOTE: When the application is running in dev mode, the account WILL NOT be checked when
the API is invoked.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/files/ReportingAccountWhitelist.java
+++ b/src/main/java/org/candlepin/subscriptions/files/ReportingAccountWhitelist.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2009 - 2019 Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.files;
+
+import org.candlepin.subscriptions.ApplicationProperties;
+
+import org.springframework.context.ResourceLoaderAware;
+import org.springframework.core.io.ResourceLoader;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+
+import javax.annotation.PostConstruct;
+
+/**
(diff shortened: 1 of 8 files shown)


mstead:
This code is ready to review, however I'm still working on adding the appropriate bits in the deployment project to add the file and configure it for prod.

mstead:
> This code is ready to review, however I'm still working on adding the appropriate bits in the deployment project to add the file and configure it for prod.

MR up for this. See card on sprint board.

src/main/java/org/candlepin/subscriptions/security/auth/AdminWithWhitelistedAccountOnly.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A security annotation ensuring that the user has an admin role and has an account that has been
+ * white listed.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@PreAuthorize("hasRole('" + IdentityHeaderAuthenticationDetailsSource.ORG_ADMIN_ROLE + "') and " +
+    "@reportAccessService.providesAccessTo(authentication)")
+public @interface AdminWithWhitelistedAccountOnly {
```
kahowell: Rather than a new annotation, I'd rather just see this applied to `@AdminOnly`

src/main/java/org/candlepin/subscriptions/files/ReportingAccountWhitelist.java
```diff
@@ -0,0 +1,69 @@
...
+        source = resourceLocation != null ? new PerLineFileSource(resourceLocation) : null;
+        this.isDevMode = props.isDevMode();
+    }
+
+    public boolean hasAccount(String account) throws IOException {
+        // Whitelist any account when running in dev mode!
+        if (isDevMode) {
+            return true;
+        }
+        return source != null && source.list().contains(account);
```
kahowell: I just realized this will cause the file to be re-read on every request. Probably not problematic due to caching, but it might be worth leaving a note that this is an area we could apply application-level caching if necessary.

mstead: I was thinking about reading the file only once on creation, but since that would force us to have to restart the app to have the list updated, I didn't bother. I will add a note.

src/main/java/org/candlepin/subscriptions/security/auth/AdminWithWhitelistedAccountOnly.java
mstead: Ack. Will make the change. I wish spring security allowed putting multiple security annotations on a method. :/

mstead: DONE

src/main/java/org/candlepin/subscriptions/files/ReportingAccountWhitelist.java
mstead: DONE

PR #81, opened 2019-11-19
Apply RHEL collector to RHEL for $arch

kahowell:
Without this, we don't apply the same hypervisor counting to RHEL for
x86, etc.


mstead:
ACK

PR #85, opened 2019-12-09
Generate a client for the Cloudigrade concurrent resource.

awood:
Cloudigrade does not do spec-first development.  They provide an OpenAPI
specification in JSON that is generated from the source code.
Unfortunately, the generated file does not work properly.  I ran into
multiple issues where OpenAPI complained about having to use
UNKNOWN_BASE_TYPE when generating a client.

Since Cloudigrade is stable and we only use a piece of the API, I've
addressed the issue by simply defining our own specification file.  We
are obligated to keep this file in sync with any changes to the
Cloudigrade API, but hopefully Cloudigrade will move toward a spec-first
development model and then we can consume a working OpenAPI definition
from them.

### DIFF
--- a/cloudigrade-client/cloudigrade-api-spec.yaml
+++ b/cloudigrade-client/cloudigrade-api-spec.yaml
@@ -1,0 +1,106 @@
+# This is a file maintained by the rhsm-subscription project that describes
+# a portion of the Cloudigrade API.
+openapi: "3.0.2"
+info:
+  title: "cloudigrade-api"
+  description: "Third-party specification for Cloudigrade API"
+  version: 1.0.0
+
+paths:
+  /concurrent:
+    description: "Operations on concurrent usages"
+    parameters:
+    - name: limit
+      required: false
+      in: query
+      description: "Number of results to return per page."
+      schema:
+        type: integer
+    - name: offset
+      required: false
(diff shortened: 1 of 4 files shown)


cloudigrade-client/cloudigrade-api-spec.yaml
```diff
@@ -0,0 +1,53 @@
...
+      summary: "List daily concurrent usages"
+      operationId: "ListDailyConcurrentUsages"
+      tags:
+        - concurrent
+      responses:
+        '200':
+          description: "The operation completed successfully"
+          content:
+            application/json:
+              schema:
```
kahowell: Doing the schema inline like this doesn't generate a model class. Can you please move to `$ref` style?

cloudigrade-client/cloudigrade-api-spec.yaml
```diff
@@ -0,0 +1,53 @@
...
+                  - vcpu
+                  - memory
+                properties:
+                  date:
+                    type: string
+                    format: date
+                  instances:
+                    type: integer
+                  instances_list:
+                    type: array
```
kahowell: Looks like there's a subobject we'll have to define too.

PR #86, opened 2019-12-20
Awood/retry collect

awood:


src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
```diff
@@ -141,4 +146,19 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
...
     }
 
+
+    @Bean(name = "collectorRetryTemplate")
+    public RetryTemplate retryTemplate() {
+        SimpleRetryPolicy retryPolicy = new SimpleRetryPolicy();
+        retryPolicy.setMaxAttempts(4);
+
+        FixedBackOffPolicy backOffPolicy = new FixedBackOffPolicy();
+        backOffPolicy.setBackOffPeriod(2000L);
```
mstead: Do you think it is worth making this configurable? I'm undecided.

mstead:
Overall, this seems fine to me and looks like it does what we talked about.

src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
kahowell: Meh, for now. Can revisit later.

mstead: Ya, that's the camp I'm in as well. Figured I'd ask for other's opinions.

mstead:
> I dig this a lot as if some accounts fail to collect, we'll log it and move on, so then we have partial success and can investigate/work on those that fail.

:+1:

PR #87, opened 2020-01-03
Add SKU column to the capacity table.

awood:


awood:
> Overall this looks good to me. I'm assuming we are Ok with null values for this field for old data?

Yeah, I don't know that we can get data to populate the old rows.  There's also no non-null constraint on the column for that reason.  That diminishes the usefulness of this column pretty substantially, but I'm not sure if there's a solution.

mstead:
@awood Once the conflicts are addressed, this is ready to go.

mstead:
> Yeah, I don't know that we can get data to populate the old rows. There's also no non-null constraint on the column for that reason. That diminishes the usefulness of this column pretty substantially, but I'm not sure if there's a solution.

ACK

kahowell:
> Yeah, I don't know that we can get data to populate the old rows. There's also no non-null constraint on the column for that reason. That diminishes the usefulness of this column pretty substantially, but I'm not sure if there's a solution.

FYI, eventually, once subscriptions have all synced, we should be able add a not-null constraint. If needed (e.g. we decide to do foreign key constraint or some such), we can force a sync of specific accounts to get there faster.

awood:
@mstead conflict resolved.

PR #90, opened 2020-01-20
Consider infrastructure_type when classifying as virtual.

awood:
Systems that report through system_profile advertise their type through
the "infrastructure_type" setting.  We need to take that into
consideration during fact normalization and during classification.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
@@ -82,10 +82,15 @@ public class FactNormalizer {
         return normalizedFacts;
     }
 
+    private boolean isVirtual(ClassifiedInventoryHostFacts hostFacts) {
+        return hostFacts.isVirtual() ||
+            "virtual".equalsIgnoreCase(hostFacts.getSystemProfileInfrastructureType());
+    }
+
     private void normalizeClassification(NormalizedFacts normalizedFacts,
         ClassifiedInventoryHostFacts hostFacts) {
         normalizedFacts.setHypervisor(hostFacts.isHypervisor());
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
```diff
@@ -82,10 +82,15 @@ public NormalizedFacts normalize(ClassifiedInventoryHostFacts hostFacts) {
         return normalizedFacts;
     }
 
+    private boolean isVirtual(ClassifiedInventoryHostFacts hostFacts) {
+        return hostFacts.isVirtual() ||
+            "virtual".equalsIgnoreCase(hostFacts.getSystemProfileInfrastructureType());
```
mstead: This code is also present in ClassificationProxyRepository.getFacts(). Could we set this property on the ClassifiedInventoryHostFacts so that this method is not needed? Not sure how complicated it would make things with another 'virtual-ish' property. Also, not sure if it is worth bothering with. Your call, just thought I'd mention it.

awood: Yes, I had thought about that, but honestly, I didn't see a very clean way of doing it.  I don't want the facts to be munged until the get to the FactNormalizer, but at the same time ClassificationProxyRepository has to do a little of the same work to see if a system is "really" virtual.  Ultimately, I'm comfortable with what's in the PR, although I recognize it's not very pretty.  If you see a more elegant way, I'm definitely up for improving it.

mstead: I'm Ok with it as it is, however, if we see us repeating this pattern again we should probably try to do something about it.

PR #92, opened 2020-01-29
Migrate tally_snapshot measurements (core, etc.) to a new table.

awood:
The measurements are now stored in snapshot_measurements and mapped with
JPA as embedded object which keeps them tightly coupled to the actual
Snapshot object (e.g. deletions cascade automatically).  The
measurements exist in a map off the snapshot where they are enumerated
by category (physical, total, hypervisor).

----
Note: the trigger functionality doesn't run during tests.  Since those are just temporary procedures that are part of the migration path, it didn't seem necessary to add them during the tests.  If others desire, I can rewrite the functions in whatever the HSQLDB equivalent is so they will be there during tests.

A side-effect of using stored procedures is that there's no real unit testing for the triggers.  I have tested them myself manually and would recommend that the reviewer do as well, but I don't know of a good way to automate it.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/roller/BaseSnapshotRoller.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/roller/BaseSnapshotRoller.java
@@ -22,6 +22,8 @@ package org.candlepin.subscriptions.tally.roller;
 
 import org.candlepin.subscriptions.db.TallySnapshotRepository;
 import org.candlepin.subscriptions.db.model.Granularity;
+import org.candlepin.subscriptions.db.model.HardwareMeasurement;
+import org.candlepin.subscriptions.db.model.HardwareMeasurementType;
 import org.candlepin.subscriptions.db.model.TallySnapshot;
 import org.candlepin.subscriptions.tally.AccountUsageCalculation;
 import org.candlepin.subscriptions.tally.ProductUsageCalculation;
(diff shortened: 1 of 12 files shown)


src/main/java/org/candlepin/subscriptions/tally/roller/BaseSnapshotRoller.java
```diff
@@ -153,68 +165,39 @@ private boolean updateMaxValues(TallySnapshot snap, ProductUsageCalculation calc
...
-        }
+        changed |= updateTotals(overrideMaxCheck, snap, calc, HardwareMeasurementType.TOTAL);
+        changed |= updateTotals(overrideMaxCheck, snap, calc, HardwareMeasurementType.PHYSICAL);
+        changed |= updateTotals(overrideMaxCheck, snap, calc, HardwareMeasurementType.HYPERVISOR);
         return changed;
     }
 
-    private boolean updatePhysicalTotals(boolean override, TallySnapshot snap, ProductUsageCalculation calc) {
+    private boolean updateTotals(boolean override, TallySnapshot snap, ProductUsageCalculation calc,
+        HardwareMeasurementType measurementType) {
```
mstead: This method is called above with each measurement type, but we are setting the same values for each type which doesn't seem correct to me. The way it is right now, we are setting the calculated overall totals for each measurement type. I think that measurement values that we are setting need to come from the appropriate calculated totals.

TOTAL:
sockets <-- calc.getTotalSockets()
cores <-- calc.getTotalCores()
instances <--  calc.getTotalInstanceCount()

PHYSICAL
sockets <-- calc.getTotalPhysicalSockets()
cores <-- calc.getTotalPhysicalCores()
instances <-- calc.getTotalPhysicalInstanceCount()

HYPERVISOR
sockets <-- calc.getTotalHypervisorSockets()
cores <-- calc.getTotalHypervisorCores()
instances <-- calc.getTotalHypervisorInstanceCount()

awood: facepalm

Should be corrected now.  Thanks for catching this.

mstead:
@awood Code looks good to me with the new changes. I'm going to spend a little time manually testing the migration and triggers.

awood:
@mstead I decided to take letter case out of the equation entirely.  Measurement types are just all upper case now.  If this works okay for you, would you mind squash merging instead of regular merging?

mstead:
> @mstead I decided to take letter case out of the equation entirely. Measurement types are just all upper case now. If this works okay for you, would you mind squash merging instead of regular merging?

Ya, this works for me, and will do.

PR #95, opened 2020-02-10
Add configuration settings to allow for a TLS connection to PostgreSQL.

awood:
Enable SSL on postgresql:
  * Set `ssl = on` in `postgresql.conf`.  Set `ssl_cert_file` and `ssl_key_file`.
  * To create a self-signed certificate:
  ```
  openssl req -new -x509 -days 365 -nodes -text -out self-signed.crt -keyout self-signed.key -subj "/CN=localhost"
  ```
  * To create a CA and a leaf certificate:
  ```
  openssl req -new -nodes -text -out root.csr -keyout root.key -subj "/CN=Test CA"
  openssl x509 -req -in root.csr -text -days 3650 -extensions v3_ca   -signkey root.key -out root.crt
  openssl req -new -nodes -text -out leaf.csr -keyout leaf.key -subj "/CN=localhost"
  openssl x509 -req -in leaf.csr -text -days 365 -CA root.crt -CAkey root.key -CAcreateserial -out leaf.crt
  ```

You can then experiment with the `root.crt` and `self-signed.crt`.  If you're using `leaf.crt` on PostgreSQLand `root.crt` isn't given as a `root-ca-location`, the connection should abort (if using the default verification mode).

Note that the first PostgreSQL connection made is for Liquibase!  It uses the same datasource, but occurs pretty early in the application initialization.

This PR is missing tests, but I'm a little undecided on how to provide them.  I'm still thinking that part over, but I'm putting this up for early review.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/validator/VerificationModeValidator.java
+++ b/src/main/java/org/candlepin/subscriptions/validator/VerificationModeValidator.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package org.candlepin.subscriptions.validator;
+
+import org.postgresql.jdbc.SslMode;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+/**
+ * Validator to ensure the verification mode option for a postgresql SSL connection is set to an acceptable
+ * value. Possible values include "disable", "allow", "prefer", "require", "verify-ca" and "verify-full". The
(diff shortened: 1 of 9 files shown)


awood:
I still haven't really figured out a great way to test this in an automated fashion.  If anyone has any ideas, I'm open to them!

src/main/java/org/candlepin/subscriptions/validator/VerificationModeValidator.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        // A null or empty value is considered invalid.
+        if (value == null || value.isEmpty()) {
+            return false;
+        }
+
+        return Arrays.asList(
```
kahowell: Thoughts on using `org.postgresql.jdbc.SslMode#of` to verify this instead of replicating the logic? Would require us to change the dep to compile, but would simplify some code...

src/main/resources/application.yaml
```diff
@@ -21,6 +21,17 @@ management:
       enabled: true
     prometheus:
       enabled: true
+rhsm-subscriptions:
+  datasource:
+    # This can be overridden elsewhere in configuration, but it's important to provide a sensible default.
+    # Without this setting, postgres' JDBC driver defaults to using $HOME/.postgresql/root.crt
+    root-ca-location: classpath:/rhsm-subscriptions-ca.crt
```
kahowell: Can we put this in the class instead, as a default for `PostgresTlsDataSourceProperties#rootCaLocation`? That way we wouldn't have to override for every new datasource...

awood: Sure.  I don't have a problem with that.

src/main/java/org/candlepin/subscriptions/validator/VerificationModeValidator.java
awood: Done.

kahowell:
I've been doing some functional testing. w/ various permutations of the following lines
```
rhsm-subscriptions.datasource.root-ca-location: file:/tmp/self-signed.crt
rhsm-subscriptions.datasource.enable-tls: true
rhsm-subscriptions.datasource.verification-mode: verify-ca
```
FWIW, I also played around with combinations of `verification-mode: disable` and the server requiring SSL via `pg_hba.conf`.

I also tested a non-matching hostname by generating a cert for "localhost2".

:+1: works quite well!

PR #96, opened 2020-02-11
Add Cloud Provider counts to snapshot record and API

mstead:
This PR includes:
- adding counts to the database
- including the counts in the API

When testing this, I hooked my local rhsm-subscription instance up to the inventory DB used by CI so that I could suck in the test systems that have the cloud provider set already. There aren't many, but served the purpose.

Remember to enable job processing in the rhsm-subscriptions config to get the job to fire. And disable when testing the API.


src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurementType.java
```diff
@@ -21,12 +21,32 @@
...
+    ALIBABA,
+    AZURE;
+
+    public static HardwareMeasurementType getCloudProvider(String name) {
+        for (HardwareMeasurementType cloudType : getCloudProviderTypes()) {
+            if (cloudType.name().equalsIgnoreCase(name)) {
+                return cloudType;
+            }
+        }
+        return null;
```
awood: Instead of returning null, I'd like to suggest that we have a method like `isSupportedCloudProvider` and the semantics would be like

```java
if (HardwareMeasurementType.isSupportedCloudProvider(normalizedFacts.getCloudProvider()) {
...
}
```

To me, this makes more sense than a comparison against null and it's also more succinct.  Plus, the implementation is just a `getCloudProviderTypes.contains(x)`

src/main/resources/liquibase/202002061140-update-copy-measurement-procedure.xml
```diff
@@ -0,0 +1,43 @@
...
+                        physical_instance_count = NEW.instance_count,
+                        physical_sockets = NEW.sockets
+                    WHERE id = NEW.snapshot_id;
+                ELSIF (NEW.measurement_type = 'HYPERVISOR') THEN
+                    UPDATE tally_snapshots SET
+                        hypervisor_cores = NEW.cores,
+                        hypervisor_instance_count = NEW.instance_count,
+                        hypervisor_sockets = NEW.sockets
+                    WHERE id = NEW.snapshot_id;
+                ELSIF (not NEW.measurement_type = ANY('{AWS,GOOGLE,AZURE,ALIBABA}'::text[])) THEN
```
kahowell: Was this change necessary as the exception causes the insert to be rolled back otherwise?

mstead: When the job runs, if it comes across a host with a cloud provider, it blows up when this procedure runs and causes the entire job to stop.

src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurementType.java
mstead: DONE

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -467,6 +470,21 @@ components:
           type: integer
           format: int32
           minimum: 0
+        cloud_instance_count:
+          description: "Total number of hosts running in the cloud, matching the report criteria on the date of this snapshot."
```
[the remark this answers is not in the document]
mstead: DONE

src/main/java/org/candlepin/subscriptions/tally/ProductUsageCalculation.java
```diff
@@ -20,99 +20,109 @@
...
-        return totalHypervisorCores;
+    public void addToTotal(int cores, int sockets, int instances) {
+        increment(HardwareMeasurementType.TOTAL, cores, sockets, instances);
     }
 
-    public int getTotalHypervisorInstanceCount() {
-        return totalHypervisorInstanceCount;
-    }
+    public void addCloudProvider(HardwareMeasurementType cloudType, int cores, int sockets, int instances) {
+        if (!HardwareMeasurementType.getCloudProviderTypes().contains(cloudType)) {
```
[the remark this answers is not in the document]
mstead: DONE

mstead:
Ready for another review. Thanks!

PR #99, opened 2020-02-25
Add the sla query parameter and include it in the tally response

mstead:
Query used by API now does sums values on the DB server. This
was done to limit the amount of data that gets post-processed
which admittedly adds complexity but should reduce the memory
footprint when using the API.

An optional API parameter (sla) has been added to limit results
to snapshots with a matching sla only.

We have opted to add a magic value for the sla parameter ('unset')
that indicates that we are filtering on any snapshot that has no
SLA set.This facilitates the requirement of being able to query
for only snapshots that have an empty SLA. Not setting a value for the
query parameter drives the "don't care" semantics of the underlying
query.

**NOTE:**
I'm not a huge fan of the magic 'unset' value for the sla API parameter as we discussed. I think I'd prefer to simple allow setting it to empty string.
sla=&...

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -59,6 +59,12 @@ paths:
           schema:
             type: string
           description: "The level of granularity to return."
+        - name: sla
+          in: query
+          schema:
+            type: string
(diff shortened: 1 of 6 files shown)


mstead:
Closed this PR because we decided to rework the solution here.

PR #100, opened 2020-02-26
Ignore hosts that have been culled in HBI

mstead:
Any host that has been culled does not get included in the produced
snapshots. A host's culled timestamp is determined via the host's
stale_timestamp as per HBI's formula:

   culled_timestamp = stale_timestamp + <offset_in_days>

The application's configuration allows us to tune <offset_in_days>
in case the HBI's above changes later. The default offset is 14 days.

NOTE:
This PR filters by stale_timestamp at the query level as well as at
the processing level. This is done so that should we move away from
querying the HBI DB directly, it will will not get missed. It also
acts as a safe guard.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "insights-host-inventory"]
 	path = insights-host-inventory
 	url = https://github.com/RedHatInsights/insights-host-inventory.git
+	branch = stable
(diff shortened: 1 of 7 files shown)


mstead:
Need to check on something here - DO NOT MERGE.

mstead:
I noticed that there's a culled_timestamp in the HBI API definition but there's only a stale_timestamp in the HBI database. I'm not sure how I missed this initially. I'm looking to how we can determine the culled_timestamp from what's in the DB.

Stay tuned.

mstead:
HBI calculates the culled_timestamp as follows:
  stale_timestamp + N days

I've made the corrections to this PR. I've made the days offset configurable since if it is changed by HBI we can quickly adjust it via configuration.

**NOTE:** I really don't like that we have to do this, but unless they are to put it on the host record, we have no choice unless we start reading event data or move to the API.

mstead:
Ready to review!

mstead:
@kahowell Try the latest changes.

PR #104, opened 2020-03-17
1811717: Filter out redundant snapshots

kahowell:
If for any reason we end up with multiple snapshots that match a given date, then this change only emits the "larger" where one of `instances` or `sockets` or `cores` is larger.


mstead:
Looks good to me.

PR #105, opened 2020-03-18
Refactor into spring profiles

kahowell:
The following profiles now exist:

- api: instance should service API requests
- capacity-ingress: instance should accept capacity records
- scheduler: instance should schedule tasks
- worker: instance should pick up tasks from kafka and run them, implies kafka-queue
- in-memory-queue: instance should queue and run tasks in-memory
- kafka-queue: instance should queue tasks to kafka, but doesn't necessarily run them

If you run `./gradlew bootRun`, it will default to:

- api
- capacity-ingress
- scheduler
- in-memory-queue

Below commands show some ways to set profiles, and the combinations I think we need.

```sh
./gradlew bootRun --args=--spring.profiles.active=api  # accept requests and consume tasks from kafka
./gradlew bootRun --args=--spring.profiles.active=capacity-ingress  # accept capacity updates
./gradlew bootRun --args=--spring.profiles.active=scheduler,worker  # schedule tasks and work them
./gradlew bootRun --args=--spring.profiles.active=worker  # pick up tasks from the kafka queue
```

Note: I don't see an immediate use for kafka-queue, but if we ever want to deploy a node which is only a producer, it would be the profile to use (along with scheduler).
Note: we will need to override profile for each deployment type in the openshift deployment configs.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
@@ -77,11 +77,6 @@ public class ApplicationProperties {
     private int accountBatchSize = 500;
 
     /**
-     * Whether the ingress endpoint is enabled on this instance of rhsm-subscriptions or not.
-     */
-    private boolean enableIngressEndpoint;
-
-    /**
      * Amount of time to cache the account list, before allowing a re-read from the filesystem.
      */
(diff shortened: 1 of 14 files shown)


mstead:
This needs to be rebased with master so that it can compile.

PR #113, opened 2020-03-26
Add DB schema and repository for AdminConfig (opt-in)

mstead:
A AccountConfig entry represents an account/org pair and
the functionality that is allowed for that pair. For example,
should data sync be enabled in tally/conduit? Or is reporting
enabled for that account in tally.

The intent here is to allow a single record to enable the
various configurable bits across both tally and conduit.

**NOTE** This PR is targeting a feature branch `feature/opt-in`


src/main/java/org/candlepin/subscriptions/db/model/config/AccountConfig.java
```diff
@@ -0,0 +1,144 @@
...
+import javax.persistence.Table;
+
+
+/**
+ * Represents the configuration for an account/org pair.
+ */
+@IdClass(AccountConfigKey.class)
+@Entity
+@Table(name = "account_config")
+public class AccountConfig implements Serializable {
```
kahowell: It would be easier to migrate to this if we split into org/account (otherwise we'll need to get the mapping to migrate).

mstead: I'm assuming you mean something like:

AccountConfig
- account_number
- sync_enabled
- reporting_enabled
- last_updated

OrgConfig
- org_id
- sync_enabled
- last_updated
(no reporting flag required for orgs at this point since conduit doesn't have that concept)

With a repository for each.

mstead:
@kahowell How's this?

src/main/java/org/candlepin/subscriptions/db/model/config/AccountConfig.java
kahowell: Perfect! It's like you read my mind :smile:

mstead:
This merge was done in error. This PR is obsolete. 

NEW: https://github.com/RedHatInsights/rhsm-subscriptions/pull/114

PR #114, opened 2020-03-30
Add DB schema and repository for AdminConfig (opt-in)

mstead:
A AccountConfig entry represents an account/org pair and
the functionality that is allowed for that pair. For example,
should data sync be enabled in tally/conduit? Or is reporting
enabled for that account in tally.

The intent here is to allow a single record to enable the
various configurable bits across both tally and conduit.

**NOTE** This PR is targeting a feature branch `feature/opt-in`

### DIFF
--- a/src/main/resources/liquibase/202003261146-add-table-for-account-config.xml
+++ b/src/main/resources/liquibase/202003261146-add-table-for-account-config.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202003261146-1" author="mstead">
+        <comment>Add table for account configuration for tally/conduit</comment>
+        <createTable tableName="account_config">
+            <column name="account_number" type="VARCHAR(255)" />
+            <column name="sync_enabled" type="BOOLEAN">
+                <constraints nullable="false" />
+            </column>
+            <column name="reporting_enabled" type="BOOLEAN">
+                <constraints nullable="false" />
+            </column>
+            <column name="opt_in_type" type="VARCHAR(255)">
+                <constraints nullable="false" />
+            </column>
+            <column name="created" type="TIMESTAMP WITH TIME ZONE">
+                <constraints nullable="false" />
+            </column>
+            <column name="updated" type="TIMESTAMP WITH TIME ZONE">
+                <constraints nullable="false" />
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet id="202003261146-2" author="mstead">
+        <addPrimaryKey constraintName="acc_num_pkey"
+                       tableName="account_config"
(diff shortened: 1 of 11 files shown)


src/main/resources/liquibase/202003261146-add-table-for-account-config.xml
```diff
@@ -0,0 +1,38 @@
...
+        <comment>Add table for account configuration for tally/conduit</comment>
+        <createTable tableName="account_config">
+            <column name="account_number" type="VARCHAR(32)" />
+            <column name="sync_enabled" type="BOOLEAN">
+                <constraints nullable="false" />
+            </column>
+            <column name="reporting_enabled" type="BOOLEAN">
+                <constraints nullable="false" />
+            </column>
+            <column name="opt_in_type" type="VARCHAR(12)">
```
kahowell: In case we add more values, it's better to make this longer (`varchar(255)`). Interestingly, I've been doing [some reading](https://dba.stackexchange.com/a/21496) that suggests just using type `text` is just as performant (and more future proof, since length is then unlimited) - and [implemented the same under the hood](https://stackoverflow.com/a/4849030). So I'd be fine with with either `varchar(255)` or `text` here.

src/main/resources/liquibase/202003271146-add-table-for-org-config.xml
```diff
@@ -0,0 +1,35 @@
...
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202003271146-1" author="mstead">
+        <comment>Add table for org based configuration.</comment>
+        <createTable tableName="org_config">
+            <column name="org_id" type="VARCHAR(32)"/>
```
kahowell: Same thoughts as previous comment about `text` or `varchar(255)`.

src/main/resources/liquibase/202003271146-add-table-for-org-config.xml
```diff
@@ -0,0 +1,35 @@
...
+            <column name="sync_enabled" type="BOOLEAN">
+                <constraints nullable="false" />
+            </column>
+            <column name="opt_in_type" type="VARCHAR(12)">
```
kahowell: Same thoughts as previous comment about `text` or `varchar(255)`.

src/main/resources/liquibase/202003261146-add-table-for-account-config.xml
mstead: DONE - Went with varchar(255) since TEXT was causing issues with primary-keys. Decided to keep it consistent.

src/main/resources/liquibase/202003271146-add-table-for-org-config.xml
mstead: DONE - Went with varchar(255) since TEXT was causing issues with primary-keys. Decided to keep it consistent.

src/main/resources/liquibase/202003271146-add-table-for-org-config.xml
mstead: DONE - Went with varchar(255) since TEXT was causing issues with primary-keys. Decided to keep it consistent.

mstead:
> LGTM. @mstead, did you want to squash before merge?

Done! Thanks for the review.

PR #117, opened 2020-04-02
Support use of the AccountConfig repository as the account source

mstead:
* AccountListSource now general interface for account lists
* File based account list source now combines sync list and
  reporting whitelist file sources.

To enable this  functionality, do not include the **_rhsm-subscriptions.accountListResourceLocation_** property in the config file. Otherwise the account sync list and the reporting sync list will be required as files.

You can manually add/update entries to the DB table (account_confg) to play with this feature.

```sql
insert into account_config (account_number, sync_enabled, reporting_enabled, opt_in_type, created, updated) values ('123456, TRUE, TRUE, 'DB', '2020-04-02 10:04:21.763-03', '2020-04-02 10:04:21.763-03');
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/task/TaskManager.java
+++ b/src/main/java/org/candlepin/subscriptions/task/TaskManager.java
@@ -22,19 +22,20 @@ package org.candlepin.subscriptions.task;
 
 import org.candlepin.subscriptions.ApplicationProperties;
 import org.candlepin.subscriptions.tally.AccountListSource;
+import org.candlepin.subscriptions.tally.AccountListSourceException;
 import org.candlepin.subscriptions.tally.UsageSnapshotProducer;
 import org.candlepin.subscriptions.task.queue.TaskQueue;
-
-import com.google.common.collect.Iterables;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Profile;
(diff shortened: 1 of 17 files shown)


src/main/java/org/candlepin/subscriptions/task/TaskManager.java
```diff
@@ -81,36 +82,38 @@ public void updateAccountSnapshots(String accountNumber) {
...
+            int accountBatchSize = appProperties.getAccountBatchSize();
+            log.info("Queuing snapshot production in batches of {}.", accountBatchSize);
 
-        int accountBatchSize = appProperties.getAccountBatchSize();
-        log.info("Queuing snapshot production for {} accounts in batches of {}.", accountList.size(),
-            accountBatchSize);
+            accountList
+                .collect(Collectors.groupingBy(a -> count.getAndIncrement() / accountBatchSize))
+                .values()
+                .forEach(accounts -> {
```
mstead, line 95: I need to find a better way of chunking up this data. As it is right here, the complete list will be read in up front and split accordingly.

@kahowell @awood Any ideas here? I'm not sure that I can accomplish this with the streaming API.

mstead: Thinking that I might have to use paging to get the result I'm after. Thoughts?

kahowell: @mstead, you could do something similar to the pattern you used in RedHatInsights/rhsm-conduit#117 ?

PR #118, opened 2020-04-07
Awood/jmx opt in

awood:


src/main/java/org/candlepin/subscriptions/jmx/OptInJmxBean.java
```diff
@@ -0,0 +1,81 @@
...
+            enableTallyReporting, enableConduitSync);
+
+        OptInConfigData data = config.getData();
+
+        String text;
+        if (Boolean.TRUE.equals(data.getOptInComplete())) {
+            text = "Completed opt in for account %s and org %s";
+        }
+        else {
+            text = "Created opt in for account %s and org %s";
```
mstead, line 77: This will never happen. optInComplete represents whether an AccountConfig/OrgConfig pair exists. Using the controller will never allow this to be false as it will create a new *Config if either doesn't exist.

src/main/java/org/candlepin/subscriptions/jmx/OptInJmxBean.java
```diff
@@ -0,0 +1,81 @@
...
+    private final OptInController controller;
+
+    public OptInJmxBean(OptInController controller) {
+        this.controller = controller;
+    }
+
+    @ManagedOperation(description = "Delete opt in configuration")
+    @ManagedOperationParameter(name = "accountNumber", description = "Red Hat Account Number")
+    @ManagedOperationParameter(name = "orgId", description = "Red Hat Org ID")
+    public void deleteOptInConfig(String accountNumber, String orgId) {
```
mstead, line 54: Perhaps rename to optOut(...)

src/main/java/org/candlepin/subscriptions/jmx/OptInJmxBean.java
```diff
@@ -0,0 +1,81 @@
...
+        controller.optOut(accountNumber, orgId);
+    }
+
+    @ManagedOperation(description = "Create an opt in configuration")
+    @ManagedOperationParameter(name = "accountNumber", description = "Red Hat Account Number")
+    @ManagedOperationParameter(name = "orgId", description = "Red Hat Org ID")
+    @ManagedOperationParameter(name = "enableTallySync", description = "Turn on Tally syncing")
+    @ManagedOperationParameter(name = "enableTallyReporting", description = "Turn on Tally reporting")
+    @ManagedOperationParameter(name = "enableConduitSync", description = "Turn on Conduit syncing")
+    public String createOptInConfig(String accountNumber, String orgId, boolean enableTallySync,
```
mstead, line 64: This method will update an existing config as well. We should rename it accordingly.

awood:
> Along with the suggested changes below, we should also add a method to allow a means for checking if a config exists for an account and org.

Added it.  The code is going to make you 
tenor

PR #119, opened 2020-04-08
Add a custom filter to prevent CSRF

kahowell:
Testing is a little bit complex. Locally I did this:

1. Add an entry to `/etc/hosts`: `127.0.0.1 testcloud.redhat.com`
2. Install nginx (dnf installable), and create a config file that looks
   like this (as test.conf):

```nginx
events {
  worker_connections 1024;
}

http {
  server {
    listen 443;
    location / {
      proxy_pass http://localhost:8080;
    }
  }
}
```

3. Run nginx: `sudo nginx -c $(pwd)/test.conf`
4. Run the service with profile api:

```sh
./gradlew bootRun --args=--spring.profiles.active=api
```

5. Note that you can't PUT without a valid Origin or Referer (e.g.):

```sh
curl -H "Origin: cloud.redhat.com" -X PUT "http://testcloud.redhat.com:443/api/rhsm-subscriptions/v1/opt-in" -H  "accept: application/vnd.api+json" -H  "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoib3JnMTIzIn19fQo"
```

vs.

```sh
curl -X PUT "http://testcloud.redhat.com:443/api/rhsm-subscriptions/v1/opt-in" -H  "accept: application/vnd.api+json" -H  "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoib3JnMTIzIn19fQo"
```

6. Stop nginx: `sudo nginx -s quit`

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/security/AntiCsrfFilter.java
+++ b/src/main/java/org/candlepin/subscriptions/security/AntiCsrfFilter.java
@@ -51,9 +51,15 @@ public class AntiCsrfFilter extends OncePerRequestFilter {
     private static Logger log = LoggerFactory.getLogger(AntiCsrfFilter.class);
 
     private final boolean disabled;
+    private final int port;
+    private final String domainSuffix;
+    private final String domainAndPortSuffix;
 
     AntiCsrfFilter(ApplicationProperties props, ConfigurableEnvironment env) {
         disabled = props.isDevMode() || Arrays.asList(env.getActiveProfiles()).contains("capacity-ingress");
+        port = props.getAntiCsrfPort();
+        domainSuffix = props.getAntiCsrfDomainSuffix();
+        domainAndPortSuffix = String.join(":", domainSuffix, Integer.toString(port));
         if (disabled) {
             log.info("Origin & Referer checking (anti-csrf) disabled.");
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/subscriptions/security/AntiCsrfFilter.java
```diff
@@ -0,0 +1,94 @@
...
+        if (referer == null) {
+            return false;
+        }
+        URI uri = URI.create(referer);
+        return uri.getHost().endsWith(".redhat.com") && (uri.getPort() == -1 || uri.getPort() == 443);
+    }
+
+    private boolean originMatches(HttpServletRequest request) {
+        String origin = request.getHeader("origin");
+        return origin != null && (origin.endsWith(".redhat.com") || origin.endsWith(".redhat.com:443"));
```
mstead, line 92: Should we make these configurable so that if we had to change anything like the port, we could tweak it a little more easily? some sort of whitelist?

kahowell: Sure. I will make the domain suffix and port configurable, with defaults. I don't think we need to handle more than one domain, so for the principle of YAGNI (you ain't gonna need it), will leave it at those two.

mstead:
This PR behaves as expected and the changes look good. Waiting for tests and I'll get it merged.

PR #123, opened 2020-04-16
Add unit tests for the orgAdminRequired configuration property.

awood:


src/test/java/org/candlepin/subscriptions/security/auth/AdminOnlyTest.java
```diff
@@ -0,0 +1,144 @@
...
+        StubResource stub = context.getBean(StubResource.class);
+        WhitelistedAccountReportAccessService mockAccess =
+            context.getBean(WhitelistedAccountReportAccessService.class);
+        when(mockAccess.providesAccessTo(any(Authentication.class))).thenReturn(false);
+
+        context.getBean(ApplicationProperties.class).setOrgAdminRequired(true);
+        assertThrows(AccessDeniedException.class, stub::reportingAdminOnlyCall);
+
+        context.getBean(ApplicationProperties.class).setOrgAdminRequired(false);
+        assertDoesNotThrow(stub::reportingAdminOnlyCall);
```
kahowell, line 112: This should still throw, as the reportingAccessService shows they're not opted-in yet.

src/test/java/org/candlepin/subscriptions/security/auth/AdminOnlyTest.java
```diff
@@ -0,0 +1,144 @@
...
```
kahowell, line 127: Should still throw, as they're not opted-in.

PR #124, opened 2020-04-17
Log AccessDenied exceptions at debug level

kahowell:


src/main/java/org/candlepin/subscriptions/exception/mapper/BaseExceptionMapper.java
```diff
@@ -57,9 +63,19 @@ public Response toResponse(T exception) {
     protected abstract Error buildError(T exception);
 
     private void logError(Error error, Throwable exception) {
-        String message = StringUtils.hasText(error.getCode()) ?
-            String.format("%s: %s", error.getCode(), error.getTitle()) : error.getTitle();
+        String message = formatMessage(error);
```
awood, line 66: This PR is fine, and I will approve it, but just a thought.  We could just do this instead and shorten things up:

```java
Error error = buildError(exception);
String message = formatMessage(error);
if (exception instanceof AccessDeniedException) {
    log.debug(message, exception);
} else {
    log.error(message, exception);
}
```

It doesn't seem like the individual `logDebug` and `logError` methods really buy us much.

But like I said, it works the way it is and is fine, so if you just want to merge it as is, just say so and I'll merge it.

kahowell: Done. (& force pushed)

PR #126, opened 2020-04-22
1824384: Handle unlimited capacity SKUs properly

kahowell:
This change handles VDC-style SKUs (those which have derived products and are socket based), and also prevents any negative values from being used for capacity.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
@@ -59,6 +59,7 @@ public class CandlepinPoolCapacityMapper {
 
         Set<String> products = productExtractor.getProducts(productIds);
         Set<String> derivedProducts = productExtractor.getProducts(derivedProductIds);
+        boolean isVirtUnlimited = isVirtUnlimited(pool);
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
```diff
@@ -73,21 +74,35 @@
             capacity.setEndDate(pool.getEndDate());
             capacity.setServiceLevel(getSla(pool));
             Long socketCapacity = getCapacityUnit("sockets", pool);
-            if (products.contains(product) && socketCapacity != null) {
+            if (products.contains(product) && socketCapacity != null && socketCapacity > 0) {
```
awood, line 77: We do these `x != && x>0` checks a lot in this method both for socketCapacity and coresCapacity.  Would it make sense to pull this out into a quick

```
public boolean isPositive(Long value) {
    return value != null && value > 0;
}
```

It wouldn't shorten things up massively, but it would make the predicates easier to read: `if (products.contains(product) && isPositive(socketCapacity))`.

I feel like this method is probably provided by Commons Lang or Guava or one of those type libraries, but a quick look didn't show anything.

kahowell: Done.

PR #130, opened 2020-04-23
Store Candlepin product ID as the SKU in the capacity table.

awood:
I've got a follow up that will go to subscription-conduit as well.  Sorry this took me awhile.  I had kind of forgotten how all the ingress stuff works.  We should probably write a doc on it :)


src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
```diff
@@ -77,6 +77,7 @@
             capacity.setBeginDate(pool.getStartDate());
             capacity.setEndDate(pool.getEndDate());
             capacity.setServiceLevel(sla);
+            capacity.setSku(pool.getName());
```
kahowell, line 80: Actually, is there some reason this:
```suggestion
            capacity.setSku(pool.getProductId());
```
shouldn't work?

PR #132, opened 2020-04-29
Move RBAC related classes underneath RBAC project.

awood:
@mstead what do you think?  There's a little redundancy with the reimplementation of the `getIdentityHeader` method but personally, I think it's worth it to keep all the RBAC stuff contained in one project.


mstead:
@awood Works for me. Fix up the checks and LGTM.

PR #133, opened 2020-04-30
Integrate RBAC permissions into application spring roles

mstead:
# PR Details
There are two key pieces to this PR.

**1) Prevent the security checks from happening multiple times per filter**

When the CSRF filter was added, security checks were being performed
multiple times per request, resulting in processing the identity header
multiple times. This was happening because we were setting the decoded
header as the principal and then replacing it with the InsightsUserPrincipal
once the header JSON was was processed. Spring security was checking
to see if the principal had changed between filters and because it was
replaced, it checked it twice.

This PR does the header parsing and principal generation in the
IdentityHeaderAuthenticationFilter. This is in turn passed to the
IdentityHeaderAuthenticationManager which simply validates the data
in the principal and throws appropriate exceptions which trickles
back to the response with the appropriate codes.

By doing this, we do not swap the principal in the context, avoiding
the failed principal equality checks resulting in the extra security
checks.
 

**2) Use RBAC client to determine the user's application roles.**

* Add an OptInFilter (tally/capacity endpoints) that checks to make sure
  that the user's current account has fully opted in. If the opt-in process
  has not been completed, an OptInRequiredException is thrown from the filter
  resulting in a 403 with an appropriate application error code
  (1004: Request was denied since opt-in has not yet occurred.)

* RBAC permissions are being fetched by the RBAC client and are being
  parsed by a RoleProvider class. This class processes the permissions
  provided by RBAC and translates them into our own application roles
  (spring based roles) that are used by our security annotations.

* The subscriptions:*:* permission is the only RBAC permission currently being
  checked. Having this permission grants the OPT_IN and REPORTING application
  roles.

  NOTE: In the permission above, 'subscriptions' is the RBAC application name
        and is configurable.

* It is safe to assume from here out that any 403 (AccessDenied) response is due
  to invalid permissions from RBAC.

# How To Test
I must admit, manually testing this is really cumbersome and requires a few workarounds and app restarts.

Start start by deploying the application with the following config properties set:
```properties
# Use the stub RBAC API
rhsm-subscriptions.rbac-service.useStub=true

# The stub rbac api will return 'subscriptions:*:*' for any user, so we set the expected
# permission  to be 'insights:*:*' to simulate a user without the appropriate permission
rhsm-subscriptions.rbacApplicationName=insights

# This workaround still exists, but let's disable it so that the role is actually checked.
rhsm-subscriptions.orgAdminOptional=false

# Make sure we are not running in dev mode
rhsm-subscriptions.dev-mode=false
```

1) Test that opt-in is checked when a request is made to tally (same applies to capacity). I created a standard ID header with account/org that has no opt-in configuration in the database.

```bash
$ curl -H "x-rh-identity: eyJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NSIsImF1dGhfdGltZSI6MH0sImFjY291bnRfbnVtYmVyIjoiNTU1NTU1In19Cg==" "http://localhost:8080/api/rhsm-subscriptions/v1/tally/products/RHEL?granularity=monthly&beginning=2019-06-19T00:00:00.000Z&ending=2019-06-19T23:59:59.999Z" | python -mjson.tool

{
    "errors": [
        {
            "code": "SUBSCRIPTIONS1004",
            "detail": "Opt-in required.",
            "status": "403",
            "title": "Access Denied"
        }
    ]
}
```
2) Opt-In will 403 because subscriptions:*:* as returned from the stub client does not match the expected insights:*:* permission.  **Note that I added an 'origin' header to make the csrf check happy.**
```bash
$ curl -X PUT -H "origin: localhost.redhat.com" -H "x-rh-identity: eyJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NSIsImF1dGhfdGltZSI6MH0sImFjY291bnRfbnVtYmVyIjoiNTU1NTU1In19Cg==" "http://localhost:8080/api/rhsm-subscriptions/v1/opt-in/" | python -mjson.tool
{
    "errors": [
        {
            "code": "SUBSCRIPTIONS1003",
            "detail": "Access is denied",
            "status": "403",
            "title": "Access Denied"
        }
    ]
}
```

3) Stop the app, comment out **rhsm-subscriptions.rbacApplicationName** property and restart the application.

4) Opting in should now work.
```bash
$ curl -X PUT -H "origin: localhost.redhat.com" -H "x-rh-identity: eyJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NSIsImF1dGhfdGltZSI6MH0sImFjY291bnRfbnVtYmVyIjoiNTU1NTU1In19Cg==" "http://localhost:8080/api/rhsm-subscriptions/v1/opt-in/" | python -mjson.tool

{
    "data": {
        "account": {
            "account_number": "555555",
            "created": "2020-04-30T10:15:28.816-03:00",
            "last_updated": "2020-04-30T15:42:10.059Z",
            "opt_in_type": "API",
            "tally_reporting_enabled": true,
            "tally_sync_enabled": true
        },
        "opt_in_complete": true,
        "org": {
            "conduit_sync_enabled": true,
            "created": "2020-04-30T10:15:28.816-03:00",
            "last_updated": "2020-04-30T15:42:10.059Z",
            "opt_in_type": "API",
            "org_id": "12345"
        }
    },
    "meta": {
        "account_number": "555555",
        "org_id": "12345"
    }
}

```

5) Stop the app and re-enable the **rhsm-subscriptions.rbacApplicationName** property and restart the application.

6) Now that opt-in is complete, check to make sure that you can not access the tally api due to permissions.
```bash
$ curl -H "x-rh-identity: eyJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NSIsImF1dGhfdGltZSI6MH0sImFjY291bnRfbnVtYmVyIjoiNTU1NTU1In19Cg==" "http://localhost:8080/api/rhsm-subscriptions/v1/tally/products/RHEL?granularity=monthly&beginning=2019-06-19T00:00:00.000Z&ending=2019-06-19T23:59:59.999Z" | python -mjson.tool

{
    "errors": [
        {
            "code": "SUBSCRIPTIONS1003",
            "detail": "Access is denied",
            "status": "403",
            "title": "Access Denied"
        }
    ]
}

```

7) Same with capacity.
```bash
$ curl -H "x-rh-identity: eyJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NSIsImF1dGhfdGltZSI6MH0sImFjY291bnRfbnVtYmVyIjoiNTU1NTU1In19Cg==" "http://localhost:8080/api/rhsm-subscriptions/v1/capacity/products/RHEL?granularity=monthly&beginning=2019-06-19T00:00:00.000Z&ending=2019-06-19T23:59:59.999Z" | python -mjson.tool

{
    "errors": [
        {
            "code": "SUBSCRIPTIONS1003",
            "detail": "Access is denied",
            "status": "403",
            "title": "Access Denied"
        }
    ]
}

```

8) Stop the app, comment out **rhsm-subscriptions.rbacApplicationName** property and restart the application.

9) Because the stub API is now returning the expected permission, both the tally and capacity APIs should be accessible to the user.
```bash
$ curl -H "x-rh-identity: eyJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NSIsImF1dGhfdGltZSI6MH0sImFjY291bnRfbnVtYmVyIjoiNTU1NTU1In19Cg==" "http://localhost:8080/api/rhsm-subscriptions/v1/tally/products/RHEL?granularity=monthly&beginning=2019-06-19T00:00:00.000Z&ending=2019-06-19T23:59:59.999Z" | python -mjson.tool

{
    "data": [
        {
            "cloud_cores": 0,
            "cloud_instance_count": 0,
            "cloud_sockets": 0,
            "cores": 0,
            "date": "2019-06-01T00:00:00Z",
            "has_data": false,
            "hypervisor_cores": 0,
            "hypervisor_instance_count": 0,
            "hypervisor_sockets": 0,
            "instance_count": 0,
            "physical_cores": 0,
            "physical_instance_count": 0,
            "physical_sockets": 0,
            "sockets": 0
        }
    ],
    "meta": {
        "count": 1,
        "granularity": "MONTHLY",
        "product": "RHEL"
    }
}

$ curl -H "x-rh-identity: eyJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NSIsImF1dGhfdGltZSI6MH0sImFjY291bnRfbnVtYmVyIjoiNTU1NTU1In19Cg==" "http://localhost:8080/api/rhsm-subscriptions/v1/capacity/products/RHEL?granularity=monthly&beginning=2019-06-19T00:00:00.000Z&ending=2019-06-19T23:59:59.999Z" | python -mjson.tool

{
    "data": [
        {
            "cores": 0,
            "date": "2019-06-01T00:00:00Z",
            "has_infinite_quantity": false,
            "hypervisor_cores": 0,
            "hypervisor_sockets": 0,
            "physical_cores": 0,
            "physical_sockets": 0,
            "sockets": 0
        }
    ],
    "meta": {
        "count": 1,
        "granularity": "monthly",
        "product": "RHEL"
    }
}

```

# Additional Testing
This same scenario can be tested using an actual RBAC service by following the instructions in the following PR to set up the client. Be sure to set the rbac applicationName property to 'insights' and comment out the property to test the invalid case (subscriptions) since the test service sets the 'insights' permissions and not subscriptions (as of now) .

https://github.com/RedHatInsights/rhsm-subscriptions/pull/121

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
+++ b/src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
@@ -21,9 +21,8 @@
 
 package org.candlepin.subscriptions.security;
 
-import org.candlepin.insights.rbac.client.RbacApi;
 import org.candlepin.insights.rbac.client.RbacApiException;
-import org.candlepin.insights.rbac.client.model.Access;
+import org.candlepin.insights.rbac.client.RbacService;
 import org.candlepin.subscriptions.ApplicationProperties;
 
 import org.slf4j.Logger;
@@ -36,7 +35,7 @@ import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedGrantedAuthoritiesWebAuthenticationDetails;
 
 import java.util.Collection;
-import java.util.LinkedList;
+import java.util.Collections;
 import java.util.List;
 
 import javax.servlet.http.HttpServletRequest;
@@ -54,41 +53,23 @@ public class IdentityHeaderAuthenticationDetailsSource implements
 
     private Attributes2GrantedAuthoritiesMapper authMapper;
     private ApplicationProperties props;
-    private RbacApi rbac;
-    private String rbacApplicationName;
     private RoleProvider roleProvider;
+    private RbacService rbacController;
 
     public IdentityHeaderAuthenticationDetailsSource(ApplicationProperties props,
-        Attributes2GrantedAuthoritiesMapper authMapper,
-        RbacApi rbac, String rbacApplicationName) {
+        Attributes2GrantedAuthoritiesMapper authMapper, RbacService rbacController) {
         this.authMapper = authMapper;
         this.props = props;
-        this.rbac = rbac;
-        this.rbacApplicationName = rbacApplicationName;
+        this.rbacController = rbacController;
 
         if (props.isDevMode()) {
             log.info("Running in DEV mode. Security will be disabled.");
         }
-        this.roleProvider = new RoleProvider(rbacApplicationName, props.isDevMode());
+        this.roleProvider = new RoleProvider(props.getRbacApplicationName(), props.isDevMode());
     }
 
     protected Collection<String> getUserRoles() {
-        List<String> permissions = new LinkedList<>();
-        if (!props.isDevMode()) {
-            try {
-                // Get all permissions for the configured application name.
-                List<Access> accessList = rbac.getCurrentUserAccess(rbacApplicationName);
-                for (Access access : accessList) {
-                    if (access.getPermission() != null) {
-                        permissions.add(access.getPermission());
-                    }
-                }
-            }
-            catch (RbacApiException e) {
-                log.warn("Unable to determine roles from RBAC service.", e);
-            }
-        }
-        return roleProvider.getRoles(permissions);
+        return roleProvider.getRoles(props.isDevMode() ? Collections.emptyList() : getPermissions());
     }
 
     @Override
@@ -112,6 +93,16 @@ public class IdentityHeaderAuthenticationDetailsSource implements
     public void setUserRoles2GrantedAuthoritiesMapper(
         Attributes2GrantedAuthoritiesMapper authMapper) {
         this.authMapper = authMapper;
+    }
+
+    private List<String> getPermissions() {
+        try {
+            return rbacController.getPermissions(props.getRbacApplicationName());
+        }
+        catch (RbacApiException e) {
+            log.warn("Unable to determine roles from RBAC service.", e);
+            return Collections.emptyList();
+        }
     }
 
 }
--- a/src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
+++ b/src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
@@ -21,27 +21,23 @@
 
 package org.candlepin.subscriptions.security;
 
-import static org.candlepin.subscriptions.security.IdentityHeaderAuthenticationFilter.*;
-
+import org.candlepin.insights.rbac.client.RbacApi;
+import org.candlepin.insights.rbac.client.RbacApiException;
+import org.candlepin.insights.rbac.client.model.Access;
 import org.candlepin.subscriptions.ApplicationProperties;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.security.authentication.AuthenticationDetailsSource;
+import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.mapping.Attributes2GrantedAuthoritiesMapper;
-import org.springframework.security.web.authentication.preauth.PreAuthenticatedCredentialsNotFoundException;
+import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedGrantedAuthoritiesWebAuthenticationDetails;
-import org.springframework.util.StringUtils;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Base64;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
+import java.util.LinkedList;
+import java.util.List;
 
 import javax.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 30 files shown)


src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
```diff
@@ -56,60 +52,56 @@
...
-            Map user = (Map) identity.getOrDefault("user", Collections.emptyMap());
-
-            if (Boolean.TRUE.equals(user.get("is_org_admin")) || props.isDevMode()) {
-                userRolesList.add(ORG_ADMIN_ROLE);
+    protected Collection<String> getUserRoles() {
+        List<String> permissions = new LinkedList<>();
+        if (!props.isDevMode()) {
+            try {
+                // Get all permissions for the configured application name.
+                List<Access> accessList = rbac.getCurrentUserAccess(rbacApplicationName);
```
awood, line 80: Would it be worth adding retry functionality here (similar to what we have in UsageSnapshotProducer)?  I'm thinking that if we get random timeouts to the RBAC service that could lead to a frustrating and inconsistent experience for the user.  And more generally, what's our policy going to be if RBAC is inaccessible for whatever reason.

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
```diff
@@ -56,60 +52,56 @@
...
+                for (Access access : accessList) {
+                    if (access.getPermission() != null) {
+                        permissions.add(access.getPermission());
+                    }
+                }
```
awood, line 85:
```suggestion
permissions = accessList.stream().map(Access::getPermission).filter(Objects::nonNull).collect(Collectors.toCollection(LinkedList::new));
```

It's even shorter if you just use `Collectors.toList()`.  Are we sure we want to use a `LinkedList` here?  The memory consumption is higher due to maintaining all the pointers, and if we allocate the ArrayList at accessList.size(), we'll avoid the hit of having to resize the list.

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationManager.java
```diff
@@ -33,91 +31,47 @@
...
-
-    public IdentityHeaderAuthenticationManager(ObjectMapper mapper) {
-        this.mapper = mapper;
-    }
-
     /**
-     * This method is a little odd.  Since the value of the x-rh-identity header is JSON, there is the
-     * potential for a JSON parsing error.  However, if we parse the JSON in
-     * {@link IdentityHeaderAuthenticationFilter}, when we throw an exception on a parse error, it is not
-     * handled very nicely.  By handling the parsing in an AuthenticationManager we follow the conventions
```
awood, line 56: Is this no longer the case since we return an empty principal now?

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
```diff
@@ -114,6 +132,20 @@ public AuthenticationEntryPoint restAuthenticationEntryPoint() {
         return new RestAuthenticationEntryPoint(new IdentityHeaderAuthenticationFailureHandler(mapper));
     }
 
+    @Bean
+    public FilterRegistrationBean<OptInFilter> optInFilterRegistration(OptInController optInController) {
```
awood, line 136: Do we want to register this under the Spring Security filter chain similarly to `identityHeaderAuthenticationFilter`?

src/test/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationFilterTest.java
```diff
@@ -0,0 +1,105 @@
...
+        String emptyJson = "e30=";
+        when(request.getHeader(RH_IDENTITY_HEADER)).thenReturn(emptyJson);
+        IdentityHeaderAuthenticationFilter filter = new IdentityHeaderAuthenticationFilter(mapper);
+        assertPrincipal(filter.getPreAuthenticatedPrincipal(request), "", "");
+    }
+
+    @Test
+    public void testMissingInternalProperty() {
+        // {"identity":{"account_number":"myaccount"}}
+        String missingInternal = "eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6Im15YWNjb3VudCJ9fQ==";
```
awood, line 72:
```suggestion
        String missingInternal = Base64.getEncoder().encodeToString("{\"identity\":{\"account_number\":\"myaccount\"}}".getBytes());
```
Having to escape all the quotes isn't ideal, but I kind of like that I can be 100% sure that the base64 going into the test is what I expect.  It also makes the tests easier to update if necessary.

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
kahowell: When I asked RBAC folks about this, they stated we should give a 403 (default to unauthorized).

I think if the RBAC call fails in a detectable way, it might be worth capturing as a prometheus metric... Thoughts?

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationFilter.java
```diff
@@ -37,27 +41,46 @@
...
+        catch (Exception e) {
+            log.error(RH_IDENTITY_HEADER + " was not valid.", e);
+            // Initialize an empty principal. The IdentityHeaderAuthenticationManager will validate it.
+            return new InsightsUserPrincipal();
         }
 
+    }
+
+    private InsightsUserPrincipal createPrincipal(byte[] decodedHeader) throws Exception {
+        // TODO The identity header should eventually get deserialized into an Object.
```
kahowell, line 72: Remove this seemingly stale TODO?

src/main/java/org/candlepin/subscriptions/security/RoleProvider.java
```diff
@@ -0,0 +1,63 @@
...
+        this.rulePrefix = rulePrefix;
+        this.devModeEnabled = devModeEnabled;
+    }
+
+    public List<String> getRoles(Collection<String> permissions) {
+        if (permissions == null) {
+            return Collections.EMPTY_LIST;
+        }
+
+        if (devModeEnabled || permissions.contains(rulePrefix + ":*:*")) {
```
kahowell, line 51: Can you please leave a comment here that `rulePrefix + ":*:*"` is typically `subscriptions:*:*`? That'll make it easier to find via `git grep` :-)

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
mstead: I originally attempted to do this, but couldn't seem to get it to filter based on URL unless it was moved out the security chain. Not sure why, but I'll make another attempt.

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
mstead: Personally, I don't think that we need a retry here. If there's issues with the RBAC service I think they'd have more issues than what our app will surface. I'm not even sure that a metric would be beneficial given how often this would be an issue. But I could be convinced. :)

src/main/java/org/candlepin/subscriptions/security/RoleProvider.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationFilter.java
mstead: DONE

src/test/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationFilterTest.java
mstead: DONE - included in next commit.

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationManager.java
mstead: Yes, this shouldn't be applicable anymore. Things seemed to behave correctly in the cases that I've tested. Please feel free to see if you can break it as in the original security PR.

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
mstead: DONE - Addressed in next commit.

mstead:
NOTE: I created the RbacService to help with sonar issue while creating beans. It didn't like the autowire. This was a good change as it allowed me to move the service under the RBAC API package allowing the application (or future apps) to use the service as a 'controller' type object.

It also now makes more sense to have the rbac-service 'applicationName' property be an app property instead of a direct RBAC property since an app could potentially required more than one permission prefix. The new name of the property is:

```properties
rhsm-subscriptions.rbacApplicationName
```

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
mstead: @awood I made another attempt at doing this, but failed. I can't seem to get the url filters to get applied. Not sure why. If this is something that you'd really like to see, let's create a card to address it so that we can get this merged.

mstead:
Changes should now be addressed.

src/main/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsSource.java
kahowell: I suppose rbac itself will be aware of issues like this, won't be hard to revisit in the future if we want, so I'm fine to not worry about metrics right now (plus we might get for free in our general request metrics).

PR #134, opened 2020-05-01
Adjustment: OpenShift subscriptions include RHEL

kahowell:
Because OpenShift subscriptions run on RHEL, and include RHEL, to simplify tracking, filter out RHEL during Tally or during Capacity calculations.

Also, because this is a change that can cause some pre-existing records to need to be removed, I changed the `PoolIngressController` accordingly.

I had to change `SubscriptionCapacityKey` to an `@EmbeddedId` to make it easier to use as a quick lookup.

Because deletion requires us to query for pre-existing records, I feel this is an area that we should monitor the performance of. Accordingly I added a few metrics:

- `rhsm-subscriptions.capacity.ingress` - timer for how long it takes to
  process an ingress request
- `rhsm-subscriptions.capacity.records_{created,updated,deleted}` -
  counts of capacity record CUD operations
- `rhsm-subscriptions.capacity.pools` - count of pools processed
- `rhsm-subscriptions.capacity.whitelisted_pools` - count of pools we
  actually did work for

To test, mock data in the DB for a subscription with RHEL capacity:

```sql
insert into subscription_capacity (account_number, product_id, subscription_id, owner_id, physical_sockets, has_unlimited_guest_sockets, begin_date, end_date, sku) VALUES ('account1', 'RHEL', 'subscription1', 'org1', 10, false, '2020-01-01 00:00:00.000000', '2023-01-01 00:00:00.000000','RH000test');
```

Then, post an update for a sub that provides both OpenShift and RHEL capacity:

```sh
curl -X POST "http://localhost:8080/api/rhsm-subscriptions/v1/ingress/candlepin_pools/org1" -H  "accept: */*" -H  "Content-Type: application/json" -d "[{\"accountNumber\":\"account1\",\"activeSubscription\":true,\"startDate\":\"2020-01-01T00:00:00Z\",\"endDate\":\"2023-01-01T00:00:00Z\",\"quantity\":1,\"productId\":\"RH000test\",\"productAttributes\":[{\"name\":\"sockets\",\"value\":\"10\"}],\"providedProducts\":[{\"productId\":\"69\"},{\"productId\":\"290\"}],\"subscriptionId\":\"subscription1\",\"type\":\"NORMAL\"}]"
```

JSON (can use swagger-ui instead of curl if desired):
```json
[
  {
    "accountNumber": "account1",
    "activeSubscription": true,
    "startDate": "2020-01-01T00:00:00Z",
    "endDate": "2023-01-01T00:00:00Z",
    "quantity": 1,
    "productId": "RH000test",
    "productAttributes": [
      {
        "name": "sockets",
        "value": "10"
      }
    ],
    "providedProducts": [
      {
        "productId": "69"
      },
      {
        "productId": "290"
      }
    ],
    "subscriptionId": "subscription1",
    "type": "NORMAL"
  }
]
```

Watch the logs/check the capacity to see that the now obsolete RHEL record is removed. Other variations can be tried if desired.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
@@ -22,6 +22,7 @@ package org.candlepin.subscriptions.capacity;
 
 import org.candlepin.subscriptions.db.model.ServiceLevel;
 import org.candlepin.subscriptions.db.model.SubscriptionCapacity;
+import org.candlepin.subscriptions.db.model.SubscriptionCapacityKey;
 import org.candlepin.subscriptions.utilization.api.model.CandlepinPool;
 import org.candlepin.subscriptions.utilization.api.model.CandlepinProductAttribute;
 import org.candlepin.subscriptions.utilization.api.model.CandlepinProvidedProduct;
@@ -33,6 +34,7 @@ import org.springframework.stereotype.Component;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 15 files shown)


kahowell:
@mstead, rebased

PR #135, opened 2020-05-05
Add usage fields to capacity and snapshot tables

kahowell:
To test, launch the application, and observe that DB changes were applied successfully.


kahowell:
397a468 to 1affde5: FIXUP: add getter/setter for usage

PR #140, opened 2020-05-22
Clean up roles and authorization

mstead:
* Removed orgAdminRequired functionality now that RBAC is in place.
* Add SWATCH_ADMIN_ROLE instead of granting opt-in and reporting roles.

NOTE:
* The ReportinAccessRequired annotation is required to support the
  account whitelist check when a report is requested. It now checks
  that the user has the SWATCH_ADMIN_ROLE as well as checks that the
  account is in the whitelist.

# Testing
See #121 and #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,29 @@ extended. Please see the
 [documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html)
 for a discussion of extension points.
 
+### RBAC
+
+rhsm-subscriptions uses an RBAC service to determine application authorization. The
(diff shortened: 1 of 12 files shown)


mstead:
> Everything looks good, but I would like to request one addition: add in some of the information from #121 and #133 into the README so that the README has a concise description of how to do development using RBAC (both against a real service and against a stub).

@awood I've updated the readme with details on how to configure for RBAC. I didn't add the details about using the debug pod since that code is not accessible externally. If you'd like to see that in there anyway, or anything else for that matter, let me know.

PR #142, opened 2020-06-04
Reconfigure Spring Security to use only one AuthenticationManager

awood:
Previously we had Spring Security configured to use one
AuthenticationManager in the IdentityHeaderAuthenticationFilter and then
another AuthenticationManager (ProviderManager) configured that
delegated to the PreAuthenticatedAuthenticationProvider.  That was not
correct.  There only needs to be one AuthenticationManager.

This patch creates a single ProviderManager which delegates to an
IdentityHeaderAuthenticationProvider (other providers can be added later
if desired).  This ProviderManager is given to the
IdentityHeaderAuthenticationFilter which is in charge of all
authentication.

## Testing
The testing process is nearly identical to the one described in #133.  There is one difference to note, however.  The procedure in #133 uses an account number of "555555".  That account number needs to be added to the reporting account white list.  I have the following in my config/application.properties

```
rhsm-subscriptions.reportingAccountWhitelistResourceLocation=file:/tmp/accounts.txt
```

And that file has "555555" in it.

With that in place, the procedure outlined should work.  (I.e. this commit should keep things functioning as usual).  For extra testing, after completing the procedure, remove "555555" from the whitelist file and make the request to tally and capacity again.  These requests should now fail due to missing reporting permissions.


kahowell:
Looks good to me. Thanks for figuring this out!

PR #145, opened 2020-06-05
Retry SonarQube quality gate check a few times

kahowell:
This should ensure that even if the webhook isn't functioning correctly,
we can still get a relatively timely finish of the pipeline.

This changes the max time to 20 minutes, with up to 4 checks with 5
minutes between each.

Note that currently we've been seeing occasional issues where the
webhook is delivered to Jenkins successfully (verified via Jenkins
logs), and yet the related job is not notified/updated.


kahowell:
Note: you can see [here](https://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/blue/organizations/jenkins/subscriptions/detail/PR-145/2/pipeline/) that the last step timed out on the initial quality gate check, and then succeeded with a retry.

@awood @mstead this should finally address our PR check instability

PR #152, opened 2020-07-10
Move config to env vars

kahowell:
I put all of the variables we currently had configured via deployment automation and a few we use regularly during development.

Configuration is still overridable via properties files, so the current `spring.config.additional-location` approach should work, until we move to just environment variables.

See README.md for documentation.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
@@ -133,7 +133,7 @@ public class ApplicationConfiguration implements WebMvcConfigurer {
     @Bean
     public AccountListSource accountListSource(ApplicationProperties applicationProperties,
         AccountConfigRepository accountConfigRepository, ApplicationClock clock) {
-        if (applicationProperties.getAccountListResourceLocation() != null) {
+        if (!applicationProperties.getAccountListResourceLocation().isEmpty()) {
             return new FileAccountListSource(
                 new FileAccountSyncListSource(applicationProperties, clock),
                 new ReportingAccountWhitelist(applicationProperties, clock)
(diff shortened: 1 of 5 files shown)


kahowell:
5bfb8b1 to 491765a: s/KAFKA_API_RECONNECT_TIMEOUT/KAFKA_API_RECONNECT_TIMEOUT_MS/

src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
```diff
@@ -133,7 +133,7 @@ public static BeanFactoryPostProcessor servletInitializer() {
     @Bean
     public AccountListSource accountListSource(ApplicationProperties applicationProperties,
         AccountConfigRepository accountConfigRepository, ApplicationClock clock) {
-        if (applicationProperties.getAccountListResourceLocation() != null) {
+        if (!applicationProperties.getAccountListResourceLocation().isEmpty()) {
```
mstead, line 136: Won't this throw a NullPointerException if not set? Perhaps we should check the null case as well?

src/main/java/org/candlepin/subscriptions/files/ProductWhitelist.java
```diff
@@ -42,7 +42,7 @@
     private final PerLineFileSource source;
 
     public ProductWhitelist(ApplicationProperties properties, ApplicationClock clock) {
-        if (properties.getProductWhitelistResourceLocation() != null) {
+        if (!properties.getProductWhitelistResourceLocation().isEmpty()) {
```
mstead, line 45: Won't this throw a NullPointerException if not set? Perhaps we should check the null case as well?

src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
kahowell: Fair enough, switched to `StringUtils.hasText` accordingly.

src/main/java/org/candlepin/subscriptions/files/ProductWhitelist.java
kahowell: Fair enough, switched to `StringUtils.hasText` accordingly.

PR #153, opened 2020-07-13
Apply access check to hosts APIs

kahowell:
To test, deploy with `rhsm-subscriptions.rbacApplicationName=insights`
and observe you cannot access the APIs.

Then deploy without the rbac application name overridden, observe that
you can access the APIs.


ntkathole:
@kahowell  Not sure if it will affect in future but I just noticed that 
When user is not opt in, error message shown on hosts API call for is 

`HTTP response body: {"errors":[{"status":"403","code":"SUBSCRIPTIONS1003","title":"Access Denied","detail":"Access is denied"}]}`

while tally and capacity APIs message shows correctly

`HTTP response body: {"errors":[{"status":"403","code":"SUBSCRIPTIONS1004","title":"Access Denied","detail":"Opt-in required."}]}`

kahowell:
@ntkathole oof, one more thing to change, luckily it's a tiny PR: #154

PR #157, opened 2020-07-23
1827686: Fix capacity handling when sla/usage not specified

kahowell:
This also makes the defaulting to "_ANY" common to the ResourceUtils method, rather than having it per-resource class. (Meaning this commit fixes the mapping for the hosts API)

(edited 2020-07-28)


ntkathole:
@kahowell will this change any existing behavior in terms of tally/capacity ?
```
Tally with "NO SLA" filter = All my tally - (Standard+Premium+Self-support)
Capacity with "NO SLA" filter = All my capacity
```

kahowell:
> @kahowell will this change any existing behavior in terms of tally/capacity ?

Actually I titled this commit/PR poorly. I'll reword to reflect better what it does which is fix a regression introduced in 5506089a267922a9fe1fa49edb6f7a31381451f0. With that commit, I unintentionally changed the behavior of "NO SLA" to act as an alias for UNSPECIFIED. (Consistent, but wrong for both Tally and Capacity :-)). The goal of this PR is to restore the proper behavior from just prior to that commit, while keeping deduplication from that commit.

PR #160, opened 2020-07-29
Fix gradle permissions issue on CI hosts

kahowell:


kahowell:
Got hint from https://github.com/gradle/gradle/issues/3318#issuecomment-360116187

PR #165, opened 2020-08-10
Remove Quartz. Rely on Spring scheduling.

awood:
The using the "purge-snapshots" and "capture-snapshots" profiles, the
Spring Boot application can be started in a "one-shot" mode to run those
jobs and then quit.  Openshift's Cron features can then start pods with
those profiles active to periodically run the jobs.  In developer mode,
the jobs are run on a cron schedule set in the application properties.
If no special profiles or modes are active, the jobs do not run at all.


src/main/java/org/candlepin/subscriptions/BootApplication.java
```diff
@@ -20,16 +20,21 @@
...
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.boot.autoconfigure.quartz.QuartzAutoConfiguration;
 
 /** Bootstrapper for Spring Boot. */
-@SpringBootApplication(exclude = { QuartzAutoConfiguration.class }) // we manually import in scheduler profile
+@SpringBootApplication
 @SuppressWarnings("checkstyle:hideutilityclassconstructor")
 public class BootApplication {
+    private static final Logger log = LoggerFactory.getLogger(BootApplication.class);
```
kahowell, line 32: Unused, please remove (along with imports) - this is what SonarCloud analysis flagged.

kahowell:
Looks good. Will hold to merge until templates are updated.

PR #167, opened 2020-08-12
Add splunk sidecar to templates

kahowell:
Based on work done originally by @patrickeasters in an internal repo.

Note I adjusted all total CPU to be below 2 cores total, and all pods
below 2000Mi total memory.


kahowell:
@mstead, rebased

PR #169, opened 2020-08-19
Replace scheduler with CronJobs for purge, tally

kahowell:
To test, observe in CI, you can adjust the schedules as desired like so:

```
oc process -f templates/rhsm-subscriptions-scheduler.yml --param=KAFKA_BOOTSTRAP_HOST=platform-mq-ci-kafka-bootstrap.platform-mq-ci.svc.cluster.local '--param=CAPTURE_SNAPSHOT_SCHEDULE=0 * * * *' '--param=PURGE_SNAPSHOT_SCHEDULE=5 * * * *' --param=IMAGE_TAG=05004c5 | oc apply -f -
```

By implementing in the scheduler template, we don't need app-interface changes (I'm arguing that the cronjobs are just reimplementations of the scheduler function).

I applied a deadline to the both the job/pod of 30 minutes.

Note that I did not apply splunk forwarder sidecar to these images as it's not designed to be used with a terminating pod (the sidecar keeps running, preventing the job from completing). Given that we're queueing up work, but not performing it, and that the completed job pods' logs are viewable in OpenShift, deferring sidecar integration for now (future Kubernetes versions may give a clean, simple solution, see kubernetes/enhancements#753).


kahowell:
@mstead FYI, this should be merged at the same time as #165 ; they're co-dependent.

kahowell:
d0f62bb to d1c9938: adjust cron schedules - got confused in the transition between Quartz expression and cron expressions.

PR #176, opened 2020-09-02
Page (rather than stream) inventory DB queries

kahowell:
To test, connect to an existing DB with some data.

Port-forward a DB from the QA environment, and configure it via:

```
RHSM_SUBSCRIPTIONS_INVENTORY_SERVICE_DATASOURCE_URL=jdbc:postgresql://localhost:8999/insights ./gradlew bootRun
```

Then, use JMX to tally an account with a large number of hosts.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
@@ -93,6 +93,18 @@ public class ApplicationProperties {
     private int accountBatchSize = 500;
 
     /**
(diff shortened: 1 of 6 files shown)


kahowell:
a1fd2d3 to 92bac5f: add order by to rhsm hypervisor query

mstead:
@kahowell Since the other fix addressed our performance issues, is this PR still required? If not we should close it and re-open if we need to.

mstead:
Closed as this is not needed at the moment. Let's re-open should we need this code.

PR #178, opened 2020-09-09
Use generated ID instead of inventory_id for Hosts

mstead:
When an ID is manually set on the Host record, spring-data
repositories will always do a merge when saving an entity.
This results in a select being executed every time a new
host is created which in not necessary since we are
deleting all hosts for an account before a tally is run.

By using a generated ID and storing the inventory_id as a
separate property, we avoid the select statements, which
improves the overall performance.

**NOTE:** I opted to delete the hosts and host_tally_buckets tables via liquibase because it was a lot more difficult to add the ID column with a generated UUID. Because of this, the tally process will need to be re-run to repopulate the host data.


mstead:
Closing this PR in favor of another related to performance improvements for host insertion.

PR #179, opened 2020-09-09
Insert host performance improvements

mstead:
This PR addresses a few items that improves the host import process.

1. When an ID is manually set on the Host record, spring-data
repositories will always do a merge when saving an entity.
This results in a select being executed every time a new
host is created which in not necessary since we are
deleting all hosts for an account before a tally is run.

By using a generated ID and storing the inventory_id as a
separate property, we avoid the select statements, which
improves the overall performance.

2. The delete all hosts by account number query has
been improved quite a bit shaving a little more time off the
overall tally run.

3. Batch insert is now enabled when persisting to the DB.

### DIFF
--- a/src/main/resources/rhsm-subscriptions.properties
+++ b/src/main/resources/rhsm-subscriptions.properties
@@ -7,6 +7,11 @@
 
 spring.jmx.enabled=${JMX_ENABLED:true}
 
+# General hibernate configurations
+spring.jpa.properties.hibernate.jdbc.batch_size=${JDBC_BATCH_SIZE:100}
+spring.jpa.properties.hibernate.order_inserts=true
+spring.jpa.properties.hibernate.order_updates=true
+
 rhsm-subscriptions.dev-mode=${DEV_MODE:false}
 rhsm-subscriptions.prettyPrintJson=${PRETTY_PRINT_JSON:false}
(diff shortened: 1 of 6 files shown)


src/main/resources/rhsm-subscriptions.properties
```diff
@@ -5,6 +5,12 @@
 # We're keeping the context-path as "/" so that the actuators we use for OKD liveness/readiness probes have a
 # fixed path.  The actual subscriptions resources will be configured to hang off of the path below.
 
+# General hibernate configurations
+spring.jpa.properties.hibernate.jdbc.batch_size=${JDBC_BATCH_SIZE:100}
+spring.jpa.properties.hibernate.order_inserts=true
+spring.jpa.properties.hibernate.order_updates=true
+#spring.jpa.properties.hibernate.batch_versioned_data=true
```
kahowell, line 12: Can you please leave a comment as to why we would uncomment this, and what default value is (looks like default is `false`?)

kahowell (changes requested):
One tiny request, but overall looks great. Kudos!

kahowell:
And needs a rebase.

src/main/resources/rhsm-subscriptions.properties
mstead: Actually, that wasn't supposed to get in there. That was an oversight on my part and would only be used if we were versioning our entities.

I'll remove it.

mstead:
@kahowell Requested changes made. Thanks for the quick review. :+1:

kahowell (approved):

PR #180, opened 2020-09-09
Upgrade spring-boot to 2.2.9

mstead:
Upgrade requires that we enable JMX since spring-boot
now disables it by default.


src/main/resources/rhsm-subscriptions.properties
```diff
@@ -5,6 +5,8 @@
 # We're keeping the context-path as "/" so that the actuators we use for OKD liveness/readiness probes have a
 # fixed path.  The actual subscriptions resources will be configured to hang off of the path below.
 
+spring.jmx.enabled=${ENABLE_JMX:true}
```
kahowell, line 8: Can we please call it `JMX_ENABLED` instead of `ENABLE_JMX`? (more consistent with the spring property name).

kahowell (changes requested):

src/main/resources/rhsm-subscriptions.properties
mstead: DONE

kahowell (approved):

PR #184, opened 2020-09-15
Support cloud in Host.hardware_type API field

mstead:
1) Removed HardwareType column on Host entity.

While the column was removed, the API's Host response
object still contains the hardware_type property so that
the UI does not have to change. This property is now
populated by the measurement_type of the Host record.

2) Track virtual during tally for non-RHEL products

This was done to prevent systems from getting classified as
TOTAL when host data is persisted.


mstead:
Closing in favor of #196

PR #187, opened 2020-09-21
1878665: No longer throw exception collection 0 socket RHEL hypervisor

mstead:
Instead of throwing an exception when collecting for a 0 socket
RHEL hypervisor, create the bucket with 0 sockets allowing it to
be filtered by the API as it currently does.

Previously, the exception was not being handled correctly leading to
an entire tally run failing for an account reporting one of these
hypervisors.

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/tally/collector/RHELProductUsageCollectorTest.java
+++ b/src/test/java/org/candlepin/subscriptions/tally/collector/RHELProductUsageCollectorTest.java
@@ -22,16 +22,20 @@ package org.candlepin.subscriptions.tally.collector;
 
 import static org.candlepin.subscriptions.tally.collector.Assertions.*;
 import static org.candlepin.subscriptions.tally.collector.TestHelper.*;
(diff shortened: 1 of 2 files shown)


src/test/java/org/candlepin/subscriptions/tally/collector/RHELProductUsageCollectorTest.java
```diff
@@ -96,13 +100,15 @@ public void testCountsForPhysicalSystem() {
     }
 
     @Test
-    public void hypervisorReportedWithNoSocketsWillRaiseException() {
-        NormalizedFacts facts = hypervisorFacts(0, 0);
-        NormalizedFacts guestFacts = guestFacts(4, 1, false);
+    public void hypervisorReportedWithNoSocketsDefaultToZero() {
```
kahowell, line 103: sonarcloud.io flagged the public modifier as unnecessary.

kahowell (changes requested):
Looks good overall. Left a comment related to the sonarcloud issue.

kahowell (approved):

PR #189, opened 2020-09-22
Apply origin/referer check to all jolokia requests

kahowell:
Even GET requests, as those can actually cause bean invocation in Jolokia.

To be consistent with the AntiCsrfFilter, I made it so that `DEV_MODE=true` causes the origin/referer check to be turned off.

To test, try running in the following configurations:

1. `DEV_MODE=true ./gradlew bootRun`: http://localhost:8080/actuator/jolokia will be fully open
2. `./gradlew bootRun`: http://localhost:8080/actuator/hawtio will be non-functional because /actuator/jolokia requests are rejected based on the default origin suffix of .redhat.com w/ port 443.
3. `RHSM_SUBSCRIPTIONS_ANTI_CSRF_DOMAIN_SUFFIX=localhost RHSM_SUBSCRIPTIONS_ANTI_CSRF_PORT=8080 ./gradlew`: http://localhost:8080/actuator/hawtio be functional

Note that there is some nuance, as hitting `/actuator/jolokia` with `GET` (when visiting with a browser), doesn't send an `Origin` header. So with the restrictor turned on, visiting /actuator/jolokia is fruitless even with the correct domain suffix/port set.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/security/JolokiaRestrictor.java
+++ b/src/main/java/org/candlepin/subscriptions/security/JolokiaRestrictor.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2009 - 2020 Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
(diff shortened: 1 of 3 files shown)


kahowell:
044088f to 2e8301e: added application.yaml changes.

kahowell:
2e8301e to 6a6fec2: rebase

awood (changes requested):
I had to rebase onto master to get this running because of the Java 11 changes, but even with the rebase, I'm not getting the behavior you describe.  Hawtio comes up and I can invoke JMX beans just fine.  The Jolokia endpoint is accessible also.

I'm not sure where (or if) I'm going wrong exactly.  Would you mind doing a double-check on your end and make sure it works as expected?  If so, we can sit down and maybe try to puzzle out what's going wrong on my end.

kahowell:
@awood I rebased locally, and it still works as described. Is it possible you have a `config/application.properties` locally or an env var where you've set `DEV_MODE`? Otherwise, let's sync up on Monday about it.

awood:
>  Is it possible you have a `config/application.properties` locally or an env var where you've set `DEV_MODE`?

D'oh

kahowell:
6a6fec2 to 948698d: rebase

kahowell:
Closed in favor of #202

PR #190, opened 2020-09-23
1827686: Fix capacity SLA/Usage UNSPECIFIED values

kahowell:
Instead of null, for consistency, they should be empty string.

Prior work in #183 doesn't fix the issue described without this change
as well.

To test try:

```
curl -X POST "http://localhost:8080/api/rhsm-subscriptions/v1/ingress/candlepin_pools/org1" -H  "accept: */*" -H  "Content-Type: application/json" -d "[{\"accountNumber\":\"account1\",\"activeSubscription\":true,\"startDate\":\"2020-01-01T00:00:00Z\",\"endDate\":\"2023-01-01T00:00:00Z\",\"quantity\":1,\"productId\":\"RH000test\",\"productAttributes\":[{\"name\":\"sockets\",\"value\":\"10\"},{\"name\":\"usage\",\"value\":\"Production\"}],\"providedProducts\":[{\"productId\":\"69\"},{\"productId\":\"290\"}],\"subscriptionId\":\"subscription1\",\"type\":\"NORMAL\"}]"
```

and observe that the SLA written in the DB is '' rather than null.
Alternatively, you can query back out the capacity for SLA unspecified
w/ the API.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/CandlepinPoolCapacityMapper.java
@@ -180,12 +180,11 @@ public class CandlepinPoolCapacityMapper {
             ServiceLevel slaValue = ServiceLevel.fromString(sla.get());
             if (slaValue == ServiceLevel.UNSPECIFIED) {
                 log.warn("Product {} has unsupported service level {}", pool.getProductId(), sla.get());
-                return null;
             }
             return slaValue;
         }
 
-        return null;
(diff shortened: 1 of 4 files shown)


awood (commented):
Do we need any database migrations to go along with this to fix existing rows?

kahowell:
> Do we need any database migrations to go along with this to fix existing rows?

Sure. Will update this PR with a migration briefly.

kahowell:
@awood please take another look when you get a chance.

awood (approved):

PR #191, opened 2020-09-29
Jenkins: Run all gradle steps inside container

kahowell:
This makes it easier to reproduce a Jenkins failure locally, and allows us to more easily update the base we are running on top of,
specifically, when we go to Java 11, it will simply be a matter of replacing `registry.access.redhat.com/ubi8/openjdk-8` with `registry.access.redhat.com/ubi8/openjdk-11`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,17 +6,17 @@ pipeline {
     stages {
         stage('Clean') {
             steps {
-                sh './gradlew --no-daemon clean'
+                sh "./podman_run.sh ./gradlew --no-daemon clean"
             }
         }
         stage('Build') {
             steps {
-                sh './gradlew --no-daemon assemble'
+                sh "./podman_run.sh ./gradlew --no-daemon assemble"
             }
         }
         stage('Unit tests') {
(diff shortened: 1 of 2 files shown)


kahowell:
@awood @mstead, Was thinking on Java 11 and decided to see if this was a viable solution for making it easy to update our Java version in use. Thoughts?

awood (approved):
This is awesome.  I like this a lot.  The only thing to maybe add is create some aliases or tasks in our buildfile so that we can execute the same podman command.

Or even just create a 'podman.sh` file that exports PODMAN_RUN and/or an alias so we can source that file in our console and then replicate a Jenkins run.

```
$ source podman.sh
# podman.sh defines PODMAN_RUN and creates an alias that equates to $PODMAN_RUN ./gradlew --no-daemon
$ jenkins-podman assemble
```

kahowell:
As suggested, I made it a script: `./podman_run.sh ./gradlew --no-daemon clean` works locally and is what also Jenkins runs.

awood (approved):

PR #195, opened 2020-10-01
Cloudigrade: skip serviceType for now

kahowell:
We plan to support serviceType in the future, but we don't yet, but cloudigrade is already collecting it, so we need to simply skip any
entries with a `service_type` other than `_ANY`.

Without this change, even when service type is unset, we end up with two values for it: `''` (unset), and `_ANY` (any/all service types). This was causing us to double count machines.


kahowell:
@ntkathole FYI this is the fix for ENT-3091

awood (approved):

PR #196, opened 2020-10-01
Make changes to system table API to support GUI

mstead:
Added DB/API fields:
* is_unmapped_guest
* is_hypervisor
* cloud_provider

HardwareType is one of 3 values:
* CLOUD
* PHYSICAL
* VIRTUALIZED

Fix broken script and set new fields

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -1077,6 +1077,15 @@ components:
           description: Latest timestamp that Subscription Watch has for this host.
           type: string
           format: date-time
+        is_unmapped_guest:
+          description: Is this host an unmapped guest (not reported by virt-who)?
+          type: boolean
+        is_hypervisor:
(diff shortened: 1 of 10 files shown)


kahowell (changes requested):
Overall, makes sense and works with some tweaks (see inline suggestion)

san7ket:
@kahowell I have a question, I see that there is this new field we are adding, `is_unmapped_guest`. Earlier, these orphan guests were treated as a separate virtual host, that consumed a sub. Now, with addition of the new field will that workflow stlll remain true, or will we simply skip these unmapped guests in tally, like we do for the mapped guests for a hypervisor.

kahowell:
@san7ket the workflow will stay the same. Our intent in providing the fields is that if the GUI wants to do something with them later, they can (e.g. if the GUI needs to distinguish between hypervisors and unmapped guests visually, it will have a flag available for that).

kahowell (approved):

PR #200, opened 2020-10-05
Bump memory request to satisfy service limit

kahowell:
New service limit ratio between limit and request is 2 or less.


kahowell:
There is a planned cluster-level change to enforce this limit.

awood (approved):

PR #202, opened 2020-10-08
Apply AntiCsrfFilter check to all Jolokia requests.

awood:
This commit adds a special version of AntiCsrfFilter that includes GET
requests.  Normally GET requests are exempted from CSRF checks as GET
requests are not intended to modify state on the server.  However, with
Jolokia even GET requests can actually cause a JMX bean invocation.

Note browsers don't send Origin or Referer headers when making GET
requests so hitting `/actuator/jolokia` will fail the CSRF filter.
Use curl instead or proxy your JMX request by using the /actuator/hawtio
endpoint.

---
# Testing
Test Requests (DEV_MODE must be false):

1. No auth, but origin for Jolokia GET: `curl -v -H "Origin: localhost:8080" -X GET "http://localhost:8080/actuator/jolokia"`
	Expect: Access allowed
	Actual: Access allowed

2. No auth, no origin for Jolokia GET: `curl -v -X GET "http://localhost:8080/actuator/jolokia"`
	Expect: Access denied - bad origin
	Actual: Access is denied but with a slightly confusing error message.  The error message returned states that the user could not be authenticated. Walking through the debugger shows the failure actually occurs on the origin match, but the error handler that gets invoked is the `IdentityHeaderAuthenticationFailureHandler`.  This is slightly annoying but I don't know that it's worth additional time investment to correct.  If reviewers feel it is worth correcting, just flag the PR as needing changes and comment accordingly.

3. Auth, but no origin for Jolokia GET: `curl -v -H "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjIzNCIsICJ1c2VyIjp7ImlzX29yZ19hZG1pbiI6dHJ1ZX0sICJpbnRlcm5hbCI6IHsib3JnX2lkIjogIm15T3JnIn19fX0K" -X GET "http://localhost:8080/actuator/jolokia"`
	Expect: Access denied - bad origin
	Actual: Access denied

3. No auth, no origin for other Actuator GET: `curl -v -X GET "http://localhost:8080/actuator/health"`
	Expect: Access allowed
	Actual: Access allowed

4. No auth, no origin for API PUT: `curl -v -X PUT "http://localhost:8080/api/rhsm-subscriptions/v1/opt-in"`
	Expect: Access denied
	Actual: Access denied

5. Auth, no origin for API PUT: `curl -v -H "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjIzNCIsICJ1c2VyIjp7ImlzX29yZ19hZG1pbiI6dHJ1ZX0sICJpbnRlcm5hbCI6IHsib3JnX2lkIjogIm15T3JnIn19fX0K" -X PUT "http://localhost:8080/api/rhsm-subscriptions/v1/opt-in"`
	Expect: Access denied - bad origin
	Actual: Access denied - bad origin

6. Auth and origin for API PUT: `curl -v -H "Origin: localhost:8080" -H "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjIzNCIsICJ1c2VyIjp7ImlzX29yZ19hZG1pbiI6dHJ1ZX0sICJpbnRlcm5hbCI6IHsib3JnX2lkIjogIm15T3JnIn19fX0K" -X PUT "http://localhost:8080/api/rhsm-subscriptions/v1/opt-in"`
	Expect: Access allowed
	Actual: Access allowed

7. Auth and no origin for API GET: `curl -v -H "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjIzNCIsICJ1c2VyIjp7ImlzX29yZ19hZG1pbiI6dHJ1ZX0sICJpbnRlcm5hbCI6IHsib3JnX2lkIjogIm15T3JnIn19fX0K" -X GET "http://localhost:8080/api/rhsm-subscriptions/v1/opt-in"`
	Expect: Access allowed
	Actual: Access allowed

8. No auth and no origin for API GET: `curl -v -X GET "http://localhost:8080/api/rhsm-subscriptions/v1/opt-in"`
	Expect: Access denied
	Actual: Access denied

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
+++ b/src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
@@ -32,6 +32,8 @@ import org.springframework.boot.actuate.autoconfigure.security.servlet.EndpointR
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.core.annotation.Order;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yaml
```diff
@@ -29,6 +29,6 @@ spring:
     active:
       - api
       - in-memory-queue
-      - capacity-ingress
+      # - capacity-ingress
```
kahowell, line 32: Unintentional commit?

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
```diff
@@ -42,144 +44,195 @@
...
+        protected void configure(HttpSecurity http) throws Exception {
+            // See https://docs.spring.io/spring-security/site/docs/current/reference/html5/#ns-custom-filters
+            // for list of filters and their order
+            http
+                .requestMatcher(EndpointRequest.to("jolokia")) // restrict this config to Jolokia
+                .csrf().disable()
+                .addFilter(identityHeaderAuthenticationFilter())
+                .addFilterBefore(getVerbIncludingAntiCsrfFilter(appProps, env), LogoutFilter.class)
+                .authorizeRequests()
+                    .requestMatchers(EndpointRequest.to("jolokia")).permitAll();
```
kahowell, line 235: See note above about /actuator/hawtio/jolokia...

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
```diff
@@ -42,144 +44,195 @@
...
+            filter.setContinueFilterChainOnUnsuccessfulAuthentication(false);
+            return filter;
+        }
+
+        @Override
+        protected void configure(HttpSecurity http) throws Exception {
+            // See https://docs.spring.io/spring-security/site/docs/current/reference/html5/#ns-custom-filters
+            // for list of filters and their order
+            http
+                .requestMatcher(EndpointRequest.to("jolokia")) // restrict this config to Jolokia
```
kahowell, line 230: Unfortunately, there is one more subpath to match. Hawtio creates a [forwarding controller](https://github.com/hawtio/hawtio/blob/c9d65be7081072ac836218c51282c91b99dacc4c/platforms/springboot/src/main/java/io/hawt/springboot/HawtioManagementConfiguration.java#L308) from `/actuator/hawtio/jolokia` (and subpaths) to `/actuator/jolokia`, so a couple quick tests that need to be addressed:
```sh
for subpath in actuator/jolokia actuator/jolokia/list actuator/hawtio/jolokia actuator/hawtio/jolokia/list; do
  test "$(curl -s http://localhost:8080/$subpath -i | head -n 1 | cut -d' ' -f2)" != "401" && echo "[FAILED] - curl -s http://localhost:8080/$subpath"
done
```
shows:
```
[FAILED] - curl -s http://localhost:8080/actuator/hawtio/jolokia
[FAILED] - curl -s http://localhost:8080/actuator/hawtio/jolokia/list
```

kahowell (changes requested):
I dig the approach and am happy to see the terminology cleanup (e.g. methods -> verbs). Couple small things to address (see inline comments).

src/main/resources/application.yaml
awood: Yep, that was an accident.

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
awood: Fixed:

Thanks for the handy shell test loop!
```
% for subpath in actuator/jolokia actuator/jolokia/list actuator/hawtio/jolokia actuator/hawtio/jolokia/list; do
  curl -H "Origin: https://localhost:8080" -s http://localhost:8080/$subpath -i | head -n 1 | cut -d' ' -f2 
done
200
200
200
200

% for subpath in actuator/jolokia actuator/jolokia/list actuator/hawtio/jolokia actuator/hawtio/jolokia/list; do
  curl -s http://localhost:8080/$subpath -i | head -n 1 | cut -d' ' -f2 
done
401
401
401
401
```

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
awood: I just removed this bit.  The other security config handles access to all the actuator stuff.

awood: Okay, I updated this with a second commit.  (Feel free to squash merge if you like).

I've kept the same access control on the actuator endpoints as before (the endpoints are referenced by name and not by specific paths).  If we want to tighten it up we can, but my previous commit was too lenient with the other endpoints.  The Jolokia access is controlled in the config class specific to the Jolokia stuff.  We can require that people be authenticated there or belong to certain roles or anything similar.

src/main/java/org/candlepin/subscriptions/security/SecurityConfig.java
```diff
@@ -42,144 +44,198 @@
...
+            return filter;
+        }
+
+        @Override
+        protected void configure(HttpSecurity http) throws Exception {
+            // See https://docs.spring.io/spring-security/site/docs/current/reference/html5/#ns-custom-filters
+            // for list of filters and their order
+            http
+                .requestMatchers(matchers ->
+                    matchers.antMatchers("/actuator/**/jolokia", "/actuator/**/jolokia/*"))
```
kahowell, line 233:
```suggestion
                    matchers.antMatchers("/actuator/**/jolokia", "/actuator/**/jolokia/**"))
```

kahowell (changes requested):
One more test I thought of: `curl -H "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjIzNCIsICJ1c2VyIjp7ImlzX29yZ19hZG1pbiI6dHJ1ZX0sICJpbnRlcm5hbCI6IHsib3JnX2lkIjogIm15T3JnIn19fX0K" http://localhost:8080/actuator/jolokia/list/foo` (i.e. a subpath of a subpath, which does happen in Jolokia's API)

kahowell (approved):

kahowell:
90d14ab to 9cc88b4: squashed

PR #203, opened 2020-10-13
1884770: Consider host as virtual when satellite maps hypervisor

mstead:
Older versions of satellite will not send the infrastructure type
to inventory, so when a host's facts are normalized, consider the
host as virtual if satellite reported the host's hypervisor UUID.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
@@ -108,7 +108,7 @@ public class FactNormalizer {
     }
 
     private boolean isVirtual(InventoryHostFacts hostFacts) {
-        return hostFacts.isVirtual() ||
+        return hostFacts.isVirtual() || StringUtils.hasText(hostFacts.getSatelliteHypervisorUuid()) ||
             "virtual".equalsIgnoreCase(hostFacts.getSystemProfileInfrastructureType());
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
```diff
@@ -108,7 +108,7 @@ private void normalizeUnits(NormalizedFacts normalizedFacts, InventoryHostFacts
     }
 
     private boolean isVirtual(InventoryHostFacts hostFacts) {
-        return hostFacts.isVirtual() ||
+        return hostFacts.isVirtual() || !StringUtils.isEmpty(hostFacts.getSatelliteHypervisorUuid()) ||
```
awood, line 111: The `hasText` method makes it so you don't have to have negation in there.  I think it reads easier. 

```suggestion
        return hostFacts.isVirtual() || StringUtils.hasText(hostFacts.getSatelliteHypervisorUuid()) ||
```

awood (changes requested):
Looks good aside from the little suggestion I made.  If you want to keep it the way it is, that's fine.  Just say so and I will approve.

src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
mstead: Done!

awood (approved):

PR #204, opened 2020-10-19
Incorporate opt-in checks into the Spring Security authorization layer

awood:
Rather than use a specific filter for the opt-in check, we can use
Spring Security's authorization facilities.  This leads to a cleaner
design and a less confusing architecture.


kahowell:
Overall I dig it. I noticed if you try to access one of the APIs without auth you get:

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class org.candlepin.subscriptions.security.InsightsUserPrincipal (java.lang.String is in module java.base of loader 'bootstrap'; org.candlepin.subscriptions.security.InsightsUserPrincipal is in unnamed module of loader 'app')
```.

Maybe just returning false if the principal is not `instanceof InsightsUserPrincipal` would work? Not picky about how to solve.

kahowell (changes requested):
See above comment

awood:
> See above comment

Should be resolved now.

kahowell (approved):

PR #205, opened 2020-10-23
Awood/better logging

awood:
* Suppress Spring Security stacktraces.  The card for the task mentions suppressing stacktraces for validation errors, but I didn't actually see any significant uses of the `@Validated` annotation.  If I'm overlooking a spot, please let me know.

* Add in a filter that places the principal in the logback Mapped Diagnostic Context (MDC) so that we can associate log messages with the user who triggered them.

* Modify log message format somewhat.  Remove some MDC references that we don't even have (likely a copy-paste error from the Candlepin logback.xml) and tidy up for format some.

## Testing
`curl -H "x-rh-identity: $(echo -n '{"identity":{"account_number":"account12345","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' 
| base64 -w 0)" "http://localhost:8080/api/rhsm-subscriptions/v1/hosts/products/RHEL?usage=Production&limit=1"`

That account is not opted-in, so an AccessDeniedException gets thrown.  With this commit, the error message appears, but there should be no accompanying stacktrace in the console log.

You can run some other tests as well.  For example, if you munge the JSON of the curl command to create invalid JSON, you'll see the error message from `org.candlepin.subscriptions.security.IdentityHeaderAuthenticationFailureHandler` has the stacktrace suppressed.  Oddly, you will still see a stacktrace emanating from `org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/].[org.candlepin.subscriptions.JaxrsApplication]` but that's not something we have control over I don't think.  Personally, I think malformed JSON in the headers is rare enough that we should sweat it too much.

### DIFF
--- a/src/main/resources/logback-spring.xml
+++ b/src/main/resources/logback-spring.xml
@@ -3,7 +3,7 @@
     <contextName>rhsm-subscriptions</contextName>
     <appender name="ConsoleAppender" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{ISO8601} [thread=%thread] [%X{requestType}=%X{requestUuid}, org=%X{org}, csid=%X{csid}] %-5p %c - %m%n</pattern>
+            <pattern>%d{ISO8601} [thread=%thread] [%-5p] [%c] %X{user}- %m%n</pattern>
         </encoder>
     </appender>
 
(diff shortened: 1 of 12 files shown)


src/main/resources/logback-spring.xml
```diff
@@ -16,7 +22,7 @@
         <totalSizeCap>100MB</totalSizeCap>
         </rollingPolicy>
         <encoder>
-            <pattern>%d{yyyy-MM-dd'T'HH:mm:ss.SSS'+0000'} [thread=%t] [level=%p] [category=%c] %m%n</pattern>
+            <pattern>%d{yyyy-MM-dd'T'HH:mm:ss.SSS'+0000'} [thread=%t] [level=%p] [category=%c] %X{user} - %m%n%ex{full, SECURITY_STACKTRACE_EVAL}  </pattern>
```
kahowell, line 25:
```suggestion
            <pattern>%d{yyyy-MM-dd'T'HH:mm:ss.SSS'+0000'} [thread=%t] [level=%p] [category=%c] %X{user} - %m%n%ex{full, SECURITY_STACKTRACE_EVAL}</pattern>
```

kahowell (changes requested):
@awood tiny fix suggested. Otherwise looks good.

awood:
> @awood tiny fix suggested. Otherwise looks good.

TYVM!

kahowell (approved):

PR #206, opened 2020-10-23
Deploy kafka, HBI, and DBs via podman-compose

kahowell:
In order to simplify our development setup, I looked at what some of the
other teams for cloud.redhat.com applications were doing, and
docker-compose/podman-compose is commonly used.

To that end, I am proposing that we replace our manual DB setup and
inventory service scripts with a `podman-compose` based set up. See
README for details.

Also, in the future if we integrate with other services, they could
feasibly be added here. Or for example, we might add cloudigrade as a
service in this way.

Note there are some huge gotchas with this approach. Specifically
`podman-compose` assumes that all services can co-exist in the same
network namespace; in other words port-conflicts are problematic.

To prevent conflicts with DB instances, a single postgresql container is
spun up and during initialization, rhsm-subscriptions and insights DBs
are created (along with users).

The overall goal I have here is that development is simply:

1. `git clone --recurse-submodules https://github.com/RedHatInsights/rhsm-subscriptions`
2. `cd rhsm-subscriptions`
3. `sudo dnf instally -y podman-compose podman java-11-openjdk-devel`
4. `podman-compose up -d`
5. `./gradlew bootRun`

A couple of non-obvious changes were made here:

1. HBI submodule updated to latest, because they added `dockerfile.dev`
2. inventory-db property change to reflect the default DB name in
   inventory: `insights` rather than `inventory`.


kahowell:
5c38464 to ab6c7d7: fix README anchor link

awood (approved):
:smiley: This is wonderful!

PR #207, opened 2020-10-26
Always sort hosts by ID initially

kahowell:
That way, there is always a consistent ordering from the API. Without this change, the DB is free to (and has according to ENT-3168), return hosts in a non-deterministic, and unfortunately, inconsistent order which order which causes issues with the table display.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resource/HostsResource.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/HostsResource.java
@@ -80,14 +80,16 @@ public class HostsResource implements HostsApi {
     public HostReport getHosts(String productId, Integer offset, Integer limit, String sla,
         String usage, Uom uom, HostReportSort sort, SortDirection dir) {
 
-        Sort sortValue = Sort.unsorted();
         Sort.Direction dirValue = Sort.Direction.ASC;
-
         if (dir == SortDirection.DESC) {
             dirValue = Sort.Direction.DESC;
         }
+        Sort.Order implicitOrder = Sort.Order.by("id");
+        Sort sortValue = Sort.by(implicitOrder);
+
         if (sort != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/subscriptions/resource/HostsResource.java
```diff
@@ -80,14 +80,19 @@ public HostsResource(HostRepository repository, PageLinkCreator pageLinkCreator)
...
+        Sort.Order implicitOrder = Sort.Order.by("id");
+        Sort sortValue;
+
         if (sort != null) {
-            sortValue = Sort.by(dirValue, SORT_PARAM_MAPPING.get(sort));
+            Sort.Order userDefinedOrder = new Sort.Order(dirValue, SORT_PARAM_MAPPING.get(sort));
+            sortValue = Sort.by(userDefinedOrder, implicitOrder);
+        }
+        else {
+            sortValue = Sort.by(implicitOrder);
```
awood, line 95:
```java
        Sort sortValue = Sort.by(implicitOrder);

        if (sort != null) {
            Sort.Order userDefinedOrder = new Sort.Order(dirValue, SORT_PARAM_MAPPING.get(sort));
            sortValue = Sort.by(userDefinedOrder, implicitOrder);
        }
```

^^^ What do you think?  Makes it a little more succinct.

awood (approved):
Small suggestion just for brevity.  Otherwise looks fine.

kahowell:
@awood  2123dcc to 5f9ef82: applied your suggestion, fixed some line length issues (force pushed).

src/main/java/org/candlepin/subscriptions/resource/HostsResource.java
kahowell: Done

awood (approved):

PR #208, opened 2020-10-29
Awood/measurement type tweak

awood:


src/main/java/org/candlepin/subscriptions/resource/HostsResource.java
```diff
@@ -56,13 +56,15 @@
...
-        HostReportSort.LAST_SEEN, "key.host.lastSeen"
-    );
+    public static final Map<HostReportSort, String> SORT_PARAM_MAPPING =
+        ImmutableMap.<HostReportSort, String>builderWithExpectedSize(5)
+        .put(HostReportSort.DISPLAY_NAME, "key.host.displayName")
+        .put(HostReportSort.CORES, "cores")
+        .put(HostReportSort.HARDWARE_TYPE, "key.host.hardwareType")
+        .put(HostReportSort.SOCKETS, "sockets")
+        .put(HostReportSort.LAST_SEEN, "key.host.lastSeen")
+        .put(HostReportSort.MEASUREMENT_TYPE, "measurement_type")
```
kahowell, line 66:
```suggestion
        .put(HostReportSort.MEASUREMENT_TYPE, "measurementType")
```

kahowell (changes requested):
@awood a few changes needed:
 - we'll need to update the trigger we've had in place in `src/main/resources/liquibase/202002061140-update-copy-measurement-procedure.xml` 
 - `bin/insert-mock-hosts` should be updated accordingly
 - `src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java` could use a test for sorting
 - `HostsResource` needs a quick fix for the field mapping (see inline comment)

kahowell (approved):

PR #209, opened 2020-11-02
Update copyrights

kahowell:
1. Update notices to beginning in 2019 rather than 2009.
2. Update the copyright regex to allow just "2020", as this is a more accurate way to begin a new class file.

Prior we were using the header copyright range we were using from Candlepin. However, that's inaccurate, so this fixes those year ranges.


awood (approved):

awood:
Some of them now say "2019 - 2019".  Personally, I don't really much care, but I thought I'd mention it in case it bothers you and you want to change it.

kahowell:
@awood, may as well fix it while I'm at it. Will push one more commit.

PR #210, opened 2020-11-03
Refactor properties to be IDEA friendly

kahowell:
Rename rhsm-subscriptions.properties to application.properties. This
enables IDE support (autocompletion, hints) for the properties file with
no additional configuration needed. Additionally, this eliminates the
need to specify @PropertySource.

Additionally, I refactored the retention properties to be picked up via
`@Component`, as for some reason, otherwise IDEA wasn't picking it up
properly.

There seem to be some limitations with the properties support, namely
`@ConfigurationProperties` annotations on `@Beans` don't seem to be
picked up, and it seems that Map objects are not well supported. I
couldn't get a few properties to be picked up by IDEA due to these
limitations:

- `rhsm-subscriptions.package_uri_mappings.org.candlepin.subscriptions`
  because it's not implemented as a type-safe config value.
- `hawtio.*` as the hawtio properties are implemented as a `Map`
  internally.


awood (approved):
Looks fine, but now we have `application.yaml` and `application.properties`.  I had originally (like when the project first started) created `rhsm-subscriptions.properties` distinctly from `application.yaml` with the idea that we would use `rhsm-subscriptions.properties` for configuration that was business logic related and `application.yaml` for more Spring related stuff (like turning off the banner, etc).   Since they kind of have the same name now, do you just want to consolidate them?

kahowell:
> do you just want to consolidate them?

Actually yeah, that'd be nice. I'm inclined to consolidate to properties rather than YAML, but could be convinced. Thoughts? @mstead, thoughts?

lindseyburnett:
> > do you just want to consolidate them?
> 
> @awood, actually yeah, that'd be nice. I'm inclined to consolidate to properties rather than YAML, but could be convinced. Thoughts? @mstead, thoughts?

#TeamYml

<goes back to lurking>

awood:
>  I'm inclined to consolidate to properties rather than YAML, but could be convinced.

I favor properties files for a single reason: the Spring boot documentation [notes ](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-external-config-yaml-shortcomings) that YAML configuration can't be loaded with the `@PropertySource` annotation.  This PR eliminates that annotation in our production code, but we still use it via the `@TestPropertySource` annotation (which I believe, based on the javadoc, has the same restriction of not working with YAML).  I think it would be a little inconsistent to load test properties via a properties file but application properties via a YAML file.  If we can figure a way to load the test configuration via YAML, I would have no objection to going all YAML.

mstead:
> > do you just want to consolidate them?
> 
> @awood, actually yeah, that'd be nice. I'm inclined to consolidate to properties rather than YAML, but could be convinced. Thoughts? @mstead, thoughts?

+1 on the consolidation. One stop shop. I'd be Ok with either format, so long as we can stick to one of them. Based on @awood comment, we can't if we go the YAML route. So properties would be fine.

kahowell:
@awood @mstead, I found a pretty convenient way to do 100% YAML and split out into profile-specific files. Unfortunately, it required some changes that I already made in #211 to profiles, so I went ahead and did that in #211 directly (see last commit of #211)... So if y'all approve of that PR, I'll close this PR in favor of #211.

kahowell:
#211 superseded this PR

PR #211, opened 2020-11-03
Refactor profiles into Configuration classes

kahowell:
The strategy is to separate component definitions into Configuration
classes for each profile:

- `api`: `ApiConfiguration`
- `capacity-ingress`: `CapacityIngressConfiguration`
- `capture-snapshots`: `CaptureSnapshotsConfiguration`
- `liquibase-only`: `LiquibaseUpdateOnlyConfiguration`
- `worker`: `TallyWorkerConfiguration`

All of these configurations are now imported centrally into
ApplicationConfiguration and ApplicationConfiguration itself
contains only beans that are common across all profiles.

I also made a Configuration for dev-mode which was a useful way to limit
the scheduling to only when dev mode is active.

In order to limit the scope of the Configuration classes, I utilized
package scanning and composition as appropriate.

This involved some refactoring, specifically, I've eliminated the
following packages:
- `org.candlepin.subscriptions.controller` (moved controllers into
  packages by function)
- `org.candlepin.subscriptions.jackson` (moved into
  `org.candlepin.subscriptions.resteasy`)

I also refactored the packages around the in-memory queue
implementation.

The task producer/consumer configurations have been reworked so that a
component that needs to produce or consume messages should work only
with abstract configurations: `TaskProducerConfiguration` and
`TaskConsumerConfiguration`, respectively. In order to unify the
consumer interface, I had to rework the in-memory queue slightly.

For the `kafka-queue` profile, there are two configurations:

- `KafkaTaskProducerConfiguration`
- `KafkaTaskConsumerConfiguration`

I removed the need for the in-memory profile by having in-memory
configurations that are only active if kafka-queue components are
missing.

This PR also reworks the task framework slightly to allow multiple task
queues (for the merge of rhsm-conduit into this codebase).

Testing
-------
I recommend launching with various profiles and confirming expected
behavior:

- `SPRING_PROFILES_ACTIVE=api ./gradlew bootRun` should not create any
  task queue and the APIs should be accessible.
- `SPRING_PROFILES_ACTIVE=capacity-ingress ./gradlew bootRun`
  should only serve the capacity ingress API.
- `SPRING_PROFILES_ACTIVE=liquibase-only ./gradlew bootRun` should
  invoke liquibase and then exit.
- `SPRING_PROFILES_ACTIVE=capture-snapshots ./gradlew bootRun` should
  capture enqueue tasks to capture snapshots and then exit.
- `SPRING_PROFILES_ACTIVE=purge-snapshots ./gradlew bootRun` should
  cause retention logic to run and then exit.
- `SPRING_PROFILES_ACTIVE=worker ./gradlew bootRun` should consume
  tasks, but not serve the API.

Appending `kafka-queue` should cause the producers/consumers to use
kafka. This can be verified by use of the podman-compose image and
commands such as:

```
podman exec rhsm-subscriptions_kafka_1 kafka-console-consumer --bootstrap-server localhost:29092 --topic platform.rhsm-subscriptions.tasks --max-messages 1
```

NOTE: looks like the podman-compose setup uses 29092, but we defaulted
kafka to 9092... Will address in a subsequent PR. You'll need to use
KAFKA_BOOTSTRAP_PORT=29092 in the meanwhile.


kahowell:
Depends on #209 and #210. Please review those first.

kahowell:
1f44b98..cb70044: push up a few properties I missed in application.yaml

kahowell:
cb70044 to eaa627e: fix test that failed due to my renaming rhsm-subscriptions.tasks.task-group to rhsm-subscriptions.tasks.topic

kahowell:
eaa627e to c3d1c3a: fix the other test due to same as above :-/

src/main/java/org/candlepin/subscriptions/files/ProductIdToProductsMapSource.java
```diff
@@ -23,16 +23,13 @@
...
-import org.springframework.stereotype.Component;
-
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
 /**
  * Loads the product ID to list of Tally products mapping from a YAML file.
  */
-@Component
```
awood, line 35: This raises an interesting question.  I see you made a bean for this class in `ProductIdMappingConfiguration.java`.  I am assuming you did that to control the creation of the bean via an `@Import` whereas with `@Component` it would be picked up universally.

What if we used [`@Conditional`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Conditional.html) and its cousins instead?  Which conditional we would use would remain to be seen, but we can use any SpEL expression or property or bean.  Just as a thought, the new `Configuration` classes that define a profile's beans could create a marker bean that the `@ConditionalOnBean` annotations could reference.  This would allow us to keep using `@Component` annotations and would cut down on the proliferation of constructing beans by hand.

awood (changes requested):
Overall I like the direction here.  There's just a few things I think we should explore first before merging.

src/main/java/org/candlepin/subscriptions/BootApplication.java
```diff
@@ -21,15 +21,16 @@
 package org.candlepin.subscriptions;
 
 import org.springframework.boot.SpringApplication;
-import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.SpringBootConfiguration;
+import org.springframework.context.annotation.Import;
 
 /** Bootstrapper for Spring Boot. */
-@SpringBootApplication
```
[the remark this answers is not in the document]
kahowell: Actually it's still applied, in ApplicationConfiguration, I'll simply move it to this level to be more idiomatic (and hopefully less confusing).

src/main/java/org/candlepin/subscriptions/DevModeConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+    "org.candlepin.subscriptions.tally.job", // for the tally job
+    "org.candlepin.subscriptions.retention" // for the retention job
+})
+public class DevModeConfiguration {
+
+    private static final Logger log = LoggerFactory.getLogger(DevModeConfiguration.class);
+
+    @Autowired
+    void logDevMode() {
+        log.info("Dev-mode enabled.");
```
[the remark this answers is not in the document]
kahowell: It works. Will push another commit with this change shortly.

src/main/java/org/candlepin/subscriptions/files/ProductIdToProductsMapSource.java
kahowell: According to Javadocs that you linked, intended use of `@Conditional` is for `Configuration` classes and `@Bean` methods only, so unfortunately, that wouldn't be supported on the Component itself.

One option is to scrap the Configuration-based approach and use `@Profile` all over the place (for all components), but I personally think that makes it much harder to grok what beans are being included in a given profile (and it's difficult to enforce if we kept package scanning enabled globally for all packages).

The approach I favor is to selectively enable component scanning on specific packages. So I will push an update that uses scanning on the files package for the `ProductIdMappingConfiguration`. (This product ID map is a little odd in that it's shared between components in unlike all other file beans). Please let me know if this works for you.

awood: Yeah, that is fine by me.  I also realized after I wrote this that having a `@Conditional` and a marker bean is actually *less* succinct and direct than using `@Profile`.  I do like using `@Component` and I think that should drop in nicely with your selective package scanning approach.  Otherwise, it's not like it's really that much of a burden to just write a simple `@Bean` method.  I can just deal with it. :)

kahowell:
0c56a14 to 46efd8b: rebase against latest develop.

@awood FYI one change I made during rebase was to move OptInController to the org.candlepin.subscriptions.security package, as this made it easier to integrate with #204's changes.

awood (approved):

awood:
@mstead Want to take a look too since this is a large-ish change?

mstead (approved):
This looks good to me. LGTM!

PR #214, opened 2020-11-06
Add test for sorting by measurement type.

awood:


src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
```diff
@@ -317,6 +320,23 @@ void testCanSortByDisplayName() {
         assertTallyHostView(found.get(0), "inventory3");
     }
 
+    @Transactional
+    @Test
+    void testCanSortByMeasurementType() {
+        Host virtualHost1 = createHost("inventoryVirtual", "accountVirtual");
+        addBucketToHost(virtualHost1, RHEL, ServiceLevel.PREMIUM, Usage.PRODUCTION,
+            HardwareMeasurementType.VIRTUAL);
```
mstead, line 328: Perhaps create a few more hosts here with different measurement types to verify that multiple results are returned and sort order can be validated. Otherwise with a single result, it'll always be in order. :)

mstead (changes requested):

src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
awood: I followed the pattern of the rest of the tests here.  If you look at the other ones they create a variety of buckets but each one has a unique tuple of product, SLA, and usage: "COOL_PROD Premium Production", "RHEL Premium Production", "RHEL Self-Support Production", etc.

I can still update the test if you like.  Let me know what you think.

mstead: Looking at the other tests, IMO they are in the same shape as the new one. As they are today, if a single result is returned, the we have only tested that the query worked and found a result, not that we have sorted on a particular column. or am I missing something?

awood: That's the way I read it as well.  Maybe we should create a card to fix them up?

awood: @kahowell Can you shed a little light on why the sorting tests just grab back one result?  I'm asking based on the git-blame output.

kahowell: @awood fair enough; as @mstead noted these tests are just ensuring that the generated JPQL/SQL works against our data model (I saw some issues with mapping API to field names during development, and wrote these tests to make sure we don't regress). I suppose even 0 entries is sufficient for that. :rofl: I'd support a separate card for making sure the sorting itself is actually working in these tests.

awood: @mstead So I think this is ready to go.  I've created a card for improving the tests: ENT-3282.

mstead (approved):

src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
mstead: :+1:

mstead:
@awood Needs rebase, then I'll merge.

awood:
> @awood Needs rebase, then I'll merge.

Fixed.

PR #215, opened 2020-11-09
Reintroduce HYPERVISOR measurement type

kahowell:
I found that we didn't include 202010291320-rename-measurement-type.xml
in changelog.xml in #208, doh!

This caused attempts to tally or retrieve tallies to fail due to
unexpected values.

However, given that we plan to start differentiating HYPERVISOR and
VIRTUAL, it no longer makes sense to just remove the value.

So instead this PR re-adds the type, and simply maps either HYPERVISOR
or VIRTUAL to the `hypervisor_sockets` or `hypervisor_cores`, etc.
values in the API for tally.

This also ensures that both HYPERVISOR and VIRTUAL are tolerated for
inserts into hardware measurements by updating the trigger

To test, use the following SQL to insert a snapshot into the DB that has
the HYPERVISOR and VIRTUAL enum values (observe that you cannot insert
VIRTUAL, until liquibase runs):

```
insert into tally_snapshots(id, product_id, account_number, granularity, owner_id, instance_count, cores, snapshot_date, sockets, physical_cores, physical_sockets, physical_instance_count, hypervisor_cores, hypervisor_sockets, hypervisor_instance_count, unit_of_measure, sla) values ( '78add7e0-74df-4908-aef6-fdc23f5f7690', 'RHEL', 'account123', 'DAILY', 'org123', 10, 10, '2020-11-09T01:01:01', 10, 0, 0, 0, 3, 3, 3, '_ANY', '_ANY');
insert into hardware_measurements(snapshot_id, measurement_type, cores, instance_count, sockets) values ('78add7e0-74df-4908-aef6-fdc23f5f7690', 'HYPERVISOR', 7, 7, 7);
insert into hardware_measurements(snapshot_id, measurement_type, cores, instance_count, sockets) values ('78add7e0-74df-4908-aef6-fdc23f5f7690', 'VIRTUAL', 3, 3, 3);
```

Call that fails without this change (but succeeds with it):

```
curl -X GET "http://localhost:8080/api/rhsm-subscriptions/v1/tally/products/RHEL?granularity=DAILY&beginning=2020-11-08T17%3A32%3A28Z&ending=2020-11-10T17%3A32%3A28Z" -H  "accept: application/vnd.api+json" -H  "x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoib3JnMTIzIn19fQo"
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurementType.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurementType.java
@@ -29,6 +29,7 @@ import java.util.List;
  */
 public enum HardwareMeasurementType {
     PHYSICAL,
+    HYPERVISOR,
     VIRTUAL,
     TOTAL,
     AWS, // AWS measured by HBI data
--- a/src/main/java/org/candlepin/subscriptions/db/model/TallySnapshot.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/TallySnapshot.java
@@ -177,11 +177,27 @@ public class TallySnapshot implements Serializable {
             snapshot.setPhysicalInstanceCount(physical.getInstanceCount());
         }
 
-        HardwareMeasurement hypervisor = this.hardwareMeasurements.get(HardwareMeasurementType.VIRTUAL);
+        HardwareMeasurement virtual = this.hardwareMeasurements.get(HardwareMeasurementType.VIRTUAL);
+        HardwareMeasurement hypervisor = this.hardwareMeasurements.get(HardwareMeasurementType.HYPERVISOR);
+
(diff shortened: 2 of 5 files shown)


awood:
> I found that we didn't include 202010291320-rename-measurement-type.xml
> in changelog.xml in #208, doh!

I would like the clarify that "we" means "me" in this case.  I am ashamed. :facepalm:

awood (approved):

kahowell:
> I would like the clarify that "we" means "me" in this case.

I reviewed it, and missed it too. :sweat_smile:

PR #216, opened 2020-11-09
Do not log stack trace on warning for unknown role

kahowell:
To test, run org.candlepin.subscriptions.tally.CloudigradeAccountUsageCollectorTest in IDEA and observe the logging output.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/CloudigradeAccountUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/CloudigradeAccountUsageCollector.java
@@ -103,7 +103,8 @@ public class CloudigradeAccountUsageCollector {
                     calculation.addCloudigrade(HardwareMeasurementType.AWS_CLOUDIGRADE, count);
                 }
                 catch (IllegalArgumentException e) {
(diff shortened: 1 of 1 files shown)


awood (approved):

src/main/java/org/candlepin/subscriptions/tally/CloudigradeAccountUsageCollector.java
```diff
@@ -103,7 +103,7 @@ private void enrichUsageWithCloudigradeData(Map<String, AccountUsageCalculation>
                     calculation.addCloudigrade(HardwareMeasurementType.AWS_CLOUDIGRADE, count);
                 }
                 catch (IllegalArgumentException e) {
-                    log.warn("Skipping cloudigrade usage: {}", usageCount, e);
+                    log.warn("Skipping cloudigrade usage: {}, due to: {}", usageCount, e.toString());
```
mstead, line 106: I'm Ok with this, but IIRC there was concern about the log messages cluttering the log. I'd suggest perhaps logging the 'usageCount' string as debug.

mstead (changes requested):
If you don't agree with my suggestion, I'm good with this change.

mstead (approved):

PR #219, opened 2020-11-17
Add event JSON schema and DB object

kahowell:
I added a gradle plugin for json schema POJO generation.

I modified our postgresql connection settings to tolerate strings in the JSONB field (other solutions were overly complex). (They get implicitly cast by postgres with `stringtype=unspecified`).

I employed two strategies in the schema:
1. When a field was an enum, '' semantically means "unset this field", while `null` or omitted means the event does not modify the field.
2. For non-enums, Optional.empty() (`null` in JSON) means "unset this field", while omitting the field will result in `null` in Java, meaning do not modify the field.

Note I preferred to use Optional on the enums as well, but couldn't find a good way to do that. In both cases, the advice to producers would be to omit the field if they don't have the info.

To test, run with `DEV_MODE=true ./gradlew bootRun`. Then navigate to http://localhost:8080/actuator/hawtio/jmx/operations?nid=root-org.candlepin.subscriptions.jmx-EventJmxBean-eventJmxBean

Try using `saveEvent` to put in an event:

```json
{"subject_id":"foobar","timestamp":"2020-12-25T00:00:00Z","account_number":"account123","event_source":"foobar","subject_type":"OpenShift Cluster"}
```

Then use `getEvent` with the event ID to fetch the event.

Try `fetchEventsInTimeRange`:

- with `accountNumber`: account123, `begin`: '2020-12-25T00:00:00Z, `end`: 2020-12-25T01:00:00Z, note that the event is returned (begin inclusive)
- with `accountNumber`: account123, `begin`: '2020-12-24T00:00:00Z, `end`: 2020-12-25T00:00:00Z, note that the event is not returned (end exclusive)

### DIFF
--- a/schemas/event.yaml
+++ b/schemas/event.yaml
@@ -78,6 +78,10 @@ properties:
           description: UOM for the measurement. (e.g. Cores, Sockets)
           type: string
           required: true
+          enum:
+            - Cores
+            - Sockets
+            - Core-seconds
     required: false
   cloud_provider:
     description: Identifier for the cloud provider of the subject.
--- a/schemas/event.yaml
+++ b/schemas/event.yaml
@@ -19,12 +19,12 @@ properties:
     description: Account identifier for the relevant account.
     type: string
     required: true
-  subject_type:
-    description: Identifier for the type of subject. (E.g. RHEL System, OpenShift Cluster)
+  service_type:
+    description: Identifier for the type of service being measured. (E.g. RHEL System, OpenShift Cluster)
     type: string
     required: true
-  subject_id:
-    description: Primary identifier for the subject, according to the source.
+  instance_id:
+    description: Primary identifier for the instance of the service, according to the event producer.
     type: string
     required: true
   timestamp:
@@ -44,17 +44,17 @@ properties:
     existingJavaType: java.util.Optional<String>
     required: false
   inventory_id:
-    description: Host-based inventory ID for the subject.
+    description: Host-based inventory ID for the service, if applicable and known.
     type: string
     existingJavaType: java.util.Optional<String>
     required: false
   insights_id:
-    description: Red Hat Insights ID for the subject.
+    description: Red Hat Insights ID for the service, if applicable and known.
     type: string
     existingJavaType: java.util.Optional<String>
     required: false
   subscription_manager_id:
-    description: subscription-manager ID for the subject.
+    description: subscription-manager ID for the service, if applicable and known.
     type: string
     existingJavaType: java.util.Optional<String>
     required: false
@@ -65,7 +65,7 @@ properties:
       type: string
     required: false
   measurements:
-    description: Capture measurements for the subject.
+    description: Captures measurements for the service instance.
     type: array
     items:
       type: object
--- a/schemas/event.yaml
+++ b/schemas/event.yaml
@@ -1,0 +1,146 @@
+$schema: http://json-schema.org/draft-07/schema#
+title: Event
+properties:
+  event_id:
+    description: UUID for the event, will be generated upon ingestion if not specified.
+    type: string
+    format: uuid
+    required: false
+  event_source:
+    description: Identifier for the system that captured the event.
+    type: string
+    required: true
+  event_type:
+    description: Source-specific identifier for the type of event (e.g. Snapshot, Delete). This allows
+      the source to encode a special semantic for the event.
+    type: string
+    required: false
+  account_number:
+    description: Account identifier for the relevant account.
+    type: string
+    required: true
+  subject_type:
+    description: Identifier for the type of subject. (E.g. RHEL System, OpenShift Cluster)
+    type: string
+    required: true
+  subject_id:
+    description: Primary identifier for the subject, according to the source.
+    type: string
+    required: true
+  timestamp:
+    description: Time that the event happened or the snapshot was captured.
+    type: string
+    format: date-time
+    required: true
+  expiration:
+    description: End of the range for the event (exclusive).
+    type: string
+    existingJavaType: java.util.Optional<java.time.OffsetDateTime>
+    format: date-time
+    required: false
+  display_name:
+    description: User-set name for the subject.
+    type: string
+    existingJavaType: java.util.Optional<String>
+    required: false
+  inventory_id:
+    description: Host-based inventory ID for the subject.
+    type: string
+    existingJavaType: java.util.Optional<String>
+    required: false
+  insights_id:
+    description: Red Hat Insights ID for the subject.
+    type: string
+    existingJavaType: java.util.Optional<String>
+    required: false
+  subscription_manager_id:
+    description: subscription-manager ID for the subject.
+    type: string
+    existingJavaType: java.util.Optional<String>
+    required: false
+  correlation_ids:
+    description: List of IDs to be used in debugging. Correlation IDs should ideally be unique.
+    type: array
+    items:
+      type: string
+    required: false
+  measurements:
+    description: Capture measurements for the subject.
+    type: array
+    items:
+      type: object
+      properties:
+        value:
+          description: Measurement value.
+          type: number
+          required: true
+        uom:
+          description: UOM for the measurement. (e.g. Cores, Sockets)
+          type: string
+          required: true
+    required: false
+  cloud_provider:
+    description: Identifier for the cloud provider of the subject.
+    type: string
+    enum:
+      - ''  # UNSPECIFIED
+      - Alibaba
+      - AWS
+      - Azure
+      - Google
+    required: false
+  hardware_type:
+    description: Identifier for the type of hardware of the subject.
+    type: string
+    enum:
+      - ''  # UNSPECIFIED
+      - Physical
+      - Virtual
+      - Cloud
+    required: false
+  hypervisor_uuid:
+    description: Subscription-manager ID of the hypervisor, as reported in virt-who mappings.
+    type: string
+    existingJavaType: java.util.Optional<String>
+    required: false
+  product_ids:
+    description: List of engineering product IDs detected on the subject.
+    type: array
+    items:
+      type: string
+    required: false
+  role:
+    description: The syspurpose role for the system.
+    type: string
+    enum:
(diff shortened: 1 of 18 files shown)


kahowell:
940a5fe to d3b2410: use test.properties in new test, rebase against develop

kahowell:
ccd0d71: Make schema more generic for future integrations
    
By removing the enums from `subject_type` and changing cores/sockets to a generic `measurement` array, we should be able to handle other types more easily in the future without modifying the schema

kahowell:
f78996a to cd51e89: rebase and squash

kahowell:
00dd06c (and few prior pushes) - fix tests post-rebase.

src/main/java/org/candlepin/subscriptions/db/model/EventRecord.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Create a new EventRecord given Event JSON.
+     *
+     * Generates a new random UUID if the Event JSON does not have a UUID.
+     *
+     * @param json event in JSON format
+     * @return EventRecord, populated with a deserialized event record
+     * @throws JsonProcessingException if the jsonString cannot be processed
+     */
+    public static EventRecord fromJson(String json) throws JsonProcessingException {
```
awood, line 80: Why are these static methods?  If EventRecord had these as instance methods, Spring could create it as a Singleton, and the ObjectMapper could be injected instead of created in a `static` block (which is awkward and makes testing harder).

schemas/event.yaml
```diff
@@ -0,0 +1,146 @@
+$schema: http://json-schema.org/draft-07/schema#
```
awood, line 1: I'm a little confused (and I may just be out of the loop here), but why aren't we using an OpenAPI definition for this?

awood (changes requested):

schemas/event.yaml
kahowell: @awood the idea was to define a JSON schema for communication from external systems to Swatch via a Kafka topic. I'm not opposed to putting it in the API spec (though it feels like we should define an (at least stubbed) API endpoint for it if we do). Thoughts?

src/main/java/org/candlepin/subscriptions/db/model/EventRecord.java
kahowell: @awood for simplicity of the factory method, though I guess it might be more idiomatic Spring to define an EventRecordFactory bean... Thoughts? I am happy to pivot if you like that approach (or want to suggest another).

lindseyburnett: I agree with the sentiment that static blocks are awkward.  

Disclaimer: I haven't made it through the rest of the PR yet for context....but here's my initial thoughts:

Did you try any of the Jackson annotations for what you're trying to do? Upon seeing references to the object mapper, my first thought was to use Jackson in general...extending StdSerializer/StdSerializer if you had to.  I think @JsonCreator might work.  Also reading your comments block `@JsonIgnoreProperties` and `@JsonInclude` came to mind for dealing with nulls/missing/empty strings I think.  (admittedly I don't remember what exactly each does off the top of my head.)

https://www.baeldung.com/jackson-annotations

src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.UUID;
+import java.util.stream.Stream;
+
+import javax.persistence.EntityNotFoundException;
+import javax.transaction.Transactional;
+
+/**
+ * Encapsulates interaction with event store.
+ */
+@Controller
```
lindseyburnett, line 42: I think the `@Service` annotation might be more appropriate here.  Traditionally `@Controller` annotation is used to indicate a class is a web controller.

src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
```diff
@@ -0,0 +1,51 @@
...
+     * Fetch a stream of events for a given account for a given time range.
+     *
+     * The events returned include those at begin and up to (but not including) end.
+     *
+     * @param accountNumber account number
+     * @param begin start of the time range (inclusive)
+     * @param end end of the time range (exclusive)
+     * @return Stream of EventRecords
+     */
+    Stream<EventRecord> findByAccountNumberAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
```
lindseyburnett, line 49: Can this be `findByAccountNumberAndTimestampBetween` or does it need to be explicit because of the inclusive/exclusive?

src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param accountNumber account identifier
+     * @param begin beginning of the time range (inclusive)
+     * @param end end of the time range (exclusive)
+     * @return stream of Event
+     */
+    public Stream<Event> fetchEventsInTimeRange(String accountNumber, OffsetDateTime begin,
+        OffsetDateTime end) {
+        return repo
+            .findByAccountNumberAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
+            accountNumber, begin, end).map(EventRecord::getEvent);
```
lindseyburnett, line 62: Is this repo method used anywhere else/does it make sense to use projections to have the method return `Event` instead?

schemas/event.yaml
beav: fwiw you'll eventually be able to enforce schemas on kafka topic via kafka itself. I don't think there is any harm in enforcing it in 2 places though.

schemas/event.yaml
```diff
@@ -0,0 +1,146 @@
...
+    type: array
+    items:
+      type: object
+      properties:
+        value:
+          description: Measurement value.
+          type: number
+          required: true
+        uom:
+          description: UOM for the measurement. (e.g. Cores, Sockets)
```
beav, line 78: can this be an enum?

src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
kahowell: @lindseyburnett it does need to be this mess because of the semantics imposed here.

Given a couple of events that represents whole hours, I'd expect them to look something like this:

| timestamp | event |
|--|--|
|2021-01-01T00:00:00Z|{"expiration":"2021-01-01T01:00:00Z,..."}
|2021-01-01T01:00:00Z|{"expiration":"2021-01-01T02:00:00Z,..."}

pseudo-sql:
```sql
select * from events where timestamp where timestamp between '2021-01-01T00:00:00Z' and '2021-01-01T01:00:00Z'
```
yields both events since SQL between is inclusive, so asking for all events in the hour is actually a bit more awkward with between... we'd have to do something like

```sql
select * from events where timestamp where timestamp between '2021-01-01T00:00:00Z' and '2021-01-01T00:59:59.999Z'
```

I personally dislike `between` for this use case for this reason.

I will leave some more explanation in the JavaDoc for this.

src/main/java/org/candlepin/subscriptions/event/EventController.java
kahowell: Good idea. I briefly looked into projections, but initially gave up early on, but I will revisit and let you know if I have any luck.

schemas/event.yaml
kahowell: @beav It *could*, but then anytime we want to onboard a new UOM, we have to update the schema. Updating a schema is not particular difficult though. :smile:  @lindseyburnett @mstead @awood thoughts?

src/main/java/org/candlepin/subscriptions/event/EventController.java
kahowell: @lindseyburnett Good catch! Will change in a follow-up commit.

schemas/event.yaml
lindseyburnett: @kahowell
Uom is enumerated (cores/sockets) in our rhsm-subscriptions-api-spec.yaml, not sure if that affects an opinion one way or another.   Are there other UOMs that are currently out there that other applications are using but swatch doesn't support yet?  How often do we anticipate onboarding new UOMs?

In general, I'm leaning towards enum now instead of later.

schemas/event.yaml
lindseyburnett: 4ql3zv

schemas/event.yaml
mstead: I think we should use an enum now. The intent of this schema is to define the expected data format. Enum here would take the guesswork out.

kahowell:
9d5db91: address review feedback.

Note I moved the ObjectMapper definition into ApplicationConfiguration and it is now shared between the API, the EventJmxBean, the EventRecord JPA converter, and the identity header parser.

9d5db91 to 4b954e7: rebase against develop

schemas/event.yaml
kahowell: Done

src/main/java/org/candlepin/subscriptions/event/EventController.java
kahowell: I had no luck. The projection idea would work well for having a subset of fields (like a DB view), but not so much for transforming a column. We can revisit if we find a nice way to implement, but for now, otherwise, tried to encapsulate all bits into that EventController.

src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
kahowell: Done

src/main/java/org/candlepin/subscriptions/event/EventController.java
kahowell: Done

kahowell:
4b954e7 to ca116e2: add test profile to ObjectMapperContextResolverTest

awood (approved):

mstead (approved):
Looks good here.

PR #226, opened 2020-11-30
Lburnett/ent 3251 unify metadata from api output

lindseyburnett:
This PR takes care of MOST of the ask of [ENT-2607](https://issues.redhat.com/browse/ENT-2607), to unify the canonical values of Granularity, Service Level, and Usage metadata.  I haven't finished Product Id yet. [ENT-2512](https://issues.redhat.com/browse/ENT-2512) goes along with this card as well.  The ask of [ENT-2512](https://issues.redhat.com/browse/ENT-2512) is to enumerate the same items.

I'd like to get feedback of what I have so far, and I'll probably put out a non-draft PR that includes the product id enumeration as well.

The "highlights" of the changes are as follows:

- Updated the api-spec.yaml file to define the enums mentioned above as schemas, and reference them in the appropriate places through the spec file
- Updated Resteasy configurations to use custom serializer/deserializer for the newly defined enums.  The deserializer ensures that the incoming query params continue to support case insensitive.  The serializer is enforcing title case when the new enums are returned in a response.
  - Added `org.apache.commons:commons-text` dependency to the project for titlecase support to prevent reinventing the wheel.  Let me know if it's preferable to implement that logic manually instead.

Something I'm not too thrilled about and is an area for improvement  (taking suggestions!) is that I've got the serializer and deserializer configured in two different manners.  For the serializer, I was able to extend StdSerializer<T> and register it as a module in the ObjectMapperContextResolver class.  When I tried to extend StdDeserializer<T> the same way, the deserialize method didn't execute as expected.  Then I found the `EnumParamConverterProvider` class, and added logic there to support the case insensitivity.  I've got some TODO flags in there to consolidate some of the code, but I'm open to a completely different approach.  An idea I had after the fact was to try to extend JsonDeserializer, but not sure how that would work out since I don't want to force every enum in a project to use a case-insensitive deserializer.  Anyway, I don't have much experience with jax-rs so I'm very much open to other ideas you guys may have.

image

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -183,17 +183,17 @@ paths:
           in: query
           required: true
           schema:
-            type: string
+            $ref: '#/components/schemas/GranularityGenerated'
           description: "The level of granularity to return."
         - name: sla
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/ServiceLevelGenerated'
           description: "Include only snapshots matching the specified service level."
         - name: usage
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/UsageGenerated'
           description: "Include only snapshots matching the specified usage."
         - name: beginning
           in: query
@@ -261,7 +261,7 @@ paths:
                 meta:
                   count: 10
                   product: RHEL
-                  granularity: DAILY
+                  granularity: Daily
                   service_level: Premium
                   usage: Production
         '400':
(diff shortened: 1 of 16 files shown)


api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -654,6 +654,17 @@ components:
...
+      type: string
+      enum: [ DAILY, WEEKLY, MONTHLY, QUARTERLY, YEARLY ]
+    ServiceLevelGenerated:
+      type: string
+      enum: [ "", Premium, Standard, Self-Support, _ANY ]
+      x-enum-varnames: [ UNSPECIFIED, PREMIUM, STANDARD, SELF_SUPPORT, ANY ]
+    UsageGenerated:
+      type: string
+      enum: [ "", Production, Development/Test, Disaster Recovery, _ANY ]
+      x-enum-varnames: [ UNSPECIFIED, PRODUCTION, DEVELOPMENT_TEST, DISASTER_RECOVERY, ANY ]
```
lindseyburnett, line 667: Preferences/Suggestions on naming conventions for these?  I had thrown the word "Generated" at the end of them to prevent the namespace collision with the enums in the db.model package...but the side effect is that the docs include the word "Generated", as seen below.

image

The reason I care about the name collision thing is because it was kind of confusing keeping them straight, and fully qualifying a package lessens readability too because clutter.  On top of that, checkstyle gets mad & requires SuppressWarning more often because of line length.

PR #228, opened 2020-12-02
Update openapi specification to create schemas for enums to mirror th…

lindseyburnett:
TL;DR

* update the open-api spec yaml file to define Granularity, Usage, ServiceLevel, and ProductId as enums
* update the resource classes accordingly
* enhance the EnumParamConverterProvider class to convert query params in a casing-agnostic manner
* create an interface for the db model enums to implement to give us a unified means of converting generated <-> db model enums
* unit tests!
* format things a million times because checkstyle still hates my formatter

Room for Improvements:

* miscellaneous inline comments on this PR
* There's no db.model for ProductId, but there doesn't look like there's a need for it.  Should we add it anyway for consistency with the other api enums?  My vote is no.

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -161,7 +161,7 @@ paths:
         in: path
         required: true
         schema:
-          type: string
+          $ref: '#/components/schemas/ProductId'
         description: "The ID for the product we wish to query"
       - name: offset
         in: query
@@ -183,17 +183,17 @@ paths:
           in: query
           required: true
           schema:
-            type: string
+            $ref: '#/components/schemas/GranularityType'
           description: "The level of granularity to return."
         - name: sla
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/ServiceLevelType'
           description: "Include only snapshots matching the specified service level."
         - name: usage
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/UsageType'
           description: "Include only snapshots matching the specified usage."
         - name: beginning
           in: query
@@ -261,7 +261,7 @@ paths:
                 meta:
                   count: 10
                   product: RHEL
-                  granularity: DAILY
+                  granularity: Daily
                   service_level: Premium
                   usage: Production
         '400':
@@ -281,7 +281,7 @@ paths:
         in: path
         required: true
         schema:
-          type: string
+          $ref: '#/components/schemas/ProductId'
         description: "The ID for the product we wish to query"
       - name: offset
         in: query
@@ -302,17 +302,17 @@ paths:
           in: query
           required: true
           schema:
-            type: string
+            $ref: '#/components/schemas/GranularityType'
           description: "The level of granularity to return."
         - name: sla
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/ServiceLevelType'
           description: "Include only capacity for the specified service level."
         - name: usage
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/UsageType'
           description: "Include only capacity for the specified usage level."
         - name: beginning
           in: query
@@ -372,7 +372,7 @@ paths:
                 meta:
                   count: 10
                   product: RHEL
-                  granularity: DAILY
+                  granularity: Daily
                   service_level: Premium
         '400':
           $ref: '#/components/responses/BadRequest'
@@ -391,7 +391,7 @@ paths:
         in: path
         required: true
         schema:
-          type: string
+          $ref: '#/components/schemas/ProductId'
         description: "The ID for the product we wish to query"
       - name: offset
         in: query
@@ -412,12 +412,12 @@ paths:
         - name: sla
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/ServiceLevelType'
           description: "Include only hosts for the specified service level."
         - name: usage
           in: query
           schema:
-            type: string
+            $ref: '#/components/schemas/UsageType'
           description: "Include only hosts for the specified usage level."
         - name: uom
           in: query
@@ -654,6 +654,31 @@ components:
           schema:
             $ref: "#/components/schemas/Errors"
   schemas:
+    # Schemas ending in "Type" are intentionally named to prevent naming conflicts with db model classes
+    GranularityType:
+      type: string
+      enum: [ Daily, Weekly, Monthly, Quarterly, Yearly ]
+    ServiceLevelType:
+      type: string
+      enum: [ "", Premium, Standard, Self-Support, _ANY ]
+    UsageType:
+      type: string
+      enum: [ "", Production, Development/Test, Disaster Recovery, _ANY ]
+    ProductId:
+      type: string
+      enum:
+        - OpenShift Container Platform
+        - RHEL
+        - RHEL Compute Node
+        - RHEL Desktop
+        - RHEL for ARM
+        - RHEL for IBM Power
+        - RHEL for IBM z
+        - RHEL for x86
+        - RHEL Server
+        - RHEL Workstation
+        - Satellite 6
+        - Satellite 6 Capsule
     Uom:
       type: string
       enum:
@@ -777,20 +802,20 @@ components:
             count:
               type: integer
             product:
-              type: string
+              $ref: '#/components/schemas/ProductId'
             service_level:
+              $ref: '#/components/schemas/ServiceLevelType'
               description: "Describes the service level that the report was made against. Not set if it wasn't
                 specified as a query parameter."
-              type: string
             usage:
+              $ref: '#/components/schemas/UsageType'
               description: "Describes the usage that the report was made against. Not set if it wasn't
                             specified as a query parameter."
-              type: string
             granularity:
+              $ref: '#/components/schemas/GranularityType'
               description: "Describes the significance of each date in the TallySnapshot list. For example if the
-                granularity is set to 'weekly', the dates in the TallySnapshot list will represent the start of a
-                seven day period."
-              type: string
+                              granularity is set to 'weekly', the dates in the TallySnapshot list will represent the start of a
+                              seven day period."
           required:
             - count
             - product
@@ -919,20 +944,20 @@ components:
             count:
               type: integer
             product:
-              type: string
+              $ref: '#/components/schemas/ProductId'
             granularity:
+              $ref: '#/components/schemas/GranularityType'
               description: "Describes the significance of each date in the CapacitySnapshot list. For example if the
                 granularity is set to 'weekly', the dates in the CapacitySnapshot list will represent the start of a
                 seven day period."
-              type: string
             service_level:
+              $ref: '#/components/schemas/ServiceLevelType'
               description: "Describes the service level that the report was made against. Not set if it wasn't
                 specified as a query parameter."
-              type: string
             usage:
+              $ref: '#/components/schemas/UsageType'
               description: "Describes the usages that the report was made against. Not set if it wasn't
                specified as a query parameter."
-              type: string
           required:
             - count
             - product
@@ -1018,15 +1043,16 @@ components:
             count:
               type: integer
             product:
-              type: string
+              $ref: '#/components/schemas/ProductId'
             service_level:
+              $ref: '#/components/schemas/ServiceLevelType'
               description: "Describes the service level that the report was made against. Not set if it wasn't
                     specified as a query parameter."
               type: string
             usage:
+              $ref: '#/components/schemas/UsageType'
               description: "Describes the usages that the report was made against. Not set if it wasn't
                    specified as a query parameter."
-              type: string
             uom:
               $ref: '#/components/schemas/Uom'
           required:
--- a/src/main/java/org/candlepin/subscriptions/db/model/Granularity.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/Granularity.java
@@ -20,6 +20,10 @@
  */
 package org.candlepin.subscriptions.db.model;
 
+import org.candlepin.subscriptions.utilization.api.model.GranularityType;
+
+import java.util.Map;
+
 /**
  * Granularity of a given snapshot.
  *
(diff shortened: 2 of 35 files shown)


api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -654,6 +654,33 @@ components:
           schema:
             $ref: "#/components/schemas/Errors"
   schemas:
+    GranularityGenerated:
```
lindseyburnett, line 657: Opinions on name for this?  I tacked on "Generated" at the end to circumvent the competing namespace with the db.model.Granularity class.  Partly because of checkstyle and lines being too long when you fully qualify the package, as well as just general confusion having two things with the same name.  The side effect is that it shows up like this in the api-doc page now.  Thoughts? Opinions?

image

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -654,6 +654,33 @@ components:
...
+      type: string
+      enum: [ Daily, Weekly, Monthly, Quarterly, Yearly ]
+      x-enum-varnames: [ DAILY, WEEKLY, MONTHLY, QUARTERLY, YEARLY ]
+    ServiceLevelGenerated:
+      type: string
+      enum: [ "", Premium, Standard, Self-Support, _ANY ]
```
lindseyburnett, line 663: The empty string displays like this...not a fan, but I might being too picky?

image

src/main/java/org/candlepin/subscriptions/db/model/EnumLookupUtil.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+        else {
+            throw new IllegalArgumentException(String
+                .format("%s is not a valid value for %s", value, Granularity.class));
+        }
+    }
+
+    String getValue();
+
+    default <T extends Enum<T>> T asOpenApiEnum() {
```
lindseyburnett, line 53: //TODO check if i still even need this anymore

src/main/java/org/candlepin/subscriptions/db/model/Granularity.java
```diff
@@ -20,17 +20,55 @@
...
-public enum Granularity {
-    DAILY,
-    WEEKLY,
-    MONTHLY,
-    QUARTERLY,
-    YEARLY
+public enum Granularity implements EnumLookupUtil {
+    DAILY("Daily") {
+        @Override
+        public GranularityGenerated asOpenApiEnum() {
```
lindseyburnett, line 35: This is producing this warning...
image

not entiiiiiirely sure that I'm concerned about it yet?  Open to suggestions.

Note: the asOpenApiEnum returns `<T extends Enum<T>> T`

src/main/java/org/candlepin/subscriptions/resource/ResourceUtils.java
```diff
@@ -149,19 +153,21 @@ static Pageable getPageable(Integer offset, Integer limit, Sort sort) {
...
+        if (Objects.isNull(usage)) {
             return Usage.ANY;
         }
-        Usage sanitizedUsage = Usage.fromString(usage);
-        // If the usage parameter is not one that we support, then throw an exception.
-        // If we don't, the query would default to UNSPECIFIED, which would be confusing.
-        if (StringUtils.hasLength(usage) && sanitizedUsage == Usage.UNSPECIFIED) {
+        Usage sanitizedUsage = Usage.fromString(usage.toString());
+
+        if (Objects.equals(sanitizedUsage, Usage.UNSPECIFIED) && StringUtils.hasLength(usage.toString())) {
```
lindseyburnett, line 162: Couple questions here

1. Is there anything downstream that is relying on a BadRequestException being thrown to handle the flow of execution?
2. Can I delete this block (as well as the sanitizeServiceLevel equivalent further down)?

If there is logic reliant on this exception being thrown, this PR will need further work.  Since the Resource.java classes are expecting a query param to be an enum now, the requests get rejected upstream, and we wouldn't have a situation where an invalid/undefined value got this far.  Which is where my second question comes from...can I delete it then.

api/rhsm-subscriptions-api-spec.yaml
awood: I like `GranularityType`, `ServiceLevelType`, etc.  `Generated`.  I think putting a comment above the schema definition about wanting to avoid naming conflicts with the DB model classes is worth having also.  Otherwise, someone might be apt to shorten it in a refactor.

api/rhsm-subscriptions-api-spec.yaml
awood: I'm okay with it.  Might be worth filing an issue with the SwaggerUI team though.  Or fix it, submit a PR to them, ????, profit.

src/main/java/org/candlepin/subscriptions/resource/ResourceUtils.java
awood: 1. I don't believe anything relies on catching that exception.  Here I think we're relying on the exception bubbling up to RESTEasy which handles it.
2. I think you can delete this, yes, since the sanitization is handled via type conversion to an enum.

api/rhsm-subscriptions-api-spec.yaml
lindseyburnett: I like it!

src/main/java/org/candlepin/subscriptions/db/model/EnumLookupUtil.java
lindseyburnett: Update: need the definition, don't need to default it.

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -654,6 +654,34 @@ components:
...
+      enum: [ Daily, Weekly, Monthly, Quarterly, Yearly ]
+      x-enum-varnames: [ DAILY, WEEKLY, MONTHLY, QUARTERLY, YEARLY ]
+    ServiceLevelType:
+      type: string
+      enum: [ "", Premium, Standard, Self-Support, _ANY ]
+      x-enum-varnames: [ UNSPECIFIED, PREMIUM, STANDARD, SELF_SUPPORT, ANY ]
+    UsageType:
+      type: string
+      enum: [ "", Production, Development/Test, Disaster Recovery, _ANY ]
+      x-enum-varnames: [ UNSPECIFIED, PRODUCTION, DEVELOPMENT_TEST, DISASTER_RECOVERY, ANY ]
```
kahowell, line 669: I'd rather not maintain the same list twice; looking at the result without `x-enum-varnames`, we get var names like specified _except_ for `EMPTY` instead of `UNSPECIFIED` and `_ANY` instead of `ANY`. In spite of the inconsistency between `EMPTY` and `UNSPECIFIED`, I prefer to omit `x-enum-varnames` (in the other cases too). Perhaps we should just rename UNSPECIFIED to EMPTY in other places? Thoughts?

lindseyburnett: I'm 100% in agreement.  I actually had done that first, added x-enum-varnames because of the unspecified/empty situation, then reverted because I was trying to minimize how invasive my changes were going to be.

I didn't really like using x-enum-varnames to begin with since it's a OpenAPI vendor extension.  Personally I like to keep everything pretty vanilla as long as possible.

src/main/java/org/candlepin/subscriptions/resteasy/EnumParamConverterProvider.java
```diff
@@ -53,34 +75,85 @@
...
+
+            Optional<Map.Entry<String, T>> result = keyValuePairs
+                .entrySet()
+                .stream()
+                .filter(kv -> value.equals(kv.getKey()))
+                .findFirst();
+
+            if (result.isPresent()) {
+                return result.get().getValue();
             }
```
kahowell, line 112: Would be more performant (untested suggestion):
```suggestion
            T result = keyValuePairs.get(value);
            if (result != null) {
                return result;
            }
```

src/main/java/org/candlepin/subscriptions/resteasy/EnumParamConverterProvider.java
```diff
@@ -53,34 +75,85 @@
...
+    /**
+     * ParamConverterProvider to enable use of enums in query providers in a case-insensitive manner
+     *
+     * @param <T>
+     */
+    static class CaseInsensitiveEnumParamConverter<T> extends EnumParamConverter<T>
+        implements ParamConverter<T> {
+
+        public CaseInsensitiveEnumParamConverter(Class<T> className) {
+            super(className);
```
kahowell, line 132: Instead, I'd suggest build a map with lowercased keys and then modify `fromString` to use `map.get(value.toLowerCase())`

kahowell (changes requested):
Overall, looks good, have a few suggestions for performance-sake.

src/main/java/org/candlepin/subscriptions/db/model/EnumLookupUtil.java
```diff
@@ -0,0 +1,55 @@
...
+
+        if (match.isPresent()) {
+            return match.get();
+        }
+        else if (Objects.nonNull(defaultEnum)) {
+            return defaultEnum;
+        }
+        else {
+            throw new IllegalArgumentException(String
+                .format("%s is not a valid value for %s", value, Granularity.class));
```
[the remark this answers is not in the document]
lindseyburnett: woops, darn rebase

src/main/java/org/candlepin/subscriptions/resteasy/EnumParamConverterProvider.java
```diff
@@ -38,12 +48,24 @@
 @Provider
 public class EnumParamConverterProvider implements ParamConverterProvider {
 
+    public static final String INVALID_VALUE_EXCEPTION_MSG = "%s is not a valid value for %s";
+
+    protected static final Collection<Class<? extends Enum<? extends Enum<?>>>> SPECIAL_ENUMS = Arrays.asList(
```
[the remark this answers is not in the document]
lindseyburnett: The classes in this collection are the openapi generated enums, which don't implement EnumLookupUtil.  This hardcoded list has been bugging me though...I'll think on a better alternative.

lindseyburnett:
Ready for review round 2!

Couple items I wanted to call out:

1. Originally I thought that dropping `x-enum-vars` was only going to affect `UNSPECIFIED` -> `EMPTY`.  I made those changes then realized that "any" as a type changes to... `ANY` -> `_ANY`.  Looking for everyone's blessing that I renamed that as well.
2. RE: the feedback @kahowell  had about detecting special enums via instanceof instead of maintaining an explicit list
Those generated enums don't implement the new StringValueEnum interface, and I'm not sure that it's a good idea to go mucking with the open api generator templates and figuring out how to only apply case insensitivity to some generated enums and not all.  Would probably require forking & writing a custom generator?  Combined with how we don't want to have to maintain a case-insensitive solution long term, that didn't sound very appealing.  Anyway, totally up for more recommendation/suggestions, but wanted to just give my two cents before I jump down that rabbit hole.

mstead:
retest this please

src/main/java/org/candlepin/subscriptions/resteasy/EnumParamConverterProvider.java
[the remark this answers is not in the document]
kahowell: How about:  `rawType.getPackageName().startsWith("org.candlepin.subscriptions.utilization.api.model")` ? That would get everything in our API easily.

src/main/java/org/candlepin/subscriptions/db/model/StringValueEnum.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Interface to reduce duplication of common methods for some enums
+ */
+public interface StringValueEnum {
```
kahowell, line 32: A bit of refactoring around the generics in this class would make it a bit more readable IMO (see following suggestions).
```suggestion
public interface StringValueEnum<T> {
```

src/main/java/org/candlepin/subscriptions/db/model/StringValueEnum.java
```diff
@@ -0,0 +1,61 @@
...
+
+    static <T extends Enum<T> & StringValueEnum> T getValueOf(Class<T> className, Map<String, T> valueEnumMap,
```
kahowell, line 34:
```suggestion
    static <T> T getValueOf(Class<T> className, Map<String, T> valueEnumMap,
```

src/main/java/org/candlepin/subscriptions/db/model/StringValueEnum.java
```diff
@@ -0,0 +1,61 @@
...
+            return defaultEnum;
+        }
+        else {
+            throw new IllegalArgumentException(String
+                .format("%s is not a valid value for %s", value, className));
+        }
+
+    }
+
+    static <T extends Enum<T> & StringValueEnum> Map<String, T> initializeImmutableMap(Class<T> enumClass) {
```
kahowell, line 53:
```suggestion
    static <T extends StringValueEnum<T>> Map<String, T> initializeImmutableMap(Class<T> enumClass) {
```

src/main/java/org/candlepin/subscriptions/db/model/StringValueEnum.java
```diff
@@ -0,0 +1,61 @@
...
+        return Arrays.stream(enumClass.getEnumConstants())
+            .collect(Collectors.toMap(t -> t.getValue().toLowerCase(), value -> value));
+    }
+
+    String getValue();
+
+    <T extends Enum<T>> T asOpenApiEnum();
```
kahowell, line 60:
```suggestion
    T asOpenApiEnum();
```

kahowell (changes requested):
Your changes from my previous feedback look great! I had a few more suggestions after looking through this again.

src/main/java/org/candlepin/subscriptions/db/model/ServiceLevel.java
```diff
@@ -33,20 +34,42 @@
...
  * system to associate it with a specific SLA requirement.
  */
-public enum ServiceLevel {
-    UNSPECIFIED(""),
-    PREMIUM("Premium"),
-    STANDARD("Standard"),
-    SELF_SUPPORT("Self-Support"),
-    ANY("_ANY");
+public enum ServiceLevel implements StringValueEnum {
+    EMPTY("") {
```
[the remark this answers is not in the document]
kahowell: To elaborate, storing it as a field as with `value` would make it much more succinct, as then `asOpenApiEnum()` would simply be something like `return openApiEnum` defined a single time.

src/main/java/org/candlepin/subscriptions/db/model/StringValueEnum.java
awood: :+1: on making the generics a little simpler

src/main/java/org/candlepin/subscriptions/db/model/ServiceLevel.java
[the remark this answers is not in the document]
lindseyburnett: Good call!

src/main/java/org/candlepin/subscriptions/db/model/StringValueEnum.java
lindseyburnett: Thanks for this!  I was trying all kinds of ways to make this more readable; my generics syntax was a little rusty :smile:

src/main/java/org/candlepin/subscriptions/db/model/Granularity.java
lindseyburnett: After latest refactor, this isn't relevant anymore

lindseyburnett:
@kahowell ready for you again!  I opted not to mess with the static nested Usage.EnumConverter and ServiceLevel.EnumConverter for JPA conversion as part of this - because Generics and all that jazz & Granularity doesn't have one, so I'm thinking it'd have to be another interface or something and not just added to StringValueEnum.

lindseyburnett:
pushed squash

kahowell (approved):

PR #230, opened 2020-12-10
lburnett/filter hosts by display name [ENT-3195]

lindseyburnett:
Add support for `display_name_contains` as an optional query param for GET `/hosts/products/{product_id}`.

Expanded the jpql `@Query` in HostRepository.  A potential item for future improvement would be to update the related `@Entity` classes and their relationships, and refactor use spring data's `Specification` api to dynamically build the WHERE expressions and not have to explicitly concern ourselves with the optional query params coming into the endpoint being null/empty/converting to "_ANY", etc.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/model/Host.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/Host.java
@@ -39,6 +39,7 @@ import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.OneToMany;
 import javax.persistence.Table;
+import javax.validation.constraints.NotNull;
 
 /**
  * Represents a reported Host from inventory. This entity stores normalized facts for a
@@ -52,16 +53,19 @@ public class Host implements Serializable {
     @GeneratedValue(strategy = GenerationType.AUTO)
     private UUID id;
 
-    @Column(name = "inventory_id")
+    @NotNull
+    @Column(name = "inventory_id", nullable = false)
     private String inventoryId;
 
     @Column(name = "insights_id")
     private String insightsId;
 
-    @Column(name = "display_name")
+    @NotNull
+    @Column(name = "display_name", nullable = false)
     private String displayName;
 
-    @Column(name = "account_number")
+    @NotNull
+    @Column(name = "account_number", nullable = false)
     private String accountNumber;
 
     @Column(name = "org_id")
--- a/src/main/resources/liquibase/202012141133-enforce-display-name-not-null.xml
+++ b/src/main/resources/liquibase/202012141133-enforce-display-name-not-null.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202012141133" author="lburnett">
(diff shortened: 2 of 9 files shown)


src/main/java/org/candlepin/subscriptions/db/model/Host.java
```diff
@@ -304,6 +307,20 @@ public void setCloudProvider(String cloudProvider) {
         this.cloudProvider = cloudProvider;
     }
 
+    @PrePersist
+    void prePersist() {
+        if (Objects.isNull(displayName)) {
+            displayName = "";
```
lindseyburnett, line 313: Theoretically, displayName shouldn't ever be null in the wild.  If somehow it is before a Host entity gets saved, default it to an empty string instead of `null`.  This is mostly to keep from having an overly-complex @Query in `HostRepository` with null & empty string checks for a scenario that doesn't happen.

Eventually we should probably enforce some constraints on the hosts.display_name column to solidify this.

Anyone think it's necessary to also attach a database update for `update hosts set display_name = '' where display_name is null`?  I wouldn't actually expect it to update any records, but figured I'd ask to just cross my T's and dot my I's.

src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
```diff
@@ -539,6 +541,50 @@ void testShouldFilterCores() {
         assertEquals(1, physical.getCores());
     }
 
+    @Transactional
+    @ParameterizedTest
+    @CsvSource({
```
lindseyburnett, line 546: At some point I'd like to look into bumping our JUnit version.  With newer minor versions, there are more options available for parameterized tests - like more defined/obvious ways of passing around null sources.

There's also an option (and available with our current version I think) to use this `@CsvSource` with an external resource file instead.  Anyone have thoughts or preference on that?  Personally, I think it's easier to read with the data in-line like this to prevent the need to open and switch between another file.

src/main/java/org/candlepin/subscriptions/db/model/Host.java
kahowell: Instead of this, I was thinking we'd just do the equivalent in liquibase:

```xml
<addNotNullConstraint tableName="hosts" columnName="display_name" defaultNullValue="" />
```

Thoughts?

src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
kahowell: I dig in-line here. I'd say a rule of thumb would be appropriate when reviewing. I would go with more than 10 lines of CSV should be in a separate file.

kahowell (changes requested):
Needs a rebase, and I'd like to see nullability changed in liquibase (see inline comment). Looks good otherwise.

src/main/resources/liquibase/202012141133-enforce-display-name-not-null.xml
```diff
@@ -0,0 +1,19 @@
...
+    <property name="table.name" value="hosts"/>
+    <property name="column.name" value="display_name"/>
+
+    <changeSet id="202012141133" author="lburnett">
+        <comment>
+            Add not null constraint to ${table.name}.${column.name} and set existing null values to empty
+            string
+        </comment>
+        <addNotNullConstraint tableName="${table.name}" columnName="${column.name}" defaultNullValue=""/>
+    </changeSet>
```
kahowell, line 17: Can you please inline table.name and column.name?

kahowell (changes requested):
Overall looks good. Tiny request (see inline)

src/main/resources/liquibase/202012141133-enforce-display-name-not-null.xml
lindseyburnett: all set

kahowell (approved):

PR #231, opened 2020-12-14
Awood/hours

awood:


src/main/java/org/candlepin/subscriptions/resource/CapacityResource.java
```diff
@@ -156,10 +169,15 @@ public CapacityReport getCapacityReport(ProductId productId, @NotNull Granularit
...
         OffsetDateTime next = OffsetDateTime.from(start);
+
+        /* Walk through the timeline by the smallest unit of granularity.  Create capacity snapshots at each
+         * point on the timeline. */
         while (next.isBefore(end) || next.isEqual(end)) {
-            result.add(createCapacitySnapshot(clock.startOfDay(next), matches));
-            next = clock.startOfDay(next.plus(offset));
+            OffsetDateTime timelinePoint = clock.tick(next, productGranularity);
+            result.add(createCapacitySnapshot(timelinePoint, matches));
+            next = clock.startOfHour(next.plus(offset));
```
awood, line 178: This is really the critical section that I want people to think about.  I'm a little torn around wondering if `next` should be incremented based on a `switch` statement rather than just by every hour.

E.g.

```
switch (productGranularity):
    case HOURLY:  next = clock.startOfHour(next.plus(offset));
    case DAILY: next = clock.startOfDay(next.plus(offset));
...
```

I'm kind of leaning that way at the moment, but I go back and forth.

src/main/resources/product_profile_registry.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - 76   # Red Hat Enterprise Linux for Scientific Computing
+  - 279  # Red Hat Enterprise Linux for Power, little endian
+  - 290  # Red Hat OpenShift Container Platform
+  - 419  # Red Hat Enterprise Linux for ARM 64
+  - 420  # Red Hat Enterprise Linux for Power 9
+  - 479  # Red Hat Enterprise Linux for x86_64 (RHEL8)
+  - 250  # Red Hat Satellite
+  - 269  # Red Hat Satellite Capsule
+finestGranularity: DAILY
+burstable: false
```
kahowell, line 17: Can you please separate the Satellite bits into a separate YAML doc (in same file)? It wasn't immediately obvious to me that the parser is intended to handle multiple YAML docs in the same file until I read the test one.

I think a few files that could be consolidated into this one, so that there is a single place to configure products:
1. `product_id_to_products_map.yaml` (datatype change from `List<String>` to `Map<String,List<String>>` would do it methinks).
2. `role_to_products_map.yaml` (new field in the profile named `syspurposeRoles`?)
3. `arch_to_product_map.yaml` (new field in the profile named `architectures`?)

Assuming no objections, that could either be done here or in a follow-on PR. Thoughts?

kahowell (changes requested):

src/main/resources/product_profile_registry.yaml
awood: Excellent idea.  I'm going to try to incorporate it into this one.  I want to redo the whole `YamlFileSource` architecture to work as factory beans.  It's been kind of a nightmare testing this since I'm having to mock the `ProductProfileRegistrySource` and then that mock needs to return a mock `ProductProfileRegistry`.

awood: I did separate the product types into 2 YAML documents.  I think, however, I'm going to do the unification of product to product data mapping in a separate PR.  This one has already taken awhile and the unification is nice to have, but it's more complicated that I initially supposed and we have bigger fish to fry at the moment.

kahowell (changes requested):
Needs a value set in `src/main/resources/application-purge-snapshots.yaml` for hourly. In discussion we came up with 70 days worth (1680).

awood:
> Needs a value set in `src/main/resources/application-purge-snapshots.yaml` for hourly. In discussion we came up with 70 days worth (1680).

Done.

kahowell:
LGTM!

PR #235, opened 2021-01-06
Add Marketplace API client

kahowell:
The overall workflow is as follows:

1. A Marketplace offering is defined.
2. The charge ID(s) from the defined offering are provided for Subscription Watch configuration
   (comes from https://sandbox.marketplace.redhat.com/en-us/commerce/edition-charge-editor).
   Because this charge ID is consistent, we should eventually create a config value for it.
   (this field is out of the scope of this change).
3. A user subscribes via the Marketplace UI.
4. A notification is sent to RH backend systems.
5. (Details TBD) Subscription Watch retrieves one or more subscription IDs to be used for integration with the Marketplace API.
6. Subscription Watch emits usage events to the Marketplace API.

The Marketplace UI can be used to verify charges: https://sandbox.marketplace.redhat.com/en-us/account/pending-charges

Needs:
  - `MARKETPLACE_URL` - e.g. `https://sandbox.marketplace.redhat.com`
  - `MARKETPLACE_API_KEY` - from https://sandbox.marketplace.redhat.com/en-us/account/service-ids, reach out to me for a secret for testing or create your own
  - `subscriptionId` - from the subscription notification (email or webhook), reach out to me to, I will provide for testing
  - `eventId` - generated, I came up locally via `uuidgen | sed 's/\-//g' | dd if=/dev/stdin of=/dev/stdout bs=24 count=1 2>/dev/null`

Additionally, `MARKETPLACE_TOKEN_REFRESH_PERIOD` controls the tolerance for the age of the temporary token used to make API calls (it defaults to 1 minute).

NOTE: according to the Marketplace team, `eventId` should actually be a GUID, and there is work planned to fix it. Currently it is a 24-char ID...

Run as:
```
DEV_MODE=true MARKETPLACE_API_KEY=$MARKETPLACE_API_KEY MARKETPLACE_URL=https://sandbox.marketplace.redhat.com SPRING_PROFILES_ACTIVE=marketplace
```

Open up https://sandbox.marketplace.redhat.com/en-us/account/pending-charges and note the current value for pending charges.

Use http://localhost:8080/actuator/hawtio/jmx/operations?nid=root-org.candlepin.subscriptions.marketplace-MarketplaceJmxBean-marketplaceJmxBean with operation
`submitUsageEvent`, give payloadJson as follows, replacing all values in the JSON of FIXME as appropriate (refer to notes above about the source of values).

```json
{
  "start": 1608218158000,
  "end": 1608218158000,
  "resourceType": "SAAS",
  "subscriptionId": "FIXME",
  "eventId": "FIXME",
  "additionalAttributes": {},
  "measuredUsage": [
    {
      "chargeId": "redhat.com:openshiftdedicated:cpu_hour",
      "value": 154
    }
  ]
}
```

Take the resulting response and note the batch ID. Verify the batch is completed via the `getUsageEventStatus` JMX operation.

Verify the resulting charge difference in https://sandbox.marketplace.redhat.com/en-us/account/pending-charges

### DIFF
--- a/marketplace-client/marketplace-client-config.json
+++ b/marketplace-client/marketplace-client-config.json
@@ -1,0 +1,8 @@
+{
+  "modelPackage": "org.candlepin.subscriptions.marketplace.api.model",
+  "apiPackage": "org.candlepin.subscriptions.marketplace.api.resources",
+  "invokerPackage": "org.candlepin.subscriptions.marketplace",
+  "groupId": "org.candlepin",
+  "artifactId": "marketplace-client",
+  "java8": true
+}
--- a/marketplace-client/marketplace-client-config.json
+++ b/marketplace-client/marketplace-client-config.json
@@ -3,6 +3,5 @@
   "apiPackage": "org.candlepin.subscriptions.marketplace.api.resources",
   "invokerPackage": "org.candlepin.subscriptions.marketplace",
   "groupId": "org.candlepin",
-  "artifactId": "marketplace-client",
-  "java8": true
+  "artifactId": "marketplace-client"
 }
--- a/src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceApiFactory.java
+++ b/src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceApiFactory.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2019 Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.marketplace;
+
+import org.candlepin.subscriptions.http.HttpClient;
+import org.candlepin.subscriptions.marketplace.api.resources.MarketplaceApi;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.FactoryBean;
+import org.springframework.stereotype.Component;
+
+/**
+ * Factory that produces marketplace API clients.
+ */
+@Component
+public class MarketplaceApiFactory implements FactoryBean<MarketplaceApi> {
+
+    private static Logger log = LoggerFactory.getLogger(MarketplaceApiFactory.class);
+
+    private final MarketplaceProperties serviceProperties;
+
+    public MarketplaceApiFactory(MarketplaceProperties serviceProperties) {
+        this.serviceProperties = serviceProperties;
+    }
+
+    @Override
+    public MarketplaceApi getObject() throws Exception {
+        ApiClient apiClient = new ApiClient();
+        apiClient.setHttpClient(HttpClient.buildHttpClient(serviceProperties, apiClient.getJSON(),
+            apiClient.isDebugging()));
+        if (serviceProperties.getUrl() != null) {
+            log.info("Marketplace service URL: {}", serviceProperties.getUrl());
+            apiClient.setBasePath(serviceProperties.getUrl());
+        }
+        else {
+            log.warn("Marketplace service URL not set...");
+        }
+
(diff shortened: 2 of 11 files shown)


kahowell:
44d6b6c to 3b6876f: rebase

kahowell:
3b6876f to b9a5bb4: rebase - squashed checkstyle changes also

ntkathole:
@kahowell Looking into description, it seems that every account/user has to configure MARKETPLACE_API_KEY and charge ID so that swatch (cloud.redhat.com) can communicate with RHM. Are we thinking of asking user to add those details in sources from cloud.redhat.com or am I missing something ?

kahowell:
@ntkathole for clarity, during deployment, `MARKETPLACE_API_KEY` is configured once for all integrations against marketplace. The user will simply have to associate their Marketplace Account ID with their instances (e.g. OpenShift clusters), and then we'll take care of looking up the subscription ID based on the Marketplace Account ID on the backend.

kahowell:
b9a5bb4 to bcc1a8c: rebase against develop

marketplace-client/marketplace-client-config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "modelPackage": "org.candlepin.subscriptions.marketplace.api.model",
+  "apiPackage": "org.candlepin.subscriptions.marketplace.api.resources",
+  "invokerPackage": "org.candlepin.subscriptions.marketplace",
+  "groupId": "org.candlepin",
+  "artifactId": "marketplace-client",
+  "java8": true
```
lindseyburnett, line 7: Java 11 now!

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceApiFactory.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright (c) 2019 Red Hat, Inc.
```
lindseyburnett, line 2: 2021

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceJmxBean.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (c) 2020 Red Hat, Inc.
```
lindseyburnett, line 2: 2021!

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceApiFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import org.candlepin.subscriptions.http.HttpClient;
+import org.candlepin.subscriptions.marketplace.api.resources.MarketplaceApi;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.FactoryBean;
+import org.springframework.stereotype.Component;
+
+/**
+ * Factory that produces marketplace API clients.
```
lindseyburnett, line 32: Is this a pattern that's being used elsewhere in the application?

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceProperties.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2020 Red Hat, Inc.
```
lindseyburnett, line 2: 2021!  Can't wait until we're using that spotless plugin...I've got it configured to do all the headers as part as the format task :)

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceService.java
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (c) 2020 Red Hat, Inc.
```
lindseyburnett, line 2: 2021

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceWorkerConfiguration.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2020 Red Hat, Inc.
```
lindseyburnett, line 2: 2021

lindseyburnett (changes requested):

marketplace-client/marketplace-client-config.json
kahowell: According to https://openapi-generator.tech/docs/generators/java/

> Use Java8 classes instead of third party equivalents. Starting in 5.x, JDK8 is the default and the support for JDK7, JDK6 has been dropped

so I'll remove this now unneeded config option :+1:

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceApiFactory.java
kahowell: Yeah, many places, I tried to refactor as much common code out as possible. See `org.candlepin.subscriptions.http.HttpClient`. Unfortunately, the generated clients don't use base classes for ApiClient or for MarketplaceApi (or similar), which makes it difficult to DRY any further. If you have ideas, happy to try them though.

lindseyburnett (approved):

PR #239, opened 2021-01-12
Add subscription entity

jdhrh:
Add subscription entity

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/PoolIngressController.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/PoolIngressController.java
@@ -22,6 +22,8 @@ package org.candlepin.subscriptions.capacity;
 
 import org.candlepin.subscriptions.capacity.files.ProductWhitelist;
 import org.candlepin.subscriptions.db.SubscriptionCapacityRepository;
+import org.candlepin.subscriptions.db.SubscriptionRepository;
+import org.candlepin.subscriptions.db.model.Subscription;
 import org.candlepin.subscriptions.db.model.SubscriptionCapacity;
 import org.candlepin.subscriptions.db.model.SubscriptionCapacityKey;
 import org.candlepin.subscriptions.utilization.api.model.CandlepinPool;
@@ -52,7 +54,8 @@ public class PoolIngressController {
 
     private static final Logger log = LoggerFactory.getLogger(PoolIngressController.class);
 
-    private final SubscriptionCapacityRepository repository;
+    private final SubscriptionCapacityRepository subscriptionCapacityRepository;
+    private final SubscriptionRepository subscriptionRepository;
     private final CandlepinPoolCapacityMapper capacityMapper;
     private final ProductWhitelist productWhitelist;
     private final Counter poolsProcessed;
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/subscriptions/capacity/PoolIngressController.java
```diff
@@ -75,13 +79,53 @@ public PoolIngressController(SubscriptionCapacityRepository repository,
...
+                    updatableSubscription.setSku(poolSku);
+                }
+                else {
+                    skuToSubscription.remove(poolSku);
+                }
+                updatableSubscription.setSubscriptionId(pool.getSubscriptionId());
+                updatableSubscription.setStartDate(pool.getStartDate());
+                updatableSubscription.setEndDate(pool.getEndDate());
+                updatableSubscription.setOwnerId(orgId);
+                updatableSubscription.setQuantity(pool.getQuantity());
```
kahowell, line 111: FYI, sometimes this will be 1:1 with the subscription quantity, and sometimes not.

https://access.redhat.com/sites/default/files/attachments/subscriptionandentitlementaccounting-v1.1.pdf has some good examples...

For example:
- given `RH0103678 (Red Hat Enterprise Linux Server, Premium (1-2 sockets) (Up to 1 guest) with Smart Management)`, the pool quantity will be equal to the subscription quantity.
- given `RS00013​ (Red Hat Ceph Storage for Management Nodes, Premium (Up to 6 Physical Nodes)`, the pool quantity would be 6 when the subscription quantity is 1.

For simplicity, let's just leave subscription quantity untouched for the short-term.

src/main/resources/liquibase/202101121122-add-subscription-table.xml
```diff
@@ -0,0 +1,21 @@
...
+        <comment>Create the subscription table</comment>
+        <createTable tableName="subscription">
+            <column name="sku" type="VARCHAR(255)"/>
+            <column name="owner_id" type="VARCHAR(255)"/>
+            <column name="subscription_id" type="VARCHAR(255)"/>
+            <column name="quantity" type="BIGINT"/>
+            <column name="start_date" type="TIMESTAMP WITH TIME ZONE"/>
+            <column name="end_date" type="TIMESTAMP WITH TIME ZONE"/>
+        </createTable>
+        <addPrimaryKey tableName="subscription" columnNames="sku, owner_id"/>
```
kahowell, line 18: Actually primary key should be the subscription_id. (A customer can have more than one subscription for a given SKU - think renewals for example).

src/main/java/org/candlepin/subscriptions/db/model/Subscription.java
```diff
@@ -0,0 +1,182 @@
...
+    @Id
+    @Column(name = "sku")
+    private String sku;
+
+    @Id
+    @Column(name = "owner_id")
+    private String  ownerId;
+
+    @Column(name = "subscription_id")
+    private String subscriptionId;
```
kahowell, line 50: `subscriptionId` should be the Id, actually. See comment on liquibase changeset for rationale.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/capacity/PoolIngressController.java
jdhrh: sounds good!

src/main/java/org/candlepin/subscriptions/db/model/Subscription.java
jdhrh: updated

src/main/java/org/candlepin/subscriptions/capacity/PoolIngressController.java
jdhrh: Removed lines where the ingress and ingress test set the quantity.

kahowell (approved):
Looks good to me!

Can you please address the new check that sonarcloud added for unnecessary uses of Mockito `eq`? https://sonarcloud.io/project/issues?id=rhsm-subscriptions&pullRequest=239&resolved=false&types=CODE_SMELL

src/test/java/org/candlepin/subscriptions/db/SubscriptionRepositoryTest.java
```diff
@@ -0,0 +1,65 @@
...
+    SubscriptionRepository subject;
+
+    @Transactional
+    @Test
+    void canInsertAndRetrieveSubscriptions() {
+        final Subscription subscription = createSubscription("1", "testsku", "123");
+        subject.saveAndFlush(subscription);
+
+        final Subscription retrieved = subject.getOne("123");
+        //final Subscription retrieved = subject.findByOwnerIdAndSku("1", "testsku");
```
lindseyburnett, line 50: delete if unneeded

lindseyburnett (changes requested):

lindseyburnett (approved):

PR #240, opened 2021-01-12
ENT-2192: added a counter for total hosts.

dmckoske:
Added a counter to record the total number of hosts processed, and register it to "rhsm-subscriptions.capacity.records_total" in Micrometer.  Please see the reference card for live test instructions.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
@@ -38,6 +38,9 @@ import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.util.StringUtils;
 
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -59,14 +62,16 @@ public class InventoryAccountUsageCollector {
     private final InventoryDatabaseOperations inventory;
     private final HostRepository hostRepository;
     private final int culledOffsetDays;
+    private final Counter totalHosts;
 
     public InventoryAccountUsageCollector(FactNormalizer factNormalizer,
         InventoryDatabaseOperations inventory, HostRepository hostRepository,
-        ApplicationProperties props) {
+        ApplicationProperties props, MeterRegistry meterRegistry) {
         this.factNormalizer = factNormalizer;
         this.inventory = inventory;
         this.hostRepository = hostRepository;
         this.culledOffsetDays = props.getCullingOffsetDays();
+        this.totalHosts = meterRegistry.counter("rhsm-subscriptions.capacity.records_total");
     }
 
     @SuppressWarnings("squid:S3776")
@@ -162,6 +167,8 @@ public class InventoryAccountUsageCollector {
                 if (!facts.isHypervisor()) {
                     hostRepository.save(host);
                 }
+
+                totalHosts.increment();
             }
         );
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -196,4 +203,7 @@ else if (facts.isVirtual() && !StringUtils.isEmpty(facts.getHypervisorUuid())) {
         return calcsByAccount;
     }
 
+    protected Counter getTotalHosts() {
+        return totalHosts;
+    }
```
kahowell, line 208: Rather than expose the `Counter` as `protected`, consider this instead: `@Autowire` the MeterRegistry in the test, and then call `registry.counter("rhsm-subscriptions.capacity.records_total")` - this will return the same counter instance.

src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
```diff
@@ -503,6 +503,25 @@ void testGuestCountIsTrackedOnHost() {
...
+        expectedHypervisorMap.put(hypervisor.getSubscriptionManagerId(),
+            hypervisor.getSubscriptionManagerId());
+        mockReportedHypervisors(targetAccounts, expectedHypervisorMap);
+
+        when(inventoryRepo.getFacts(eq(targetAccounts), anyInt()))
+                .thenReturn(Arrays.asList(hypervisor).stream());
+
+        Map<String, AccountUsageCalculation> calcs = collector.collect(NON_RHEL_PRODUCTS, targetAccounts);
+        assertEquals(1, calcs.size());
+        assertEquals(1, collector.getTotalHosts().count());
```
kahowell, line 522: Rather than assert the count as `1`, it looks like you'll need to capture the counter value before `collect`, and then assert that the diff is `1`. Otherwise, other tests that run before this one invalidate the test. :-)

kahowell (changes requested):

src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
dmckoske: Thanks, that solves a problem I was having with this test.

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
dmckoske: Okay, good.  I will rewrite this.

src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
```diff
@@ -503,6 +507,27 @@ void testGuestCountIsTrackedOnHost() {
         assertEquals(2, savedHypervisor.getNumOfGuests().intValue());
     }
 
+    @Test
+    public void testTotalHosts() {
```
kahowell, line 511:
```suggestion
    void testTotalHosts() {
```

dmckoske: Will do.

src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
```diff
@@ -228,7 +232,7 @@ public void testTallyCoresAndSocketsOfRhelWhenInventoryFoundForAccount() throws
     }
 
     @Test
-    void testTallyForMultipleSlas() {
+    public void testTallyForMultipleSlas() {
```
kahowell, line 235:
```suggestion
    void testTallyForMultipleSlas() {
```

kahowell (approved):

PR #246, opened 2021-01-14
Pull OpenShift Metrics from Prometheus

mstead:
This PR adds the ability to query a Prometheus instance for openshift metrics and persist the metrics as Events to the rhsm-subscriptions database.

A metering package was created to contain all things metering. It is structured in a way that should allow adding other prometheus based metering fairly easily in the future.
      
The general flow for metrics updates:
- process initiated via JMX or job run.
- a per account task message is sent to a kafka topic with the details of the metrics update that is to be performed.
- when a message is received by the worker, the metrics are pulled from prometheus via the
  Prometheus client, translated to an Event and stored in the database.

Profiles
--------
I tried to make profiles and configuration for the metering components a little more clear and predictible.
I deviated a little from the existing pattern because of some of the issues that I ran into. I'm
OK with re-aligning if it is an issue.

Three new profiles were added:
- _openshift-metering-jmx_: Loads the JMX endpoints for pulling metrics
- _openshift-metering-job_: Launches the metrics gathering process and exits.
- _openshift-metering-worker_: Monitors the openshift metrics queue for tasks and performs the actual work on a per account basis.

**NOTE:** The JMX endpoints are activated along with the openshift-metering-worker profile but can be deployed on their own. This is useful if you'd like to queue up a bunch of messages to be processed once the worker comes online.
                             
                             
Prometheus HTTP Client
------------------
A prometheus client is being generated by an openapi schema that defines both the QueryAPI
as well as the RangeQueryAPI. At this point in time, the prometheus project does not have
an openapi schema for there API so we had to create our own.

Querying Prometheus Metrics
---------------------------
Metrics are being pulled from prometheus following a similar pattern as the Thanos Querier web UI, in that
given a time range, end date and a resolution/step, we get metrics back for a given query. The metering component,
allows tweaking the time range and resolution via configuration.

There are a few key settings and parameters that should be mentioned:
- **End Date:** The last time point to include a matching metric.
- **Time Range** (in minutes): Defines a period of time before the 'end date' that defines the range of the query. For example, a range of 60 minutes would yeild: START_DATE = END_DATE - 60 minutes
- **Resolution/step**: Defines how the metric data should be partitioned/combined. For example, if you query with a range of an hour, you could specify a resolution of 600 seconds (10 mins) and the results would be the max sample taken (if it exists) for each 10 minute block in the 1h time range.

The time range and resolution values are configurable to allow us to customize the results we get. By default, we use a 1 hour time range and a 1 hour resolution.

Testing This PR
---------------
**Reach out if you need the details for the prometheus (telemeter) instance as I didn't want to make the details public. I used one from DataHub**

1. Determine the URL and Auth token of the prometheus server you are connecting to.
2. Launch a local kafka instance.
3. Launch the openshift metering worker (JMX bean is loaded by default with the worker profile). Be sure to substitute
   your values for <YOUR_PROMETHEUS_INSTANCE_URL> and <YOUR_AUTH_TOKEN>.

```bash
$ DEV_MODE=true RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_RANGE_IN_MINUTES=180 PROM_AUTH_TOKEN=<YOUR_AUTH_TOKEN> PROM_URL=<YOUR_PROMETHEUS_INSTANCE_URL> ./gradlew clean bootRun --args="--spring.profiles.active=openshift-metering-worker,kafka-queue"

```
**NOTE**: I'm setting the RANGE_IN_MINUTES configuration value to 3h because the data on datahub's telemeter instance seems to be a few hours behind. This makes sure that any metric's gatered from NOW (any endpoint that does not require an end_date parameter) will yeild results.

4. Visit: http://localhost:8080/actuator/hawtio/jmx/ and navigate to the org.candlepin.subscriptions.metering.jmx where you should  find the OpenShiftJmxBean.
5. Test all of the operations to ensure that data can be retrieved. Event data should be persisted in the DB when run.
   * Turn debugging on to see more info
   * Use account 6274079 as this should yeild data.
   * Use an end date of 2021-01-10T09:12:28Z paired with a range of 3600
6. Test that the job runs (ensure that you have an opted in account)
```bash
DEV_MODE=true RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_RANGE_IN_MINUTES=180 PROM_AUTH_TOKEN=$(echo $OS_AUTH_TOKEN) PROM_URL="https://telemeter-lts.datahub.redhat.com/api/v1" ./gradlew clean bootRun --args="--spring.profiles.active=openshift-metering-job,kafka-queue"
```

**NOTE:** This can also be tested without Kafka, but it is better that we test with Kafka enabled since that is how we will be deploying it.


beav:
what is the kafka topic used for?

prometheus-client/prometheus-query-api-spec.yaml
```diff
@@ -0,0 +1,140 @@
+# This is a file maintained by the rhsm-subscriptions project and describes
+# a portion of the Prometheus Query API.
```
beav, line 2: Not as part of this PR, but after things are working can the query API team take over this file?

beav:
is this PR ok to be reviewed? it has an impressive amount of work done so far, but is still receiving updates.

prometheus-client/prometheus-query-api-spec.yaml
mstead: The problem right now is that this definition contains some hacks to work around a limitation of the OpenAPI Java code generator that we use. It doesn't properly generate code to support dynamic response objects required by the GET /query endpoint. Even if they took this over and defined it the 'correct' way, until that support is added to the generator, we wouldn't really be able to use it.

That being said, it might be a good starting point.

This was really just hacked together to get us going. I'm really not a fan of the end result. It is a little hacky, but works.

mstead:
> is this PR ok to be reviewed? it has an impressive amount of work done so far, but is still receiving updates.

Yes, this is ready for review, but there are still a couple issues that need to get addressed around the join in the PromQL and skipping the Eval SLA.

src/main/resources/application-openshift-metering-worker.yaml
```diff
@@ -0,0 +1,15 @@
...
+  # Use Spring Resource notation for this (e.g. "classpath:" or "file:")
+  account-list-resource-location: ${ACCOUNT_LIST_RESOURCE_LOCATION:}
+  account-batch-size: ${ACCOUNT_BATCH_SIZE:1}
+
+  metering:
+    prometheus:
+      metric:
+        openshift:
+          metricPromQL: >-
+            group(subscription_labels{ebs_account='%s', support =~ 'Premium|Standard|Self-Support|None'})by(_id,ebs_account,support)*on(_id)
```
ntkathole, line 11: @mstead Not sure if this will affect us, but when I use this query against time at which cluster's subscription settings was updated, then I see error as below

`Error executing query: found duplicate series for the match group {_id="6d7a300f-0e78-4f14-90ba-bd2ccfab6296"} on the left hand-side of the operation: [{_id="6d7a300f-0e78-4f14-90ba-bd2ccfab6296", ebs_account="6661312", support="Standard"}, {_id="6d7a300f-0e78-4f14-90ba-bd2ccfab6296", ebs_account="6661312", support="Premium"}];many-to-many matching not allowed: matching labels must be unique on one side`

Screenshot from 2021-02-08 10-19-20

mstead: > @mstead Not sure if this will affect us, but when I use this query against time at which cluster's subscription settings was updated, then I see error as below
> 
> `Error executing query: found duplicate series for the match group {_id="6d7a300f-0e78-4f14-90ba-bd2ccfab6296"} on the left hand-side of the operation: [{_id="6d7a300f-0e78-4f14-90ba-bd2ccfab6296", ebs_account="6661312", support="Standard"}, {_id="6d7a300f-0e78-4f14-90ba-bd2ccfab6296", ebs_account="6661312", support="Premium"}];many-to-many matching not allowed: matching labels must be unique on one side`
> 
> Screenshot from 2021-02-08 10-19-20

@ntkathole I'd love for you to show me how to reproduce this. Can you help?

mstead: Nevermind -- I have a reproducer. :)

api/rhsm-subscriptions-api-config.json
```diff
@@ -4,5 +4,5 @@
   "invokerPackage": "org.candlepin.subscriptions.utilization",
   "groupId": "org.candlepin",
   "artifactId": "rhsm-subscriptions",
-  "java8": true
+  "java8": "true"
```
kahowell, line 7: Option unneeded after rebase

cloudigrade-client/cloudigrade-client-config.json
```diff
@@ -4,5 +4,5 @@
   "invokerPackage": "org.candlepin.subscriptions.cloudigrade",
   "groupId": "org.candlepin",
   "artifactId": "cloudigrade-client",
-  "java8": true
+  "java8": "true"
```
kahowell, line 7: Option unneeded after rebase

container-compose.yml
```diff
@@ -1,14 +1,6 @@
...
 version: '3.1'
 services:
-  db:
-    image: docker.io/postgres
-    environment:
-      - POSTGRES_HOST_AUTH_METHOD=trust
-    volumes:
-      - ./init_dbs.sh:/docker-entrypoint-initdb.d/init_dbs.sh:z
-    ports:
-      - "5432:5432"
```
kahowell, line 11: Accidental commit?

prometheus-client/prometheus-client-config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "modelPackage": "org.candlepin.subscriptions.prometheus.model",
+  "apiPackage": "org.candlepin.subscriptions.prometheus.resources",
+  "invokerPackage": "org.candlepin.subscriptions.prometheus",
+  "groupId": "org.candlepin",
+  "artifactId": "prometheus-client",
+  "java8": "true"
+}
```
kahowell, line 8: Nitpick: newline missing

prometheus-client/prometheus-query-api-spec.yaml
```diff
@@ -0,0 +1,140 @@
...
+        error:
+          type: string
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: token
+
+security:
+  - bearerAuth: []
```
kahowell, line 140: Nitpick: missing newline

rbac-client/rbac-client-config.json
```diff
@@ -4,5 +4,5 @@
   "invokerPackage": "org.candlepin.subscriptions.rbac",
   "groupId": "org.candlepin",
   "artifactId": "rbac-client",
-  "java8": true
+  "java8": "true"
```
kahowell, line 7: Option unneeded after rebase

src/main/java/org/candlepin/subscriptions/metering/MeteringConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Top level config class for metering. Beans are defined/configured
+ * on a profile basis via the profile package. This class serves as
+ * an application entry point for the metering components.
+ */
+@Configuration
+@ComponentScan({"org.candlepin.subscriptions.metering.profile"})
```
kahowell, line 32: Why not just scan `org.candlepin.subscriptions.metering` here?

src/main/java/org/candlepin/subscriptions/metering/MeteringEventFactory.java
```diff
@@ -0,0 +1,99 @@
...
+        OffsetDateTime measuredTime, Double measuredValue) throws EventCreationException {
+        Event event = new Event()
+            .withEventSource(OPENSHIFT_CLUSTER_EVENT_SOURCE)
+            .withEventType(OPENSHIFT_CLUSTER_EVENT_TYPE)
+            .withServiceType(OPENSHIFT_CLUSTER_SERVICE_TYPE)
+            .withAccountNumber(accountNumber)
+            .withInstanceId(clusterId)
+            .withTimestamp(measuredTime)
+            .withDisplayName(Optional.of(clusterId))
+            .withUsage(Usage.PRODUCTION) // Inferred
```
kahowell, line 73: Let's not infer here please. I noticed the appropriate label already exists, so let's use it.

src/main/java/org/candlepin/subscriptions/metering/MeteringEventFactory.java
```diff
@@ -0,0 +1,99 @@
...
+            .withMeasurements(Arrays.asList(new Measurement().withUom(Uom.CORES).withValue(measuredValue)));
```
kahowell, line 74: Can use `List.of` instead of `Arrays.asList` now.

src/main/java/org/candlepin/subscriptions/metering/jmx/OpenshiftJmxBean.java
```diff
@@ -0,0 +1,174 @@
...
+    @ManagedOperationParameter(name = "accountNumber", description = "Red Hat Account Number")
+    @ManagedOperationParameter(
+        name = "endDate",
+        description = "The end date for metrics gathering (default: now). i.e 2018-03-20T09:12:28Z"
+    )
+    @ManagedOperationParameter(
+        name = "rangeInMinutes",
+        description = "Period of time (before the end date) to start metrics gathering. Must be >= 0."
+    )
+    public void performCustomOpenshiftMeteringForAccount(String accountNumber, String endDate,
```
kahowell, line 89: For sanity, let's please validate the dates passed are rounded to whole hours...

src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringControllerTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+import java.math.BigDecimal;
+import java.time.OffsetDateTime;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+
+
+@SpringBootTest
+@ActiveProfiles("openshift-metering-worker,test")
```
kahowell, line 57:
```suggestion
@ActiveProfiles({"openshift-metering-worker", "test"})
```

src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.util.StringUtils;
+
+import java.net.URLEncoder;
+import java.time.OffsetDateTime;
+
+
+@SpringBootTest
+@ActiveProfiles("openshift-metering-worker,test")
```
kahowell, line 44:
```suggestion
@ActiveProfiles({"openshift-metering-worker", "test"})
```

kahowell (changes requested):
Two nitpicks throughout:
- year in copyright should be 2021.
- When representing OpenShift as a String, we capitalize as "OpenShift" (don't care too much about var names)

Note: I intend to make another pass and functionally verify some things tomorrow, but wanted to go ahead and get this feedback to you.

src/main/java/org/candlepin/subscriptions/metering/profile/OpenShiftWorkerProfile.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Bean(name = "openshiftMetricRetryTemplate")
+    public RetryTemplate openshiftRetryTemplate(PrometheusMetricsPropeties metricProperties) {
+        SimpleRetryPolicy retryPolicy = new SimpleRetryPolicy();
+        retryPolicy.setMaxAttempts(metricProperties.getOpenshift().getMaxAttempts());
+
+        FixedBackOffPolicy backOffPolicy = new FixedBackOffPolicy();
+        backOffPolicy.setBackOffPeriod(2000L);
+
+        RetryTemplate retryTemplate = new RetryTemplate();
```
kahowell, line 60: Please replace with exponential backoff policy with max interval of 30 minutes.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/metering/profile/OpenShiftWorkerProfile.java
beav: can the max be configurable via env var? IMO 30 min is pretty long

prometheus-client/prometheus-client-config.json
lindseyburnett: don't need the "java8":"true" anymore, right?

api/rhsm-subscriptions-api-config.json
mstead: DONE - Available in next commit.

cloudigrade-client/cloudigrade-client-config.json
mstead: DONE - Available in next commit.

container-compose.yml
mstead: DONE - Reverted in next commit.

prometheus-client/prometheus-client-config.json
mstead: DONE - Available in next commit.

prometheus-client/prometheus-query-api-spec.yaml
mstead: DONE - Available in next commit.

rbac-client/rbac-client-config.json
mstead: DONE - Addressed in next commit.

src/main/java/org/candlepin/subscriptions/metering/MeteringConfiguration.java
mstead: @kahowell I did it this way because I wanted the profile itself to dictate what to include. Chasing around what beans get automatically loaded, and in what circumstances, is a PITA sometimes.

If you feel strongly about this, let me know and I'll make the change.

src/main/java/org/candlepin/subscriptions/metering/MeteringEventFactory.java
mstead: DONE - Available in next commit.

src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringControllerTest.java
mstead: DONE - Addressed in next commit.

src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusServiceTest.java
mstead: DONE - Addressed in next commit.

src/main/java/org/candlepin/subscriptions/metering/MeteringConfiguration.java
kahowell: I was just curious. Given we plan to adjust our pattern around beans in the future, not worried about it.

src/main/java/org/candlepin/subscriptions/metering/jmx/OpenshiftJmxBean.java
mstead: DONE - Available in next commit.

src/main/java/org/candlepin/subscriptions/metering/MeteringEventFactory.java
mstead: DONE - Available in next commit.

src/main/java/org/candlepin/subscriptions/metering/profile/OpenShiftWorkerProfile.java
mstead: DONE - Available in next commit.

mstead:
> * year in copyright should be 2021.

DONE - Next commit.

mstead:
> * When representing OpenShift as a String, we capitalize as "OpenShift" (don't care too much about var names)

I renamed a class and changed any with the wrong case in logging/comments that I could find.

mstead:
@kahowell I've update this PR with the new query that bartek has suggested. It seems to address the join issue, but I think that we need to further investigate the case where the sla has changed over the period that we are querying against. I wasn't able to find an example account to demonstrate this. I'm wondering if the result will be a 'rolled up' support label or not. I wasn't able to catch Bartek to ask him. We should still find this out.

I will rebase the two-query solution with this branch, so we have it as something to fall back on if issues are found with the query once QA gets their hands on it.

mstead:
For quick reference, here's the new query:
{{{
max(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) without(environment, region)*on(_id) group_right(prometheus) group(subscription_labels{ebs_account='%s'}) by (_id,ebs_account, support, usage)
}}}

A couple downsides are that we need to add any new labels we want or sla filters here. Could make parameters for them but I don't think it is worth it.

mstead:
This should now be good for final review!

beav:
> A couple downsides are that we need to add any new labels we want or sla filters here. Could make parameters for them but I don't think it is worth it.

Agreed, it seems like something that could be done as a future fix easily enough if needed.

ntkathole:
> For quick reference, here's the new query:
> 
> `max(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) without(environment, region)*on(_id) group_right(prometheus) group(subscription_labels{ebs_account='%s'}) by (_id,ebs_account, support, usage) `
> 
> A couple downsides are that we need to add any new labels we want or sla filters here. Could make parameters for them but I don't think it is worth it.

@mstead 

- Currently I see duplication issue is fixed with this query. This query returns now 2 records when subscription settings are updated (Did we handle this in our code ?).  
Screenshot from 2021-02-15 19-21-40

- Just want to confirm : Original query had "capacity_physical_cpu_cores:min:5m" while new query has max("capacity_physical_cpu_cores:max:5m"). Are both equivalent ?

mstead:
> > For quick reference, here's the new query:
> > `max(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) without(environment, region)*on(_id) group_right(prometheus) group(subscription_labels{ebs_account='%s'}) by (_id,ebs_account, support, usage) `
> > A couple downsides are that we need to add any new labels we want or sla filters here. Could make parameters for them but I don't think it is worth it.
> 
> @mstead
> 
> * Currently I see duplication issue is fixed with this query. This query returns now 2 records when subscription settings are updated (Did we handle this in our code ?).

Yes, this is handled in code. It will result in metric events getting persisted for the same cluster with different SLAs.  We should check the data points though to make sure that there is no overlap. This can be done by looking at the json response in that web UI.

For example, we should see that the the last metric we get for SLA Standard is just before the first metric for Premium.

I can double check this when I'm back from PTO tomorrow if you do not have time to look today.

>   Screenshot from 2021-02-15 19-21-40
> * Just want to confirm : Original query had "capacity_physical_cpu_cores:min:5m" while new query has max("capacity_physical_cpu_cores:max:5m"). Are both equivalent ?

Yes, this is correct.

mstead:
@ntkathole one thing that I noticed in your screen shot is that you are using the 'Console' tab. This will use prometheus' Query API instead of the Range API that we are using in code. If you instead use the 'Graph' tab, it will allow you to test that query over a period of time.

src/main/resources/application-openshift-metering-worker.yaml
```diff
@@ -0,0 +1,16 @@
...
+    prometheus:
+      metric:
+        openshift:
+          metricPromQL: >-
+            max(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) without(environment, region)*on(_id)
+            group_right(prometheus) group(subscription_labels{ebs_account='%s', support=~'Premium|Standard|Self-Support|None'})
+            by (_id,ebs_account, support, usage)
+      client:
+        token: ${PROM_AUTH_TOKEN:}
+        url: ${PROM_URL:https://localhost/api/v1}
```
kahowell, line 16: Can you please add env vars for backoff?

kahowell (changes requested):

src/main/resources/application-openshift-metering-worker.yaml
mstead: Yep, will do. Totally forgot about the env var.

beav: thanks! I could see folks wanting 1 minute max backoff, like if you have to sit around and watch if it works after making a fix. I could also see 30 min though to not spam the API :) the env var lets us change it as needed.

mstead:
> Yes, this is handled in code. It will result in metric events getting persisted for the same cluster with different SLAs. We should check the data points though to make sure that there is no overlap. This can be done by looking at the json response in that web UI.
> 
> For example, we should see that the the last metric we get for SLA Standard is just before the first metric for Premium.
> 
> I can double check this when I'm back from PTO tomorrow if you do not have time to look today.

@ntkathole I looked into my concern here, and the query seems to be returning exactly what I'd expect.

new_query_result


Here's a small snippet from the data we are storing once the results come back. I've targeted the points at which the sla and usage labels start to change. Looks right to me. 

```bash
 6661312  | 2021-02-16 00:35:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Production"       |  12.0

 6661312  | 2021-02-16 00:40:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Production"       |  12.0

 6661312  | 2021-02-16 00:45:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Production"       |  12.0

 6661312  | 2021-02-16 00:50:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Development/Test" | 12.0

 6661312  | 2021-02-16 00:55:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Development/Test" |  12.0

 6661312  | 2021-02-16 01:00:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Development/Test" | 12.0

 6661312  | 2021-02-16 01:05:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Development/Test" |  12.0

 6661312  | 2021-02-16 01:10:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Development/Test" | 12.0

 6661312  | 2021-02-16 01:15:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Development/Test" | 12.0

 6661312  | 2021-02-16 01:20:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Standard"    | "Development/Test" | 12.0

 6661312  | 2021-02-16 01:25:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Premium"     | "Production"       |  12.0

 6661312  | 2021-02-16 01:30:00-04 | d0dbc245-76a4-4f1f-917b-dcdba23ccdf3 | "Premium"     | "Production"       |  12.0

```

ntkathole:
> > Yes, this is handled in code. It will result in metric events getting persisted for the same cluster with different SLAs. We should check the data points though to make sure that there is no overlap. This can be done by looking at the json response in that web UI.
> > For example, we should see that the the last metric we get for SLA Standard is just before the first metric for Premium.
> > I can double check this when I'm back from PTO tomorrow if you do not have time to look today.
> 
> @ntkathole I looked into my concern here, and the query seems to be returning exactly what I'd expect.

Awesome. Thank you. :+1:

src/main/resources/application-openshift-metering-worker.yaml
mstead: DONE.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/task/PrometheusMetricsTaskManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Transactional
+    public void updateOpenshiftMetricsForAllAccounts(OffsetDateTime start, OffsetDateTime end) {
+        try (Stream<String> accountStream = accountListSource.syncableAccounts()) {
+            log.info("Queuing OpenShift metrics update for all configured accounts.");
+            accountStream.forEach(account -> updateOpenshiftMetricsForAccount(account, start, end));
+            log.info("Done queuing updates of OpenShift metrics");
+        }
+        catch (AccountListSourceException e) {
+            throw new TaskManagerException("Could not list accounts for openshift metrics update.", e);
```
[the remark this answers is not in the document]
mstead: DONE

mstead:
@kahowell I think that's it for this PR.

kahowell (approved):

beav:
:tada: thanks all

PR #248, opened 2021-01-20
Set ACCOUNT_BATCH_SIZE globally, filter larger messages

kahowell:
To test, try:

```
SPRING_PROFILES_ACTIVE=capture-snapshots,kafka-queue ACCOUNT_BATCH_SIZE=40 ./gradlew bootRun
```

followed by:

```
SPRING_PROFILES_ACTIVE=worker,kafka-queue ACCOUNT_BATCH_SIZE=1 ./gradlew bootRun
```

and observe the log messages about ignored messages, beginning with
"Skipping message"

This test also illustrates that the ACCOUNT_BATCH_SIZE is set properly
for both worker and capture-snapshots profiles.


mstead:
This looks good and works as expected and as discussed. 

Would be nice at some point to see if we can find a way to move the offset of our consumers and groups to the end without having to pull each message down and filter it. Nice workaround for now though!

PR #250, opened 2021-01-21
ENT-3395: google checkstyle poc 

lindseyburnett:
See https://github.com/google/google-java-format for intellij plugin.

`./gradlew spotlessCheck` will check the files and dump correctly formatted versions into build/spotless directory.
`./gradlew spotlessApply` applies the formatting changes.

Example of what a failed spotlessCheck looks like during the build pipeline: https://rhsm-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/blue/organizations/jenkins/subscriptions/detail/PR-250/11/pipeline


build.gradle
```diff
@@ -45,11 +45,28 @@ ext {
     avro_version = "1.8.2"
 }
 
+spotless {
+    java {
+        enforceCheck false
```
lindseyburnett, line 50: By default, `./gradlew build` will fail if the code style is violated.  Setting the flag to false suppresses that, so we can have a standalone build step in jenkins just for the code style.  This is what this branch is doing now, where I replaced the checkstyle stage in the Jenkinsfile with a spotless stage.

build.gradle
```diff
@@ -45,11 +45,28 @@ ext {
...
+        googleJavaFormat()
+        importOrder('org.candlepin','com','org','net','ch','io','liquibase','java','javax','') // or importOrderFile
```
lindseyburnett, line 52: The google format intellij plugin doesn't take care of import order, so you'd have to import their formatter .xml yourself to get the google style import order (which i'm not a huge fan of....just one giant list and i'm pretty sure it starts using asterisks in imports way sooner than i would personally choose to).  This line is totally optional though, and I lifted the order from what was previously defined in our checkstyle config.

PR #251, opened 2021-01-23
Unify product data into product profile registry.

awood:
All product related YAML files are now collected into one product
profile registry that is accessed via the ProductProfileRegistry class.
This class allows for access to individual ProductProfiles as well as
providing aggregate operations such as collecting syspurpose information
across all profiles.


awood:
I'd like some suggestions on the naming used in the profile registry.  I'm not thrilled with the names "productIds" and the attributes "id" and "product" but haven't really come up with anything better.  Likewise with "products" under "syspurposeRoles".  Also, let me know what you think about memoizing the ProductIdToProductsMap, RoleToProductMap, and ArchToProductMap.  Right now those values are only calculated in constructors or bean definitions so they aren't getting run over and over.  But memoizing the results might future-proof against running these calculations redundantly.  Of course, it probably only takes microseconds to do those calculations since registry elements don't really ever have more than a dozen elements.  Maybe I am prematurely optimizing.

src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
```diff
@@ -42,33 +41,33 @@
...
 public class InventoryAccountUsageCollectorTest {
 
     private static final String TEST_PRODUCT = "RHEL";
     public static final Integer TEST_PRODUCT_ID = 1;
     private static final String NON_RHEL = "OTHER PRODUCT";
     public static final Integer NON_RHEL_PRODUCT_ID = 2000;
 
-    public static final List<String> RHEL_PRODUCTS = Collections.singletonList(TEST_PRODUCT);
-    public static final List<String> NON_RHEL_PRODUCTS = Collections.singletonList(NON_RHEL);
+    public static final Set<String> RHEL_PRODUCTS = new HashSet<>(Arrays.asList(TEST_PRODUCT));
```
lindseyburnett, line 69: Sine we're on java 11 now, I think you can do
`Set.of(TEST_PRODUCT)`

lindseyburnett (approved):
Thinking out loud here rather than suggesting any type of change, but I'm wondering if there's a better option for maintaining this product profile registry.  Maybe in the database or something and we load into memory at startup?  Throwing that out there as an example, not that I'm necessarily suggesting that particular option.

The yaml file approach totally works for now, but I'm anticipating that bloating and being weird to try to maintain as we onboard more products to swatch.

Thoughts @awood , @kahowell , @mstead , @dmckoske ?

awood:
> The yaml file approach totally works for now, but I'm anticipating that bloating and being weird to try to maintain as we onboard more products to swatch.

I did think about this some while I was writing it.  I'm not sure what the tipping point is, but with where we are now, I felt like setting up the ORM is going to be overkill.  Additionally, we'd have to define all the product information in `INSERT` statements and then read it in.  To me, a very obvious line would be if product information were provided to us via service or external database.

src/main/resources/product_profile_registry.yaml
```diff
@@ -1,22 +1,88 @@
...
-  - 279  # Red Hat Enterprise Linux for Power, little endian
-  - 290  # Red Hat OpenShift Container Platform
-  - 419  # Red Hat Enterprise Linux for ARM 64
-  - 420  # Red Hat Enterprise Linux for Power 9
-  - 479  # Red Hat Enterprise Linux for x86_64 (RHEL8)
+  - id: 68 # Red Hat Enterprise Linux Desktop
+    products:
+      - RHEL
+      - RHEL for x86
+      - RHEL Desktop
```
kahowell, line 7: Originally, I was thinking:

```yaml
68: # Red Hat Enterprise Linux Desktop
  - RHEL
  - RHEL for x86
  - RHEL Desktop
```

but now that I'm seeing this, I dig how the fields are explicit (rather than having an Object with arbitrary key/value types).

:+1: :100:

awood: I went that way to begin with, but getting SnakeYaml to deserialize that would have required a lot of customization.  (I tried doing that customization too but wasn't successful)

kahowell (approved):

PR #252, opened 2021-01-25
Make InventoryAccountCollector remove only stale data

kahowell:
There were a couple optimizations to be had to make this performant:

 - Made the HostTallyBuckets work properly with the `equals/hashCode` methods
 - Gather all stale host records and remove them at the end

I also had to clear the buckets list to be recalculated each time (though you can see that Hibernate does not issue extraneous DML).

I made a couple of changes to the `insert-mock-hosts` script to make testing easier (see below).

Testing
-------
First, for each, clear out some data in rhsm-subscriptions:

```sql
truncate tally_snapshots cascade;
truncate hosts cascade;
truncate account_config;
```

Then, use the `insert-mock-hosts` script to set up one of the test cases below:

A few large accounts (10 accounts with 100,000 hosts each in HBI):

```
bin/insert-mock-hosts --hbi --clear
bin/insert-mock-hosts --hbi --num-physical 100000 --account acct_ --num-accounts 10 --clear
bin/insert-mock-hosts --opt-in --account acct_ --num-accounts 10
```

Lots of small accounts (10,000 accounts with 10 hosts each in HBI):

```
bin/insert-mock-hosts --hbi --clear
bin/insert-mock-hosts --hbi --num-physical 10 --account acct_ --num-accounts 10000 --clear
bin/insert-mock-hosts --opt-in --account acct_ --num-accounts 10000
```

To run the test, start the service via `./gradlew bootRun`. Then, once started up, make note of the time, use `TallyJmxBean.tallyConfiguredAccounts()` via http://localhost:8080/actuator/hawtio.

Make note of the final timestamp.

It's worth repeating the test, as subseqent tallies have no DML to do.

If desired you can simulate the last_seen being updated by running sql against the insights DB:

```sql
update hosts set modified_on='2102-01-01';
```

In my testing, I saw the timings get about twice as fast for the large accounts, and about twice as fast for the second; even better is the subsequent runs (with no new buckets) were about 3 times faster (or more).

You can also see the SQL if curious by setting the following properties in `config/application.properties` (please disable for timings):

```
logging.level.org.hibernate.SQL=DEBUG
logging.level.org.hibernate.type=TRACE
spring.jpa.properties.hibernate.format_sql=true
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/model/Host.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/Host.java
@@ -361,7 +361,9 @@ public class Host implements Serializable {
 
     public void addBucket(HostTallyBucket bucket) {
         bucket.getKey().setHost(this);
-        getBuckets().add(bucket);
+        if (!getBuckets().contains(bucket)) {
+            getBuckets().add(bucket);
+        }
     }
 
     public void removeBucket(HostTallyBucket bucket) {
--- a/src/main/java/org/candlepin/subscriptions/db/model/Host.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/Host.java
@@ -26,12 +26,12 @@ import org.candlepin.subscriptions.tally.facts.NormalizedFacts;
 
 import java.io.Serializable;
 import java.time.OffsetDateTime;
-import java.util.ArrayList;
 import java.util.EnumMap;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 
 import javax.persistence.CascadeType;
@@ -131,7 +131,7 @@ public class Host implements Serializable {
         orphanRemoval = true,
         fetch = FetchType.EAGER
     )
-    private List<HostTallyBucket> buckets = new ArrayList<>();
+    private Set<HostTallyBucket> buckets = new HashSet<>();
 
     @Column(name = "is_unmapped_guest")
     private boolean isUnmappedGuest;
@@ -338,15 +338,11 @@ public class Host implements Serializable {
         this.lastSeen = lastSeen;
     }
 
-    public List<HostTallyBucket> getBuckets() {
-        if (this.buckets == null) {
-            this.buckets = new ArrayList<>();
-        }
-
+    public Set<HostTallyBucket> getBuckets() {
         return buckets;
     }
 
-    public void setBuckets(List<HostTallyBucket> buckets) {
+    public void setBuckets(Set<HostTallyBucket> buckets) {
         this.buckets = buckets;
     }
 
@@ -361,9 +357,7 @@ public class Host implements Serializable {
 
     public void addBucket(HostTallyBucket bucket) {
         bucket.getKey().setHost(this);
-        if (!getBuckets().contains(bucket)) {
-            getBuckets().add(bucket);
-        }
+        getBuckets().add(bucket);
     }
 
     public void removeBucket(HostTallyBucket bucket) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -196,4 +201,27 @@ else if (facts.isVirtual() && !StringUtils.isEmpty(facts.getHypervisorUuid())) {
...
+        return accounts.stream()
+            .map(hostRepository::findByAccountNumber)
+            .flatMap(List::stream)
+            .collect(Collectors.toList());
+    }
+
+    private Host getOrCreateHost(Map<String, Host> inventoryHostMap, InventoryHostFacts hostFacts,
+        NormalizedFacts facts) {
+
+        Host existingHost = inventoryHostMap.remove(hostFacts.getInventoryId().toString());
```
awood, line 214: :worried: Not sure how I feel about the side-effects of this method.  Is there a way we could break this up into two steps?  Something like 

```
getOrCreateHost(update(inventoryHostMap, hostFacts), hostFacts, facts);
```

Where `update` just does the removal?  It's not as pretty to look at, but it makes it more obvious that the map is being mutated.

awood (changes requested):

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
kahowell: @awood @lindseyburnett I've refactored a bit. Please take a look when you get a chance.

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -115,7 +115,15 @@ public InventoryAccountUsageCollector(FactNormalizer factNormalizer,
...
 
-                Host host = new Host(hostFacts, facts);
+                Host existingHost = inventoryHostMap.remove(hostFacts.getInventoryId().toString());
+                Host host = existingHost == null ? new Host(hostFacts, facts) : existingHost;
+                if (existingHost != null) {
+                    host.getBuckets().clear(); // ensure we recalculate to remove any stale buckets
+                }
+                else {
+                    host.populateFieldsFromHbi(hostFacts, facts);
+                }
```
awood, line 125: The old code was
```java
    Host existingHost = inventoryHostMap.remove(hostFacts.getInventoryId().toString());
    Host host;
    if (existingHost == null) {
        host = new Host(hostFacts, facts);
    }
    else {
        host = existingHost;
        host.getBuckets().clear(); // ensure we recalculate to remove any stale buckets
        host.populateFieldsFromHbi(hostFacts, facts);
    }
    return host;    
```

I'm not sure that this new code is equivalent.  Specifically, in the old code if `existingHost` was non-null, `populateFieldsFromHbi` would run.  In the new code, it's only run if `existingHost` is null.

Maybe I am misreading because it is close to lunch and I am hungry.  I wouldn't be opposed to just dropping in the old code inline here (but I would add a comment above the assignment to `existingHost` highlighting that the map is being mutated).

kahowell: @awood thanks, my refactor was just bad. I did a fixup (see 2f18d7f), and rebased while I was at it, due to test failures.

src/main/java/org/candlepin/subscriptions/db/model/Host.java
```diff
@@ -277,7 +286,9 @@ public HostTallyBucket addBucket(String productId, ServiceLevel sla, Usage usage
 
     public void addBucket(HostTallyBucket bucket) {
         bucket.getKey().setHost(this);
-        getBuckets().add(bucket);
+        if (!getBuckets().contains(bucket)) {
```
lindseyburnett, line 289: Why not make this a Set<HostTallyBucket> instead?

lindseyburnett (approved):

kahowell:
05cc3eb to 6b91db8: rebase against develop

src/main/java/org/candlepin/subscriptions/db/model/Host.java
kahowell: I dig it! Will try that. Commit incoming.

awood (approved):

PR #254, opened 2021-01-27
Create dependabot.yml

awood:
Part of ENT-3387.  I'm submitting this PR for group discussion around using Dependabot.

Dependabot automatically submits PR requests to update libraries.  Personally, I would prefer a less automated system that provides a dashboard of dependencies with a list of which ones are out of date and which ones have open CVEs.  I do see some utility in using Dependabot even if it's a little different from my ideal.

See the [documentation](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates) for all the available configuration options.

Sidenote: Github [does not currently support Gradle](https://github.com/isaacs/github/issues/1620) in its Dependency Graph section (underneath the Insights tab).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+    target-branch: "develop"


kahowell:
Here's my thoughts: we already have other (RH internal) security processes for identifying libraries with vulnerabilities, so I'd rather avoid the whole mess that comes with maintaining a generated pom.xml if possible. 

Going forward though, if there are other compelling features, I'm not _strongly_ opposed to exploring pom.xml generation, so I'm happy to revisit if we find other reasons to generate pom.xml and check-in.

My proposal is we merge this PR as-is, and revisit again when we find some other compelling reason to look at the dep graph.

I would like to hear what @mstead @lindseyburnett @dmckoske thoughts are too :smile:

dmckoske:
If I hadn't seen this kind of thing done before, I would find it absurd or overkill.  But I have, and it does work despite the required infrastructure: we would need a small application and database space to pull this off.  The reason for the database is to record and track the presence or absence of libraries and their versions.  This way, you can generate a pom.xml (or whatever your target is) of any version or configuration and audit the library version history as needed.  

If this seems like overkill (and it may), then one alternative is multiple versions of the pom.xml, reflecting the upgrade of different library modules in succession.  Library versions which compel upgrades to their dependencies would record those upgrades at the same time.  

I'm game for either of these or something entirely else.  To me, the main question is how much time we can afford to spend on it.

lindseyburnett:
It might be because I'm not really sure where to find the listing of security vulnerabilities that we get from our internal processes...but I think I like the idea of alerts and/or PR's to kind of see immediately that there's an update that we can perform.

That said, I'm always hesitant when anything starts to try to automatically apply updates or generate things.  I think my ideal is the same as @awood described as his ideal - an alert that then allows you to action on it.

Since we don't think that exists, I don't hate the idea of just the PR getting created, because you either click merge to action on it, or we reject the PR (although, I'd like to know if dependabot would just keep recreating a PR for the same library).

As far as the pom generation goes...If we start generating pom.xml files, I'd probably start advocating that we move away from gradle and towards maven as a build tool.

Admittedly, I'm not sure I entirely follow @dmckoske's recommendation about archiving pom.xml files in a database, but at face value making an application to maintain historical dependency catalogs does feel like significant overkill since we've already got processes in place for identifying vulnerabilities.

awood:
> although, I'd like to know if dependabot would just keep recreating a PR for the same library

```
OK, I won't notify you again about this release, but will get in touch when a new version is available.

If you change your mind, just re-open this PR and I'll resolve any conflicts on it.
```

Is dependabot's response when you close a PR.

As far as archiving `pom.xml` files.  Could we not just check them in to version control (generate them automatically when we tag a release) and then track the history that way?

barnabycourt:
In regards to checking in pom.xml files. If we were to go that far, why would we not just convert from gradle to maven as a build tool? Having two sets of build artifacts checked in tends to confuse IDEs and developers new to a project.

awood:
> In regards to checking in pom.xml files. If we were to go that far, why would we not just convert from gradle to maven as a build tool? Having two sets of build artifacts checked in tends to confuse IDEs and developers new to a project.

Well, we could just drop the `pom.xml` in a lower directory so it wouldn't show up at the top and point Dependabot to it.  I will say that I am weakly opposed to generating `pom.xml`s.  We have internal processes that apprise us of dependency changes and I am content to rely on them.

Overall I guess my feeling would be, we merge this and then by policy, each Dependabot PR gets a card for review.  Dependabot isn't a magician and it can make some bad suggestions so we can't just go

kahowell (approved):
I created https://issues.redhat.com/browse/ENT-3523 to capture the idea we may revisit the dependency graph issue in the future. I'm not seeing any objection to trying the configuration in the PR, so merging it.

PR #257, opened 2021-01-29
ENT-3424: Up max http header size to 48000

kahowell:
Also add a template param for this, so we can adjust without a code update.


kahowell:
Huge thanks to @lindseyburnett for helping me track this down yesterday!

PR #258, opened 2021-02-03
initial commit for subscription sync controller

jdhrh:


jdhrh:
Currently `@MockBean` fails in the tests with the `SubscriptionRepository` initializing to null. I'm unsure what is going on there since it doesn't appear to be an injection error but rather `@MockBean` not working here despite working in other tests such as `PoolIngressControllerTest`

src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
```diff
@@ -23,20 +23,30 @@
...
      * Object a set of subscriptions
      * @param ownerId the ownerId of the subscriptions
      * @param subscriptionIds the list of subscriptionIds to filter on
      * @return a list of subscriptions with the specified ownerId and a subscriptionId from the provided list
      */
-    List<Subscription> findByOwnerIdAndSubscriptionIdIn(String ownerId,
-        List<String> subscriptionIds);
+    @Query("SELECT s FROM Subscription s where s.endDate > CURRENT_TIMESTAMP " +
+        "AND s.ownerId = :ownerId AND s.subscriptionId IN :subscriptionIds")
+    List<Subscription> findActiveByOwnerIdAndSubscriptionIdIn(@Param("ownerId") String ownerId,
```
lindseyburnett, line 46: Can a query method be used here instead of the hardcoded query?
Something like:
`findByEndDateAfterAndOwnerIdAndSubscriptionIdIn(Calendar.getInstance().getDate(), ownerId, subscriptionId)`

lindseyburnett: Unsure of how this will work with the OffsetDateTime type

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -0,0 +1,110 @@
...
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Optional;
+
+/**
+ * Update subscriptions from subscription service responses.
+ */
+@Component
+public class SubscriptionSyncController {
```
lindseyburnett, line 37: Thoughts on renaming this to `SubscriptionSyncService`?  This would be considered the "service layer" in the eyes of spring boot.  @kahowell

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -0,0 +1,110 @@
...
+                newSub.setStartDate(convertLongTime(subscription.getEffectiveStartDate()));
+            }
+            if (subscription.getEffectiveEndDate() != null) {
+                newSub.setEndDate(convertLongTime(subscription.getEffectiveEndDate()));
+            }
+            subscriptionRepository.save(newSub);
+        }
+    }
+
+    private static boolean updateSubscription(Subscription dto,
```
lindseyburnett, line 87: I'd like to see this get renamed/broken apart or something.  It's unclear what the method is meant to do, as it's name indicates it's updating a subscription - but it's returning a boolean.  Line 90 is returning true without updating the subscription entity at all.  The next few lines update the entity, but then return a false.  

It looks like it's intended to be a "does the subscription need to be updated" kind of flag...while having a side effect of updating the entity class anyway prior to it being saved on line 68.

Do these two private methods need to be static?

src/main/resources/liquibase/changelog.xml
```diff
@@ -37,5 +37,6 @@
     <include file="liquibase/202011130945-add-events-table.xml" />
     <include file="liquibase/202101081600-tweak-offering.xml" />
     <include file="liquibase/202101121122-add-subscription-table.xml" />
+    <include file="liquibase/202102031030-tweak-subscription.xml" />
```
lindseyburnett, line 40: Can we give this a more descriptive name? `update-subscription-table-pkey.xml`

lindseyburnett (changes requested):

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
jdhrh: No issue with splitting them up though I'd prefer to keep them static as they are not bound to the instance state and java has significant performance differences between static and member calls.

src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
jdhrh: OffsetDateTime's super-precision seems to be trimmed by the DB as well which makes precise comparison tricky too. That's why I took the JPQL current_timestamp there, but this does create the query annotation need which I definitely get wanting to avoid. @awood @kahowell preferences or thoughts on a clean way to handle this? The unit tests currently show this issue well as once something is persisted to the db, fractions of a millisecond are dropped as java's offsetdatetime is more precise than the db stores.

src/main/resources/liquibase/changelog.xml
jdhrh: Done. I'll push this with my next commit to the branch.

src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
lindseyburnett: Wonder if we can do something cheeky with extending OffsetDateTime or something and use it as a db converter, applying a different DateTimeFormatter.ofPattern without the milliseconds.

Or maybe there's a @Temporal annotation we can leverage.

If there isn't an obvious straightforward solution for this, my vote is to keep it the way you have it and we can mull over it later.  Probably not worth spinning wheels on it right now.

lindseyburnett (approved):

src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
kahowell: Personally, I'm fine with what's written, FWIW, I think the precision loss wouldn't cause significant change in behvaior.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
kahowell: I'm not opposed generally, but I think we should do this kind of renaming across the code-base outside the scope of this PR if so.

src/main/resources/liquibase/202102031030-tweak-subscription.xml
```diff
@@ -0,0 +1,12 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="202101081600-1" author="jharriso">
+        <comment>Change primary key</comment>
+        <dropPrimaryKey tableName="subscription" />
+        <addPrimaryKey tableName="subscription" columnNames="subscription_id,start_date" />
+    </changeSet>
+
+</databaseChangeLog>
```
kahowell, line 12: nitpick: newline

src/test/java/org/candlepin/subscriptions/subscription/SubscriptionSyncControllerTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
+
+import java.time.OffsetDateTime;
+import java.util.Optional;
+
+@SpringBootTest
+@ActiveProfiles("worker,test")
```
kahowell, line 37:
```suggestion
@ActiveProfiles({"worker", "test"})
```

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -0,0 +1,110 @@
...
+        if (dto.getEffectiveStartDate() != null) {
+            entity.setStartDate(
+                convertLongTime(dto.getEffectiveStartDate())
+            );
+        }
+        if (dto.getEffectiveEndDate() != null) {
+            entity.setEndDate(
+                convertLongTime(dto.getEffectiveEndDate())
+            );
+        }
```
kahowell, line 101: Was thinking about this, and I think we need to keep start date out of this, because imagine the following scenario:

1. Subscription state is synced:
```yaml
start: 2020-01-01
end: 2020-12-31
quantity: 10
```
-> creates a record that looks like that.
2. Subscription state changes to
```yaml

start: 2020-01-01
end: 2020-12-31
quantity: 5
```
-> saves as two subscription records, one with endDate > now()
3. sync is called again with the same data, active subscription has start != 2020-01-01... -> a third subscription record created...

It'd be ideal to unit test this scenario.

src/test/java/org/candlepin/subscriptions/subscription/SubscriptionSyncControllerTest.java
kahowell: this will fix the merge CI failure

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
jdhrh: So would it instead be reasonable to set the end date to now and create a new record with the updated info if quantity or start date changes? That way our "active data" is only the most current information we have.

kahowell: Actually, I guess similar issue happens with start date, so we might have to drop both from the check.

kahowell (approved):
LGTM

PR #259, opened 2021-02-04
Add MetricUsageCollector

kahowell:
Given that this is not yet integrated generally, unit tests will have to suffice for now.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/AccountRepository.java
+++ b/src/main/java/org/candlepin/subscriptions/db/AccountRepository.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2021 Red Hat, Inc.
+ *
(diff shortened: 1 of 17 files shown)


src/main/java/org/candlepin/subscriptions/db/AccountRepository.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) 2019 Red Hat, Inc.
```
lindseyburnett, line 2: 2021

kahowell:
Note: I marked all the deprecated code as WONTFIX in sonarqube, and created https://issues.redhat.com/browse/ENT-3545 to address later on.

lindseyburnett (approved):

kahowell:
b440f952 to 380e3a93: rebase against develop

PR #270, opened 2021-02-08
Update to spring boot 2.4.2

kahowell:
I had to replace org.springframework.kafka.support.serializer.ErrorHandlingDeserializer2 with org.springframework.kafka.support.serializer.ErrorHandlingDeserializer (apparently these have the same semantics as Javadocs suggest for a while ErrorHandlingDeserializer2 just subclassed the other).

I removed some unneeded dependency version overrides, namely:
  - hamcrest
  - junit
  - micrometer-registry-prometheus
  - mockito
  - snakeyaml

In doing so the spring dep bom updated them for us :-)

I had to separate profiles used in `@ActiveProfiles` annotations (e.g. `{"api", "test"}` instead of `"api,test"`. Apparently older spring boot tolerated either.

Closes #266

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,19 +29,14 @@ plugins {
 ext {
     swagger_annotations_version = "1.5.21"
     swagger_ui_version = "3.20.8"
-    prometheus_version = "1.1.2"
     resteasy_version = "3.6.2.Final"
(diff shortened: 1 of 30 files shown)


awood (approved):
:fireworks: :tada:

lindseyburnett (approved):
:+1:

PR #281, opened 2021-02-09
ENT-3430: Fix regression: allow multiple subscriptions per SKU

kahowell:
Small oversight, we were grouping existing subscriptions by SKU rather than by subscription ID, which causes problems when there is more than one sub for a given SKU.


kahowell:
1ca58320 to e7ae96af: rebase against develop

lindseyburnett (approved):

PR #287, opened 2021-02-10
Stop overriding postgresql and spring-data versions

kahowell:
Note that they are included in spring-boot-dependencies... With the
override removed, they are automatically updated to latest versions.

Closes #278
Closes #279


lindseyburnett (approved):

lindseyburnett:
:+1: that's better...I was wondering why even though we bumped to 2.4, the JPA library wasn't matching

PR #295, opened 2021-02-12
AutoConfigureDatabase one of the two classes using AccountConfigRepos…

lindseyburnett:
…itory is apparently making it so the tests from one don't bleed into the other.  Don't even think of annotating both of them though because it breaks again???


mstead (commented):
Works for me. All tests pass.

PR #296, opened 2021-02-13
Perform OpenShift metrics gathering via two PromQL queries

mstead:
First pull the subscription_lables metrics that contain our metadata,
from prometheus and key the label sets by cluster_id and timestamp.

Next the OpenShift cluster cores metrics are pulled by account and an
Event object is created/persisted if a label set was found matching
the cluster ID and timestamp of the metric value.

Note:
  While this approach isn't optimal, it was done in order to get around
  a join issue caused by the original query when labels such as 'support'
  changes for a cluster.

**Testing**
Testing should be exactly the same as the instructions in #246 .

In fact, we should be seeing the same results from both branches, which would actually be a good test.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
@@ -35,14 +35,17 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.retry.support.RetryTemplate;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.util.StringUtils;
 
 import io.micrometer.core.annotation.Timed;
 
 import java.math.BigDecimal;
 import java.time.OffsetDateTime;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 
 /**
@@ -79,62 +82,21 @@ public class PrometheusMeteringController {
         MetricProperties openshiftProperties = metricProperties.getOpenshift();
         openshiftRetry.execute(context -> {
             try {
-                log.info("Collecting OpenShift metrics");
+                // Reset the start/end dates to ensure they span a complete hour.
+                // NOTE: If the prometheus query step changes, we will need to adjust this.
+                OffsetDateTime queryStart = clock.startOfHour(start);
+                OffsetDateTime queryEnd = clock.endOfHour(end);
 
-                QueryResult metricData = prometheusService.runRangeQuery(
-                    // Substitute the account number into the query. The query is expected to
-                    // contain %s for replacement.
-                    String.format(openshiftProperties.getMetricPromQL(), account),
-                    // Reset the start/end dates to ensure they span a complete hour.
-                    // NOTE: If the prometheus query step changes, we will need to adjust this.
-                    clock.startOfHour(start),
-                    clock.endOfHour(end),
-                    openshiftProperties.getStep(),
-                    openshiftProperties.getQueryTimeout()
-                );
-
-                if (StatusType.ERROR.equals(metricData.getStatus())) {
-                    throw new MeteringException(
-                        String.format("Unable to fetch OpenShift metrics: %s", metricData.getError())
-                    );
+                log.info("Collecting labels for OpenShift metrics...");
+                Map<LabelSetKey, Map<String, String>> labels = getSubscriptionLabels(account, queryStart,
+                    queryEnd);
+                if (labels.isEmpty()) {
+                    log.info("No subscription labels found. No events will be generated.");
                 }
 
-                // Given the possibility of a very large number of events, we will batch persist them
-                // to provide the ability to tweak them.
-                List<Event> events = new LinkedList<>();
-                int eventCount = 0;
-                for (QueryResultDataResult r : metricData.getData().getResult()) {
-                    Map<String, String> labels = r.getMetric();
-                    String clusterId = labels.get("_id");
-                    String sla = labels.get("support");
-                    String usage = labels.get("usage");
-
-                    // For the openshift metrics, we expect our results to be a 'matrix'
-                    // vector [(instant_time,value), ...] so we only look at the result's getValues()
-                    // data.
-                    for (List<BigDecimal> measurement : r.getValues()) {
-                        BigDecimal time = measurement.get(0);
-                        BigDecimal value = measurement.get(1);
-
-                        Event event = MeteringEventFactory.openShiftClusterCores(account, clusterId,
-                            sla, usage, clock.dateFromUnix(time), value.doubleValue());
-                        events.add(event);
-                        eventCount++;
-
-                        if (events.size() >= openshiftProperties.getEventBatchSize()) {
-                            log.info("Saving {} events", events.size());
-                            eventController.saveAll(events);
-                            events.clear();
-                        }
-                    }
-                }
-
-                // Flush the remainder
-                if (!events.isEmpty()) {
-                    log.debug("Saving remaining events: {}", events.size());
-                    eventController.saveAll(events);
-                }
-                log.info("Created {} Events for OpenShift metrics.", eventCount);
+                log.info("Collecting OpenShift metrics and generating events");
+                int eventCount = generateEventsFromMetrics(labels, account, queryStart, queryEnd);
+                log.info("Created {} OpenShift metric events for account {}.", eventCount, account);
                 return null;
             }
             catch (Exception e) {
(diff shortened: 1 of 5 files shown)


mstead:
**THIS PR SHOULD ONLY BE MERGED IF THE LATEST UPDATED PROMQL QUERY DOES NOT SUFFICE!!!!**

Please see: #246 for more details.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -76,65 +79,23 @@ public PrometheusMeteringController(ApplicationClock clock, PrometheusMetricsPro
...
-
-                if (StatusType.ERROR.equals(metricData.getStatus())) {
-                    throw new MeteringException(
-                        String.format("Unable to fetch OpenShift metrics: %s", metricData.getError())
-                    );
-                }
+                // Reset the start/end dates to ensure they span a complete hour.
+                // NOTE: If the prometheus query step changes, we will need to adjust this.
+                OffsetDateTime queryStart = clock.startOfHour(start);
+                OffsetDateTime queryEnd = clock.endOfHour(end);
```
lindseyburnett, line 87: Is this something that we could use a Duration or Period class for?  Admittedly not familiar enough with the clock to know if this is a silly recommendation or not.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -76,65 +79,23 @@ public PrometheusMeteringController(ApplicationClock clock, PrometheusMetricsPro
...
-                        eventCount++;
-
-                        if (events.size() >= openshiftProperties.getEventBatchSize()) {
-                            log.info("Saving {} events", events.size());
-                            eventController.saveAll(events);
-                            events.clear();
-                        }
-                    }
+                log.info("Collecting labels for OpenShift metrics...");
+                Map<LabelSetKey, Map<String, String>> labels = getSubscriptionLabels(account, queryStart,
```
lindseyburnett, line 90: Because we're all Java 11-ified now, you can replace that type with `var` if you wanted to...total readability/preference thing

```java
var labels = getSubscriptionLabels(...
```

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -76,65 +79,23 @@ public PrometheusMeteringController(ApplicationClock clock, PrometheusMetricsPro
...
                 }
 
-                // Flush the remainder
-                if (!events.isEmpty()) {
-                    log.debug("Saving remaining events: {}", events.size());
-                    eventController.saveAll(events);
-                }
-                log.info("Created {} Events for OpenShift metrics.", eventCount);
+                log.info("Collecting OpenShift metrics and generating events");
+                int eventCount = generateEventsFromMetrics(labels, account, queryStart, queryEnd);
```
lindseyburnett, line 97: Wondering if the log statement above this should be the first line of the `generateEventsFromMetrics` method to ensure that the message always goes with its logic

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -145,4 +106,159 @@ public void collectOpenshiftMetrics(String account, OffsetDateTime start, Offset
...
+            });
+        });
+
+        return labelData;
+    }
+
+    private int generateEventsFromMetrics(Map<LabelSetKey, Map<String, String>> labelSets,
+        String account, OffsetDateTime start, OffsetDateTime end) {
+        OpenshiftMetricProperties openshiftProperties = metricProperties.getOpenshift();
+        log.info("Collecting metrics!");
```
lindseyburnett, line 145: aha! RE: my other comment around line 96/97 about the log statement being paired with this method

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -145,4 +106,159 @@ public void collectOpenshiftMetrics(String account, OffsetDateTime start, Offset
...
+
+                if (events.size() >= openshiftProperties.getEventBatchSize()) {
+                    log.info("Saving {} events", events.size());
+                    eventController.saveAll(events);
+                    events.clear();
+                }
+            }
+        }
+
+        // Flush the remainder
```
lindseyburnett, line 204: why do we flush any remaining events?

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -145,4 +106,159 @@ public void collectOpenshiftMetrics(String account, OffsetDateTime start, Offset
...
+        }
+
+        if (!labelSets.containsKey(new LabelSetKey(clusterId, time))) {
+            // This could potentially happen if the metric existed but the
+            // Labels weren't reported yet.
+            log.warn("Found OpenShift metric but could not associate cluster {} " +
+                "with any labels. Time: {} Value: {}", clusterId, time, value);
+            return false;
+        }
+        return true;
```
lindseyburnett, line 227: Curious about everyone's opinion on methods with multiple exit points.

https://hackerchick.com/religious-war-48293-single-vs-multiple/ 

I probably would have wrote this something like this (**DISCLAIMER** completely untested and hastily refactored in notepad*)

```java
    private boolean isMetricDataValid(Map<LabelSetKey, Map<String, String>> labelSets,
        String clusterId, BigDecimal time, BigDecimal value) {

        boolean isTimeValuePairValid = Objects.nonNull(time) && Objects.nonNull(value)

        boolean isMetricExistingAndLabelsNotReported = !labelSets.containsKey(new LabelSetKey(clusterId, time));

        return isTimeValuePairValid && !isMetricExistingAndLabelsNotReported;
     }
```

src/test/java/org/candlepin/subscriptions/metering/OpenShiftMetricsHelper.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2009 - 2019 Red Hat, Inc.
```
lindseyburnett, line 2: 2021

src/main/resources/application-openshift-metering-worker.yaml
```diff
@@ -7,10 +7,8 @@ rhsm-subscriptions:
     prometheus:
       metric:
         openshift:
-          metricPromQL: >-
-            max(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) without(environment, region)*on(_id)
-            group_right(prometheus) group(subscription_labels{ebs_account='%s', support=~'Premium|Standard|Self-Support|None'})
-            by (_id,ebs_account, support, usage)
+          metricPromQL: group(subscription_labels{ebs_account='%s',support=~'Premium|Standard|Self-Support|None'})by(_id,ebs_account)*on(_id) group_right(ebs_account) cluster:usage:workload:capacity_physical_cpu_cores:min:5m
```
lindseyburnett, line 10: Just an observation more than anything else...but sometimes we have application properties using camelCase convention, and other times we use dash-case

src/test/java/org/candlepin/subscriptions/metering/OpenShiftMetricsHelper.java
```diff
@@ -0,0 +1,86 @@
...
+import java.math.BigDecimal;
+import java.util.List;
+
+/**
+ * A utility class to help with creating prometheus API response data.
+ */
+public class OpenShiftMetricsHelper {
+
+    private OpenShiftMetricsHelper() {
+
```
lindseyburnett, line 38: I'm surpised Sonar isn't upset that this doesn't have a comment inside of it explaining why its empty

src/test/java/org/candlepin/subscriptions/metering/OpenShiftMetricsHelper.java
```diff
@@ -0,0 +1,86 @@
...
+               new QueryResultData()
+                   .resultType(ResultType.MATRIX)
+                   .addResultItem(dataResult)
+            );
+    }
+
+    public static QueryResult buildCoresMetricQueryResult(String account, String clusterId,
+        List<List<BigDecimal>> timeValueTuples) {
+        QueryResultDataResult dataResult = new QueryResultDataResult()
+            .putMetricItem("_id", clusterId)
```
lindseyburnett, line 58: CTRL+F on this page has seven results of `"_id"`...I know at least some of them are probably duplicates because of the diffs....but is it worth making this a constant field?

src/test/java/org/candlepin/subscriptions/metering/OpenShiftMetricsHelper.java
```diff
@@ -0,0 +1,86 @@
...
+            .putMetricItem("ebs_account", account);
+
+        // NOTE: A tuple is [unix_time,value]
+        timeValueTuples.forEach(tuple -> dataResult.addValuesItem(tuple));
```
lindseyburnett, line 62: Would you consider this self-documenting code if we renamed to `unixTimeValueTuples` and then dropped the comment on line 61?

src/test/java/org/candlepin/subscriptions/metering/OpenShiftMetricsHelper.java
```diff
@@ -0,0 +1,86 @@
...
+           new QueryResultData()
+               .resultType(ResultType.MATRIX)
+               .addResultItem(dataResult)
+        );
+    }
+
+    public static QueryResultDataResult buildLabelDataResult(String account, String clusterId,
+        String serviceLevel, String usage, List<List<BigDecimal>> timeValueTuples) {
+        QueryResultDataResult dataResult = new QueryResultDataResult()
+            .putMetricItem("_id", clusterId)
```
lindseyburnett, line 76: Already made the comment about `"_id"` specifically, but seeing them alongside these other metric items...is an enum in order?

src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringControllerTest.java
```diff
@@ -77,29 +77,43 @@
...
+        String expectedMetricQuery = String.format(props.getOpenshift().getMetricPromQL(), expectedAccount);
+
         QueryResult errorResponse = new QueryResult();
         errorResponse.setStatus(StatusType.ERROR);
         errorResponse.setError("FORCED!!");
 
-        QueryResult good = buildOpenShiftClusterQueryResult(expectedAccount, expectedClusterId, expectedSla,
-            expectedUsage, List.of(List.of(new BigDecimal(12312.345), new BigDecimal(24))));
+        QueryResult metrics = buildCoresMetricQueryResult(expectedAccount, expectedClusterId,
+            List.of(List.of(new BigDecimal(12312.345), new BigDecimal(24))));
```
lindseyburnett, line 89: :+1:  for `List.of` :smile:

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
lindseyburnett: If yes, there are also some unit tests that could do the same.

mstead:
Closing this PR in favor of #246 only. 

**Let's keep the branch around until QE has a chance to run 246 through its paces.**

PR #293, opened 2021-02-12
ENT-3260: enhance tally snapshot controller

lindseyburnett:
Implement CaptureMetricsSnapshotTask and expose via TallyJmxBean

Adds the ability to initiate a hourly tally for a given account between a given time frame.  You must use the kafka-queue profile, there is no corresponding in-memory queue implementation.  Producing snapshots emits a kafka message of List<TallySummary>

PR #297, opened 2021-02-15
ENT-3263: prometheus latency setting

lindseyburnett:
Subtract duration from start & end times during MetricUsageCollector.collect to account for prometheus latency.

Depends on #293.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -97,6 +97,11 @@ allprojects {
     dependencies {
         implementation 'javax.annotation:javax.annotation-api:1.3.2'
 
+        compileOnly "org.projectlombok:lombok"
+        annotationProcessor "org.projectlombok:lombok"
+        testCompileOnly "org.projectlombok:lombok"
+        testAnnotationProcessor "org.projectlombok:lombok"
+
         testCompile "org.springframework.boot:spring-boot-starter-test"
         testCompile "org.springframework:spring-test"
 
--- a/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
@@ -20,6 +20,7 @@
  */
 package org.candlepin.subscriptions.tally;
 
+import org.candlepin.subscriptions.ApplicationProperties;
 import org.candlepin.subscriptions.db.AccountRepository;
 import org.candlepin.subscriptions.db.model.Account;
 import org.candlepin.subscriptions.db.model.HardwareMeasurementType;
@@ -37,9 +38,11 @@ import org.candlepin.subscriptions.metering.MeteringEventFactory;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.HashSet;
@@ -64,6 +67,7 @@ public class MetricUsageCollector {
     private final AccountRepository accountRepository;
     private final EventController eventController;
     private final ProductConfig productConfig;
+    private final ApplicationProperties applicationProperties;
 
     /**
      * Encapsulates all per-product information we anticipate putting into configuration used by this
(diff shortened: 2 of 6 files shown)


dmckoske (commented):
Looks good to me.

build.gradle
```diff
@@ -36,6 +36,7 @@ ext {
     kafka_avro_serializer_version = "6.1.0"
     avro_version = "1.10.1"
     webjars_locator_version = "0.40"
+    lombok_version="1.18.16"
```
kahowell, line 39: Methinks this line unnecessary; 99% sure spring-boot bom brings in a lombok version itself.

build.gradle
```diff
@@ -97,6 +98,12 @@ allprojects {
     dependencies {
         implementation 'javax.annotation:javax.annotation-api:1.3.2'
 
+        compileOnly "org.projectlombok:lombok:$lombok_version"
+        annotationProcessor "org.projectlombok:lombok:$lombok_version"
+
+        testCompileOnly "org.projectlombok:lombok:$lombok_version"
+        testAnnotationProcessor "org.projectlombok:lombok:$lombok_version"
+
```
kahowell, line 106: Can drop the `:$lombok_version` per previous comment.

build.gradle
```diff
@@ -159,6 +166,8 @@ release {
 }
 
 dependencies {
+
+    implementation "org.projectlombok:lombok:$lombok_version"
```
kahowell, line 170: Can drop the `:$lombok_version` per previous comment.

src/main/resources/application.yaml
```diff
@@ -97,4 +97,4 @@ rhsm-subscriptions:
       tasks:
         topic: ${OPENSHIFT_METERING_TASK_TOPIC:platform.rhsm-subscriptions.openshift-metering-tasks}
         kafka-group-id: ${OPENSHIFT_METERING_TASK_GROUP_ID:openshift-metering-task-processor}
-
+  prometheus-latency-duration: ${PROMETHEUS_LATENCY_DURATION:PT4H}
```
kahowell, line 100: I prefer the "more readable" format the Spring Boot supports (https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration) just "4h". I had to google PT4H, so regardless, I learned something :-)

kahowell (changes requested):

build.gradle
lindseyburnett: I'll test some more things out...100% sure that IntelliJ couldn't find the lombok to import when I added the first annotation.

src/main/resources/application.yaml
mstead: :+1:

mstead: Could have used this in the metrics PR. :)

lindseyburnett: Well would you look at that....didn't know there even was a more readable version.  Just a heads up on this though...while spring magic makes this work a property...junit doesn't have the same magic for the @ParameterizedTest I made.

```
Error converting parameter at index 1: Failed to convert String "0h" to type java.time.Duration
org.junit.jupiter.api.extension.ParameterResolutionException: Error converting parameter at index 1: Failed to convert String "0h" to type java.time.Duration
	at org.junit.jupiter.params.ParameterizedTestMethodContext.parameterResolutionException(ParameterizedTestMethodContext.java:273)
```

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -99,17 +103,39 @@ public ServiceLevel getDefaultSla() {
...
         this.eventController = eventController;
         this.productConfig = new ProductConfig();
+        this.applicationProperties = applicationProperties;
+
+    }
+
+    protected OffsetDateTime adjustTimeForLatency(OffsetDateTime dateTime, Duration adjustmentAmount) {
+
+        return dateTime.toZonedDateTime().minus(adjustmentAmount).toOffsetDateTime();
+
```
kahowell, line 120: Thought about it some, and I think we should just do some validation of the ranges passed to these methods in hit tasks, and save the adjustment exclusively for ENT-3542. I find it a little confusing to apply it to anything _but_ the scheduler.

kahowell (changes requested):

build.gradle
```diff
@@ -159,6 +164,7 @@ release {
 }
 
 dependencies {
+    implementation "org.projectlombok:lombok"
```
kahowell, line 167: @lindseyburnett  just noticed this line is also unnecessary.

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
lindseyburnett: I agree.  Will adjust this on the ENT-3542 branch (#299)

PR #300, opened 2021-02-18
Fix bean issue with default profiles

mstead:
Duplicate bean name was causing a startup issue when running with default profiles.

## Testing
Deploy as follows:

```bash
./gradlew clean bootRun
```

Before this patch, the application will not start and complains about eventController bean.

Application should start after this patch is applied.


dmckoske (approved):
Looks good to me, given all checks pass.

lindseyburnett (approved):

PR #301, opened 2021-02-18
ENT-3563: Comment out updating subscription logic to prevent constraint violation

lindseyburnett:


dmckoske (approved):
Looks good. Open a card in the backlog and let Kevin know.

lindseyburnett:
> Looks good. Open a card in the backlog and let Kevin know.

It's in the backlog already as ENT-3573.  I'll update the comment in the code to specify as such.

kahowell (approved):

PR #302, opened 2021-02-18
ENT-3562: null while collecting existing snapshot of account.

dmckoske:
Fixing a file where this occurs, adding some backstop code for future
occurrences (credit lburnett)

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/model/Host.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/Host.java
@@ -20,6 +20,7 @@
  */
 package org.candlepin.subscriptions.db.model;
 
+import jdk.jfr.DataAmount;
 import org.candlepin.subscriptions.inventory.db.model.InventoryHostFacts;
 import org.candlepin.subscriptions.json.Measurement;
 import org.candlepin.subscriptions.tally.facts.NormalizedFacts;
@@ -177,8 +178,15 @@ public class Host implements Serializable {
         this.subscriptionManagerId = inventoryHostFacts.getSubscriptionManagerId();
         this.guest = normalizedFacts.isVirtual();
         this.hypervisorUuid = normalizedFacts.getHypervisorUuid();
-        this.measurements.put(Measurement.Uom.CORES, normalizedFacts.getCores().doubleValue());
-        this.measurements.put(Measurement.Uom.SOCKETS, normalizedFacts.getSockets().doubleValue());
+
+        if(normalizedFacts.getCores() != null) {
+            this.measurements.put(Measurement.Uom.CORES, normalizedFacts.getCores().doubleValue());
+        }
+
+        if(normalizedFacts.getSockets() != null) {
+            this.measurements.put(Measurement.Uom.SOCKETS, normalizedFacts.getSockets().doubleValue());
+        }
+
         this.isHypervisor = normalizedFacts.isHypervisor();
         this.isUnmappedGuest = normalizedFacts.isVirtual() && normalizedFacts.isHypervisorUnknown();
         this.cloudProvider = normalizedFacts.getCloudProviderType() == null ?
--- a/src/main/java/org/candlepin/subscriptions/db/model/Host.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/Host.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Red Hat, Inc.
+ * Copyright (c) 2019 - 2021 Red Hat, Inc.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -20,7 +20,6 @@
  */
 package org.candlepin.subscriptions.db.model;
 
-import jdk.jfr.DataAmount;
 import org.candlepin.subscriptions.inventory.db.model.InventoryHostFacts;
 import org.candlepin.subscriptions.json.Measurement;
 import org.candlepin.subscriptions.tally.facts.NormalizedFacts;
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
```diff
@@ -100,6 +102,14 @@ public void produceSnapshotsForAccounts(List<String> accounts) {
...
             }
         }
+        catch (NullPointerException e) {
+            StringWriter sw = new StringWriter();
+            PrintWriter pw = new PrintWriter(sw);
+            e.printStackTrace(pw);
+            log.error("Could not collect existing usage snapshots for accounts {}, stacktrace: {}",
+                accounts, pw);
+            return;
+        }
```
dmckoske, line 112: This from Lindsey.  Thanks, Lindsey.

kahowell: I think this style of logging is fine to use in a one-off branch to verify an issue.

However, it's not the idiomatic way to log a stack trace w/ slf4j. We're actually already doing that... we should track down why the *existing* logging statement below:

```java
log.error("Could not collect existing usage snapshots for accounts {}", accounts, e);
```
didn't emit the stack trace (out of scope for this fix - I made https://issues.redhat.com/browse/ENT-3574 for this). For now, please remove these changes.

src/main/java/org/candlepin/subscriptions/db/model/Host.java
```diff
@@ -177,8 +177,10 @@ public void populateFieldsFromHbi(InventoryHostFacts inventoryHostFacts,
         this.subscriptionManagerId = inventoryHostFacts.getSubscriptionManagerId();
         this.guest = normalizedFacts.isVirtual();
         this.hypervisorUuid = normalizedFacts.getHypervisorUuid();
-        this.measurements.put(Measurement.Uom.CORES, normalizedFacts.getCores().doubleValue());
-        this.measurements.put(Measurement.Uom.SOCKETS, normalizedFacts.getSockets().doubleValue());
+        this.measurements.put(Measurement.Uom.CORES,
+            (normalizedFacts.getCores() == null ? 0.0 : normalizedFacts.getCores().doubleValue()));
+        this.measurements.put(Measurement.Uom.SOCKETS,
+            (normalizedFacts.getSockets() == null ? 0.0 : normalizedFacts.getSockets().doubleValue()));
```
kahowell, line 183: I would rather see us check these values and *not* call `this.measurements.put` if the value is `null`. Simple `if` checks would suffice.

I would also like to see a unit test for this, so that we don't regress on either cores or sockets.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
kahowell: FWIW, the reason the stack trace was not logged could be as simple as https://stackoverflow.com/a/9387598, meaning there would have been at least one earlier instance where the trace was logged.

dmckoske: Fair.  I will update this and the new ticket, and remove this block.

src/main/java/org/candlepin/subscriptions/db/model/Host.java
dmckoske: Very good, I'll change this and add a unit test.

src/test/java/org/candlepin/subscriptions/db/model/HostTest.java
```diff
@@ -0,0 +1,109 @@
...
+        assertEquals(host.getHypervisorUuid(), normalizedFacts.getHypervisorUuid());
+        assertEquals(2, host.getMeasurements().size());
+        assertTrue(host.isHypervisor());
+        assertEquals(host.getCloudProvider(), normalizedFacts.getCloudProviderType().name());
+        assertEquals(host.getLastSeen(), inventoryHostFacts.getModifiedOn());
+        assertEquals(host.getHardwareType(), normalizedFacts.getHardwareType());
+    }
+
+    private InventoryHostFacts getInventoryHostFactsFull() {
+        inventoryHostFacts.setInventoryId(UUID.randomUUID());
```
lindseyburnett, line 85: It's weird to have a "get" method with side effects - setting the attributes of a shared object - at which point you don't even need to return the object because it updated at the class level and can be accessed as such.

I would prefer to see one of the following options:

1.  Remove inventoryHostFacts and normalizedFacts as class level variables, and put the contents of these two methods local to the test they're being used for, since the methods only get invoked in one place.
2.  Remove inventoryHostFacts and normalizedFacts as class level variables, and make these helper methods return a new instance of InventoryHostFacts/NormalizedFacts
```java
private InventoryHostFacts getNewInventoryHostFactsFull() {
        InventoryHostFacts inventoryHostFacts = new InventoryHostFacts();
        inventoryHostFacts.setInventoryId(UUID.randomUUID());
        // .....
       return inventoryHostFacts;
```
3.  If you wanted all tests to share the same instance of these objects (doesn't seem like that's what we want to do), you could initialize them with a method annotated with `@Before`

lindseyburnett (changes requested):

src/test/java/org/candlepin/subscriptions/db/model/HostTest.java
dmckoske: I'll take option 2 above, since it's easy and results in smaller blocks of code than option 1.  Option 3 doesn't work with the first test with all-null fields.  New PR forthcoming.

lindseyburnett (approved):

lindseyburnett (approved):

PR #303, opened 2021-02-19
Fix JPA mapping to address NPE during account merge

kahowell:
Prior to the change, if a host had tally buckets added, then they would not receive the proper host_id populated. Apparently, it not intended to map the `@ManyToOne` in an embedded ID. See the javadoc for `@MapsId` for a proper example. I tried several variations; `@MapsId` isn't picked up if applied to the `@EmbeddedId`.

I also noted that the HashMap for instances was causing extra unnecessary update statements, so I tweaked its JPA config too.


mstead:
:+1:  Nice fix and good catch on the update statements!

mstead (approved):
Looks good to me.

PR #304, opened 2021-02-19
Tweak MetricUsageCollector Strings

kahowell:
1. Product ID was too long
2. The service type didn't match what we used in the metrics gathering
   phase.


mstead (approved):
Noticed this yesterday. Was about to put a PR up for this but you beat me to it! :)

PR #308, opened 2021-02-19
Filter out hosts with null inventory IDs when collecting from inventory

mstead:
Now that we are storing host records from more than just HBI,
inventory ID may be null. Filter is so that we do not get
duplicate keys during the stream mapping.

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
+++ b/src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
@@ -527,6 +527,35 @@ public class InventoryAccountUsageCollectorTest {
         assertEquals(1, counter.count() - initialCount);
     }
 
+    @Test
+    void accountsWithNullInventoryIdFiltered() {
+        String account = "A1";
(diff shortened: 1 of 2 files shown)


src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorTest.java
```diff
@@ -527,6 +527,35 @@ void testTotalHosts() {
...
 
+    @Test
+    void accountsWithNullInventoryIdFiltered() {
+        String account = "A1";
+        List<Integer> products = Arrays.asList(TEST_PRODUCT_ID);
+        List<String> targetAccounts = Arrays.asList(account);
+        InventoryHostFacts host = createRhsmHost(account, "O1", products, 12, 3, "",
+            OffsetDateTime.now());
+
+        when(hostRepo.findByAccountNumber(eq("A1"))).thenReturn(List.of(
```
kahowell, line 538: sonarcloud fix:
```suggestion
        when(hostRepo.findByAccountNumber("A1")).thenReturn(List.of(
```

kahowell (changes requested):

kahowell (approved):

PR #310, opened 2021-02-19
Awood/metric deploy

awood:
Openshift templates for deploying Prometheus metering.

### DIFF
--- a/templates/rhsm-subscriptions-scheduler.yml
+++ b/templates/rhsm-subscriptions-scheduler.yml
@@ -19,6 +19,10 @@ parameters:
     value: 0 3 * * *
   - name: CAPTURE_SNAPSHOT_SCHEDULE
     value: 0 1 * * *
+  - name: OPENSHIFT_METERING_SCHEDULE
+    value: 0 * * * *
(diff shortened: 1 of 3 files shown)


templates/rhsm-subscriptions-scheduler.yml
```diff
@@ -19,6 +19,10 @@ parameters:
     value: 0 3 * * *
   - name: CAPTURE_SNAPSHOT_SCHEDULE
     value: 0 1 * * *
+  - name: OPENSHIFT_METERING_SCHEDULE
+    value: 0 * * * *
+  - name: OPENSHIFT_METERING_RANGE
+    value: '180'
```
kahowell, line 25: Why 180 here? I expected to keep the default of 60?

kahowell (changes requested):

templates/rhsm-subscriptions-scheduler.yml
awood: That was a setting for testing.  Just pushed a fix.

kahowell (approved):

PR #319, opened 2021-02-26
Retain a year's worth of daily snapshots

awood:
Because of leap year, we would have needed to do at least 366, but I
like round numbers so I went with 370.  A little buffer might be helpful
anyway.

### DIFF
--- a/src/main/resources/application-purge-snapshots.yaml
+++ b/src/main/resources/application-purge-snapshots.yaml
@@ -1,8 +1,18 @@
 rhsm-subscriptions:
   tally-retention-policy:
+    # 70 days worth
     hourly: ${TALLY_RETENTION_HOURLY:1680}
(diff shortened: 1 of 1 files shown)


awood:
TODO: Update internal documentation after this is merged: "Retention" on "Subscription Watch" page.

awood:
@kahowell Do the Jenkins magic strings e.g. ("retest this please") work here?  I think the PR is fine but Jenkins went kaput and for some reason I can't restart it.

awood:
retest this please

kahowell:
@awood, for retest it's Details -> Login -> Re-run; however, we're in the middle of a Jenkins migration, so it might be best to start at the top level pipeline job (i.e. https://stage-jenkins-csb-smqe.cloud.paas.psi.redhat.com/job/Subscriptions%20PR%20Check/ and specifically choose the branch or PR and re-run that way).

PR #320, opened 2021-03-01
ENT-3385: Perform summation for rollups on hourly snapshots

lindseyburnett:
collab w/ @kahowell...+1 for writing up this great description :+1: 

---

``

This adds a new class CombiningSnapshotStrategy, which works somewhat
similarly to pre-existing logic, except that it loads more snapshots in
order to (re-)calculate the rollups. For example, it loads enough hourly
snapshots to calculate any affected daily snapshots. The strategy takes
an argument for the function to use during rollup, so it should be
straightforward to retrofit the existing logic into the new class.

Existing UsageSnapshotProducer was renamed MaxSeenSnapshotStrategy.

Testing
-------
Insert some test data:
```sql
insert into events(id, account_number, timestamp, data) values ('033a45b7-e8c4-440a-b0bb-76c3f3f67d85','account123', '2021-01-01T00:00Z', '{"account_number":"account123","event_id":"033a45b7-e8c4-440a-b0bb-76c3f3f67d85","service_type":"OpenShift Cluster","instance_id":"4e8534b5-42b5-4f00-9969-29046e2b1986","measurements": [{"uom":"Cores","value":4.0}]}');
insert into events(id, account_number, timestamp, data) values ('a66ac2a1-17c0-4bc3-bc29-0430953d626c','account123', '2021-01-01T01:00Z', '{"account_number":"account123","event_id":"a66ac2a1-17c0-4bc3-bc29-0430953d626c","service_type":"OpenShift Cluster","instance_id":"4e8534b5-42b5-4f00-9969-29046e2b1986","measurements": [{"uom":"Cores","value":2.0}]}');
insert into events(id, account_number, timestamp, data) values ('fad48795-ae0a-4674-b864-822ddfb799d0','account123', '2021-01-01T02:00Z', '{"account_number":"account123","event_id":"fad48795-ae0a-4674-b864-822ddfb799d0","service_type":"OpenShift Cluster","instance_id":"4e8534b5-42b5-4f00-9969-29046e2b1986","measurements": [{"uom":"Cores","value":4.0}]}');
```

Then trigger a tally: http://localhost:8080/actuator/hawtio/jmx/operations?nid=root-org.candlepin.subscriptions.jmx-TallyJmxBean-tallyJmxBean
`tallyAccountByHourly(String, String, String)`, `account`: account123, `startDateTime`: 2021-01-01T00:00:00Z, `endDateTime`: 2021-01-01T12:00:00Z

Verify the resulting tally snapshots (including rollups):
```sql
select account_number,snapshot_date,uom,value from tally_snapshots s join tally_measurements m on s.id=m.snapshot_id where measurement_type='PHYSICAL' order by sla,usage,snapshot_date asc;
```


src/main/java/org/candlepin/subscriptions/db/model/TallySnapshot.java
```diff
@@ -49,6 +55,10 @@
 /**
  * Model object to represent pieces of tally data.
  */
+@ToString
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
```
dmckoske, line 61: The four Lombok annotations above are the way we should proceed with new models, for sure.  If we can get away with in for existing models, all the better.

src/main/java/org/candlepin/subscriptions/db/model/TallySnapshot.java
```diff
@@ -281,14 +293,26 @@ public void setUsage(Usage usage) {
...
+            return false;
+        }
+        TallySnapshot that = (TallySnapshot) o;
+        return Objects.equals(snapshotDate, that.snapshotDate) &&
+            Objects.equals(productId, that.productId) &&
+            Objects.equals(ownerId, that.ownerId) &&
+            Objects.equals(accountNumber, that.accountNumber) &&
+            serviceLevel == that.serviceLevel &&
+            usage == that.usage &&
+            granularity == that.granularity;
```
dmckoske, line 310: I'd recommend replacing this long concatenation with StringBuilder, per https://javapapers.com/java/java-string-vs-stringbuilder-vs-stringbuffer-concatenation-performance-micro-benchmark/.  It may not be so critical for this particular code, but it's a habit we should get into.

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotNaturalKey.java
```diff
@@ -0,0 +1,55 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+import java.time.OffsetDateTime;
+
+/**
+ * The business key for a TallySnapshot.
+ */
+@Data
+@AllArgsConstructor
```
dmckoske, line 37: No lombok Builder?

dmckoske (approved):
Looks good, I'd just like some further discussion on how we're using lombok.

src/main/java/org/candlepin/subscriptions/db/model/TallySnapshot.java
lindseyburnett: We're not building a string here, it's conditionals for overriding the `equals` method.

dmckoske: Okay, I see now.

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotNaturalKey.java
lindseyburnett: I'm still of the philosophy of only including what you're going to use.

dmckoske: Right, but you could use it.  I see the constructor syntax elsewhere.

lindseyburnett: You need an all args constructor for a builder, but you don't have to have a builder to use an all args constructor.

kahowell: @dmckoske let's revisit general patterns for lombok later on, for now since this PR doesn't need it.

awood (changes requested):
Overall, I'd like to see more comments.  This is a highly technical PR (kudos, btw) and I think more comments would help me (and future maintainers) grok it faster.

awood:
@lindseyburnett Thanks for working on the readability of the code.  I would like to see some actual discussion of it though.  Javadoc on all the major methods would be fine.  Just a short paragraph describing the algorithm.  "This method gathers all snapshots at the finest level of granularity and..."

Specifically, I am a little clueless as why there are 2 `produceRollups` methods and a little unclear why both DAILY and HOURLY granularity are always operative as opposed to just the finest granularity appropriate to the product.

lindseyburnett:
> @lindseyburnett Thanks for working on the readability of the code. I would like to see some actual discussion of it though. Javadoc on all the major methods would be fine. Just a short paragraph describing the algorithm. "This method gathers all snapshots at the finest level of granularity and..."
> 
> Specifically, I am a little clueless as why there are 2 `produceRollups` methods and a little unclear why both DAILY and HOURLY granularity are always operative as opposed to just the finest granularity appropriate to the product.

Understood, I hear ya.  I wasn't quite finished yet...I plucked the low-hanging fruit of refactoring while mulling over how to articulate the explanation in comments.

awood (approved):

PR #318, opened 2021-02-25
Add Liquibase changeset to apply unique index and index to events table

awood:

PR #322, opened 2021-03-01
ENT-3587: Properly create/update events when processing OpenShift metrics

mstead:
When metric data is being processed, update existing events and
delete any that are stale from a previous run. If the metering
process is re-run then event data should be in sync afterwards.

Prior to this change, duplicate events would be created as if
the metering process was run twice.

https://issues.redhat.com/browse/ENT-3587

**NOTE:** I've merged the required database changes from #318 into this branch so they can be merged together.

**TESTING**

1) See https://docs.engineering.redhat.com/display/ENT/Prometheus+Metrics for getting an auth token for thanos. I like to export it into a variable so my startup command remains the same.
```bash
export PROM_TOKEN=<YOUR_TOKEN_FROM_ABOVE>
```

2) After getting your token, deploy the app. Using  a step of 300 will create way more events for the period than our normal 1h step when the collection runs. We will make sure they get purged later.
```bash
$ DEV_MODE=true RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_STEP=300 RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_RANGE_IN_MINUTES=480 PROM_AUTH_TOKEN="$(echo $PROM_TOKEN)" PROM_URL="https://telemeter-lts.datahub.redhat.com/api/v1" ./gradlew clean bootRun --args="--spring.profiles.active=openshift-metering-worker"
```
3) Hit the JMX endpoint to start collection for the account.
```bash
curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=openshiftJmxBean,type=OpenShiftJmxBean","operation":"performOpenshiftMeteringForAccount(java.lang.String)","arguments":["1460290"]}'
```

4) Watch the logs to see that events were added, and check the count in the DB.
```bash
psql -U rhsm-subscriptions rhsm-subscriptions -c "select count(*) from events;"
 count 
-------
   440

```

5) Stop the application and re-launch as follows ensuring that your database isn't wiped out. This time we use a 1h step to reduce the number of metrics pulled from prometheus.
```bash
$ DEV_MODE=true RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_STEP=3600 RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_RANGE_IN_MINUTES=480 PROM_AUTH_TOKEN="$(echo $PROM_TOKEN)" PROM_URL="https://telemeter-lts.datahub.redhat.com/api/v1" ./gradlew clean bootRun --args="--spring.profiles.active=openshift-metering-worker"
```

6) Hit the JMX endpoint again
```bash
curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=openshiftJmxBean,type=OpenShiftJmxBean","operation":"performOpenshiftMeteringForAccount(java.lang.String)","arguments":["1460290"]}'
```
NOTE:
```bash
2021-03-01 14:43:24,120 [thread=pool-6-thread-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusService] - Fetching metrics from prometheus: 2021-03-01T10:00Z -> 2021-03-01T18:59:59.999999999Z [Step: 3600]
2021-03-01 14:43:26,391 [thread=pool-6-thread-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusMeteringController] - Persisted 41 events for OpenShift metrics.
2021-03-01 14:43:26,391 [thread=pool-6-thread-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusMeteringController] - Deleting 399 stale OpenShift metric events.
2021-03-01 14:43:26,515 [thread=pool-6-thread-1] [INFO ] [org.candlepin.subscriptions.metering.task.OpenShiftMetricsTask] - Openshift metrics task complete.
```

7) Watch the logs to see that events were added/purged, and check the count in the DB to make sure that it reflects the deletions/updates.
```bash
psql -U rhsm-subscriptions rhsm-subscriptions -c "select count(*) from events;"

 count 
-------
    41
(1 row)
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
+++ b/src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
@@ -34,6 +34,7 @@ import java.util.stream.Stream;
  * @see EventRecord
  * @see org.candlepin.subscriptions.json.Event
  */
+@SuppressWarnings({"linelength", "indentation"})
 public interface EventRecordRepository extends JpaRepository<EventRecord, UUID> {
 
     /**
@@ -52,4 +53,23 @@ public interface EventRecordRepository extends JpaRepository<EventRecord, UUID> 
      */
     Stream<EventRecord> findByAccountNumberAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
         String accountNumber, OffsetDateTime begin, OffsetDateTime end);
+
+    /**
+     * Fetch a stream of events for a given account, event type and event source for a given time range.
+     *
+     * The events returned include those at begin and up to (but not including) end.
+     *
+     * NOTE: this query does not use `between` since between semantics are inclusive. e.g. `select * from
+     * events where timestamp between '2021-01-01T00:00:00Z' and '2021-01-01T01:00:00Z` would match events
+     * at midnight UTC and 1am UTC.
+     *
+     * @param accountNumber account number
+     * @param eventSource event source
+     * @param eventType event type
+     * @param begin start of the time range (inclusive)
+     * @param end end of the time range (exclusive)
+     * @return Stream of EventRecords
+     */
+    Stream<EventRecord> findByAccountNumberAndEventSourceAndEventTypeAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
+        String accountNumber, String eventSource, String eventType, OffsetDateTime begin, OffsetDateTime end);
 }
--- a/src/main/java/org/candlepin/subscriptions/db/model/EventKey.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/EventKey.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2021 Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.db.model;
+
+import org.candlepin.subscriptions.json.Event;
+
+import lombok.Getter;
+
+import java.time.OffsetDateTime;
+import java.util.Objects;
+
+/**
+ * Defines the unique constraints for an EventRecord. This object is primarily used for creating
+ * lookup tables while processing Events pulled from the DB.
+ */
+@Getter
+public class EventKey {
+
+    private String accountNumber;
+    private String eventType;
+    private String eventSource;
+    private String instanceId;
+    private OffsetDateTime timestamp;
+
+    public EventKey(String accountNumber, String eventSource, String eventType, String instanceId,
+        OffsetDateTime timestamp) {
+        Objects.requireNonNull(accountNumber, "EventKey requires a non null 'accountNumber'.");
+        this.accountNumber = accountNumber;
+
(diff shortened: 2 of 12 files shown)


src/main/java/org/candlepin/subscriptions/db/model/EventKey.java
```diff
@@ -0,0 +1,87 @@
...
+@Getter
+public class EventKey {
+
+    private String accountNumber;
+    private String eventType;
+    private String eventSource;
+    private String instanceId;
+    private OffsetDateTime timestamp;
+
+    public EventKey(String accountNumber, String eventType, String eventSource, String instanceId,
```
kahowell, line 43: Can you please swap the order to be `eventSource`, then `eventType`, to be consistent with the repository method?

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -99,8 +101,17 @@ public void collectOpenshiftMetrics(String account, OffsetDateTime start, Offset
...
+                    MeteringEventFactory.OPENSHIFT_CLUSTER_EVENT_TYPE,
+                    start,
+                    end
+                );
+                log.debug("Found {} existing events.", existing.size());
+
                 // Given the possibility of a very large number of events, we will batch persist them
-                // to provide the ability to tweak them.
+                // to provide the ability to tweak the size later.
                 List<Event> events = new LinkedList<>();
```
kahowell, line 115: Rather than use a `List` here, for consistency, I think you should use `Map<EventKey, Event>`.  That way, we can handle the case where there are conflicting events in the same batch from Prometheus (last-in wins). I think we'll also have to drop the batching mechanism for this (though I think it likely wasn't doing too much for us since this is all in one transaction anyways).

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -116,8 +127,8 @@ public void collectOpenshiftMetrics(String account, OffsetDateTime start, Offset
                         BigDecimal time = measurement.get(0);
                         BigDecimal value = measurement.get(1);
 
-                        Event event = MeteringEventFactory.openShiftClusterCores(account, clusterId,
-                            sla, usage, clock.dateFromUnix(time), value.doubleValue());
+                        Event event = createOrUpdateEvent(existing, account, clusterId, sla, usage, time,
+                            value);
                         events.add(event);
                         eventCount++;
```
kahowell, line 133: Because this `eventCount` may be inaccurate when processing conflicting events (see `Map` suggestion), this var will need to be replaced by `events.size()`.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/db/model/EventKey.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
```diff
@@ -52,4 +53,23 @@
...
+     * at midnight UTC and 1am UTC.
+     *
+     * @param accountNumber account number
+     * @param eventSource event source
+     * @param eventType event type
+     * @param begin start of the time range (inclusive)
+     * @param end end of the time range (exclusive)
+     * @return Stream of EventRecords
+     */
+    Stream<EventRecord> findByAccountNumberAndEventSourceAndEventTypeAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
```
lindseyburnett, line 73: I can't wait until we roll out JPA specifications/criteria to the whole project :D

kahowell (approved):

PR #325, opened 2021-03-02
ENT-3592: fix DataIntegrityViolation.

dmckoske:
Added a line to HostTallyBucket.setHost() which sets the hostID
of the HostBucketKey.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
@@ -106,6 +106,7 @@ public class HostTallyBucket implements Serializable {
 
     public void setHost(Host host) {
         this.host = host;
+        this.key.setHostId(host.getId());
     }
 
     @Override
--- a/src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
+++ b/src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
@@ -106,7 +106,7 @@ public class HostTallyBucket implements Serializable {
 
     public void setHost(Host host) {
         this.host = host;
-        this.key.setHostId(host.getId());
+        this.key.setHostId(host == null ? null : host.getId());
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/test/java/org/candlepin/subscriptions/db/model/HostTallyBucketTest.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2019 Red Hat, Inc.
```
lindseyburnett, line 2: 2021

lindseyburnett (changes requested):

src/test/java/org/candlepin/subscriptions/db/model/HostTallyBucketTest.java
dmckoske: Done.

lindseyburnett (approved):

src/test/java/org/candlepin/subscriptions/db/model/HostTallyBucketTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import java.util.UUID;
+
+public class HostTallyBucketTest {
+
+    @Test
+    void setHostId() {
```
kahowell, line 34: I would prefer a more descriptive test name and a comment of why we care about the behavior; how about:
```suggestion
    // this behavior necessary to avoid duplicate keys with hostId null
    void testBucketSetHostAlsoSetsBucketKeyHostId() {
```
?

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
```diff
@@ -106,6 +106,7 @@ public Host getHost() {
 
     public void setHost(Host host) {
         this.host = host;
+        this.key.setHostId(host.getId());
```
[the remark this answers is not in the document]
dmckoske: Good point, will do.

dmckoske (commented):
I've made changes per your recommendations.  Please have a second look.

kahowell (approved):

PR #327, opened 2021-03-03
MON-1507: Update promql query for OpenShift

mstead:
This PR addresses a few things:
* Updated the PromQL query to closer represent the core hours metrics.
* Addresses some query encoding issues that were uncovered when the query was updated.
* Because we are taking the average over an hour, we need to set the event
date to the start of the hour that the query is averaging over. For
example, the query might return a metric at time of 3pm, but the value
is actually the average b/w 2pm and 3pm. We set the event timestamp to
the start 2pm and have it expire at 3pm. We adjust this by the configured
step of the query.

**HOW TO TEST**
The steps on #322 is a good way to test this.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/metering/MeteringEventFactory.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/MeteringEventFactory.java
@@ -61,7 +61,7 @@ public class MeteringEventFactory {
      * @return a populated Event instance.
      */
     public static Event openShiftClusterCores(String accountNumber, String clusterId, String serviceLevel,
-        String usage, OffsetDateTime measuredTime, Double measuredValue) {
+        String usage, OffsetDateTime measuredTime, OffsetDateTime expired, Double measuredValue) {
         return new Event()
             .withEventSource(OPENSHIFT_CLUSTER_EVENT_SOURCE)
             .withEventType(OPENSHIFT_CLUSTER_EVENT_TYPE)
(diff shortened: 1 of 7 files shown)


ntkathole:
@mstead I see some results difference in old query and new query. Is this difference expected?

Steps I followed:

1. Initially had 2 clusters in accounts. (Time - 2021-03-04 09:03:00)
2. Changed 0e94471b-f9bb-4b30-8d6e-8ad6c3539392 cluster's subscription settings (from premium -> standard and production -> Dev/Test) in OCM. (Time - 2021-03-04 09:05:33)
3. Observed 3 clusters in telemetry with new query from 2021-03-04 09:05:33 to 2021-03-04 10:01:07 while old query always shows 2 clusters

Telemetry result:

Screenshot from 2021-03-04 15-36-38
Screenshot from 2021-03-04 14-45-37

kahowell:
@ntkathole, yeah it's expected. Whenever the parameter is changed, we'll have two conflicting views for that hour: one for the old SLA/usage settings and one for the new SLA/usage settings (we have to pick just one to represent the hour). #322 should ensure that we only create a single record in the DB. Which one is somewhat arbitrary (but I think should be deterministic?). Future work might prioritize specific values based on PM input (or future work might find a way to split up the hour in a smarter way).

ntkathole:
> Whenever the parameter is changed, we'll have two conflicting views for that hour: one for the old SLA/usage settings and one for the new SLA/usage settings (we have to pick just one to represent the hour). #322 should ensure that we only create a single record in the DB. Which one is somewhat arbitrary (but I think should be deterministic?). Future work might prioritize specific values based on PM input (or future work might find a way to split up the hour in a smarter way

Thank you. It makes it clear :+1:

kahowell (approved):

PR #329, opened 2021-03-04
Awood/marketplace id provider

awood:


src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
```diff
@@ -49,4 +50,12 @@
     @Query("SELECT s FROM Subscription s where s.endDate > CURRENT_TIMESTAMP " +
         "AND s.subscriptionId = :subscriptionId")
     Optional<Subscription> findActiveSubscription(@Param("subscriptionId") String subscriptionId);
+
+    @Query("SELECT s FROM Subscription s WHERE s.ownerId = :ownerId AND " +
```
kahowell, line 54: Actually we'll want to fetch by account

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdCollector.java
```diff
@@ -0,0 +1,45 @@
...
+    public static final String DUMMY = "DUMMY";
+    private final MarketplaceApi marketplaceApi;
+
+    @Autowired
+    public MarketplaceSubscriptionIdCollector(MarketplaceApi marketplaceApi) {
+        this.marketplaceApi = marketplaceApi;
+    }
+
+    public String fetchSubscriptionId(String orgId, Key usageKey) {
+        return DUMMY;
```
kahowell, line 43: @awood, can you read from an env var for short term? Even something with `System.getenv()` would be fine, since we don't plan to do long term (though I'm fine if you want to go ahead and place in our config file per norm).

src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
```diff
@@ -49,4 +50,12 @@
...
         "AND s.subscriptionId = :subscriptionId")
     Optional<Subscription> findActiveSubscription(@Param("subscriptionId") String subscriptionId);
+
+    @Query("SELECT s FROM Subscription s WHERE s.accountNumber = :accountNumber AND " +
+        "s.sku = ALL (SELECT DISTINCT o.sku FROM Offering o WHERE " +
+        ":#{#key.usage} = o.usage AND " +
+        ":#{#key.sla} = o.serviceLevel AND " +
+        ":#{T(java.lang.Integer).parseInt(#key.productId)} MEMBER OF o.productIds)")
+    Optional<Subscription> findSubscriptionByAccountAndUsageKey(@Param("accountNumber") String accountNumber,
+        @Param("key") UsageCalculation.Key usageKey);
```
kahowell, line 60: I think we should find all matching subscriptions and then pick one in the logic. Example use case: Marketplace subscription is purchased for 1 year, then another is bought the next year. We need to select the appropriate one for that timeframe. Also in the future we may need to support multiple active, so may as well fetch all matches for now.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
```diff
@@ -0,0 +1,77 @@
...
+            if (!StringUtils.hasText(s.getMarketplaceSubscriptionId())) {
+                String missingId = collector.fetchSubscriptionId(accountNumber, usageKey);
+                s.setMarketplaceSubscriptionId(missingId);
+                subscriptionRepo.save(s);
+            }
+            return Optional.of(s.getMarketplaceSubscriptionId());
+        }
+
+        return Optional.empty();
+    }
```
kahowell, line 76: I think this logic needs to be reworked slightly. Instead of applying missing marketplace subscription ID information to an existing subscription record, it should fetch a whole subscription record from the marketplace subscription ID collector (I guess we should rename to MarketplaceSubscriptionCollector). 

If any matches already exist in our DB:
 - pick the first one.
 - log a warning if more than one exists.

else:

 - find matches via the MarketplaceSubscriptionCollector
 - save them to the DB
 - pick the first
 - log a warning if more than one match exists

It also occurs to me now that this method should accept a start date and end date and should filter out subscriptions that aren't valid for the passed range.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
awood: Okay, I've updated this, but it does change the semantics of the code some (discussed on IRC but recording here for posterity).  Originally, the code assumed there are Subscriptions that might be missing marketplace subscription IDs and updated those subscriptions with missing IDs by fetching via MarketplaceSubscriptionIdCollector.  Now, the code looks for subscriptions and if it finds them, **it assumes the marketplace ID is present**.  If no matches are present, subscriptions are fetched via MarketplaceSubscriptionCollector (which ultimately is going to make a call to the Marketplace API), the results are persisted into the subscriptions table, and then we query the database again.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Class responsible for communicating with the Marketplace API and fetching the subscription ID.
+ */
+@Component
+public class MarketplaceSubscriptionCollector {
+    public static final String DUMMY = "DUMMY";
```
kahowell, line 42: Vestigial constant?

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
```diff
@@ -0,0 +1,91 @@
...
+            subscriptions.forEach(syncController::syncSubscription);
+            result = subscriptionRepo.findSubscriptionByAccountAndUsageKey(accountNumber, usageKey);
+            result = filterByDateRange(result, rangeStart, rangeEnd);
+        }
+
+        if (result.size() > 1) {
+            log.warn("Multiple subscriptions found for account {} with key {}. Selecting first result",
+                accountNumber, usageKey);
+        }
+        return result.get(0).getMarketplaceSubscriptionId();
```
kahowell, line 82: Need to still handle the case that `result.isEmpty()`

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceProperties.java
```diff
@@ -39,6 +39,12 @@
      */
     private String apiKey;
 
+    /**
+     * TEMPORARY setting for returning a dummy subscription ID. Remove once
+     * MarketplaceSubscriptionCollector is fully implemented.
+     */
+    private String dummyId = "DUMMY";
```
kahowell, line 46: Can you please alias to the env var `MARKETPLACE_DUMMY_ID` in `src/main/resources/application-marketplace.yaml`?

kahowell (changes requested):
I'm going to do some smoketesting with this merged into the worker from #347, may have a little more additional feedback after that.

src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
```diff
@@ -49,4 +50,12 @@
     @Query("SELECT s FROM Subscription s where s.endDate > CURRENT_TIMESTAMP " +
         "AND s.subscriptionId = :subscriptionId")
     Optional<Subscription> findActiveSubscription(@Param("subscriptionId") String subscriptionId);
+
+    @Query("SELECT s FROM Subscription s WHERE s.accountNumber = :accountNumber AND " +
+        "s.sku = ALL (SELECT DISTINCT o.sku FROM Offering o WHERE " +
+        ":#{#key.usage} = o.usage AND " +
+        ":#{#key.sla} = o.serviceLevel AND " +
+        ":#{T(java.lang.Integer).parseInt(#key.productId)} MEMBER OF o.productIds)")
```
kahowell, line 58: @awood FYI I am realizing on doing some smoketesting that the productIds here are different.... key.productId is a swatchProductId, while Offering.productId is an engineering productId... We'll have to reconcile this somehow... will discuss internally.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
```diff
@@ -0,0 +1,107 @@
...
+        Assert.isTrue(Usage._ANY != usageKey.getUsage(), "Usage cannot be _ANY");
+        Assert.isTrue(ServiceLevel._ANY != usageKey.getSla(), "Service Level cannot be _ANY");
+
+        String productId = usageKey.getProductId();
+        ProductProfile profile = profileRegistry.findProfileForSwatchProductId(productId);
+        Set<String> roles = profile.getRolesBySwatchProduct().getOrDefault(productId, Collections.emptySet());
+
+        List<Subscription> result =
+            subscriptionRepo.findSubscriptionByAccountAndUsageKey(accountNumber, usageKey, roles);
+        result = filterByDateRange(result, rangeStart, rangeEnd);
```
kahowell, line 79: @awood, I was just thinking... if we do one more filter here: filter out any subs without a marketplace subscription ID, it'll probably be a bit more robust. I'm thinking we might encounter a situation where there are some very similar looking non-marketplace subs we'd want to filter out.

awood: Updated.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
```diff
@@ -0,0 +1,65 @@
...
+    public List<Subscription> fetchSubscription(String orgId, Key usageKey) {
+        var s = new Subscription();
+        var ref = new ExternalReference();
+        ref.setSubscriptionID(properties.getDummyId());
+
+        Map<String, ExternalReference> refMap = new HashMap<>();
+        refMap.put(SubscriptionDtoUtil.MARKETPLACE, ref);
+        s.setExternalReferences(refMap);
+
+        return Collections.singletonList(s);
```
kahowell, line 63:
```suggestion
        return Collections.emptyList();
```
Otherwise we'll populate some data we'd need to remove later. FWIW, if we did populate data here there are a bunch of fields we'd need to populate (quantity, startDate, endDate, etc).

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
```diff
@@ -0,0 +1,120 @@
...
+    private List<Subscription> filterByMissingMarketplaceSubscriptionId(List<Subscription> result) {
+        return result.stream()
+            .filter(x -> StringUtils.hasText(x.getMarketplaceSubscriptionId()))
+            .collect(Collectors.toList());
+    }
+
+    private List<Subscription> filterByDateRange(List<Subscription> result, OffsetDateTime rangeStart,
+        OffsetDateTime rangeEnd) {
+        return result.stream()
+            .filter(x -> rangeStart.isBefore(x.getStartDate()) && rangeEnd.isAfter(x.getEndDate()))
```
kahowell, line 117:
```suggestion
        return result.stream().filter(
            x -> (rangeStart.isEqual(x.getStartDate()) || rangeStart.isAfter(x.getStartDate())) &&
            (rangeEnd.isEqual(x.getEndDate()) || rangeEnd.isBefore(x.getEndDate())))
```

We need to handle the edges of the ranges appropriately... Crude ascii:
```
1. snapshot range neatly inside:
| subscription      |
  | snapshot range |

rangeStart.isAfter(x.startDate) == true
rangeEnd.isBefore(x.endDate) == true

2. snapshot range "just fits"...
| subscription       |
| snapshot range     |

x.startDate == rangeStart
x.endDate == rangeEnd
```
(Other permutations left as an exercise for the reader).

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
```diff
@@ -0,0 +1,120 @@
...
+        if (result.isEmpty()) {
+            /* If we are missing the subscription, call out to the MarketplaceSubscriptionCollector
+               to fetch from Marketplace.  Sync all those subscriptions. Query again. */
+            var subscriptions = collector.fetchSubscription(accountNumber, usageKey);
+            subscriptions.forEach(syncController::syncSubscription);
+            result = fetchSubscriptions(accountNumber, usageKey, roles, rangeStart, rangeEnd);
+        }
+
+        if (result.isEmpty()) {
+            return Optional.empty();
```
kahowell, line 89:
```suggestion
            return Optional.of(properties.getDummyId());
```
Here is where I would like to see us return the dummy ID instead of populating a dummy subscription.

kahowell (changes requested):

kahowell:
I've been using some mocked data for easy smoke testing:
```sql
insert into offering(sku, role, sla, usage) values ('SKU01234', 'osd', 'Premium', 'Production');

insert into subscription(account_number, sku, subscription_id, quantity, start_date, end_date, marketplace_subscription_id)
values ('account123', 'SKU01234', 'SUB01234', 1, '2021-01-01', '2021-01-02', 'marketplace-1');

insert into subscription(account_number, sku, subscription_id, quantity, start_date, end_date, marketplace_subscription_id)
values ('account123', 'not-existent', 'SUB-none', 1, '2021-01-02', '2021-01-04', 'marketplace-1');
```

and I threw this method locally into a JMX bean for easy smoketesting:

```java
    @ManagedOperation(description = "Find marketplace subscription ID.")
    public String findMarketplaceSubscriptionId(String accountNumber, String productId, String sla, String usage, String begin,
        String end) {
        UsageCalculation.Key usageKey = new UsageCalculation.Key(productId, ServiceLevel.fromString(sla),
            Usage.fromString(usage));
        try {
            return provider.findSubscriptionId(accountNumber, usageKey, OffsetDateTime.parse(begin),
                OffsetDateTime.parse(end)).toString();
        }
        catch (Exception e) {
            log.info("uh-oh", e);
            throw e;
        }
    }
```

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
awood: Yeah, that was an artifact of the earlier approach where we were just fetching IDs instead of entire subscription objects.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
```diff
@@ -0,0 +1,136 @@
...
+        SubscriptionRepository subscriptionRepo, SubscriptionSyncController syncController,
+        ProductProfileRegistry profileRegistry, MarketplaceProperties properties,
+        MeterRegistry meterRegistry) {
+        this.collector = collector;
+        this.subscriptionRepo = subscriptionRepo;
+        this.syncController = syncController;
+        this.profileRegistry = profileRegistry;
+        this.properties = properties;
+        this.missingSubscriptionCounter = meterRegistry.counter("rhsm-conduit.missing.subscription");
+        this.ambiguousSubscriptionCounter = meterRegistry.counter("rhsm-conduit.ambiguous.subscription");
```
kahowell, line 76: For clarity:
```suggestion
        this.missingSubscriptionCounter = meterRegistry.counter("rhsm-subscriptions.marketplace.missing.subscription");
        this.ambiguousSubscriptionCounter = meterRegistry.counter("rhsm-subscriptions.marketplace.ambiguous.subscription");
```

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
```diff
@@ -0,0 +1,136 @@
...
+            .filter(x -> StringUtils.hasText(x.getMarketplaceSubscriptionId()))
+            .collect(Collectors.toList());
+    }
+
+    private List<Subscription> filterByDateRange(List<Subscription> result, OffsetDateTime rangeStart,
+        OffsetDateTime rangeEnd) {
+        return result.stream()
+            // !isAfter => rangeStart is either equal to or before getStartDate
+            // !isBefore => rangeEnd is either equal to or after getEndDate
+            .filter(x -> !rangeStart.isAfter(x.getStartDate()) && !rangeEnd.isBefore(x.getEndDate()))
```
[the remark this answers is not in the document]
awood: Ah, thanks.  I got it backwards :facepalm:

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java
awood: double :man_facepalming:

kahowell:
@awood needs a rebase, and since #350 has landed, can you please go ahead and integrate the provider? Should just need to inject the ID provider into the payload mapper and remove `org.candlepin.subscriptions.marketplace.MarketplacePayloadMapper#findSubscriptionId` and replace it with a call to the ID provider's method.

kahowell (approved):

PR #357, opened 2021-03-22
ENT-2662: Add product service

redsaz:
[ENT-2662](https://issues.redhat.com/browse/ENT-2662)

This adds the bare minimum of the product services API to use in the creation of Offerings. This includes the endpoint for fetching the product tree when given a SKU, and the endpoint for fetching engineering products associated with the list of one or more SKUs.

### Testing

There are only unit tests for the ProductService, no integration testing has been performed, nor are there integration tests. The unit tests cover when the SKU is found and when it is not found.

`StubProductApi` is the dummy version of the API. Currently it does nothing but throw `ApiException` if used, which is fine as ProductService is not yet hooked in to anything. StubProductApi *can* read from the classpath (say, `test/java/src/org/candlepin/subscription/product) when needing to use real engprod or tree data for later testing.


dmckoske (approved):
I'm still getting up to speed on what everything is and how it works.  The code looks solid and I look forward to others comments.

src/main/java/org/candlepin/subscriptions/product/StubProductApi.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private <T> T readJsonResource(String resName, Class<T> clazz) throws ApiException {
+        try (InputStream res = StubProductApi.class.getResourceAsStream(resName)) {
+            return MAPPER.readValue(res, clazz);
+        }
+        catch (IOException e) {
+            throw new ApiException(e);
+        }
+
+    }
```
kahowell, line 77: Could you please:
- [ ] throw mocked examples into the codebase and document it (comment here is fine)?
- [ ] document (internally?) how to produce these files from real data; add a pointer in comments of where to look for this documentation.
- [ ] handle gracefully missing files (return empty/null object?)

src/main/resources/application-worker.yaml
```diff
@@ -24,3 +24,9 @@ rhsm-subscriptions:
     keystore: file:${SUBSCRIPTION_KEYSTORE:}
     keystorePassword: changeit
     maxConnections: ${SUBSCRIPTION_MAX_CONNECTIONS:100}
+  product:
+    useStub: ${STUB_PRODUCT:true}
+    url: http://${PRODUCT_HOST:localhost}:${PRODUCT_PORT:8080}/svcrest/product/v3
+    keystore: file:${PRODUCT_KEYSTORE:}
+    keystorePassword: changeit
```
kahowell, line 31: Can you please expose the keystore password as an environment variable?
```suggestion
    keystorePassword: ${PRODUCT_KEYSTORE_PASSWORD:changeit}
```

src/main/resources/application-worker.yaml
```diff
@@ -24,3 +24,9 @@ rhsm-subscriptions:
     keystore: file:${SUBSCRIPTION_KEYSTORE:}
     keystorePassword: changeit
     maxConnections: ${SUBSCRIPTION_MAX_CONNECTIONS:100}
+  product:
+    useStub: ${STUB_PRODUCT:true}
```
kahowell, line 28: Two nitpicks: Let's default this to false instead, and rename env var to `PRODUCT_USE_STUB` for consistency (realizing the subscription product could use these changes too... :-/ appreciate if you want to go ahead and tackle that here too, but also happy to just make a tech debt card for it.
```suggestion
    useStub: ${PRODUCT_USE_STUB:false}
```

src/main/resources/application-worker.yaml
```diff
@@ -24,3 +24,9 @@ rhsm-subscriptions:
...
+    url: http://${PRODUCT_HOST:localhost}:${PRODUCT_PORT:8080}/svcrest/product/v3
```
kahowell, line 29: Let's just go ahead and default to something more useful for us:
```suggestion
    url: https://${PRODUCT_HOST:product.qa.api.redhat.com}:${PRODUCT_PORT:443}/svcrest/product/v3
```
Notably, this is *https*, and the QA endpoint with TLS port.

kahowell (changes requested):
Overall, looks good, and I really dig the direction you've gone with the stubs, that could be quite useful :-) Left some feedback, and please address the sonarcloud issues: https://sonarcloud.io/project/issues?id=rhsm-subscriptions&pullRequest=357&resolved=false

src/main/java/org/candlepin/subscriptions/product/StubProductApi.java
redsaz: Can do! I originally had two products pulled from QA-env's product service, so I can put it back. As for the final bullet, I should make this stub product api behave like the real product service anyway, which will succeed and return a null object anyway, so I'll do that :)

redsaz: I've completed the changes:

- "Mocked" examples (these files are actually from real product data from QA, but are stripped to only the data we use from the service). These files are located in src/main/resources/org/candlepin/subscriptions/product/ so if you'd rather them be somewhere other than main, let me know.
- I've provided a document on how to produce these files, accessible internally at https://source.redhat.com/groups/public/teamnado/blog/grabbing_real_sku_tree_and_engprod_oid_data_for_swatch_stubproductapi
- Missing files will return a null object. This better aligns to what the real product service does for these calls anyway, so that's nice :)

src/main/resources/application-worker.yaml
redsaz: Done. I also applied this change to the subscription config above.

src/main/resources/application-worker.yaml
redsaz: I **did** change this to be `PRODUCT_USE_STUB` as requested, but I **did not** switch this to `false`. When I did, various tests failed since a keystore file is also needed to be used, which this doesn't provide.

src/main/resources/application-worker.yaml
redsaz: I thought the following would be even more useful, maybe?

```java
url: ${PRODUCT_URL:https://product.qa.api.redhat.com/svcrest/product/v3}
```

(this was also done for the subscription config)
If you disagree and there is more benefit to providing only the host and port, rather than the entire base URL, let me know and I'll use the provided suggestion verbatim.

src/main/resources/application-worker.yaml
kahowell: Ah, make sure that the test uses `@ActiveProfiles("test")` and then set `useStub` to `true` in `src/test/resources/application-test.yaml`

src/main/resources/application-worker.yaml
kahowell: :+1:

kahowell (changes requested):
Left a suggestion about `useStub` `false` by default. The only other tiny suggestion I have is rather than store the stubbed product responses in `org/candlepin/subscriptions/product` folder, simplify to `product-stub-data`.

kahowell (approved):

PR #361, opened 2021-03-23
Use fully qualified names for parameterized tests

awood:
The default test name for a parameterized test is just the index number
and the arguments which does not actually tell you which specific test
method is being run.  The reasoning behind this is that test runners are
expected to print the test method name and then underneath in a
tree-like fashion, print the individual instances of that test.  Our
test runner doesn't follow this behavior making it difficult to tell
what method generated a particular parameterized test.  This patch
corrects the issue by mandating a more verbose name.

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
+++ b/src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
@@ -21,6 +21,7 @@
 package org.candlepin.subscriptions.db;
 
 import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.params.ParameterizedTest.*;
 
 import org.candlepin.subscriptions.db.model.HardwareMeasurementType;
 import org.candlepin.subscriptions.db.model.Host;
@@ -473,7 +474,7 @@ class HostRepositoryTest {
     }
 
     @Transactional
-    @ParameterizedTest
+    @ParameterizedTest(name = DISPLAY_NAME_PLACEHOLDER + " " + DEFAULT_DISPLAY_NAME)
     @MethodSource("org.candlepin.subscriptions.db.HostRepositoryTest#instanceSortParams")
     void canSortByInstanceBasedSortMethods(String sort) {
         Pageable page = PageRequest.of(0, 1, Sort.by(sort));
@@ -486,7 +487,7 @@ class HostRepositoryTest {
     }
 
     @Transactional
-    @ParameterizedTest
+    @ParameterizedTest(name = DISPLAY_NAME_PLACEHOLDER + " " + DEFAULT_DISPLAY_NAME)
     @MethodSource("org.candlepin.subscriptions.db.HostRepositoryTest#tallyViewSortParams")
     void canSortByHostBasedSortMethods(String sort) {
         Pageable page = PageRequest.of(0, 1, Sort.by(sort));
@@ -595,7 +596,7 @@ class HostRepositoryTest {
     }
 
     @Transactional
-    @ParameterizedTest
+    @ParameterizedTest(name = DISPLAY_NAME_PLACEHOLDER + " " + DEFAULT_DISPLAY_NAME)
     @CsvSource({
         "'',3",
         "banana,1",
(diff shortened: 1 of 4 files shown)


mstead (changes requested):
I love the concept here, but something isn't lining up for me.

Testing your changes out on the CLI I get:
```
org.candlepin.subscriptions.db.HostRepositoryTest

  Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
  Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
  Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
  Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
  Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
  Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
```

Testing on develop I get:
```
org.candlepin.subscriptions.db.HostRepositoryTest

  Test [1] sort=host.displayName PASSED
  Test [2] sort=cores PASSED
  Test [3] sort=host.hardwareType PASSED
  Test [4] sort=sockets PASSED
  Test [5] sort=host.lastSeen PASSED
  Test [6] sort=measurementType PASSED
```

I think this PR needs something else. I feel like I want a mix of both.

```canSortByHostBasedSortMethods(String)  sort=host.displayName PASSED```

Can we achieve this somehow? Is there a way that we can make this easier to apply without having to remember the format or go copy-paste it?

**Comment by awood** [comment]
> I think this PR needs something else. I feel like I want a mix of both.

The reason I wrote this is because in the IDE, the current format is not helpful:

Selection_008

You can't even tell what the top level test is.  I didn't realize it printed the tree format correcting on the CLI.  I'll see about if I can split the difference.

**Comment by awood** [comment]
I am going to close this.  The problem is actually with Gradle!  It's being tracked at https://github.com/gradle/gradle/issues/5975  Essentially, running the unit tests through Gradle in IntelliJ lops off the portion of the test tree that tells you what method a parameterized test derived from.  Running the unit test through JUnit in IntelliJ and everything is fine.

So the quick work around is just to create a "Run All Tests" run configuration that uses the JUnit run configuration.  As you can see from the screenshot, the test tree is constructed and you can actually tell what unit test a failing parameterized test belongs too.  Easy enough hack for a fairly uncommon problem.  Hopefully Gradle will fix the issue soon anyway.

In the meantime, we could probably remove the names that are already applied to some of the parameterized tests.

Selection_009

**Comment by awood** [comment]
> Testing your changes out on the CLI I get:
> 
> ```
> org.candlepin.subscriptions.db.HostRepositoryTest
> 
>   Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
>   Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
>   Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
>   Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
>   Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
>   Test canSortByHostBasedSortMethods(String) REDHAT_PWNS PASSED
> ```

P.S. @mstead How did you invoke Gradle here?  When I run `./gradlew :test --tests "HostRepositoryTest"` my results look like

```
  Test testReturnsGuestsOfHypervisor() PASSED                                                                                                                    
  Test testFindHostsWhenUsageIsDifferent() PASSED                                                                                                                
  Test testShouldFilterSockets() PASSED                                                                                                                          
  Test testFindHostsWhenAccountIsDifferent() PASSED                                                                                                              
  Test [1] displayNameSubstring=, expectedResults=3 PASSED                                                                                                       
  Test [2] displayNameSubstring=banana, expectedResults=1 PASSED
  Test [3] displayNameSubstring=rang, expectedResults=1 PASSED
  Test [4] displayNameSubstring=an, expectedResults=2 PASSED
  Test [5] displayNameSubstring=celery, expectedResults=0 PASSED
...
```

and the actual name of the parameterized test, `testDisplayNameFilter` is nowhere to be found.  

It'd be nice to have it display properly on the CLI too. :rofl:

PR #363, opened 2021-03-24
Jenkins: Reduce number of old job runs to 10

kahowell:
In order to waste less disk space.


kahowell:
unneeded (for now at least)

PR #365, opened 2021-03-25
ENT-3684: Set DB pool size to 25 for workers

kahowell:
The health check fails when it can't acquire a DB connection in a sufficient time.

The number of threads (24) plus 1. This way, even if all the kafka threads are busy, the health check will still be able to acquire a connection.

Testing Hikari connection params
--------------------------------

Easiest thing is to set both the env vars and then check at http://localhost:8080/actuator/hawtio/jmx/operations?nid=root-com.zaxxer.hikari-HikariDataSource-rhsmSubscriptionsDataSource:

```
DATABASE_CONNECTION_TIMEOUT_MS=5000 DATABASE_MAX_POOL_SIZE=50 ./gradlew bootRun
```

Testing TLS support
-------------------

Because this change removes some code that was written in #95 to support TLS, we need to make sure we still have a way to use TLS...

1. Run the app as usual, note that the changes don't cause issues:

```
./gradlew bootRun
```

2. Try enabling TLS and observe that the app refuses to connect, since the server doesn't support TLS:

```
DATABASE_SSL_MODE=verify-full ./gradlew bootRun
```

3. Next, stop your existing DB instance (e.g. `podman pod stop rhsm-subscriptions`).

4. Create an SSL cert for postgres:

```
mkdir certs
chmod 777 certs  # permissions get confusing here!
podman run --rm -ti -v $(pwd)/certs:/certs:z docker.io/postgres /bin/bash -c 'gosu postgres openssl req -new -x509 -days 365 -nodes -text -out /certs/server.crt -keyout /certs/server.key -subj "/CN=db.example.com"'
```

5. Start postgres with the server cert configured.

```
podman run --rm -ti --network=host \
  -v $(pwd)/certs/server.crt:/server.crt:z \
  -v $(pwd)/certs/server.key:/server.key:z \
  -v $(pwd)/init_dbs.sh:/docker-entrypoint-initdb.d/init_dbs.sh:z \
  -e POSTGRES_HOST_AUTH_METHOD='trust' \
  docker.io/postgres \
  -c ssl=on \
  -c ssl_key_file=/server.key \
  -c ssl_cert_file=/server.crt
```

6. Try connecting without specifying the cert. Observe that the app fails:

```
DATABASE_SSL_MODE=verify-full ./gradlew bootRun
```

7. Connect while specifying both the cert and the mode. Observe the the connection fails because the cert isn't valid for hostname `localhost`.

```
DATABASE_SSL_CERT=certs/server.crt DATABASE_SSL_MODE=verify-full ./gradlew bootRun
```

8. Connect with verify mode set to `require` (does not do hostname checking), observe success:

```
DATABASE_SSL_CERT=certs/server.crt DATABASE_SSL_MODE=require ./gradlew bootRun
```

9. Now restore your environment (restart postgres; e.g. `podman pod start rhsm-subscriptions`)


mstead (approved):
Looks good and tests well.

PR #349, opened 2021-03-17
ENT-3635: Get OpenShift metering accounts from Prometheus

mstead:
## Details
[ENT-3635](https://issues.redhat.com/browse/ENT-3635)

When OpenShift metrics collection is triggered for all accounts
(via the job or jmx), the account list is pulled from Prometheus using the
subscription_labels metric, matching on the billing_model="marketplace"
label, and will be included in the list of accounts that has
OpenShift metering enabled.

When an OpenShift metering task is executed for an account, we first
make sure that it is opted in, and then gather the meterics.

## Testing
Currently there doesn't seem to be any accounts in subscription_labels that have a _billing_model_ label set to _marketplace_. As such, you will have to modify the PromQL so that it will actually match some of them.

First you'll need to set up access to a Prometheus server. See https://docs.engineering.redhat.com/display/ENT/Prometheus+Metrics for getting an auth token for thanos. I like to export it into a variable so my startup command remains the same.

Deploy the application as follows. We override the PromQL here so that we pull back accounts with billing_model  of standard.
```bash
DEV_MODE=true RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_ENABLED_ACCOUNT_PROM_Q_L="group(min_over_time(subscription_labels{ebs_account != '', billing_model='standard'}[1h])) by (ebs_account)" PROM_URL="http://localhost:8082/api/v1" ./gradlew clean bootRun --args="--spring.profiles.active=openshift-metering-worker"
```

Test pulling the metrics for ALL matching accounts. 
```bash
$ curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=openshiftJmxBean,type=OpenShiftJmxBean","operation":"performOpenshiftMetering()","arguments":[]}'
```

The logs should show something like the following as the accounts are getting processed.
```
--- SNIP ---
2021-03-17 14:21:46,931 [thread=pool-6-thread-11] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusMeteringController] - Ensuring marketplace account 5481602 has been set up for syncing/reporting.
2021-03-17 14:21:46,932 [thread=pool-6-thread-19] [INFO ] [org.candlepin.subscriptions.metering.task.OpenShiftMetricsTask] - Running Openshift metrics update for account: 6459587
[...]

PR #367, opened 2021-03-25
Tweak cronjob for queuing metrics gathering

kahowell:
1. Apply token-refresher sidecar to metrics cronjob
    
Follow-on to #349.
    
To test, run the CI deploy job against this branch (only the scheduler needs to be deployed), and then tweak the metrics cronjob schedule to the near-future (hint: `oc edit CronJob/openshift-metrics-cron`).

Then verify via logs that the tasks were queued up as expected.

2. Add `OPENSHIFT_ENABLED_ACCOUNT_PROMQL` to make account promql easier to use; override to use `standard` billing_model for now.

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -105,8 +105,8 @@ rhsm-subscriptions:
       metric:
         openshift:
           enabledAccountPromQL: >-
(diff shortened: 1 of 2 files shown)


mstead (approved):
Looks good. I've deployed all this on CI and it runs fine.

PR #368, opened 2021-03-25
work around generator encoding reserve characters by not making them …

jdhrh:
…parameters

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/subscription/StubSearchApi.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/StubSearchApi.java
@@ -32,7 +32,19 @@ import java.util.List;
 public class StubSearchApi extends SearchApi {
 
     @Override
-    public List<Subscription> searchSubscriptions(String criteria, String options) throws ApiException {
+    public Subscription getSubscriptionById(String id) throws ApiException {
+        return createData();
+    }
+
+    @Override
+    public List<Subscription> searchSubscriptionsByAccountNumber(String accountNumber, Integer index,
+        Integer pageSize) throws ApiException {
+        return Collections.singletonList(createData());
+    }
+
+    @Override
+    public List<Subscription> searchSubscriptionsByOrgId(String orgId, Integer index, Integer pageSize)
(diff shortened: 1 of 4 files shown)


mstead:
Not sure if this is the same sort of issue, but we had encoding issues with the PromQL query parameter for the generated Prometheus client. I created an API wrapper class that did some encoding before calling the API. I was gonna take a similar approach as you did here, but I wasn't a fan of having to bake stuff into the schema file.

If you are interested, here's the wrapper.

https://github.com/RedHatInsights/rhsm-subscriptions/blob/develop/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusService.java

jdhrh:
@mstead unfortunately I cannot do such a thing here. The issue is not that I need to escape but the opposite. The subscription service uses an obscure web technology which requires `;` and `=` to not be encoded. The generated code will indiscriminately encode any argument passed in like so:
```java
String localVarPath = "/search/criteria;web_customer_id={orgId}/options;products=ALL;isCustomer=false;firstResultIndex={index};maxResults={pageSize}".replaceAll("\\{format\\}","json")
      .replaceAll("\\{" + "orgId" + "\\}", apiClient.escapeString(orgId.toString()));
```
if we were to pass the criteria as a path parameter rather than these smaller, more specific details, we would end up with something like 
`criteria;web_customer_id%3D123/options;products%3DALL%3BisCustomer%3Dfalse%3BfirstResultIndex%3D0%3BmaxResults%3D10` which will throw an error parsing within the subscription service. Unless there are control characters which encodes to `;` and `=`, we cannot rely on the generated apis to function as desired as PathSegment is not within the openapi standard of serialization.

mstead:
> e will indiscriminately encode any argument passed in like so:

Ha! That's what I missing. Brutal! :)

mstead:
Fix up the conflicts and I'll give it another look over.

mstead (approved):
Looks good to me. Unfortunate that you had to go this route. 

Waiting on integration tests before merge.

PR #350, opened 2021-03-17
ENT-3264: Marketplace Payload Mapper

lindseyburnett:
# Prereq
* Run with `marketplace` profile enabled
* Use debug log level

# Test option 1
* Invoke the `submitTallySummary` JMX endpoint with this payload
(org.candlepin.subscriptons.marketplace -> MarketplaceJmxBean -> Operations -> submitTallySummary(String))

```javascript
{
    "account_number": "account123",
    "tally_snapshots": [
        {
            "id": "c204074d-626f-4272-aa05-b6d69d6de16a",
            "snapshot_date": 1615406400,
            "product_id": "OpenShift-metrics",
            "usage": "Production",
            "sla": "Premium",
            "granularity": "Hourly",
            "tally_measurements": [
                {
                    "hardware_measurement_type": "TOTAL",
                    "uom": "Cores",
                    "value": 36
                },
                {
                    "hardware_measurement_type": "PHYSICAL",
                    "uom": "Cores",
                    "value": 36
                }
            ]
        }
    ]
}
```

# Test option 2
ORRRRR

If you have TallySummary messages in a kafka topic, the MarketplaceWorker will pick them up now.

# Results
You should see a debug log statement with the UsageEvent payload
```
2021-03-19 16:09:03,168 [thread=marketplace-worker-0-C-1] [DEBUG ] [org.candlepin.subscriptions.marketplace.MarketplacePayloadMapper] - UsageRequest class UsageRequest {
    data: [class UsageEvent {
        start: 1615402800
        end: 1615406400
        subscriptionId: bananas2c017c5f-f1b1-4f3b-8e30-b2137f9eaffe
        eventId: 8779f11c-481c-4cee-86f8-0fee0ab19bee
        additionalAttributes: null
        measuredUsage: [class UsageMeasurement {
            metricId: redhat.com:openshiftdedicated:cpu_hour
            value: 12.0
        }]
    }]
}

```
# Expected errors

An ugly stacktrace about `RESTEASY003720: path param MARKETPLACE_URL` is expected, unless you have it set in `application-marketplace.yml` or have the environment variable set.

```
[...]

PR #369, opened 2021-03-25
ENT-3341, ENT-3744: Implement MarketplaceSubscriptionIdCollector

lindseyburnett:
# Testing
* Make sure there are offerings in the offering table that matches our SKUs of interest.  See JIRA ticket comments for insert statements for testing
* Find an account to use that has subscriptions (https://webadmin.corp.qa.redhat.com/subscription-admin).  See JIRA ticket comments for account numbers I used
* Get OCP or OSD events into your table for the associated account number
* Make sure the account_config table has your account opted in
* Kick off an hourly tally for a time frame that encompasses the events in your event table

Assuming that your subscriptions table was empty to start with, after this logic executes you will see an entry in the rhsm-subscription.subscription table that matches a subscription fetched from the subscription service
```bash
curl --cert-type P12 --cert nonprod-insights.jks:"$(cat jks_pass)" 'https://subscription.qa.api.redhat.com/svcrest/subscription/v5/search/criteria;oracle_account_number=account123/options;products=ALL;showExternalReferences=true'
```

If you were to execute this again, without clearing your subscription table, the logic shouldn't execute a GET request to the subscription service this time around, because that data is in the swatch database now.

# app properties tips

```yaml
# these settings currently reside in the application-worker.yaml file
rhsm-subscriptions:
  subscription:
    useStub: false
    url: https://subscription.qa.api.redhat.com/svcrest/subscription/v5
    keystore: nonprod-insights.jks
    keystorePassword:
# turns on logging for HTTP events to help verify when the subscription service is/isn't getting called
logging:
  level:
    org:
      apache:
        http: DEBUG
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
+++ b/src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
@@ -27,6 +27,7 @@ import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
 
+import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
@@ -53,10 +54,12 @@ public interface SubscriptionRepository extends JpaRepository<Subscription,
     Optional<Subscription> findActiveSubscription(@Param("subscriptionId") String subscriptionId);
 
     @Query("SELECT s FROM Subscription s WHERE s.accountNumber = :accountNumber AND " +
-        "s.sku = ALL (SELECT DISTINCT o.sku FROM Offering o WHERE " +
-        ":#{#key.usage} = o.usage AND " +
+        "s.sku = ALL (SELECT DISTINCT o.sku FROM Offering o WHERE " + ":#{#key.usage} = o.usage AND " +
         ":#{#key.sla} = o.serviceLevel AND " +
-        "o.role IN :#{#roles})")
-    List<Subscription> findSubscriptionByAccountAndUsageKey(@Param("accountNumber") String accountNumber,
-        @Param("key") UsageCalculation.Key usageKey, @Param("roles") Set<String> roles);
+        "o.role IN :#{#roles}) AND s.startDate <= :rangeStart AND s.endDate >= :rangeEnd AND " +
+        "s.marketplaceSubscriptionId IS NOT NULL AND s.marketplaceSubscriptionId <> ''")
+    List<Subscription> findSubscriptionByAccountAndUsageKeyAndStartDateAndEndDateAndMarketplaceSubscriptionId(
+        @Param("accountNumber") String accountNumber, @Param("key") UsageCalculation.Key usageKey,
+        @Param("roles") Set<String> roles, @Param("rangeStart") OffsetDateTime rangeStart,
+        @Param("rangeEnd") OffsetDateTime rangeEnd);
 }
--- a/src/main/java/org/candlepin/subscriptions/marketplace/MarketplacePayloadMapper.java
+++ b/src/main/java/org/candlepin/subscriptions/marketplace/MarketplacePayloadMapper.java
@@ -23,6 +23,7 @@ package org.candlepin.subscriptions.marketplace;
 import org.candlepin.subscriptions.db.model.HardwareMeasurementType;
 import org.candlepin.subscriptions.db.model.ServiceLevel;
 import org.candlepin.subscriptions.db.model.Usage;
+import org.candlepin.subscriptions.exception.ErrorCode;
 import org.candlepin.subscriptions.files.ProductProfileRegistry;
 import org.candlepin.subscriptions.json.TallySnapshot;
 import org.candlepin.subscriptions.json.TallySummary;
@@ -129,16 +130,16 @@ public class MarketplacePayloadMapper {
             .filter(this::isSnapshotPAYGEligible).collect(Collectors.toList());
 
         List<UsageEvent> events = new ArrayList<>();
-
         for (TallySnapshot snapshot : eligibleSnapshots) {
             String productId = snapshot.getProductId();
-            OffsetDateTime snapshotDate = snapshot.getSnapshotDate();
-            String eventId = snapshot.getId().toString();
 
             // call MarketplaceIdProvider.findSubscriptionId once available
             UsageCalculation.Key usageKey = new UsageCalculation.Key(productId,
                 ServiceLevel.fromString(snapshot.getSla().toString()),
                 Usage.fromString(snapshot.getUsage().toString()));
+
+            OffsetDateTime snapshotDate = snapshot.getSnapshotDate();
+            String eventId = snapshot.getId().toString();
 
             /*
             This will need to be updated if we expand the criteria defined in the
@@ -152,13 +153,8 @@ public class MarketplacePayloadMapper {
             var subscriptionIdOpt = idProvider.findSubscriptionId(
                 tallySummary.getAccountNumber(), usageKey, startDate, snapshotDate);
 
-            /*
-            The //NOSONAR flag is Suppressing warning java:S2583 - Conditionally executed code should be
-            reachable.  Since findSubscriptionId is stubbed out right now, it's never returning Optional
-            .empty(), so this if statemant can't be evaluated to true, which is making sonar unhappy.
-             */
-            if (subscriptionIdOpt.isEmpty()) { //NOSONAR
-                log.error("Couldn't find subscription id. Error code here eventually?");
+            if (subscriptionIdOpt.isEmpty()) {
+                log.error("{}", ErrorCode.SUBSCRIPTION_SERVICE_MARKETPLACE_ID_LOOKUP_ERROR);
                 continue;
             }
 
--- a/src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
@@ -20,32 +20,45 @@
  */
 package org.candlepin.subscriptions.marketplace;
 
-import org.candlepin.subscriptions.marketplace.api.resources.MarketplaceApi;
+import org.candlepin.subscriptions.subscription.SubscriptionService;
 import org.candlepin.subscriptions.subscription.api.model.Subscription;
-import org.candlepin.subscriptions.tally.UsageCalculation.Key;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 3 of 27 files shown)


src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
```diff
@@ -20,32 +20,57 @@
...
+     */
+    public List<Subscription> requestSubscriptions(String accountNumber) {
+        try {
+            var subscriptions = subscriptionService.getSubscriptionsByAccountNumber(accountNumber);
+            return filterNonApplicableSubscriptions(subscriptions);
+        }
+        catch (ApiException e) {
+            log.error("Exception occurred when making GET request to subscription service via account " +
+                "number", e);
+            return Collections.emptyList();
```
kahowell, line 66: I'd rather throw an exception here.

kahowell (changes requested):
A few small-ish requests:

- [ ] please go ahead and move the subscription properties to application.yaml since they'll be used in more than one place.
- [ ] subscription url will need `https` rather than http
- [ ] we should go ahead and name the stub property `SUBSCRIPTION_USE_STUB` for consistency
- [ ] we need template updates for the SUBSCRIPTION_HOST, SUBSCRIPTION_PORT properties as well as the keystore and truststore properties (I prefer re-using `RHSM_KEYSTORE`/`RHSM_KEYSTORE_PASSWORD` for this). Good references would be the way these are handled in `templates/rhsm-conduit.yaml`... Looking at this, I'm realizing we might need to do a small bit of work to get the proper truststore set up. WIll also be needed in #389, so I will swarm this with you.
- [ ] we should remove the DUMMY ID stuff (or at least make it a dev-only kind of thing)

And a quick note for anyone else reviewing, this requires data in the offering table to test:

```sql
insert into offering(sku, product_name, role, sla, usage) values ('MW00210HR', 'OpenShift Container Platform', 'ocp', 'Premium', 'Production');
insert into offering(sku, product_name, role, sla, usage) values ('MW01370', 'OpenShift Dedicated', 'osd', 'Premium', 'Production');
```
FWIW, I'm using instructions for #350 otherwise to test along with

```
DEV_MODE=true \
  RHSM_KEYSTORE=$RHSM_KEYSTORE \
  RHSM_KEYSTORE_PASSWORD=$RHSM_KEYSTORE_PASSWORD \
  RHSM_TRUSTSTORE=$RHSM_TRUSTSTORE \
  RHSM_TRUSTSTORE_PASSWORD=$RHSM_TRUSTSTORE_PASSWORD \
  SUBSCRIPTION_HOST=subscription.qa.api.redhat.com \
  SUBSCRIPTION_PORT=443 \
  SUBSCRIPTION_USE_STUB=false \
  ./gradlew bootRun
```

May have a touch more feedback tomorrow, but wanted to go ahead and get the bulk of it to you.

lindseyburnett:
@kahowell, pushed some updates.  It addresses your feedback **except** for

- truststore changes
- dummy id stuff

Also, I added a retryTemplate to the SubscriptionService if you could check that out too.

Thanks!

dmckoske (changes requested):

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
```diff
@@ -20,32 +20,49 @@
...
 
 /**
  * Class responsible for communicating with the Marketplace API and fetching the subscription ID.
  */
 @Component
 public class MarketplaceSubscriptionCollector {
-    @SuppressWarnings("java:S1068") // Unused field; Remove after implementing
-    private final MarketplaceApi marketplaceApi;
+
+    private static final String IBMMARKETPLACE = "ibmmarketplace";
```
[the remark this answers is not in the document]
kahowell: Having it as a constant here is fine. Let's avoid the anti-pattern of a separate constants file (e.g. see https://www.baeldung.com/java-constants-good-practices#2-a-large-global-constants-class for rationale on why that's a bad practice)

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionService.java
```diff
@@ -50,16 +54,48 @@ public Subscription getSubscriptionById(String id) throws ApiException {
...
+        // consistency?
+        Set<Subscription> total = new HashSet<>();
+        do {
+            List<Subscription> subscriptionsByAccountNumber = getSubscriptionsByAccountNumber(accountNumber,
+                index, pageSize);
+            latestResultCount = subscriptionsByAccountNumber.size();
+            total.addAll(subscriptionsByAccountNumber);
+            index = index + pageSize;
+        } while (latestResultCount != 0);
+
```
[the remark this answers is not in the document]
kahowell: The way the service functions is that if you ask for a page that doesn't exist, you get an empty list, so when result count is 0, you're done. Actually, we could avoid some unneeded calls by changing the condition (we also know we're done if the list we got back isn't our page size):

```java
        } while (latestResultCount == pageSize);
```

src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
```diff
@@ -59,4 +60,15 @@
         "o.role IN :#{#roles})")
     List<Subscription> findSubscriptionByAccountAndUsageKey(@Param("accountNumber") String accountNumber,
         @Param("key") UsageCalculation.Key usageKey, @Param("roles") Set<String> roles);
+
+    @Query("SELECT s FROM Subscription s WHERE s.accountNumber = :accountNumber AND " +
+        "s.sku = ALL (SELECT DISTINCT o.sku FROM Offering o WHERE " +
+        ":#{#key.usage} = o.usage AND " +
+        ":#{#key.sla} = o.serviceLevel AND " +
+        "o.role IN :#{#roles}) AND s.startDate > :rangeStart AND s.endDate < :rangeEnd AND " +
```
kahowell, line 68:
```suggestion
        "o.role IN :#{#roles}) AND s.startDate <= :rangeStart AND s.endDate >= :rangeEnd AND " +
```
Subscription needs to start before the usage can be reported against, and it must end after the usage period end.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionServiceProperties.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.Setter;
+import lombok.ToString;
+
+/**
+ * Additional properties related to the Subscription Service
+ */
+@Getter
+@Setter
+@ToString
+public class SubscriptionServiceProperties {
```
kahowell, line 33: Can you please extend org.candlepin.subscriptions.http.HttpClientProperties ? Then you can just use without a qualifier in org.candlepin.subscriptions.subscription.SubscriptionServiceConfiguration and intellij won't get confused about whether rhsm-subscriptions.subscription refers to SubscriptionServiceProperties or HttpClientProperties (cause it'll naturally be both :smile:).

IOW, after you do this, we won't need to have any mention of HttpClientProperties in SubscriptionServiceConfiguration.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplacePayloadMapper.java
```diff
@@ -128,18 +128,26 @@ protected boolean isSnapshotPAYGEligible(TallySnapshot snapshot) {
...
             .filter(this::isSnapshotPAYGEligible).collect(Collectors.toList());
 
-        List<UsageEvent> events = new ArrayList<>();
+        // Abandon ship if the usageKey isn't for one that's relevant to us, otherwise we get a whole
+        // bunch of missing subscription spam
+        List<TallySnapshot> premiumProductionSnapshots = eligibleSnapshots.stream().filter(x -> {
+            ServiceLevel sla = ServiceLevel.fromString(x.getSla().toString());
+            Usage usage = Usage.fromString(x.getUsage().toString());
+            return Objects.equals(ServiceLevel.PREMIUM, sla) && Objects.equals(Usage.PRODUCTION, usage);
+        }).collect(Collectors.toList());
```
kahowell, line 137: Let's not do this. When all the other bits are sorted out (mainly the billing_type bits) we shouldn't be receiving data for non-Premium or non-Production usage, and if something is wrong and we do receive that data, it'd be best to be alerted about it. Additionally, future requirements may introduce non-Premium or non-Production PAYG.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionServiceProperties.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Number of times we should try requesting info from the Subscription Service if something fails.
+     */
+    private int maxAttempts = 4;
+
+    /**
+     * The initial sleep interval between retries when retrying fetching info from the Subscription Service
+     */
+    private long backOffInitialInterval = 1000L;
```
kahowell, line 43: Can you please define these two in the application.yaml? Also, I prefer using `Duration` for the backOffInitialInterval.

src/main/resources/application.yaml
```diff
@@ -114,3 +114,9 @@ rhsm-subscriptions:
   prometheus-latency-duration: ${PROMETHEUS_LATENCY_DURATION:4h}
   hourly-tally-offset: ${HOURLY_TALLY_OFFSET:60m}
   metric-lookup-range-duration: ${METRIC_LOOKUP_RANGE:24h}
+  subscription:
+    useStub: ${SUBSCRIPTION_USE_STUB:true}
```
kahowell, line 118: Can you please default this to false (for consistency)?

src/main/resources/application.yaml
```diff
@@ -114,3 +114,9 @@ rhsm-subscriptions:
...
+    url: ${SUBSCRIPTION_URL:https://subscription.qa.api.redhat.com/svcrest/subscription/v5}
+    keystore: file:${SUBSCRIPTION_KEYSTORE:}
+    keystorePassword: ${SUBSCRIPTION_KEYSTORE_PASSWORD:changeit}
+    maxConnections: ${SUBSCRIPTION_MAX_CONNECTIONS:100}
```
kahowell, line 122: Can you please populate these env vars in `templates/marketplace-worker.yaml`. They should replace `MARKETPLACE_DUMMY_ID`?

For populating `SUBSCRIPTION_KEYSTORE` and `SUBSCRIPTION_KEYSTORE_PASSWORD`, `templates/rhsm-conduit.yaml` -> `RHSM_KEYSTORE`, etc. is a good reference (we will use the same secret).

kahowell (changes requested):

kahowell (approved):

PR #371, opened 2021-03-25
Fix null additionalAttributes with Marketplace

kahowell:
Before this change:

```
Caused by: org.candlepin.subscriptions.marketplace.ApiException: {"status":"failed","message":"Validation errors exist","errorDetails":[{"path":"$.data[0].additionalAttributes","type":"type","message":"$.data[0].additionalAttributes: null found, object expected"}],"data":[]}
```

After this change, marketplace API calls succeed.


dmckoske (approved):
Looks good.

lindseyburnett (approved):

PR #372, opened 2021-03-25
Map times on the snapshot as milliseconds since epoch

kahowell:
Otherwise, we end up sending usage from a long, long time ago...


dmckoske (approved):
Looks good.

mstead (approved):
Looks good.

PR #373, opened 2021-03-26
ENT-3609: Add verification and metrics for marketplace batches

kahowell:
[ENT-3609: Address marketplace API call failures](https://issues.redhat.com/browse/ENT-3609)

The strategy is to log and move on whenever there is an issue, as well as to capture metrics.

Three metrics were added:
- `rhsm_subscriptions_marketplace_batch_rejected_total`
- `rhsm_subscriptions_marketplace_batch_unverified_total`
- `rhsm_subscriptions_marketplace_batch_accepted_total`

(Prometheus names above; see MarketplaceProducer for micrometer names)

I replaced the mapper stubbed boolean that was stale with a boolean to toggle batch verification (in case we want to turn it off for some reason).

Testing
-------

Unit testing covers all scenarios, it's hard to test locally, here's an example of testing failure scenario though:

```
MARKETPLACE_MAX_ATTEMPTS=1 MARKETPLACE_URL=http://localhost:8999 ./gradlew bootRun
```

then manually emit to "marketplace" an event:

```
curl -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.marketplace:name=marketplaceJmxBean,type=MarketplaceJmxBean","operation":"submitTallySummary(java.lang.String)","arguments":["{}"]}' http://localhost:8080/actuator/jolokia
```

Now verify the counter shows 1.0:

```
curl http://localhost:8080/actuator/prometheus | grep -e '^rhsm_subscriptions_marketplace_batch_rejected_total'
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceProducer.java
+++ b/src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceProducer.java
@@ -22,6 +22,7 @@ package org.candlepin.subscriptions.marketplace;
 
 import org.candlepin.subscriptions.exception.ErrorCode;
 import org.candlepin.subscriptions.exception.SubscriptionsException;
+import org.candlepin.subscriptions.marketplace.api.model.BatchStatus;
 import org.candlepin.subscriptions.marketplace.api.model.StatusResponse;
 import org.candlepin.subscriptions.marketplace.api.model.UsageEvent;
 import org.candlepin.subscriptions.marketplace.api.model.UsageRequest;
@@ -34,7 +35,14 @@ import org.springframework.retry.support.RetryTemplate;
 import org.springframework.stereotype.Service;
 
 import io.micrometer.core.annotation.Timed;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceProducer.java
```diff
@@ -46,27 +54,93 @@
...
+                unverifiedCounter.increment();
+            }
+        });
+    }
+
+    private String verifyBatchId(String batchId) {
+        try {
+            StatusResponse response = marketplaceService.getUsageBatchStatus(batchId);
+            String status = Objects.requireNonNull(response.getStatus());
+            if ("inprogress".equals(status)) {
```
mstead, line 113: Let's make ```inprogress``` status string a constant.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceProducer.java
```diff
@@ -46,27 +54,93 @@
...
+                // throw an exception so that retry logic re-checks the batch
+                throw new MarketplaceUsageSubmissionException(response.getMessage(), status);
+            }
+            else if (!"accepted".equals(status)) {
```
mstead, line 117: Let's make ```accepted``` status string a constant.

src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceProducer.java
```diff
@@ -46,27 +54,93 @@
...
+                e
+            );
+        }
     }
 
     private StatusResponse tryRequest(UsageRequest usageRequest) {
         try {
             StatusResponse status = marketplaceService.submitUsageEvents(usageRequest);
             log.debug("Marketplace response: {}", status);
+            if (!Arrays.asList(SUCCESSFUL_SUBMISSION_STATUSES).contains(status.getStatus())) {
```
mstead, line 141: If we are just going to convert ```SUCCESSFUL_SUBMISSION_STATUSES``` to a list each time, we might as well just create the static final List above.

```java
private static final List<String> SUCCESSFUL_SUBMISSION_STATUSES = List.of("accepted", "inprogress");
```

mstead (changes requested):
Couple of small items, but functionally sound.

dmckoske (approved):
I second Mike's review points and I also agree, it looks very good.

mstead (approved):

PR #374, opened 2021-03-26
ENT-3699: Fix duplicate key violation when sla changes during metrics regather

kahowell:


mstead (approved):
This fixes the issue. Nice to see a test around it! :+1: 

Just need to fix the checkstyle issues.

PR #385, opened 2021-04-01
Use same metricId for OCP (temporarily)

kahowell:
To unblock dev-instance of marketplace-worker

### DIFF
--- a/src/main/resources/product_profile_registry.yaml
+++ b/src/main/resources/product_profile_registry.yaml
@@ -117,3 +117,5 @@ marketplaceMetrics:
     uom: Cores
     swatchProductIds:
       - OpenShift-dedicated-metrics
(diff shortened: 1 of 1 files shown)


ntkathole:
@kahowell  is https://sandbox.marketplace.redhat.com/en-us/products/openshift-container-platform-stage not useful ?

kahowell:
@ntkathole once #369 lands, it'll be useful. In the super short term, we're mapping back to a single hard-coded subscription ID for testing purposes, so we need to choose either OCP or OSD.

ntkathole:
ACK :+1:

dmckoske (approved):
Looks good.  Once this is checked in, I will need to rebase for ENT-3716.

PR #388, opened 2021-04-05
ENT-3716: Match PAYG subscriptions by offering product name rather th…

dmckoske:
...rather than role.  This is similar to ENT-3610, which uses mostly unit testing to test it.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/files/ProductProfile.java
+++ b/src/main/java/org/candlepin/subscriptions/files/ProductProfile.java
@@ -61,6 +61,7 @@ public class ProductProfile {
     private Set<SubscriptionWatchProduct> products;
     private Set<SyspurposeRole> syspurposeRoles;
     private Set<MarketplaceMetric> marketplaceMetrics;
+    private Set<SwatchProductByOfferingProductName> swatchProductsByOfferingProductName;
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/subscriptions/files/ProductProfile.java
```diff
@@ -92,29 +96,13 @@ public ProductProfile(String name, Set<SubscriptionWatchProduct> products, Granu
         setProducts(products);
     }
 
-    public String getName() {
```
awood, line 95: Why did these get stripped out?

awood (changes requested):
Overall, it looks okay; I just have questions around why a lot of getters and setters got trimmed out of `ProductProfile`

src/main/java/org/candlepin/subscriptions/files/ProductProfile.java
dmckoske: I replaced them with the Lombok annotations.

lindseyburnett: @awood, it looks like a previous PR made the lombok change but these weren't deleted as part of that

awood (approved):

PR #389, opened 2021-04-06
ENT-3667: lookup orgId when performing auto-opt-in

kahowell:
I also added a few corresponding methods to the opt-in JMX bean.

The lookup only happens if the account doesn't already have an opt-in record.

Testing
-------

1. Clear the account_config and org_config tables:

```sql
truncate table account_config;
truncate table org_config;
```

2. Run the application with modified PromQL and connected to the QA user
   service - replace `$RHSM_KEYSTORE` and `$RHSM_KEYSTORE_PASSWORD`:

Run a proxy to prometheus per (internal) https://docs.engineering.redhat.com/display/ENT/Connecting+to+Observatorium+to+Fetch+Prometheus+Metrics

```
DEV_MODE=true \
  OPENSHIFT_MAX_ATTEMPTS=1 \
  RHSM_SUBSCRIPTIONS_METERING_PROMETHEUS_METRIC_OPENSHIFT_ENABLED_ACCOUNT_PROM_Q_L="group(min_over_time(subscription_labels{ebs_account != '', billing_model='standard'}[1h])) by (ebs_account)" \
  PROM_URL=http://localhost:8082/api/v1 \
  USER_HOST=user.qa.api.redhat.com \
  RHSM_KEYSTORE=$RHSM_KEYSTORE \
  RHSM_KEYSTORE_PASSWORD=$RHSM_KEYSTORE_PASSWORD \
  ./gradlew bootRun
```

3. Invoke the JMX API to get the service to tally accounts.

```
curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=openshiftJmxBean,type=OpenShiftJmxBean","operation":"performOpenshiftMetering()","arguments":[]}'
```

It is normal to see some warnings about account not being found due to data differences between prometheus and the user API.

4. View resulting records in `account_config` and `org_config` tables.

```sql
select * from org_config;
select * from account_config;
````

5. Try some of the new methods on the OptInJmxBean if desired (http://localhost:8080/actuator/hawtio) Hint: you can use the log messages to get some sample org IDs and account numbers.

### DIFF
--- a/rhsm-client/src/main/java/org/candlepin/subscriptions/conduit/rhsm/client/X509ApiClientFactoryConfiguration.java
+++ b/rhsm-client/src/main/java/org/candlepin/subscriptions/conduit/rhsm/client/X509ApiClientFactoryConfiguration.java
@@ -22,6 +22,7 @@
 package org.candlepin.subscriptions.conduit.rhsm.client;
 
 import org.apache.http.conn.ssl.DefaultHostnameVerifier;
+import org.springframework.util.StringUtils;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -110,7 +111,7 @@ public class X509ApiClientFactoryConfiguration {
     }
 
     public boolean usesDefaultTruststore() {
-        return getTruststoreFile() == null;
+        return !StringUtils.hasText(getTruststoreFile());
     }
 
     public String toString() {
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
@@ -20,7 +20,6 @@
  */
 package org.candlepin.subscriptions.metering.service.prometheus;
 
-import org.candlepin.subscriptions.db.AccountConfigRepository;
 import org.candlepin.subscriptions.db.model.EventKey;
 import org.candlepin.subscriptions.db.model.config.OptInType;
 import org.candlepin.subscriptions.event.EventController;
@@ -30,6 +29,7 @@ import org.candlepin.subscriptions.metering.MeteringException;
 import org.candlepin.subscriptions.prometheus.model.QueryResult;
 import org.candlepin.subscriptions.prometheus.model.QueryResultDataResult;
 import org.candlepin.subscriptions.prometheus.model.StatusType;
+import org.candlepin.subscriptions.security.OptInController;
 import org.candlepin.subscriptions.util.ApplicationClock;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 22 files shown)

### CODE REVIEW

File: rhsm-client/src/main/java/org/candlepin/subscriptions/conduit/rhsm/client/X509ApiClientFactoryConfiguration.java [hunk 1]
```diff
@@ -110,7 +111,7 @@ public boolean usesClientAuth() {
     }
 
     public boolean usesDefaultTruststore() {
-        return getTruststoreFile() == null;
+        return !StringUtils.hasText(getTruststoreFile());
```
kahowell, line 114: Note: needed this to reuse the RHSM_KEYSTORE without specifying truststore.

awood (changes requested):

awood:
Your test worked great for me once I got the TLS stuff sorted out!

rhsm-client/src/main/java/org/candlepin/subscriptions/conduit/rhsm/client/X509ApiClientFactoryConfiguration.java
kahowell: I'm not trying to combine the stores. Sorry for the confusion.

I wanted to prevent the proliferation of so many aliases for the same keystore, so I decided to re-use the same env var that is already used for the RHSM API client. However, currently there's some buggy logic: if you set `RHSM_KEYSTORE`, then `RhsmApiFactory` logic falls into the branch where it uses `X509ApiClientFactory`. `X509ApiClientFactory` currently checks `usesDefaultTruststore` and without this change, because we define truststore file as `truststore-file: ${RHSM_TRUSTSTORE:}`, it is never null. Then when the logic tries to load a truststore, it fails. This isn't an issue so far in practice because we always specify both together. However, this PR makes it valid to specify just the keystore (but not the truststore)... This change was simply to fix that bug.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -171,6 +169,15 @@ public void collectOpenshiftMetrics(String account, OffsetDateTime start, Offset
         });
     }
 
+    private void ensureOptIn(String account) {
```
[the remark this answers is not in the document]
kahowell: :+1:

awood (approved):

kahowell:
783209c8 to 96d74d22: rebase

PR #390, opened 2021-04-06
exposing subscription jmx bean and fleshing out controller

jdhrh:


src/main/java/org/candlepin/subscriptions/subscription/jmx/SubscriptionJmxException.java
```diff
@@ -0,0 +1,30 @@
...
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.subscription.jmx;
+
+/**
+ * JMX Bean catch-all for the subscription jmx bean's exceptions.
+ */
+public class SubscriptionJmxException extends Exception {
```
kahowell, line 26: Not for this PR, but a general thought: we should figure out a way to do this more generally (either a single JmxException class) or figure out a better way to accomplish our goal. (We were using a checked exception to ensure that we simply followed a pattern of logging the error before returning to the client, as otherwise we don't have a record of the exception in our logs). And it definitely hasn't fulfilled the intended purpose, looking at other Jmx beans, we're pretty inconsistent :-/.

src/main/resources/application.yaml
```diff
@@ -114,3 +114,8 @@ rhsm-subscriptions:
   prometheus-latency-duration: ${PROMETHEUS_LATENCY_DURATION:4h}
   hourly-tally-offset: ${HOURLY_TALLY_OFFSET:60m}
   metric-lookup-range-duration: ${METRIC_LOOKUP_RANGE:24h}
+
+  subscriptions:
+    tasks:
+      topic: ${SUBSCRIPTION_TASK_TOPIC:platform.subscriptions.subscriptions-tasks}
```
kahowell, line 120:
```suggestion
      topic: ${SUBSCRIPTION_TASK_TOPIC:platform.rhsm-subscriptions.subscriptions-tasks}
```

src/test/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProviderTest.java
```diff
@@ -54,7 +56,7 @@
 
 @SpringBootTest
 @ExtendWith(MockitoExtension.class)
-@ActiveProfiles({"marketplace", "test"})
+@ActiveProfiles({"marketplace", "worker", "test"})
 class MarketplaceSubscriptionIdProviderTest {
```
kahowell, line 60: Changes in this file unintentionally included?

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionTaskQueueConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Task queue configuration for Subscription tasks.
+ */
+@Configuration
+public class SubscriptionTaskQueueConfiguration {
```
kahowell, line 34: Please fold this into `CapacityIngressConfiguration`. capacity-ingress is the profile for the worker that currently handles requests from subscription-conduit; I'd rather have that deployment refactored over time to handle all subscription syncing work than introduce yet another deployment type.

kahowell (changes requested):

src/test/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProviderTest.java
jdhrh: No, without the worker profile, this complains about missing dependencies for injection since the rest of this code was added.

PR #392, opened 2021-04-08
Expose Kafka consumers over JMX, add seeking params

kahowell:
`KAFKA_SEEK_OVERRIDE_END` when set to true will seek the stream to the very end.
`KAFKA_SEEK_OVERRIDE_TIMESTAMP` when set to an OffsetDateTime, will seek the stream to that timestamp.

The intended usage of the new params is to set them, do a deployment, and then remove them.

Additionally, there are a few methods exposed to seek at runtime.

I changed the ack mode to `RECORD` because I realized its semantic is consistent with what we were doing anyways.

I also consolidated the properties for the tally summary messages into the same properties as the other task queues.

I renamed some task consumer beans to reduce confusion when viewing via hawtio.

Testing
-------

If you have work in kafka topics, from previous testing, yay! Otherwise, queue up some work via JMX (exercise left for the reader).

If desired, you can see spring kafka logging by setting `LOGGING_LEVEL_ORG_SPRINGFRAMEWORK_KAFKA=DEBUG`...

```
LOGGING_LEVEL_ORG_CANDLEPIN_SUBSCRIPTIONS_UTIL=DEBUG ./gradlew bootRun
```

Try out of the seek methods as desired.

Restart the process to see that the new commits are persistent.

Additionally, try

```
KAFKA_SEEK_OVERRIDE_END=true LOGGING_LEVEL_ORG_CANDLEPIN_SUBSCRIPTIONS_UTIL=DEBUG ./gradlew bootRun
```

and

```
KAFKA_SEEK_OVERRIDE_TIMESTAMP=2022-01-01T00:00:00Z LOGGING_LEVEL_ORG_CANDLEPIN_SUBSCRIPTIONS_UTIL=DEBUG ./gradlew bootRun
```

or any variations desired.

Note it can be difficult to test the scenario of a mid-flight process stop; `ctrl+c` will initiate a graceful shutdown. I had some luck (but not consistently) with adding Thread.sleep(5000) and killing the process a few times. This can make the AckMode change a bit tricky to verify. (The spring kafka logging does show when commits are made though).


mstead (changes requested):
This is a great feature to help us get out of tricky situations. However, I did notice an issue.

I used the seek to end functionality to get past some stuck UPDATE_SNAPSHOTS messages. All is well. Then I re-ran the app without the seek to end env variable, and the tasks started again.

Please see if you can verify the same thing happens on your side.

awood:
It would also be nice to update the README or README-kafka with instructions on how this is meant to be used.

There's a lot of cruft in the README-kafka file so we might just want to zap most of it.

kahowell:
7d144fa to 638d45a: rebase against develop

kahowell:
@mstead I added a hack to force a commit of the seeked offset...

This hack uses the ability to detect when the listener is idle, along with a utility bean: KafkaConsumerRegistry, which gives the listener access to the raw Kafka consumer.

I set the idle timeout to 5s; I don't see a reason currently to set differently, but I went ahead and added it as an env var (but did not add to templates).

With this change you'll see messages right before the commits happen:

```
2021-05-04 13:49:46,306 [thread=rhsm-conduit-task-processor-0-C-1] [INFO ] [org.candlepin.subscriptions.util.SeekableKafkaConsumer] - Committing offset for [platform.rhsm-conduit.tasks-0]
2021-05-04 13:49:46,306 [thread=rhsm-subscriptions-task-processor-0-C-1] [INFO ] [org.candlepin.subscriptions.util.SeekableKafkaConsumer] - Committing offset for [platform.rhsm-subscriptions.tasks-0]
2021-05-04 13:49:46,306 [thread=openshift-metering-task-processor-0-C-1] [INFO ] [org.candlepin.subscriptions.util.SeekableKafkaConsumer] - Committing offset for [platform.rhsm-subscriptions.openshift-metering-tasks-0]
2021-05-04 13:49:46,306 [thread=marketplace-worker-0-C-1] [INFO ] [org.candlepin.subscriptions.util.SeekableKafkaConsumer] - Committing offset for [platform.rhsm-subscriptions.tally-0]
```

@awood I updated the README files as requested.

mstead:
Seek to end seems to work great.

However, there appears to be an edge case with the seek to date functionality. Give the following steps a try and see if you can reproduce.

## Reproducer
Run the app so that just the metering JMX is enabled. This will allow you to queue up some messages without them getting run.

```bash
$ DEV_MODE=true SPRING_PROFILES_ACTIVE=openshift-metering-jmx,kafka-queue ./gradlew bootRun
```

Queue up a handful of messages
```bash
$ for i in {1..5}; do curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=openshiftJmxBean,type=OpenShiftJmxBean","operation":"performCustomOpenshiftMeteringForAccount(java.lang.String, java.lang.String, java.lang.Integer)","arguments":["6364898", null, 50000]}'; done
```

Wait a minute or two and queue up another single update.
```bash
$ curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=openshiftJmxBean,type=OpenShiftJmxBean","operation":"performCustomOpenshiftMeteringForAccount(java.lang.String, java.lang.String, java.lang.Integer)","arguments":["6364898", null, 50000]}'
```

Stop the app and re-launch  specifying a timestamp between the big block of messages and the single one at the end. Make sure that you have an incorrect prometheus URL so that you can force an error on the last one.
```bash
$ KAFKA_SEEK_OVERRIDE_TIMESTAMP=2021-05-13T15:02:00Z PROM_URL="http://NOTAVALIDURL:8082/api/v1" DEV_MODE=true ./gradlew clean bootRun
```

Failure will happen processing the single message.

Next run the app with the correct prometheus URL. I was expecting just the last message to be processed because of the initial seek to timestamp, but all messages were processed again.

```bash
$ PROM_URL="http://NOTAVALIDURL:8082/api/v1" DEV_MODE=true ./gradlew clean bootRun
```

kahowell:
@mstead I see what's happening w/ your reproducer...

My hack to commit the offsets relies on the idle callback (which only happens when the consumer polls and there is no pending message). Since we have retries inside of the task, the idle callback doesn't happen until the retries are exhausted (only then does the consumer become idle - I confirmed by reducing number of attempts `OPENSHIFT_MAX_ATTEMPTS=4`).

After attempts are exhausted, the following happens:

```
2021-05-14 13:23:15,696 [thread=openshift-metering-task-processor-0-C-1] [ERROR] [org.candlepin.subscriptions.metering.task.OpenShiftMetricsTask] - Problem running task: OpenShiftMetricsTask
   [ stack trace omitted for brevity ]
2021-05-14 13:23:20,706 [thread=openshift-metering-task-processor-0-C-1] [INFO ] [org.candlepin.subscriptions.util.SeekableKafkaConsumer] - Committing offset for [platform.rhsm-subscriptions.openshift-metering-tasks-0]
```

If the app is restarted, the consumer has its position immediately after the failed message (which would have happened even without the manual commit).

Unfortunately, I don't see a good solution, as the commit *must* happen inside the consumer thread, and the idle callback is the only good hook we have into that thread without modifying the logic inside easy consumer listener method manually.

My 2 cents is we'll just have to live with the limitation and be aware of it.

mstead:
ACK - I'm Ok with that.

mstead (approved):

PR #394, opened 2021-04-09
Make unique hawtio paths for new workers

kahowell:
Not pressing, overridden in the meanwhile in app-interface.

### DIFF
--- a/templates/marketplace-worker.yml
+++ b/templates/marketplace-worker.yml
@@ -10,7 +10,7 @@ metadata:
 
 parameters:
   - name: HAWTIO_BASE_PATH
-    value: /app/rhsm-subscriptions/actuator/hawtio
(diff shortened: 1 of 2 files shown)


awood (approved):
When we deploy with these profiles are we running under different servlet contexts?

kahowell:
@awood, yeah, each DeploymentConfig is a totally separate process altogether. The base path override is important because we are proxied by turnpike, and there is a convention for turnpike paths where the path should start with `/app/$name_of_app` in order to allow multiple cloud.redhat.com apps to coexist without too much headache. Unfortunately, the spring boot hawtio actuator is not easily relocated to a different path (though we could alternatively have moved the whole `/actuator` path...); we added this override in #175; sadly I did not put a lot of explanation as to why this is necessary there. Happy to discuss further, I'm not sure if I'm emitting coherent explanations this late on a Friday :smile:

PR #397, opened 2021-04-09
ENT-3737: Update metricIds for OCP and OSD

kahowell:
Testing
-------
This is really just a data update, but if desired, you can repeat
testing steps from #350 to see the metric applied.


mstead (approved):

dmckoske (commented):
Looks good to me.

PR #401, opened 2021-04-13
ENT-3461: Update gradle to 6.8.3

lindseyburnett:


awood (approved):
This works, but our old version of checkstyle is causing problems.  Luckily, we figured out a fix for this awhile back but never applied it for some reason.  I'm going to push that commit to this branch.

kahowell (approved):

PR #413, opened 2021-04-16
ENT-3339: Implement EventPurger

lindseyburnett:
- Have events in rhsm-subscriptions.events table
Purge EventRecord entries on the same schedule as the PurgeSnapshotsJob

- Pick a duration for how far back you want to purge events and set rhsm-subscriptions.metering.prometheus.metric.eventRetentionDuration
- Change the cron schedule for rhsm-subscriptions.jobs.purge-snapshot-schedule so the job gets ran
- After a cycle of the job completes (log statements in the console), check your events table and confirm that expected  events were deleted

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
+++ b/src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
@@ -72,4 +72,10 @@ public interface EventRecordRepository extends JpaRepository<EventRecord, UUID> 
      */
     Stream<EventRecord> findByAccountNumberAndEventSourceAndEventTypeAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
         String accountNumber, String eventSource, String eventType, OffsetDateTime begin, OffsetDateTime end);
+
+    /**
+     * Delete old event records given a cutoff date
+     * @param cutoffDate Dates BEFORE this timestamp get deleted
+     */
+    void deleteEventRecordsByTimestampBefore(OffsetDateTime cutoffDate);
 }
--- a/src/main/java/org/candlepin/subscriptions/retention/EventRecordsRetentionProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/retention/EventRecordsRetentionProperties.java
@@ -18,21 +18,20 @@
  * granted to use or replicate Red Hat trademarks that are incorporated
  * in this software or its documentation.
  */
-package org.candlepin.subscriptions.metering.service.prometheus;
+package org.candlepin.subscriptions.retention;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
 
 import lombok.Getter;
 import lombok.Setter;
 
-/**
- * Properties related to all metrics that are to be gathered from the prometheus service.
- */
+import java.time.Duration;
+
 @Getter
 @Setter
-@ConfigurationProperties(prefix = "rhsm-subscriptions.metering.prometheus.metric")
-public class PrometheusMetricsProperties {
-
-    private MetricProperties openshift = new MetricProperties();
-
+@Component
+@ConfigurationProperties(prefix = "rhsm-subscriptions.event-retention-policy")
+public class EventRecordsRetentionProperties {
+    private Duration eventRetentionDuration = Duration.ofDays(90L);
 }
--- a/src/main/java/org/candlepin/subscriptions/retention/PurgeEventRecordsJob.java
+++ b/src/main/java/org/candlepin/subscriptions/retention/PurgeEventRecordsJob.java
@@ -18,21 +18,32 @@
  * granted to use or replicate Red Hat trademarks that are incorporated
  * in this software or its documentation.
  */
-package org.candlepin.subscriptions.metering.service.prometheus;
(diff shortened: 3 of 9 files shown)


src/main/java/org/candlepin/subscriptions/retention/EventRecordsRetentionProperties.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.context.annotation.Configuration;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.time.Duration;
+
+@Getter
+@Setter
+@Configuration
```
awood, line 33: Should this be `@Component`?  None of our other `...Properties` classes are marked as `@Configuration`

src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
```diff
@@ -72,4 +72,10 @@
      */
     Stream<EventRecord> findByAccountNumberAndEventSourceAndEventTypeAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
         String accountNumber, String eventSource, String eventType, OffsetDateTime begin, OffsetDateTime end);
+
+    /**
+     * Delete old event records given a cutoff date
+     * @param cutoffDate Dates BEFORE this timestamp get deleted
+     */
+    void deleteEventRecordsByTimestampBefore(OffsetDateTime cutoffDate);
```
awood, line 80: This is fairly minor, but is it possible to return the number of records deleted?  I was thinking it'd be nice to have in the log messages "Blah blah job deleted N records" so that you can tell if something actually happened.

src/main/java/org/candlepin/subscriptions/retention/TallyRetentionController.java
```diff
@@ -63,8 +76,17 @@ public void cleanStaleSnapshotsForAccount(String accountNumber) {
...
-            repository.deleteAllByAccountNumberAndGranularityAndSnapshotDateBefore(accountNumber,
+            tallySnapshotRepository.deleteAllByAccountNumberAndGranularityAndSnapshotDateBefore(accountNumber,
                 granularity, cutoffDate);
         }
     }
+
+    public void purgeOldEventRecords() {
+        Duration eventRetentionDuration = eventRecordsRetentionProperties.getEventRetentionDuration();
+
+        log.info("Purging event records older than Duration {}", eventRetentionDuration);
```
awood, line 87: I would suggest maybe changing this to something like:

```
OffsetDateTime cutoff = OffsetDateTime.now().truncatedTo(ChronoUnit.DAYS).minus(eventRetentionDuration);
log.info("Purging all event records prior to {}", cutoff);
eventRecordRepository.deleteEventRecordsByTimestampBefore(cutoff);
```

since the truncation to `ChronoUnit.DAYS` means that if you tried to do the math yourself by looking at the log entry timestamp and the value of the Duration, it wouldn't add up.

awood (changes requested):
Looks pretty good.  I think though that we'll need to define an OpenShift cronjob in the template to run this though.  Right now, it will only run in dev mode since `@Scheduled` only happens when Spring is in charge of scheduling.

src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
lindseyburnett: Agreed, I actually had something like this at first.  I can't do it using spring data "out of the box", but you could do something where you do the delete in two parts:

```java
var results = repo.findAllByTimestampBefore(timestamp);
int numberToDelete = results.size();

repo.delete(results);

log.info("deleted {} event records", numberToDelete);
```

We could technically explicitly define the Query to get the number back too, but I've been trying to avoid explicitly defining queries using the annotation like this - so we can use spring data to enhance them later if we need to.
```java
@Modifying
@Query("delete from EventRecords where timestamp Before :cutoffDate")
public int deleteByTimestampBefore(OffsetDateTime cutoffDate);
```

I'm interested in everyones' opinions on how we want to handle deletes like this.

@awood @kahowell  @mstead @dmckoske

awood: IMO, if Spring Data doesn't do it out of the box, then I don't think it's important enough to implement custom functionality.

awood (approved):

dmckoske (approved):
Thanks, awood for your review.

src/main/java/org/candlepin/subscriptions/retention/PurgeEventRecordsJob.java
```diff
@@ -0,0 +1,58 @@
...
+    public void run() {
+        Duration eventRetentionDuration = prometheusMetricsProperties.getEventRetentionDuration();
+
+        log.info("Starting PurgeEventRecordsJob for records older than {} duration.", eventRetentionDuration);
+
+        retentionController.purgeEventRecords(eventRetentionDuration);
+
+        log.info("PurgeEventRecordsJob complete.");
+
+    }
```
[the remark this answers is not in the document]
lindseyburnett: I hear ya, I'm going for consistency at this point - I think our jobs/jobs profiles/scheduled things are due for a overhaul ing eneral.

PR #417, opened 2021-04-21
ENT-3773: If two or more subscriptions exist in Marketplace, pick the newest

dmckoske:
Testing forthcoming.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
+++ b/src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
@@ -57,7 +57,8 @@ public interface SubscriptionRepository extends JpaRepository<Subscription,
         "s.sku = ALL (SELECT DISTINCT o.sku FROM Offering o WHERE " + ":#{#key.usage} = o.usage AND " +
         ":#{#key.sla} = o.serviceLevel AND " +
         "o.role IN :#{#roles}) AND s.startDate <= :rangeStart AND s.endDate >= :rangeEnd AND " +
-        "s.marketplaceSubscriptionId IS NOT NULL AND s.marketplaceSubscriptionId <> ''")
+        "s.marketplaceSubscriptionId IS NOT NULL AND s.marketplaceSubscriptionId <> '' " +
+        "ORDER BY s.startDate DESC")
     List<Subscription> findSubscriptionByAccountAndUsageKeyAndStartDateAndEndDateAndMarketplaceSubscriptionId(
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionCollector.java
```diff
@@ -56,9 +59,19 @@ public MarketplaceSubscriptionCollector(SubscriptionService subscriptionService)
...
 
     protected List<Subscription> filterNonApplicableSubscriptions(List<Subscription> subscriptions) {
+
+        Comparator<Subscription> byEffectiveStartDate =
+            Comparator.comparing(Subscription::getEffectiveStartDate);
+
         return subscriptions.stream()
             .filter(sub -> !Objects.isNull(sub.getExternalReferences()))
             .filter(sub -> !Objects.isNull(sub.getExternalReferences().get(IBMMARKETPLACE)))
+            .collect(Collectors.groupingBy(sub -> sub.getSubscriptionNumber() == null ?
```
dmckoske, line 69: This uses Subscription.subscriptionNumber to group subscriptions and determine the ones with the latest date by subscription number.  Please let me know if subscriptionNumber is the right field.  If not, I'll change this.

kahowell (changes requested):
@dmckoske this needs to be reworked. The MarketplaceSubscriptionCollector's purpose is to sync relevant subscription data into the application. It doesn't actually _choose_ which subscription ID is used if there are multiple. I would point you to https://github.com/RedHatInsights/rhsm-subscriptions/blob/05a4e15b7ffc17ac255cd4a13d4d8fb33f2f42e7/src/main/java/org/candlepin/subscriptions/marketplace/MarketplaceSubscriptionIdProvider.java#L104-L109 where we actually detect that there is more than one subscription, and then proceed to select the first one.

The semantic for "newest" should be having latest starting date... we shouldn't need to do anything special with subscriptionNumber.

There are two options:
1. (preferred) Change the query in https://github.com/RedHatInsights/rhsm-subscriptions/blob/05a4e15b7ffc17ac255cd4a13d4d8fb33f2f42e7/src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java#L56-L60 to order by `Subscription.startDate desc`. That way, the "first" subscription in the resulting list will be the "newest".
2. (backup) Sort the returned List<Subscription> afterwards in-memory in org.candlepin.subscriptions.marketplace.MarketplaceSubscriptionIdProvider#fetchSubscriptions.

kahowell (approved):

PR #421, opened 2021-04-22
ENT-3808: RHM pending charge using and SWatch core hours not matching up in test account

lindseyburnett:


src/main/java/org/candlepin/subscriptions/marketplace/MarketplacePayloadMapper.java
```diff
@@ -148,13 +146,11 @@ protected boolean isSnapshotPAYGEligible(TallySnapshot snapshot) {
             This will need to be updated if we expand the criteria defined in the
             isSnapshotPAYGEligible method to allow for Granularities other than HOURLY
              */
-            OffsetDateTime startDate = snapshotDate.minus(Duration.of(1, ChronoUnit.HOURS));
-
-            long start = startDate.toInstant().toEpochMilli();
+            long start = snapshotDate.toInstant().toEpochMilli();
             long end = snapshotDate.toInstant().toEpochMilli();
```
kahowell, line 150: We still want `end` to be an hour out from `start` (i.e. plus one hour). Otherwise the records show as starting and ending at the same time.

kahowell (changes requested):

kahowell (approved):

PR #426, opened 2021-04-26
ENT-3395: Switch to google java style guide

lindseyburnett:
Introduce diffplug/spotless gradle plugin to enforce google-java-format code style.

Gradle `build` task will not immediately fail due to violating code style, but Jenkins build will fail if the code style is violated.  This is configurable, but tried to make the behavior mirror how checkstyle was doing it previously.

Running `./gradlew :spotlessCheck` will let you know if there are any style violations.  You can automatically format your code to adhere to the google-java-format specifications a few different ways.  

1. Running the gradle task `./gradlew :spotlessApply`
2. Running the code formatter through your IDE
Instructions for IntelliJ and google-java-format can be found here: https://github.com/google/google-java-format#intellij-android-studio-and-other-jetbrains-ides 

Pay special attention to the part about the import order, and that you need to import the code style profile in addition to the google-java-format jetbrains plugin.  I checked in a copy of to the config/codestyle directory.
Be sure to turn off any other code style profiles or plugins that may be formatting your code for you (ie EditorConfig).

The most significant change here is that this gradle plugin doesn't produce a violations html report like checkstyle used to do.  I'm thinking that we could later expand on this to get that functionality back (and maybe even get checkstyle to hook up to use the google-java-format style even).  I opted not to tackle that right now, and think we should wait until we refactor our build file to use non-deprecated syntax and whatnot.


build.gradle
```diff
@@ -39,6 +39,18 @@ ext {
     generatedMetamodels = "build/generated/sources/annotationProcessor/java/main/org/candlepin/subscriptions"
 }
 
+spotless {
+    java {
+        target project.fileTree(project.rootDir) {
+            include '**/*.java'
+            exclude '**/build/**'
+        }
+        enforceCheck false
```
lindseyburnett, line 48: This line is what allows `./gradlew build` task to be successful, even if there is a code style violation

build.gradle
```diff
@@ -39,6 +39,18 @@ ext {
...
+        googleJavaFormat()
+        licenseHeaderFile "${projectDir}/config/codestyle/HEADER.txt"
+        toggleOffOn()
```
lindseyburnett, line 51: This enables a cool feature that lets you specify code that you don't want to violate rules or be reformatted

Read more here: https://github.com/diffplug/spotless/tree/main/plugin-gradle#spotlessoff-and-spotlesson

build.gradle
kahowell: Wanna put that in a comment above the line itself for clarity?

build.gradle
kahowell: Same deal, wanna put these useful details in comments in `build.gradle`?

build.gradle
```diff
@@ -39,6 +39,20 @@ ext {
...
+            exclude '**/build/**'
+        }
+        //allows build task to be successful, even if there is a code style violation
+        enforceCheck false
+        googleJavaFormat()
+        licenseHeaderFile "${projectDir}/config/codestyle/HEADER.txt"
+        //lets you specify code that you don't want to violate rules or be reformatted
+        toggleOffOn()
+    }
+}
```
kahowell, line 55: @lindseyburnett few tiny changes I'd like to suggest:
1. move this to the `allprojects` section; you'll need to manually apply the plugin there too.
2. drop `target`, we can use the built-in source set detection
3. use `${rootDir}` instead of `${projectDir}` since we keep some projects in subdirs.

I applied this locally and was able to repro your spotlessApply results 1:1. Snippet below:

```gradle
    apply plugin: "com.diffplug.spotless"
    spotless {
        java {
            enforceCheck false
            googleJavaFormat()
            licenseHeaderFile "${rootDir}/config/codestyle/HEADER.txt"
            toggleOffOn()
        }
    }
```

There are a couple extra benefits to this too; namely:
1. Can run `spotlessCheck`/`spotlessApply` on the subprojects alone if wanted (e.g. `./gradlew rbac-client:spotlessCheck`)
2. No need to manually exclude files (e.g. I have `.idea/fileTemplates/internal/Class.java` locally that I'd need to exclude with the previous way).

lindseyburnett: spotlessApply worked as expected, but the check step of it is checking generated code in the `build` directory

`./gradlew clean build spotlessCheck`

```
The following files had format violations:
build/generated-sources/js2p/org/candlepin/subscriptions/json/Event.java
...
Violations also present in 21 other files.
```

PR #431, opened 2021-05-04
Fix hourly tally CronJob schedule expression

kahowell:
Cron expressions in k8s are:

```
minute hour day_of_month month day_of_week
```

whereas w/ Spring `@Scheduled` they are:

```
second minute hour day_of_month month day_of_week
```

So the current cronjob schedule was set for nightly rather than for
hourly.


awood (commented):
:facepalm: 

I think Kubernetes actually [supports](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) `@hourly`.  Do you want to just use that and avoid this issue in the future?

kahowell:
@awood sure! I dig that idea a lot. Done & amended.

awood (approved):

PR #434, opened 2021-05-04
ENT-2658: Add offering sync controller

redsaz:
This adds the start of OfferingSyncController. Currently this will only sync offerings, it will not pull updated offerings from upstream, as that will arrive in a later story, ENT-2664.

Currently the controller is disabled, via an option added to ApplicationProperties.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
```diff
@@ -0,0 +1,78 @@
...
+   * Offering with the matching SKU is updated with the given Offering.
+   *
+   * @param newState the updated Offering
+   */
+  public void syncOffering(Offering newState) {
+    String sku = newState.getSku();
+    if (newState.getProductIds() == null || newState.getProductIds().isEmpty()) {
+      LOGGER.warn("The given sku=\"{}\" contained no engineering products. Skipping sync.", sku);
+      return;
+    }
```
kahowell, line 55: We actually do care about SKUs not providing content. I had to go look at the MW01484 to be sure. So maybe a better filter would just be against `PRODUCT_CATEGORY`?

kahowell (changes requested):
Overall looks good! I want to adjust the criteria we use for filtering so that we don't skip SKUs such as MW01484 though.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
redsaz: I thought there was a requirement on this to deny syncing offerings that had no engprod OIDs, but such a requirement is not in ENT-2658. It turns out that the requirement was in ENT-2664: "Validation: we should log a warning for products with no engineering product IDs (and not store any information)." Since there is an offering with no engProds, should I remove this check entirely?

kahowell: Just to reiterate an earlier conversation for visibility, let's implement as originally scoped, and we'll revisit with a tweak after the whole feature is done.

PR #435, opened 2021-05-04
ENT-3818: Enhance the tally response to include total_core_hours.

dmckoske:
Tests forthcoming.


kahowell (changes requested):
Overall looks good, though here is a gap we need to address: we should calculate the `total_core_hours` value in the non-running totals format version too. (i.e. given your example, I should also see 24 total core hours when `use_running_totals_format` is false or null).

kahowell (approved):

PR #444, opened 2021-05-17
ENT-3308: Populate displayName via openshift.cluster_id

kahowell:
Note that there's RHSM API work that needs to happen before this can be tested with real data (see linked issue in Jira).


kahowell:
9b7aada to 6d90a85: remove unneeded `eq` per sonarcloud.

dmckoske (approved):
Looks good, a straightforward change.  I have no other comments or recommendations.

PR #445, opened 2021-05-18
ENT-3824: Support subscriptions:reports:read permission

kahowell:
I also did some light refactoring around the RBAC properties, primarily pulling all rbac properties into a single properties class.

Testing
-------

Launch and run the following steps with two different configurations:

1. `RBAC_USE_STUB=true ./gradlew bootRun` (stub returns
   `subscriptions:*:*`).
2. `RBAC_USE_STUB=true RBAC_STUB_PERMISSIONS=subscriptions:reports:read
   ./gradlew bootRun`.

(you can try other arguments to RBAC_STUB_PERMISSIONS if desired)

Verify that the following two APIs still work (this hits both the security SPEL expressions we use):

1. Opt-in:
```
curl -X PUT \
  'http://localhost:8080/api/rhsm-subscriptions/v1/opt-in?enable_tally_sync=true&enable_tally_reporting=true&enable_conduit_sync=true' \
  -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoib3JnMTIzIn19fQo='
```

2. Check tally:
```
curl \
  'http://localhost:8080/api/rhsm-subscriptions/v1/tally/products/OpenShift%20Container%20Platform?granularity=Daily&beginning=2017-07-21T17%3A32%3A28Z&ending=2017-07-21T17%3A32%3A28Z&use_running_totals_format=false' \
  -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoib3JnMTIzIn19fQo='
```

Launch w/ `RBAC_USE_STUB=true ./gradlew bootRun`. By default, the stub returns `subscriptions:*:*` for permissions. Verify that the opt-in endpoint continues to function:

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/http/HttpClientProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/http/HttpClientProperties.java
@@ -21,8 +21,10 @@
 package org.candlepin.subscriptions.http;
 
 import java.io.File;
+import lombok.Data;
 
 /** HTTP service client configuration. */
+@Data
 public class HttpClientProperties {
 
   /** Use a stub of the service. */
@@ -46,52 +48,4 @@ public class HttpClientProperties {
 
   /** Certificate authenticate file password */
   private char[] keystorePassword;
-
-  public boolean isUseStub() {
-    return useStub;
-  }
-
-  public void setUseStub(boolean useStub) {
-    this.useStub = useStub;
-  }
-
-  public String getUrl() {
-    return url;
-  }
-
-  public void setUrl(String url) {
-    this.url = url;
-  }
-
-  public String getToken() {
-    return token;
-  }
-
-  public void setToken(String token) {
-    this.token = token;
-  }
-
-  public int getMaxConnections() {
-    return maxConnections;
-  }
-
-  public void setMaxConnections(int maxConnections) {
-    this.maxConnections = maxConnections;
-  }
-
-  public File getKeystore() {
-    return keystore;
-  }
-
-  public void setKeystore(File keystorePath) {
-    this.keystore = keystorePath;
-  }
-
-  public char[] getKeystorePassword() {
-    return keystorePassword;
-  }
-
-  public void setKeystorePassword(char[] keystorePassword) {
-    this.keystorePassword = keystorePassword;
-  }
 }
--- a/src/main/java/org/candlepin/subscriptions/http/HttpClientProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/http/HttpClientProperties.java
@@ -21,10 +21,12 @@
 package org.candlepin.subscriptions.http;
 
 import java.io.File;
-import lombok.Data;
+import lombok.Getter;
+import lombok.Setter;
 
 /** HTTP service client configuration. */
-@Data
+@Getter
+@Setter
 public class HttpClientProperties {
 
   /** Use a stub of the service. */
--- a/src/main/java/org/candlepin/subscriptions/rbac/RbacProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/rbac/RbacProperties.java
@@ -21,12 +21,14 @@
 package org.candlepin.subscriptions.rbac;
 
 import java.util.List;
-import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
 import org.candlepin.subscriptions.http.HttpClientProperties;
 
 /** RBAC-specific properties */
-@Data
+@Getter
+@Setter
 @EqualsAndHashCode(callSuper = true)
 public class RbacProperties extends HttpClientProperties {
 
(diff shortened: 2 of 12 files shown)


kahowell:
38a072e to 0426c89: run spotlessApply

src/main/java/org/candlepin/subscriptions/http/HttpClientProperties.java
```diff
@@ -21,8 +21,10 @@
 package org.candlepin.subscriptions.http;
 
 import java.io.File;
+import lombok.Data;
 
 /** HTTP service client configuration. */
+@Data
```
dmckoske, line 27: Consistent with previous work, this should be Getter and Setter, since those are the only methods you're replacing.  The aspect of Data that we're avoiding is RequiredArgsConstructor.

src/main/java/org/candlepin/subscriptions/rbac/RbacProperties.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.candlepin.subscriptions.rbac;
+
+import java.util.List;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import org.candlepin.subscriptions.http.HttpClientProperties;
+
+/** RBAC-specific properties */
+@Data
```
dmckoske, line 29: See comment above.

src/test/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsServiceTest.java
```diff
@@ -60,6 +64,13 @@ void testAdminRoleGranted() throws Exception {
     assertRoles(false, RoleProvider.SWATCH_ADMIN_ROLE);
   }
 
+  @Test
+  void testReportReaderRoleGranted() throws RbacApiException {
```
dmckoske, line 68: The semantics here are a little strange.  You have an assertion of false, but your test name suggests you're testing which roles were granted.  Either rename it, explain it in comments, or add an assertion concerning which rol(s) you expect to see.

dmckoske (changes requested):

src/test/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsServiceTest.java
```diff
@@ -20,16 +20,20 @@
...
 
-import static org.hamcrest.MatcherAssert.*;
-import static org.junit.jupiter.api.Assertions.*;
-import static org.mockito.ArgumentMatchers.*;
-import static org.mockito.Mockito.*;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.Mockito.when;
```
awood, line 27: I'm not going to ask you to change it, but I like the static wildcard imports better.  Maybe we should figure out a team style to agree on for this.

src/test/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsServiceTest.java
awood: The assertion isn't false; it's the parameter to `extractRoles` that's false.  The assertion is "assert that the returned collection from extractRoles contains SWATCH_REPORT_READER."

awood (approved):

src/main/java/org/candlepin/subscriptions/http/HttpClientProperties.java
kahowell: Done (see latest commit)

src/test/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsServiceTest.java
dmckoske: I like that as well, but the style guide -- and I think sonarcloud -- object.

dmckoske (approved):

src/test/java/org/candlepin/subscriptions/security/IdentityHeaderAuthenticationDetailsServiceTest.java
lindseyburnett: Google Style is what drives this import order/style.  It doesn't bother me enough to think its worth deviatiating from those out of the box rules, because I usually collapse the imports in my IDE anyway.  Willing to be convinced otherwise though

PR #451, opened 2021-05-26
Make compatible with Spring Boot 2.5

kahowell:
1. Because spring-boot 2.5 comes with newer liquibase, needed to change how we fetch the logger in custom tasks.

2. Since newer liquibase comes with liquibase hub, I added a system property to disable it (otherwise we get an unnecessary warning in our logs).

3. Newer promtheus client surfaces a conflict in the naming of one of our custom metrics that was actually a copy-paste error anyways. Startup failed otherwise.

4. Add `jackson-module-kotlin` as a dep in order to avoid warnings. `spring-kafka` upgrade added kotlin deps, causing messages like:

```
WARN org.springframework.http.converter.json.Jackson2ObjectMapperBuilder - For Jackson Kotlin classes support please add "com.fasterxml.jackson.module:jackson-module-kotlin" to the classpath
```

After these changes are applied we can merge #449 and #447.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -212,6 +212,8 @@ dependencies {
     compile "org.springframework.retry:spring-retry"
     compile "org.springframework:spring-context-support"
     compile "org.springframework.kafka:spring-kafka"
+    // the following dep is necessary to avoid jackson kotlin warnings
+    compile "com.fasterxml.jackson.module:jackson-module-kotlin"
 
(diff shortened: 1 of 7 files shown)


awood (approved):

kahowell:
9f01e9ec to 04b1694f: change metric name in test too

PR #466, opened 2021-06-14
Update release instructions

kahowell:


kahowell:
Self-merging since it's simply a README update.

PR #467, opened 2021-06-15
ENT-2664 Add logic to convert upstream products into an Offering

redsaz:
The mechanism that performs the conversion is called MidProduct, because
it sits in the middle between the product types from upstream and the
Offering type of Swatch.

MidProduct is in charge of knowing which product trees and eng products
to pull in order to return an Offering.

The testing for MidProduct+OfferingSyncController is done with
OfferingSyncControllerTest. It uses a specific set of actual SKUs and
engOIDs, stripped down to only the data used by Swatch, to test normal
cases and corner cases. Examples of corner cases are SKUs with derived
products or SKUs with unlimited cores or sockets.

Closes ENT-2664


src/main/java/org/candlepin/subscriptions/db/model/Offering.java
```diff
@@ -102,19 +102,19 @@
 
   /** Effective physical CPU cores capacity per quantity of subscription to this offering. */
   @Column(name = "physical_cores")
-  private int physicalCores;
+  private Integer physicalCores;
```
redsaz, line 105: Fields in Offering are changed from their primitive types to their object types, since the entries are nullable in the DB.

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
```diff
@@ -0,0 +1,376 @@
...
+import org.candlepin.subscriptions.product.api.model.RESTProductTree;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Provides an easier way to translate an operational product, its children, and their derived
+ * products into an Offering. The upstream product data are put into this intermediate structure
+ * which can then be "merged down" into its ultimate Offering form.
+ */
+public class MidProduct {
```
redsaz, line 48: Rather than converting directly from upstream's representation of products into swatch's Offering type, I introduced a representation that sits between the two. It stores all of the fields and attributes in a Map so that it is easier to collapse all of the products down into a single offering. It is designed to be a single point to contain all of the business logic necessary to create an Offering based on upstream data. It is somewhat similar to CandlepinPoolCapacityMapper in several ways.

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
```diff
@@ -0,0 +1,376 @@
...
+    return skus;
+  }
+
+  private void calcCoresAndSocketsForOffering(Offering offering) {
+    String ifl = attrs.get(Attr.IFL);
+    // ifl takes precedence over cores, so don't set cores if ifl exists.
+    String cores = (ifl == null) ? attrs.get(Attr.CORES) : null;
+    String sockets = attrs.get(Attr.SOCKET_LIMIT);
+
+    String virtIfl = attrs.get(Attr.__X_VIRTUAL_IFL);
```
redsaz, line 267: I could not find any real-world examples of a derived SKU with an IFL attr. For the sake of symmetry I added the IFL logic to both the physical and virtual cores calculations, even though in reality I could only find examples of physical cores being affected.

Therefore, if this is an example of YAGNI then I can remove it.

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
```diff
@@ -0,0 +1,376 @@
...
+    offer.mapAttributes(parent);
+
+    // For each child, merge its unconflcting information into the parent.
+    children.stream().map(MidProduct::createFromProduct).forEach(offer::merge);
+
+    return offer;
+  }
+
+  private Offering toOffering() {
+    if (conflicts.length() != 0) {
```
redsaz, line 137: When the initial convert-into-Offering logic was written, I wanted to make sure that the "merge attrs+fields from multiple products into a single Offering" logic made sense. I wrote it so it would throw an exception whenever two or more products had different values for the same attr or field.

As it is now written, it will still collect those collisions, but it will only log the collisions, and will create an Offering regardless.

If this seems like it won't be useful, let me know and I can remove it. It is straightforward enough to remove.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
```diff
@@ -64,6 +94,20 @@ public void syncOffering(Offering newState) {
     // taskManager.reconcileCapacityForOffering(...); //NOSONAR
   }
 
+  private void sortOidsAndSkus(Optional<Offering> oOffering) {
```
redsaz, line 97: Lombok is used for the Offering impl, and uses whatever Lombok creates for the equals and hashcode impls. I didn't want to erronously update an Offering when the only difference was in the ordering of the items in the childSkus and productIds lists.

If we want to do something else other than normalizing these lists by sorting them, let me know.

src/test/java/org/candlepin/subscriptions/product/OfferingSyncControllerTest.java
```diff
@@ -20,28 +20,66 @@
...
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.context.TestConfiguration;
 import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Primary;
 import org.springframework.test.context.ActiveProfiles;
 
-@SpringBootTest
+@SpringBootTest(classes = {OfferingSyncControllerTest.TestProductConfiguration.class})
```
redsaz, line 51: Since MidProduct is only ever used by OfferingSyncController, it seemed more reasonable and convenient to treat OfferingSyncController and MidProduct as a single unit, hence way there is only one unit test file, rather than two. However, I'm willing to split the MidProduct tests out into their own unit test file if there are strong enough opinions on it.

src/test/java/org/candlepin/subscriptions/product/OfferingSyncControllerTest.java
```diff
@@ -108,4 +146,157 @@ void testSyncOfferingNoProductIdsShouldPersist() {
     // Then it should still persist, since there are Offerings that we need that have no eng prods.
     verify(repo).save(sku);
   }
+
+  @Test
+  void testGetUpstreamOfferingForOcpOffering() {
```
redsaz, line 151: Rather than feeding the tests fake product data, the tests use real product data, stripped down to only the fields and attrs that are actually used in creating an Offering. There are benefits to this, but again if there are strong opinions against this technique then let me know.

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
```diff
@@ -0,0 +1,383 @@
...
+
+  private void calcCoresAndSocketsForOffering(Offering offering) {
+    String ifl = attrs.get(Attr.IFL);
+    // ifl takes precedence over cores, so don't set cores if ifl exists.
+    String cores = (ifl == null) ? attrs.get(Attr.CORES) : null;
+    String sockets = attrs.get(Attr.SOCKET_LIMIT);
+
+    String virtIfl = attrs.get(Attr.X_VIRTUAL_IFL);
+    // likewise for virtual cores, virtual ifl takes precedence if it exists.
+    String virtCores = attrs.get(Attr.X_VIRTUAL_CORES);
```
kahowell, line 277: Comment suggests this should be:
```suggestion
    // likewise for virtual cores, virtual ifl takes precedence if it exists.
    String virtCores = (virtIfl == null) ? attrs.get(Attr.X_VIRTUAL_CORES) : null;
```
?

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
redsaz: Sounds good! Will do.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
redsaz: That was my preference as well but wasn't sure if there was a reason we wanted List. I will update Offering to use Set.

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
redsaz: Ah, yes! Good catch, you are right.

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
```diff
@@ -0,0 +1,383 @@
...
+
+  private MidProduct(String sku) {
+    this.sku = sku;
+  }
+
+  private String sku;
+  private SortedSet<String> children = new TreeSet<>();
+  private SortedSet<Integer> engOids = new TreeSet<>();
+  private EnumMap<Attr, String> attrs = new EnumMap<>(Attr.class);
+  private StringBuilder conflicts = new StringBuilder(0);
```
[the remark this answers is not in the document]
redsaz: Sounds like a good idea, I'll update as suggested.

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
```diff
@@ -0,0 +1,383 @@
...
+    if (virtualCores != null) {
+      derived.putIfNoConflict(Attr.X_VIRTUAL_CORES, virtualCores);
+      derived.attrs.remove(Attr.CORES);
+    }
+
+    String virtualSockets = derived.attrs.get(Attr.SOCKET_LIMIT);
+    if (virtualSockets != null) {
+      derived.putIfNoConflict(Attr.X_VIRTUAL_SOCKETS, virtualSockets);
+      derived.attrs.remove(Attr.SOCKET_LIMIT);
+    }
```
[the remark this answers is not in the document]
redsaz: Re-reading org.candlepin.subscriptions.capacity.CandlepinPoolCapacityMapper#handleSockets I see now where I deviated. Thankfully this simplifies quite a few things so it shouldn't be too much effort to fix.

src/main/java/org/candlepin/subscriptions/product/UpstreamProductData.java
```diff
@@ -67,30 +69,27 @@
...
-  /** Maps opProd attribute codes to MidProduct.Attrs of the same name. */
+  /** Maps opProd attribute codes to UpstreamProductData.Attrs of the same name. */
   private static final Map<String, Attr> CODE_TO_ENUM =
       Arrays.stream(Attr.values())
-          .collect(Collectors.toMap(Attr::name, attr -> attr, (a1, a2) -> a1));
+          .collect(Collectors.toUnmodifiableMap(Attr::name, attr -> attr, (a1, a2) -> a1));
 
-  private MidProduct(String sku) {
+  /** Ignore any product attribute whose value is in this set. */
+  private static final Set<String> ATTRIBUTE_DISALLOWED_VALUES = Set.of("n/a", "0", "n", "none");
```
redsaz, line 82: I peeked at a few Candlepin pools to see what some of the attributes looked like between a normal pool and a guest derived pool. I noticed a "socket = 0" attribute was missing in the derived sku, so I went back and re-checked the Candlepin Adapter conversion rules. I had forgotten that there is a set of "empty" attribute values which, if Candlepin Adapters encounters, will ignore the attribute outright.

The result is that UpstreamProductData will now also ignore these attributes as well,

kahowell: :+1:

src/test/java/org/candlepin/subscriptions/product/OfferingSyncControllerTest.java
kahowell: :+1:

src/main/java/org/candlepin/subscriptions/product/MidProduct.java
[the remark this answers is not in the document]
kahowell: Exactly.

kahowell (approved):
LGTM. Will merge once next CI run passes.

PR #468, opened 2021-06-16
ENT-3973: Fix TagProfile loading from classpath

kahowell:
Using `ResourceUtils.getFile()` for a file that is on the classpath, and not the filesystem won't work. Doh!

Testing
-------
`./gradlew assemble` from `develop`, run `java -jar build/libs/rhsm-subscriptions-1.0.53-SNAPSHOT.jar`, observe message w/

```
java.io.FileNotFoundException: class path resource [tag_profile.yaml] cannot be resolved to absolute file path because it does not reside in the file system: jar:file:/deployments/rhsm-subscriptions-1.0.53-SNAPSHOT.jar!/BOOT-INF/classes!/tag_profile.yaml
```

Repeat steps with this branch, observe that the app starts without errors.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/files/ProductMappingConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/files/ProductMappingConfiguration.java
@@ -20,13 +20,12 @@
  */
 package org.candlepin.subscriptions.files;
 
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
+import java.io.IOException;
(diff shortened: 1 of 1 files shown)


dmckoske (approved):
Good catch.  This oughta do it.

kahowell:
Note you'll need to ensure you're running proper java version (i.e. if `java -version` doesn't show a Java 11 version, you'll need to either address that via `update-alternatives` or run via `/usr/lib/jvm/java-11/bin/java`).

PR #471, opened 2021-06-21
ENT-3841: Add tests for instance-hour events

kahowell:
Already handled by existing code, so this change is unit test only


mstead (commented):
Looks good to me. Waiting for test runs to complete.

PR #474, opened 2021-06-23
ENT-3871/ENT-3978: Update metrics task to use tag profile

mstead:
This PR combines changes for:
[ENT-3871](https://issues.redhat.com/browse/ENT-3871)
[ENT-3978](https://issues.redhat.com/browse/ENT-3978)

### Added a PromQL query builder

The QueryBuilder uses Spring Expression Language (SPEL) to apply
variables to a configured query template.

A query template is expected to be a well formed PromQL query.
Query templates are to be defined as a map (templateKey: promql)
in a configurationfile and can be be accessed via
PrometheusMetricsProperties:

   rhsm-subscriptions.metering.prometheus.metric.queryTemplates

A template can include template variables via a SPEL expressions.
These variables are substituted when the query is built and template
values are defined in the QueryDescriptor passed to the builder
at build time. Every property defined in the descriptor object will
accessible as template variables when the query is built. Variable
names are specified in dot notation and can be nested to access any
value defiend in the descriptor.

A QueryDescriptor contains two key fields that can be accessed in
a query template.
* **tag**: Comes from a configured TagMetric
* **runtime**: A generic collection of runtime args passed at runtime.

While the example below is not PromQL, an example of accessing
descriptor values would be as follows:

```
Metric #{tag.metricId} looked up for Account #{runtime[account]}
```

NOTE:
Template variables can contain other SPEL expressions, but it is important
note that we support 3 levels of nesting out of the box. This is not a
supported feature of SPEL. The number of levels we support was made
configurable, but I can't see us needing to change it.

###  Update metrics task to dynamically run query

* When the MeteringJob runs, pull the supported tags
  and UOMs from the tag profile definition.
* Defined a default query template for metrics.
* Add query parameters to TagMetric definitions that automatically
  get applied to the prometeus query based on the tag/uom when the task runs.
* Apply appropriate tag properties to events when they are generated
  from prometheus metrics.

### Testing
1. Configure access to telemeter.
```
podman run --rm -d -ti -p 8082:8080 quay.io/observatorium/token-refresher:master-2021-02-05-5da9663 --oidc.client-secret=$(cat ~/.token-refresher-stage) --oidc.client-id=observatorium-subwatch-staging --oidc.issuer-url=https://sso.redhat.com/auth/realms/redhat-external --url=https://observatorium.api.stage.openshift.com
```

2. Run the app
```
USER_USE_STUB=true PROM_URL="http://localhost:8082/api/metrics/v1/telemeter/api/v1" DEV_MODE=true ./gradlew clean bootRun
```

3. Kick off metrics gathering for OpenShift-metrics and observe in the logs that a task is created for each account for metering Cores. Verify that events are created.
```
curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=meteringJmxBean,type=MeteringJmxBean","operation":"performCustomMetering(java.lang.String, java.lang.String, java.lang.Integer)","arguments":["OpenShift-metrics", null, 1440]}'
```

4. In an ideal test environment, data would exist in prometheus that would allow us to properly test each product, but sadly in stage prometheus there is currently only data for OpenShift-metrics. We can test that Events are getting created for OpenShift-dedicated-metrics though, if we modify the tag_profile.yaml. Modify the yaml file and kick off metrics gathering for OpenShift-dedicated-metrics and observe in the logs that a task is created for each account for metering BOTH Cores AND Instance-hours. Verify that events are created for cores and instance-hours.
```yaml
# Modify the prometheusMetadataMetric query param in tag_profile.yaml so that the product filter is osd

# OSD metrics
- tag: OpenShift-dedicated-metrics
    metricId: redhat.com:openshift_dedicated:4cpu_hour
    uom: CORES
    queryParams:
      prometheusMetric: cluster:usage:workload:capacity_physical_cpu_cores:max:5m
      prometheusMetadataMetric: subscription_labels{product="ocp", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None"}
  - tag: OpenShift-dedicated-metrics
    metricId: redhat.com:openshift_dedicated:cluster_hour\
    uom: INSTANCE_HOURS
    queryParams:
      prometheusMetric: group(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) by (_id)
      prometheusMetadataMetric: subscription_labels{product="ocp", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None"}
```
```
curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=meteringJmxBean,type=MeteringJmxBean","operation":"performCustomMetering(java.lang.String, java.lang.String, java.lang.Integer)","arguments":["OpenShift-dedicated-metrics", null, 1440]}'
```

Tip:
Query the event data as follows to clean up the output.
```sql
select instance_id, account_number, event_type, data->>'sla', data->>'role', data->>'usage', data->>'measurements' from events;
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
@@ -62,7 +62,7 @@ public class PrometheusMeteringController {
   private final PrometheusService prometheusService;
   private final EventController eventController;
   private final ApplicationClock clock;
-  private final PrometheusMetricsProperties promMetricsProperties;
+  private final PrometheusMetricsProperties prometheusMetricsProperties;
   private final RetryTemplate openshiftRetry;
   private final OptInController optInController;
   private final QueryBuilder prometheusQueryBuilder;
@@ -70,7 +70,7 @@ public class PrometheusMeteringController {
 
   public PrometheusMeteringController(
       ApplicationClock clock,
-      PrometheusMetricsProperties promMetricsProperties,
+      PrometheusMetricsProperties prometheusMetricsProperties,
       PrometheusService service,
       QueryBuilder queryBuilder,
       EventController eventController,
@@ -78,7 +78,7 @@ public class PrometheusMeteringController {
       OptInController optInController,
       TagProfile tagProfile) {
     this.clock = clock;
-    this.promMetricsProperties = promMetricsProperties;
+    this.prometheusMetricsProperties = prometheusMetricsProperties;
     this.prometheusService = service;
     this.prometheusQueryBuilder = queryBuilder;
     this.eventController = eventController;
@@ -95,7 +95,7 @@ public class PrometheusMeteringController {
   @Transactional
   public void collectMetrics(
       String tag, Uom metric, String account, OffsetDateTime start, OffsetDateTime end) {
-    Optional<TagMetric> tagMetric = promMetricsProperties.getTagMetric(tag, metric);
+    Optional<TagMetric> tagMetric = prometheusMetricsProperties.getTagMetric(tag, metric);
     if (tagMetric.isEmpty()) {
       throw new UnsupportedOperationException(
           String.format("Unable to find TagMetric for tag %s and metric %s!", tag, metric));
@@ -108,7 +108,7 @@ public class PrometheusMeteringController {
     }
 
     MetricProperties metricProps =
-        promMetricsProperties.getSupportedMetricsForProduct(tag).get(metric);
+        prometheusMetricsProperties.getSupportedMetricsForProduct(tag).get(metric);
 
     // Reset the start/end dates to ensure they span a complete hour.
     // NOTE: If the prometheus query step changes, we will need to adjust this.
@@ -146,8 +146,8 @@ public class PrometheusMeteringController {
                     // We need to shift the start and end dates by the step, to account for the
                     // shift in the event start date when it is created. See note about eventDate
                     // below.
-                    startDate.minusSeconds(promMetricsProperties.getOpenshift().getStep()),
-                    endDate.minusSeconds(promMetricsProperties.getOpenshift().getStep()));
+                    startDate.minusSeconds(prometheusMetricsProperties.getOpenshift().getStep()),
+                    endDate.minusSeconds(prometheusMetricsProperties.getOpenshift().getStep()));
             log.debug("Found {} existing events.", existing.size());
 
             Map<EventKey, Event> events = new HashMap<>();
@@ -173,7 +173,7 @@ public class PrometheusMeteringController {
                 // actually represents the end of the measured period. The start of the event
                 // should be at the beginning.
                 OffsetDateTime eventDate =
-                    eventTermDate.minusSeconds(promMetricsProperties.getOpenshift().getStep());
+                    eventTermDate.minusSeconds(prometheusMetricsProperties.getOpenshift().getStep());
 
                 Event event =
                     createOrUpdateEvent(
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
@@ -131,7 +131,8 @@ public class PrometheusMeteringController {
 
             if (StatusType.ERROR.equals(metricData.getStatus())) {
               throw new MeteringException(
-                  String.format("Unable to fetch OpenShift metrics: %s", metricData.getError()));
+                  String.format(
+                      "Unable to fetch %s %s metrics: %s", tag, metric, metricData.getError()));
             }
 
             Map<EventKey, Event> existing =
@@ -154,7 +155,7 @@ public class PrometheusMeteringController {
               String usage = labels.get("usage");
               // NOTE: Role comes from the product label despite its name. The values set here
               //       are NOT engineering or swatch product IDs. They map to the roles in the
-              //       product profile. For openshift, the values will be 'ocp' or 'osd'.
+              //       tag profile. For openshift, the values will be 'ocp' or 'osd'.
               String role = labels.get("product");
 
               // For the openshift metrics, we expect our results to be a 'matrix'
@@ -190,14 +191,16 @@ public class PrometheusMeteringController {
             }
 
             eventController.saveAll(events.values());
-            log.info("Persisted {} events for OpenShift metrics.", events.size());
+            log.info("Persisted {} events for {} {} metrics.", events.size(), tag, metric);
 
             // Delete any stale events found during the period.
             deleteStaleEvents(existing.values());
             return null;
           } catch (Exception e) {
             log.warn(
-                "Exception thrown while updating OpenShift metrics. [Attempt: {}]: {}",
+                "Exception thrown while updating {} {} metrics. [Attempt: {}]: {}",
+                tag,
+                metric,
                 context.getRetryCount() + 1,
                 e.getMessage());
             throw e;
@@ -242,7 +245,7 @@ public class PrometheusMeteringController {
     if (event == null) {
       event = new Event();
     }
-    MeteringEventFactory.updateOpenShiftClusterCores(
+    MeteringEventFactory.updateMetricEvent(
         event,
         account,
         metricId,
@@ -260,7 +263,7 @@ public class PrometheusMeteringController {
 
   private void deleteStaleEvents(Collection<Event> toDelete) {
     if (!toDelete.isEmpty()) {
-      log.info("Deleting {} stale OpenShift metric events.", toDelete.size());
+      log.info("Deleting {} stale metric events.", toDelete.size());
       eventController.deleteEvents(toDelete);
     }
   }
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
@@ -173,7 +173,8 @@ public class PrometheusMeteringController {
                 // actually represents the end of the measured period. The start of the event
                 // should be at the beginning.
                 OffsetDateTime eventDate =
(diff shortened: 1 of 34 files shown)


src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
```diff
@@ -35,50 +42,80 @@
 @ConfigurationProperties(prefix = "rhsm-subscriptions.metering.prometheus.metric")
 public class PrometheusMetricsProperties {
 
-  public static final String OPENSHIFT_PRODUCT_PROFILE_ID = "OpenShift";
+  @Autowired TagProfile tagProfile;
```
kahowell, line 45: I'm not a fan of field injection. Can we use constructor injection here?

Could use the constructor to store the result of `tagProfile.measurementsByTag(productTag).forEach(metric -> metrics.put(metric, openshift));` instead of recalculating on each call to `getSupportedMetricsForProduct`.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
```diff
@@ -35,50 +42,80 @@
...
-    return Map.of(Uom.CORES, openshift);
-  }
 
-  // ENT-3835 will change the data structures used here and should refactor this method as needed
-  public Collection<String> getMetricsEnabledProductProfiles() {
-    return List.of(OPENSHIFT_PRODUCT_PROFILE_ID);
+    Map<Uom, MetricProperties> metrics = new EnumMap<>(Uom.class);
+    tagProfile.measurementsByTag(productTag).forEach(metric -> metrics.put(metric, openshift));
+    return metrics;
   }
```
kahowell, line 68: I'd rather see this method in TagProfile

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
```diff
@@ -35,50 +42,80 @@
...
-    // NOTE(khowell): doesn't make sense for a given product profile (e.g. OSD) to have different
-    // queries per-metric. Grabbing the first non-empty one for now.
-    return getSupportedMetricsForProduct(productProfileId).values().stream()
-        .map(MetricProperties::getEnabledAccountPromQL)
-        .filter(StringUtils::hasText)
-        .findFirst()
-        .orElseThrow();
+  public Collection<String> getMetricsEnabledProductTags() {
+    return tagProfile.getTagsWithPrometheusEnabledLookup();
   }
```
kahowell, line 72: Can this method just be inlined?

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
```diff
@@ -35,50 +42,80 @@
...
+    if (!StringUtils.hasText(productTag) || Objects.isNull(metric)) {
+      return Optional.empty();
+    }
+
+    return tagProfile.getTagMetrics().stream()
+        .filter(x -> productTag.equals(x.getTag()) && metric.equals(x.getUom()))
+        .findFirst();
+  }
+
+  public Optional<TagMetaData> getTagMetadata(String productTag) {
```
kahowell, line 112: I'd rather see this method in TagProfile, along with precomputation (produce view in constructor).

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -54,20 +61,23 @@
   private final PrometheusService prometheusService;
   private final EventController eventController;
   private final ApplicationClock clock;
-  private final PrometheusMetricsProperties metricProperties;
+  private final PrometheusMetricsProperties promMetricsProperties;
```
kahowell, line 64: Nitpick: for consistency:
```suggestion
  private final PrometheusMetricsProperties prometheusMetricsProperties;
```
(similar should be done throughout file).

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+  private String buildQuery(String template, QueryDescriptor descriptor) {
+    ExpressionParser parser = new SpelExpressionParser();
+    StandardEvaluationContext context = new StandardEvaluationContext(descriptor);
+
+    // Only allow nested expressions based on a config setting. We need to do this
+    // to prevent potential infinite recursion.
+    String query = template;
+    for (int i = 0; i < metricsProperties.getTemplateParameterDepth(); i++) {
+      // Silly hack to make sonar happy.
+      assert query != null;
```
kahowell, line 79: Rather than this maybe we should throw an exception if the query ever comes back as `null`?

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryDescriptor.java
```diff
@@ -0,0 +1,51 @@
...
+ * Describes the variables to be applied to a query template. Within a template, these variables can
+ * be utilized as follows:
+ *
+ * <p>#{tag.metricId} #{runtime[yourCustomProperty]}
+ */
+@Getter
+public class QueryDescriptor {
+
+  /** Any variables that should be provided by the tag configuration. */
+  private TagMetric tag;
```
kahowell, line 38: Clever :-). I find the naming scheme a touch confusing though. A tag may have more than one metricId. How do you feel about renaming to "metric" instead of "tag"?

src/main/resources/tag_profile.yaml
```diff
@@ -30,20 +30,26 @@ tagMetrics:
   - tag: OpenShift-metrics
     metricId: redhat.com:openshift_container_platform:cpu_hour
     uom: CORES
-    prometheusMetric: cluster:usage:workload:capacity_physical_cpu_cores:max:5m
-    prometheusMetadataMetric: subscription_labels
+    queryParams:
+      product: ocp
+      prometheusMetric: cluster:usage:workload:capacity_physical_cpu_cores:max:5m
+      prometheusMetadataMetric: subscription_labels{product="#{tag.queryParams[product]}", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None"}
```
kahowell, line 36: I'd like to see this restructured a bit:

```suggestion
      prometheusMetadataMetric: subscription_labels
```

and then in the query template:

```
max(sum_over_time(#{tag.queryParams[prometheusMetric]}[1h:5m]) / 13.0) by (_id)
            * on(_id) group_right
            min_over_time(#{tag.queryParams[prometheusMetadataMetric]}{product="#{tag.queryParams[product]}", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None"}[1h])
```

That way account query template can be:

```
group(min_over_time(#{#{tag.queryParams[prometheusMetadataMetric]}}{product='#{tag.queryParams[product]}', ebs_account != '', billing_model='marketplace'}[1h]
```
.
Also, note that we're planning on make the product label optional, so that will have some impact here, but would rather handle outside of scope of this PR.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
mstead: I agree, but I think we should wait to see what comes out of the MetricProperties/PrometheusMeteringProperties refactor in ENT-4005. I really don't like bringing MetricProperties into TagProfile.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
mstead: Agreed on this. An oversight -- thanks!!!

DONE

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMetricsProperties.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryBuilder.java
mstead: DONE

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryDescriptor.java
mstead: Works for me. Done.

src/main/resources/tag_profile.yaml
mstead: DONE - This really cleans things up a lot. 

> Also, note that we're planning on make the product label optional, so that will have some impact here, but would rather handle outside of scope of this PR.

ACK. Don't think this will be a big deal when we have to consider this.

kahowell (approved):
LGTM, will merge once CI passes

PR #475, opened 2021-06-24
Sgunta/ent 2659 sync subscription

snehagunta:
- Implements the story https://issues.redhat.com/browse/ENT-2659


src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
```diff
@@ -152,6 +152,6 @@
   /** Additional properties related to the Subscription Service */
   private SubscriptionServiceProperties subscription = new SubscriptionServiceProperties();
 
-  /** If enabled, will sync Offerings with the upstream product service. */
-  private boolean offeringSyncEnabled = false;
+  /** If enabled, will sync Subscriptions with the upstream subscription service. */
+  private boolean subscriptionSyncEnabled = false;
```
redsaz, line 156: Both `offeringSyncEnabled` and `subscriptionSyncEnabled` should exist, right? They're two independent (though related) things I think.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
```diff
@@ -0,0 +1,115 @@
...
+    reconcileSubscriptionCapacities(
+        newCapacities, subscription.getSubscriptionId(), subscription.getSku());
+  }
+
+  private Collection<SubscriptionCapacity> mapSubscriptionToCapacities(Subscription subscription) {
+
+    Offering offering = offeringRepository.getById(subscription.getSku());
+    Set<String> products =
+        productExtractor.getProducts(
+            offering.getProductIds().stream().map(Object::toString).collect(Collectors.toSet()));
```
redsaz, line 76: Something seems a bit "extra" with this. That is:

1. offering already has a non-null set of Integers of productIds.
2. `productExtractor#getProducts` only takes a collection of Strings.
3. So this will convert the `Set<Integer>` into `Set<String>` and pass it to `getProducts`
4. What that method will do to those strings is make sure they're non-null and a valid integer and turn them back into Integers
5. It proceeds grabbing the products as necessary.

With some minor additions to CapacityProductExtractor, it is able to serve clients that have either a collection of Strings or a collection of Integers, like so:

```java
// In CapacityProductExtractor, getProducts splits into two methods:

  public Set<String> getProducts(Collection<String> productIds) {
    return getProductsByIds(
        productIds.stream()
            .map(CapacityProductExtractor::parseIntSkipUnparseable)
            .filter(Objects::nonNull));
  }

  public Set<String> getProductsByIds(Stream<Integer> productIds) {
    Set<String> products =
        productIds
            .map(engProductIdToSwatchProductIdsMap::get)
            .filter(Objects::nonNull)
            .flatMap(Set::stream)
            .collect(Collectors.toSet());

    if (setIsInvalid(products)) {
      // Kick out the RHEL products since it's implicit with the RHEL-included product being there.
      products = products.stream().filter(matchesRhel().negate()).collect(Collectors.toSet());
    }

    return products;
  }
```

Then, to use it, the above code becomes:

```java
Set<String> products = productExtractor.getProductsByIds(offering.getProductIds().stream());
```

Is it *worth* doing the above. One could argue that some clarity is added and redundancy reduced. Another would argue that CapacityProductExtractor is tested and works, so why change it.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
```diff
@@ -0,0 +1,115 @@
...
+
+    Collection<SubscriptionCapacity> toSave = new ArrayList<>();
+    Map<SubscriptionCapacityKey, SubscriptionCapacity> existingCapacityMap =
+        subscriptionCapacityRepository.findByKeySubscriptionId(subscriptionId).stream()
+            .collect(Collectors.toMap(SubscriptionCapacity::getKey, Function.identity()));
+
+    if (productWhitelist.productIdMatches(sku)) {
+      newCapacities.forEach(
+          newCapacity -> {
+            toSave.add(newCapacity);
```
redsaz, line 94: It looks like this method would either save all or delete all. Do we need to consider `SubscriptionCapacity`s that exist and are unchanged from their old value, and leave them untouched?

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
```diff
@@ -0,0 +1,115 @@
...
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.candlepin.subscriptions.capacity.files.ProductWhitelist;
+import org.candlepin.subscriptions.db.OfferingRepository;
+import org.candlepin.subscriptions.db.SubscriptionCapacityRepository;
+import org.candlepin.subscriptions.db.model.*;
```
redsaz, line 34: I'm not sure if the swatch team has a standard on star imports, but for teamnado at least, it is not allowed. You may want to bump up the default IntelliJ value from 4 to 99 or something like that.

src/main/java/org/candlepin/subscriptions/capacity/PoolIngressController.java
```diff
@@ -29,15 +29,15 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.transaction.Transactional;
+
```
redsaz, line 34: Make sure to run `./gradlew spotlessApply` before checking in (or run it while doing a clean build), this blank line and the indenting on line 64 would be corrected.

src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
snehagunta: They are, but during our chat yesterday, Kevin said I could rename the `offeringSyncEnabled` as it's currently not being used afaik. It can probably be added back in when used.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: I translated whatever is being done today to reconcile capacities in the pool ingress controller. The idea is to add or update any new capacities and remove stale ones. I could add a check to leave unchanged capacities untouched, but imo it adds unnecessary complexity to the implementation with not a lot of value. The current approach assumes that the latest state is the correct state. Does that make sense?

src/main/java/org/candlepin/subscriptions/capacity/PoolIngressController.java
snehagunta: I did run that. That validation is part of the check that runs for every commit. Not sure why it did not remove this line.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: I am curious as to the purpose of that rule, but I have added the specific imports anyway.

redsaz: I don't know if there is such a rule or not for this project. But for explanations on why star imports should be avoided, there is [this stackoverflow answer](https://stackoverflow.com/a/147461) and a blog post "[import on demand is evil](http://www.javadude.com/posts/20040522-import-on-demand-is-evil/)". Admittedly the "no star imports" rule isn't without controversy, as the [second-most-popular answer](https://stackoverflow.com/a/149282) from the same stackoverflow question above says that you should prefer star imports instead.

The main purpose of the rule is so that if one updates a dependency, and that library happens to add a class whose name is already star imported elsewhere, then the project has a compilation failure. I don't know how often it happens, but apparently it does.

Ultimately I think this is ultimately a team-standard thing. For this project, star imports appear to be fine: They are used quite a bit in the test classes, and all of the Roller classes are doing a static star import for Granularity.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
```diff
@@ -0,0 +1,118 @@
...
+          newCapacity -> {
+            toSave.add(newCapacity);
+            SubscriptionCapacity oldVersion = existingCapacityMap.remove(newCapacity.getKey());
+            if (oldVersion != null) {
+              capacityRecordsUpdated.increment();
+            } else {
+              capacityRecordsCreated.increment();
+            }
+          });
+      subscriptionCapacityRepository.saveAll(toSave);
```
lindseyburnett, line 105: Do you know what happens in the event that this `saveAll` throws an exception?  Where I'm going with this is that I'm wondering if we need a `@Transactional` on this method or not, to "revert" the `capacityRecordsUpdated` and `capacityRecordsCreated` values that have already been incremented.  

Alternatively wondering if we might able to leverage java stream's `groupingBy` collector to be like "Ok here's a list of the capacities that existed in the database before, here's another list of the capacities that didn't.  Save all the capacities, then just count the number that fit either category.

```java
var dividedCapacities = newCapacities.groupingByMagic();

subscriptionCapacityRepository.saveAll(newCapacities);

// i think that grouping by gives some obnoxious map, so this pseudocode about arrays isn't even a little accurate on how to actually do it
capacityRecordsUpdated.increment(dividedCapacities[0].size());
capacityRecordsCreated.increment(dividedCapacities[1].size());

```

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    Collection<SubscriptionCapacity> toDelete = new ArrayList<>(existingCapacityMap.values());
```
lindseyburnett, line 108: Ah, ok so I see what was up with the map that you were removing elements from....my comment about groupingBy doesn't take that into consideration at all

src/main/java/org/candlepin/subscriptions/db/model/Subscription.java
```diff
@@ -83,4 +84,14 @@ public SubscriptionCompoundId() {
...
+
+  public void endSubscription() {
+    endDate = OffsetDateTime.now();
+  }
+
+  public boolean quantityHasChanged(long newQuantity) {
+    return this.getQuantity() != newQuantity;
+  }
+
+  // TODO: Add builder to throw exceptions if required fields are missing //NOSONAR
```
lindseyburnett, line 96: What's up with this TODO that was also tagged with `//NOSONAR`?  Is this something that should be completed as part of ENT-2659 or is there another story that this should be part of?  If it's the latter, I'd rather see this TODO as part of the description in a JIRA card that addresses it instead of a comment here.

cc: @kahowell

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -32,73 +33,85 @@
...
+    this.subscriptionService = subscriptionService;
+    this.capacityReconciliationController = capacityReconciliationController;
     this.clock = clock;
   }
 
   @Transactional
   public void syncSubscription(Subscription subscription) {
-    final Optional<org.candlepin.subscriptions.db.model.Subscription> entity =
+
+    // TODO: add metrics for subscriptions created and updated //NOSONAR
```
lindseyburnett, line 54: Same question as previous comment about if there's a corresponding JIRA item for this

cc: @kahowell

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -32,73 +33,85 @@
...
+    final Optional<org.candlepin.subscriptions.db.model.Subscription> maybePresent =
```
lindseyburnett, line 55: I prefer variables names like `subscriptionOpt` so when you're reading further down (like line 60ish), `subscriptionOpt.isPresent()` is a bit more descriptive than `maybePresent.isPresent()`

lindseyburnett (changes requested):

src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
kahowell: We need only a single var. We either want to sync subscription/offering information from RHIT services, or we want to use candlepin pools.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
kahowell: We have generally been against * imports ... except for in unit tests for classes such as Mockito and Jupiter/Junit assertions. I personally wouldn't have even noticed though :smile:

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private final OfferingRepository offeringRepository;
+  private final ProductWhitelist productWhitelist;
+  private final CapacityProductExtractor productExtractor;
+  private final SubscriptionCapacityRepository subscriptionCapacityRepository;
+  private final Counter capacityRecordsCreated;
+  private final Counter capacityRecordsUpdated;
+  private final Counter capacityRecordsDeleted;
+
+  public CapacityReconciliationController(
```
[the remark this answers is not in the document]
kahowell: It also prevents sonar from flagging as having too many params.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
kahowell: I'm fine with all the above suggestions. If the class-based projection suggestion starts to look like a time sink, then feel free to defer/descope.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: Removed the * imports.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
kahowell: :+1: from my PoV

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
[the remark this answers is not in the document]
snehagunta: Added annotation.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
snehagunta: Makes sense. Renamed.

src/main/java/org/candlepin/subscriptions/db/model/Subscription.java
kahowell: +1, I would add it's nice to put the Jira reference here if you do end up pushing out to a backlog item.

kahowell:
Overall looks good. Didn't have many other comments to add that @lindseyburnett didn't already cover. I did notice unit tests are failing because of changes from `List` to `Set` in Offering productIds.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: The change that @redsaz suggested makes sense to me and I made that update. 

As for using class based projections, I can't help feel that it's slightly over engineering, specifically for this use case. I am not against adding the projections but it does involve adding separate classes for the different types of projections we would want and that feels like something that needs to be looked at across the project instead of a single scenario. That way, we could have a list of the most common projections used that are more domain specific and can be reused. Also we are fetching a single offering which isn't usually that much of a concern in terms of size (imo that's the motivation for using projections, correct me if I am wrong). That's my 2c but I am open to changing my mind.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: Removed the wildcard imports.

src/main/java/org/candlepin/subscriptions/db/model/Subscription.java
snehagunta: Added subtasks to ENT-2659 as discussed. I need to add the //NOSONAR tag to avoid sonarqube check failures.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
snehagunta: Added a Jira subtask.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
lindseyburnett: In regards to "something that needs to be looked at across the project instead of a single scenario", I agree we should look at it across the project eventually, but if we chose to use projections we'd probably want to work that in incrementally instead.

Using projections (or even just simplifying some of the named query methods...I think you can do something like `Stream<String> findOffering_SkuBySku(String sku)` which would still cut down on the amount of data we're loading from the database.  If fetching a single offering doesn't do a ton of unnecessary eager fetching on other tables, I'm fine with leaving it the way it is and then revisiting projections if slowness/memory consumption becomes a problem.

src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
snehagunta: @kahowell In that case, can I leave this variable as is and I will add a story to control the offering sync via this config?

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: +1 to starting somewhere. I meant that we need to talk as a group about the possible projections that make sense fro the domain perspective. 

Looking into what the table design in like for offerings, looks like there is a join on sku between offering and sku_oid and the productIds are fetched eagerly. Still not a lot of data afaik, a sku has upto 10 engIds on an average. 

That being said, I will go ahead and add a ProductID projection, as it seems pretty straightforward and is an optimization without a lot of cost.

snehagunta: @lindseyburnett after looking closely I realized that I do need the full offering entity because if you look at line 84, I do `products.stream()
        .map(product -> from(subscription, offering, product))
        .collect(Collectors.toList());` which basically builds subscription capacity like this

`.serviceLevel(offering.getServiceLevel())
        .usage(offering.getUsage())
        .sku(offering.getSku())
        .physicalSockets(offering.getPhysicalSockets())
        .virtualSockets(offering.getVirtualSockets())
        .virtualCores(offering.getVirtualCores())
        .physicalCores(offering.getPhysicalCores())`

src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
kahowell: :+1:

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: @lindseyburnett Are there any changes that you recommend that I make here? I

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
snehagunta: `reconcileCapacityForSubscription` is only ever called by `syncSubscription` which is executed within a transaction, but it is also a public method, so I added `@Transactional` to capacity reconciliation as well to ensure it always executes within a transaction.

src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
lindseyburnett: @snehagunta  nope, I'm good...it was all related to projections and stuff what we're tabling

lindseyburnett (approved):

kahowell (approved):

PR #494, opened 2021-07-08
ENT-3815: Add retry to MeteringJob

mstead:
Since we require that the job runs each hour, if a required
service should go down when the job is run, we should retry
the operation to ensure that we do not miss a metrics update.

This was done in an attempt to avoid issues such as the one
outlined in: ENT-3815

### Testing
A simple way of testing retry is to stop the token refresher pod and start the app with the job profile. The job will fail and will begin to retry. Restart the refresher pod and the job should start working correctly.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/task/PrometheusMetricsTaskManager.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/task/PrometheusMetricsTaskManager.java
@@ -34,6 +34,7 @@ import org.candlepin.subscriptions.task.queue.TaskQueue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.retry.support.RetryTemplate;
 import org.springframework.stereotype.Component;
 
 /** Produces task messages related to pulling metrics back from Telemeter. */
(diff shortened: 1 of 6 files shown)


mstead:
Found something that I want to change with this PR so moved it back to draft state.

mstead:
> Found something that I want to change with this PR so moved it back to draft state.

This has been addressed and the PR is now ready for review.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/task/PrometheusMetricsTaskManager.java
```diff
@@ -89,24 +91,41 @@ private void queueMetricUpdateForAccount(
   @Transactional
   public void updateMetricsForAllAccounts(
       String productTag, OffsetDateTime start, OffsetDateTime end) {
+    updateMetricsForAllAccounts(productTag, start, end, Optional.empty());
```
kahowell, line 94: Instead of using optional here, you could just use `RetryTemplate.builder().maxAttempts(1).build()`, then that would avoid the need to have the if statement. Thoughts?

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/task/PrometheusMetricsTaskManager.java
awood: :+1:

mstead: DONE

kahowell (approved):

PR #507, opened 2021-07-14
Sgunta/ent 3823 add subscription number to subscription table

snehagunta:
- add subscription_number to subscription table and subscription entity


redsaz (commented):
LGTM!

src/main/java/org/candlepin/subscriptions/db/model/SubscriptionCapacity.java
```diff
@@ -40,6 +35,8 @@
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 @Getter
 @Setter
+@EqualsAndHashCode
+@ToString
```
lindseyburnett, line 39: We can replace a lot of these annotations with the `@Data` annotation

https://projectlombok.org/features/Data

lindseyburnett (changes requested):

lindseyburnett (approved):

PR #510, opened 2021-07-19
ENT-3936:  lburnett/swatch-core-library-3936

lindseyburnett:
Move "easy to move" classes that will be shared by various services when they're refactored into their own services.

Introduces new "swatch-core" subdirectory with its own build file for this.  Include the swatch-core project in the build file of the new services.


src/main/java/org/candlepin/subscriptions/util/ApplicationClock.java
```diff
@@ -41,166 +31,12 @@
  * <p>All start* methods return the time at midnight - 2019-04-19 00:00:00.0 All end* methods return
  * the max time of the day - 2019-04-19 23:59:59.999999999Z
  */
-public class ApplicationClock {
+@NoArgsConstructor
+public class ApplicationClock extends CoreApplicationClock {
```
kahowell, line 35: I'd rather keep the base class named "ApplicationClock", and have the enhanced version named "ApiApplicationClock". This'll also reduce scope of changes :-)

swatch-core/schemas/event.yaml
```diff
@@ -0,0 +1,153 @@
+$schema: http://json-schema.org/draft-07/schema#
```
kahowell, line 1: This should have been flagged as a move by git. I'm thinking you need to amend... something like `git rm -rf schemas; git commit --amend` should work

swatch-core/schemas/tally_summary.yaml
```diff
@@ -0,0 +1,71 @@
+$schema: http://json-schema.org/draft-07/schema#
```
kahowell, line 1: See comment about event.yaml

kahowell (changes requested):
Overall, works for me and looks good. Have a few small tweaks I'm suggesting.

src/main/java/org/candlepin/subscriptions/util/ApplicationClock.java
lindseyburnett: Moved ApplicationClock back to the base class because we'll be refactoring this more in depth when we break out which future services need which methods of the ApplicationClock

kahowell (approved):

PR #521, opened 2021-07-30
API for subscription table

snehagunta:
- Create SubscriptionResource to implement SubscriptionAPI
- Add SubscriptionTableController to accept inputs from resource and generate a SkuCapacityReport
- Add SubscriptionCapacityView that creates a view that's a join between SubscriptionCapacity, Subscription and Offering
- Add SubscriptionCapacityRepository that uses a SubscriptionCapacityViewSpecification to find capacities
- Add tests at all layers

PR #525, opened 2021-08-04
Fix socket and cores calculation include quantity

snehagunta:
- Capacity calculation changes made after ENT-2659 and ENT-2660 introduced a bug where SubscriptionCapacity did not take multiply subscription quantity.
- This PR fixes that bug


kahowell:
Note, I'm good with the small fix that's part of this PR :+1: , will merge once feedback on #521 lands.

kahowell (approved):

PR #530, opened 2021-08-11
Move ProductRegistry and related classes to its own package

awood:


swatch-core/src/main/java/org/candlepin/subscriptions/files/FileSourceConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.files;
+
+import org.springframework.context.annotation.ComponentScan;
+
+/** Configuration that provides the product ID to product map. */
```
kahowell, line 25: This comment no longer accurate with this change.

kahowell (changes requested):

kahowell (approved):

PR #500, opened 2021-07-09
Migrate rhsm-conduit code to swatch-system-conduit

kahowell:
https://issues.redhat.com/browse/ENT-3937
    
The following are notable details:
    
To run conduit, use `./gradlew swatch-system-conduit:bootRun` from the root of the checkout, or use `../gradlew bootRun` from the swatch-system-conduit subfolder.
    
`orgsync` still exists as a profile, and it activates the job to schedule up syncs for all opted in orgs.
    
`Dockerfile` builds a separate image for the conduit, but must be build from the root. (e.g. `podman build . -f swatch-system-conduit/Dockerfile`).

Testing
---------
Run via `DEV_MODE=true RHSM_USE_STUB=true ./gradlew swatch-system-conduit:bootRun`

Go to http://localhost:8080/actuator/hawtio and invoke some methods on `rhsmConduitJmxBean`.

Verify the inventory messages at http://localhost:3030/kafka-topics-ui (topic: `platform.inventory.host-ingress`)

PR #538, opened 2021-08-16
Fix issues with missing beans

kahowell:
Mostly issues introduced in #500.

To reproduce, first you need a very specific set of env vars.

capacity-ingress has issues, because it has different ports for management (8080) and ingress (8443) when deployed in our OpenShift environments. `MANAGEMENT_SERVER_PORT=8081` mimics this.

marketplace deployment had issues related to missing beans from the `capacity` package.

Testing
-------

```
MANAGEMENT_SERVER_PORT=8081 DEV_MODE=false SPRING_PROFILES_ACTIVE=capacity-ingress ./gradlew :bootRun
```

fails to start without this PR.

Others have issues because `SubscriptionServiceConfiguration` had `@Profile("capacity-ingress")` added.

```
DEV_MODE=false SPRING_PROFILES_ACTIVE=marketplace ./gradlew :bootRun
```

fails to start without this PR.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
@@ -53,6 +53,7 @@ import org.candlepin.subscriptions.util.LiquibaseUpdateOnlyConfiguration;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.Primary;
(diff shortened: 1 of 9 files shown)


src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
```diff
@@ -84,6 +85,7 @@
   // NOTE(khowell): actually not needed in marketplace worker
   RhsmSubscriptionsDataSourceConfiguration.class,
 })
+@ComponentScan("org.candlepin.subscriptions.util")
```
awood, line 88: I don't object to this but it's a little different from what we're doing elsewhere where we create a XYZConfiguration class in the XYZ package with the `ComponentScan` on it.  If you feel strongly about doing it the way you have it, I'm fine with that, but I thought it might be nice to remain consistent.

awood (approved):

src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
kahowell: I can do that. Hopefully we'll be out of this mess soon via refactoring :-)

PR #553, opened 2021-08-24
Docker setup for rhsm-subscriptions

ksynvrit:
This PR is to add Docker-compose as an option for local development with the main code base.

- Added Docker folder structure
- Created Readme for Docker
- Add comment to root README.md


kahowell:
@ksynvrit, please see #542 and #543, where we made some updates last week. Do these work for you? I'd rather not carry around an extra copy of the compose file. We can reopen this PR if you have further changes you want to apply to the README.

PR #559, opened 2021-08-27
ENT-4259 Periodically sync offerings

redsaz:
Offerings will periodically sync via specifications in templates/rhsm-subscriptions-scheduler.yml. The job will run every 24-hours and enqueues all SKUs listed in the allowlist to be synced with upstream. Because there was yet no kafka message tasking set up to sync offerings, that is added as part of this commit.

To test the enqueueing functionality:
- Create a product allowlist file with a handful of entries. Example
  allowlist.csv file:
```
MW0030
MW01484
MW01485
RH00604F5
RH00618F5
RH0180191
RH3413336
```
- Point to that allowlist.csv file in src/main/resources/application-worker.yaml:
```
...
rhsm-subscriptions:
  product-whitelist-resource-location: file:/path/to/allowlist.csv
  ...
```
- Make sure swatch-core-src/main/resources/swatch-core/application.yaml, product section, has the correct url, keystore, and keystorePassword
- Start a local instance like you would normally
- Kick off the sync all offerings operation using the  [Offering JMX Bean](http://localhost:8080/actuator/hawtio/jmx/operations?nid=root-org.candlepin.subscriptions.product-OfferingJmxBean-offeringJmxBean)
- Logs should show offerings getting synced

Currently there is no way to test the cron operation local as that is an OpenShift config.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
@@ -20,14 +20,21 @@
  */
 package org.candlepin.subscriptions.product;
 
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
+import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import org.candlepin.subscriptions.capacity.files.ProductWhitelist;
 import org.candlepin.subscriptions.db.OfferingRepository;
 import org.candlepin.subscriptions.db.model.Offering;
+import org.candlepin.subscriptions.task.TaskQueueProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.kafka.core.KafkaTemplate;
 import org.springframework.stereotype.Component;
 
 /** Update {@link Offering}s from product service responses. */
@@ -39,15 +46,24 @@ public class OfferingSyncController {
   private final OfferingRepository offeringRepository;
   private final ProductWhitelist productAllowlist;
   private final ProductService productService;
+  private final Timer enqueueAllTimer;
+  private final KafkaTemplate<String, OfferingSyncTask> offeringSyncKafkaTemplate;
+  private final String offeringSyncTopic;
 
   @Autowired
   public OfferingSyncController(
       OfferingRepository offeringRepository,
       ProductWhitelist productAllowlist,
-      ProductService productService) {
+      ProductService productService,
+      MeterRegistry meterRegistry,
+      KafkaTemplate<String, OfferingSyncTask> offeringSyncKafkaTemplate,
+      @Qualifier("offeringSyncTasks") TaskQueueProperties taskQueueProperties) {
     this.offeringRepository = offeringRepository;
     this.productAllowlist = productAllowlist;
     this.productService = productService;
+    this.enqueueAllTimer = meterRegistry.timer("swatch_offering_sync_enqueue_all");
+    this.offeringSyncKafkaTemplate = offeringSyncKafkaTemplate;
+    this.offeringSyncTopic = taskQueueProperties.getTopic();
   }
 
   /**
@@ -90,9 +106,34 @@ public class OfferingSyncController {
     // taskManager.reconcileCapacityForOffering(...); //NOSONAR
   }
 
+  /**
+   * Enqueues all offerings listed in the product allowlist to be synced with upstream.
+   *
+   * @return number of enqueued products
+   */
+  public int syncAllOfferings() {
+    Timer.Sample enqueueTime = Timer.start();
+
+    Set<String> products = productAllowlist.allProducts();
+    products.forEach(this::enqueueOfferingSyncTask);
+
+    Duration enqueueDuration = Duration.ofNanos(enqueueTime.stop(enqueueAllTimer));
+    int numProducts = products.size();
+    LOGGER.info(
+        "Enqueued numOfferingSyncTasks={} to sync offerings from upstream in enqueueTimeMillis={}",
+        numProducts,
+        enqueueDuration.toMillis());
+
+    return numProducts;
+  }
+
   // If there is an existing offering in the DB, and it exactly matches the latest upstream
   // version then return true. False means we should sync with the latest upstream version.
   private boolean alreadySynced(Optional<Offering> persisted, Offering latest) {
     return persisted.isPresent() && Objects.equals(persisted.get(), latest);
+  }
+
+  private void enqueueOfferingSyncTask(String sku) {
+    offeringSyncKafkaTemplate.send(offeringSyncTopic, new OfferingSyncTask(sku));
   }
 }
--- a/src/main/java/org/candlepin/subscriptions/tally/job/OfferingSyncConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/job/OfferingSyncConfiguration.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.tally.job;
+
+import org.candlepin.subscriptions.spring.JobRunner;
+import org.candlepin.subscriptions.task.queue.TaskProducerConfiguration;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Profile;
+
+/** A class to hold all job related configuration. */
+@Configuration
+@Profile("offering-sync")
+@Import(TaskProducerConfiguration.class)
+@ComponentScan({"org.candlepin.subscriptions.product"})
+public class OfferingSyncConfiguration {
+  @Bean
+  JobRunner jobRunner(OfferingSyncJob job, ApplicationContext applicationContext) {
+    return new JobRunner(job, applicationContext);
+  }
+}
--- a/templates/rhsm-subscriptions-scheduler.yml
+++ b/templates/rhsm-subscriptions-scheduler.yml
@@ -25,6 +25,8 @@ parameters:
     value: 0 10 * * *
   - name: METERING_SCHEDULE
     value: 0 * * * *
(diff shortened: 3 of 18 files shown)


src/main/java/org/candlepin/subscriptions/product/lombok.config
```diff
@@ -0,0 +1 @@
+lombok.anyconstructor.addconstructorproperties = true
```
mstead, line 1: Curious as to why this was done?

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
```diff
@@ -90,9 +106,34 @@ public void syncOffering(Offering newState) {
     // taskManager.reconcileCapacityForOffering(...); //NOSONAR
   }
 
+  /**
+   * Enqueues all offerings listed in the product allowlist to be synced with upstream.
+   *
+   * @return number of enqueued products
+   */
+  public int syncAllOfferings() {
```
mstead, line 114: Might be a good idea to get the timing for this process into prometheus so that we can monitor it over time. This can be done by annotating the method and giving it some descriptive metric name. Something like this:

```
@Timed("rhsm-subscriptions.offering.sync")
```

src/main/java/org/candlepin/subscriptions/tally/job/OfferingSyncConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Profile;
+
+/** A class to hold all job related configuration. */
+@Configuration
+@Profile("offering-sync")
+@Import(TaskProducerConfiguration.class)
+@ComponentScan({"org.candlepin.subscriptions.tally.job", "org.candlepin.subscriptions.product"})
```
mstead, line 36: Curious as to why we need to include "org.candlepin.subscriptions.tally.job" as a scan path?

templates/rhsm-subscriptions-scheduler.yml
```diff
@@ -713,3 +715,122 @@ objects:
...
+                        secretKeyRef:
+                          name: rhsm-db
+                          key: db.name
+                    - name: INVENTORY_DATABASE_HOST
+                      valueFrom:
+                        secretKeyRef:
+                          name: host-inventory-db-readonly
+                          key: db.host
+                    - name: INVENTORY_DATABASE_DATABASE
+                      valueFrom:
```
mstead, line 783: Doesn't seem like these INVENTORY_* variables are required for this job, or are they? Could this be copy/paste issues?

mstead (changes requested):
Please add some "How To Test" steps to this PR.

Overall the code looks decent, just a couple questions about a few things. :+1:

src/main/java/org/candlepin/subscriptions/product/lombok.config
redsaz: tl;dr: It lets the OfferingSyncTask class be immutable AND deserializable with Jackson with minimal coding effort.

Long version:

Glad you asked! This is done solely for the OfferingSyncTask class, whose code is in its entirety here:
```
package org.candlepin.subscriptions.product;

import lombok.NonNull;
import lombok.Value;

@Value
public class OfferingSyncTask {
  @NonNull private final String sku;
}
```

This is the message that gets submitted to kafka and eventually read back into swatch when an offering needs to be synced (asynchronously). This was such a simple class, with a single field, so I didn't think it would be wise to put the `@Builder` annotation on it. It's also nice if classes are immutable, and since there was the one field, the constructor was good enough, and no need for a setter.

However, there was one hitch: Jackson. In order to deserialize the message from the kafka topic, jackson needs at least one of the following:

* setter methods for all the fields
* A `@JsonCreator` annotation on a constructor or a static method, and `@JsonProperty("nameOfField")` for each param so that Jackson knows which param maps to which json field (this *can* be done with Lombok, but is clunky to do.)
* A `@ConstructorProperties({"param1Name", "param2Name", etc}) annotation on the constructor, so that Jackson knows which param maps to which json field

So, by adding this setting config, Lombok will automatically add the field names to the `@ConstructorProperties` annotation, which happens to match the json field names, which allows us to have an immutable "value" class, and STILL be able to use it with Jackson.

mstead: Ok, so putting the _lombok.config_ file in the package dir, ensures this is only applied to files in that package?

The reason I initially asked is that these little details are being hidden away in a config file that I wasn't really expecting.

I believe that the same thing can be achieved with annotations on the class itself.
```java
@RequiredArgsConstructor(onConstructor_=@ConstructorProperties({"sku"}))
@Value
public class OfferingSyncTask {
  @NonNull private final String sku;
}
```

redsaz: > putting the lombok.config file in the package dir, ensures this is only applied to files in that package?

Correct. I *thought* about applying it project-level since it seemed like an innocuous change, but that seemed too far reaching and out of the scope of the PR, so I limited the scope as much as possible.

> these little details are being hidden away in a config file that I wasn't really expecting.

I didn't want to create the config file, because as you say, it is getting hidden and unexpected. Based on what I read online, including stack overflow and the Lombok docs themselves, this is the preferred way to do it. I **WISH** Lombok offered a class annotation to make this explicit, like `@UseConstructorProperties` or something. 

> the same thing can be achieved with annotations on the class itself.

That's not a bad idea! The only thing to keep in mind doing it that way is we'd need to make sure that whenever a new field is added, that we go back an update the custom annotation, AND keep it in the same order as the fields in the class. It seemed less friendly to potential future maintainers of the class than lombok.config file.


There's several potential solutions we could pick from:

1. Add a comment on the class to make this not hidden, and to explain why: `// Note: using package-level lombok.config file to add @ConstructorProperties to constructor for Jackson deserialization`
2. Remove the lombok.config file and use the annotation addition you described, preferably with a note that says if any fields are added, that they must also be added to the annotation addition in the correct order.
3. Don't use Lombok here, make a vanilla immutable POJO with the `@ConstructorProperties` as necessary.
4. Keep using Lombok, but forget about this whole immutability thing, use a paramless ctor, use setters, and move on with life

My vote is for Option 1, [to leave a note](https://youtu.be/eNZsWIzEhP4). However you don't have to twist my arm if one of the other options is preferred.

mstead: > There's several potential solutions we could pick from:
> 
>     1. Add a comment on the class to make this not hidden, and to explain why: `// Note: using package-level lombok.config file to add @ConstructorProperties to constructor for Jackson deserialization`
> 
>     2. Remove the lombok.config file and use the annotation addition you described, preferably with a note that says if any fields are added, that they must also be added to the annotation addition in the correct order.
> 
>     3. Don't use Lombok here, make a vanilla immutable POJO with the `@ConstructorProperties` as necessary.
> 
>     4. Keep using Lombok, but forget about this whole immutability thing, use a paramless ctor, use setters, and move on with life
> 
> 
> My vote is for Option 1, [to leave a note](https://youtu.be/eNZsWIzEhP4). However you don't have to twist my arm if one of the other options is preferred.

All valid points! I think this is a great question to ask the team as I can see this as being a pattern we will follow again!

Also, for the record, the original solution isn't something that I feel strongly against. I just thought that it was very ODD that you could achieve it via a config file and not at the annotation level.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
redsaz: On one hand, using the `@Timed` annotation is nice because it is the least amount of coding. I used the Timer micrometer class based on a [recommendation from a previous PR](https://github.com/RedHatInsights/rhsm-subscriptions/pull/551#discussion_r694216471) (on periodically syncing subscriptions) because it would still do the logging to prometheus, AND also because I could log an event that we could then splunk.

Since the already-in-use Timer will get the timing into prometheus, is it alright to keep this as is?

src/main/java/org/candlepin/subscriptions/tally/job/OfferingSyncConfiguration.java
redsaz: Huh! Good point. I guess we don't need it. Deleted. I want to say this configuration class was originally located elsewhere which is why it had it? Maybe. Anyway like I said, it is deleted and tested as still working.

templates/rhsm-subscriptions-scheduler.yml
redsaz: That's a good question, I don't know if they're required or not. This was copied nearly verbatim from rhsm-subscriptions-cron-purge since it seemed close enough to the cron that this needs to have. If you'd like I can tear it out.

src/main/java/org/candlepin/subscriptions/product/lombok.config
redsaz: Ah gotcha! In that case, I'll do this:

- add this as a topic to the next swatch dev meeting for discussion
- add the comment to the class so that it at least reduces some of the surprise of what that config file is about

If, in the dev meeting, the consensus points to a better alternative, I'll do a followup PR to get this updated accordingly. Thanks again for the discussion!

mstead: Works for me!

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
mstead: > On one hand, using the `@Timed` annotation is nice because it is the least amount of coding. I used the Timer micrometer class based on a [recommendation from a previous PR](https://github.com/RedHatInsights/rhsm-subscriptions/pull/551#discussion_r694216471) (on periodically syncing subscriptions) because it would still do the logging to prometheus, AND also because I could log an event that we could then splunk.
> 
> Since the already-in-use Timer will get the timing into prometheus, is it alright to keep this as is?

This is fine as is. Somehow I totally missed that you were using a timer here. My bad.

templates/rhsm-subscriptions-scheduler.yml
mstead: > That's a good question, I don't know if they're required or not. This was copied nearly verbatim from rhsm-subscriptions-cron-purge since it seemed close enough to the cron that this needs to have. If you'd like I can tear it out.

If the subscription sync code doesn't need to look at the inventory database, then I don't think it is required. I'm not close enough to that code to know for sure, but to me it doesn't seem like the two would be related.

@kahowell Any idea if the INVENTORY_* env vars are required here?

mstead:
@redsaz Thanks for adding the testing steps. This helps a lot.

templates/rhsm-subscriptions-scheduler.yml
kahowell: The INVENTORY_* env vars shouldn't be required here. Please remove them; however, if you find this breaks the deployment, I'm happy to punt for now and fix with a tech-debt card later. (We have other deployments that have these vars when they don't need to).

redsaz: Sounds good! Deleted.

redsaz:
Per requests from @awood I added a default schedule for the subscription sync. I also went ahead and added a default schedule for offering sync as well, since it would have the same issue.

src/main/java/org/candlepin/subscriptions/product/lombok.config
awood: Honestly, I'd rather just skip Lombok on this class than have a config file in a strange location.

Or just have Lombok add `ConstructorProperties` throughout the code if that means we could put the config file in `META-INF`, `src/main/resources` or some other sensible location.  It looks like it's basically harmless to have `ConstructorProperties` there since Lombok will be responsible for keeping track of the parameter ordering and the like.

mstead: > Honestly, I'd rather just skip Lombok on this class than have a config file in a strange location.

After revisiting this, I think I agree with Al. It is such a simple POJO that I don't see what the gain is from using Lombok here at all. Good topic for dev sync.

mstead (approved):

mstead:
This looks good and tested well.

PR #561, opened 2021-08-30
Skip syncing a subscription if it's not on the allow list

snehagunta:
- PoolIngressController does not reconcile capacities if the sku is not on the product allowlist
- To match, we need to not add a subscription or reconcile capacity for subscription sync as well
- Because the sku will never be added to the offering table because it's not on the allowlist and will cause exceptions while creating subscriptions and capacities

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
@@ -25,13 +25,19 @@ import io.micrometer.core.instrument.Timer;
 import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import javax.transaction.Transactional;
 import lombok.extern.slf4j.Slf4j;
 import org.candlepin.subscriptions.capacity.CapacityReconciliationController;
+import org.candlepin.subscriptions.capacity.files.ProductWhitelist;
+import org.candlepin.subscriptions.db.OfferingRepository;
 import org.candlepin.subscriptions.db.SubscriptionRepository;
 import org.candlepin.subscriptions.db.model.OrgConfigRepository;
+import org.candlepin.subscriptions.exception.ErrorCode;
+import org.candlepin.subscriptions.exception.ExternalServiceException;
 import org.candlepin.subscriptions.subscription.api.model.Subscription;
+import org.candlepin.subscriptions.subscription.api.model.SubscriptionProduct;
 import org.candlepin.subscriptions.task.TaskQueueProperties;
 import org.candlepin.subscriptions.util.ApplicationClock;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -43,41 +49,67 @@
...
     this.syncSubscriptionsByOrgKafkaTemplate = syncSubscriptionsByOrgKafkaTemplate;
   }
 
   @Transactional
   public void syncSubscription(Subscription subscription) {
+    String sku = sku(subscription);
+
+    if (!productWhitelist.productIdMatches(sku)) {
+      log.info(
+          "Sku {} not on allowlist, skipping subscription sync for subscriptionId: {} in org: {} ",
```
redsaz, line 97: It is suggested to follow RH's [splunk logging standards](https://source.redhat.com/departments/it/splunk/splunk_wiki/splunk_logging_standards), specifically for [logging key-value pairs](https://source.redhat.com/departments/it/splunk/splunk_wiki/splunk_logging_standards#jive_content_id_Use_Clear_KeyValue_Pairs) when possible.

This makes searches easier and splunk can automatically extract things we can look for.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -133,6 +165,7 @@ void syncSubscriptions(String orgId, int offset, int limit) {
     int pageSize = limit + 1;
     List<Subscription> subscriptions =
         subscriptionService.getSubscriptionsByOrgId(orgId, offset, pageSize);
+    log.info("Got subscriptions of size {} from external service ", subscriptions.size());
```
redsaz, line 168: I fully admit this is the nitpickiest of comments: It may read "easier" if applying the previously mentioned splunk standards, and listing the org would be good too: "Fetched numSubs={} for orgId={} from external service."

OTOH, this same information and phrasing is found further down on line 178, so maybe its okay anyway.

redsaz (approved):

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -43,41 +49,67 @@
...
+    if (!productWhitelist.productIdMatches(sku)) {
+      log.info(
+          "Sku {} not on allowlist, skipping subscription sync for subscriptionId: {} in org: {} ",
+          sku,
+          subscription.getId(),
+          subscription.getWebCustomerId());
+      return;
+    }
+
+    if (offeringRepository.findById(sku).isEmpty()) {
```
kahowell, line 104: Theoretically this can be more performant (Spring Data should generate a `select count` rather than fetching whole entity):

```suggestion
    if (!offeringRepository.existsById(sku)) {
```

kahowell (changes requested):
Tiny suggestion :smile:

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
snehagunta: I did not know that we had splunk logging standards.Such goodness. I made the update for the whole class.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
snehagunta: Updated.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
snehagunta: Makes sense. Fixed.

awood (changes requested):
Would you mind doing an interactive rebase to squash some of the commits together?  E.g. merging the commits where you run spotless apply together with the more meaningful commits?  Actually with 98 lines of net changes you could probably just squash everything together and it would be fine.

snehagunta:
@awood I updated the commit history as requested.

PR #564, opened 2021-09-02
Update kafka topic names for sub sync and capacity reconciliation

redsaz:
This makes the topic names follow a pattern established for other topics.

The topic names are [merged into platform-mq](https://github.com/RedHatInsights/platform-mq/pull/148). This will make the names match in swatch.


kahowell (changes requested):
Failing test w/ easy-fix:
```
sed -i s/platform.rhsm-subscriptions.capacity.reconcile/platform.rhsm-subscriptions.capacity-reconcile/ src/test/java/org/candlepin/subscriptions/capacity/CapacityReconciliationControllerTest.java
```

redsaz:
Arg, I can't believe I forgot to run the tests before committing. Thanks for the quick fix! Tested, should be good now.

kahowell (approved):

PR #567, opened 2021-09-07
Add logic for vCPU counting of virtual x86 cpu.

ksynvrit:
Add system profile arch to begin leverage for vCPU counting.
Add logic for for x86 virtual guests to get a more accurate count of vCPUs.
Add test case to test method within the normalize workflow.
Refactor vCPU counting to account for odd number value of cpu.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
@@ -224,7 +224,7 @@ public class FactNormalizer {
         virtualFacts.getSystemProfileCoresPerSocket() * virtualFacts.getSystemProfileSockets();
 
     var threadsPerCore = 2.0;
-    return Double.valueOf(Math.ceil(cpu / threadsPerCore)).intValue();
+    return (int) Math.ceil(cpu / threadsPerCore);
   }
 
   private void getProductsFromProductIds(
--- a/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
@@ -206,8 +206,10 @@ public class FactNormalizer {
       normalizedFacts.setCores(
           hostFacts.getSystemProfileCoresPerSocket() * hostFacts.getSystemProfileSockets());
     }
-    if (hostFacts.getSystemProfileArch().equals("x86_64")
-        && hostFacts.getSystemProfileInfrastructureType().equals("virtual")) {
+    if ("x86_64".equals(hostFacts.getSystemProfileArch())
+        && HardwareMeasurementType.VIRTUAL
+            .toString()
+            .equalsIgnoreCase(hostFacts.getSystemProfileInfrastructureType())) {
       var effectiveCores = calculateVirtualCPU(hostFacts);
       normalizedFacts.setCores(effectiveCores);
     }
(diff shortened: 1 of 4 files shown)


src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
```diff
@@ -206,9 +206,25 @@ private void normalizeSystemProfileFacts(
       normalizedFacts.setCores(
           hostFacts.getSystemProfileCoresPerSocket() * hostFacts.getSystemProfileSockets());
     }
+    if (hostFacts.getSystemProfileArch().equals("x86_64")
```
lindseyburnett, line 209: I would switch the order you're checking for equals in to protect against null pointers.

```
"x86_64".equals(hostFacts.getSystemProfileArch())
```

Same for the next conditional with `virtual`.

@kahowell, do we have any enums for these strings that would be appropriate to use here instead of strings?

src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
```diff
@@ -206,9 +206,25 @@ private void normalizeSystemProfileFacts(
...
   }
 
+  private Integer calculateVirtualCPU(InventoryHostFacts virtualFacts) {
+    //  For x86, guests: if we know the number of threads per core and its greater than one,
+    //  then we divide the number of cores by that number.
+    //  Otherwise we divide by two.
+    int cpu =
+        virtualFacts.getSystemProfileCoresPerSocket() * virtualFacts.getSystemProfileSockets();
+
+    // If we have the actual threads per core, you can use it here.
```
lindseyburnett, line 224: We could probably stub out  the variable for this here and eliminate the comment.

```
var threadsPerCore = 2.0;

return Double.valueOf(Math.ceil(cpu /threadsPerCore)).intValue();
```

lindseyburnett (changes requested):

src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
ksynvrit: Hey LIndsey, thanks for the feedback. I''l look around the code base I remember seeing a enum, that I could use. For sure I'll have to create one for the arch

src/main/java/org/candlepin/subscriptions/tally/facts/FactNormalizer.java
```diff
@@ -206,9 +206,27 @@ private void normalizeSystemProfileFacts(
...
+            .toString()
+            .equalsIgnoreCase(hostFacts.getSystemProfileInfrastructureType())) {
+      var effectiveCores = calculateVirtualCPU(hostFacts);
+      normalizedFacts.setCores(effectiveCores);
+    }
     getProductsFromProductIds(normalizedFacts, hostFacts.getSystemProfileProductIds());
   }
 
+  private Integer calculateVirtualCPU(InventoryHostFacts virtualFacts) {
+    //  For x86, guests: if we know the number of threads per core and its greater than one,
```
san7ket, line 220: About finding the threads per core count, this is future work right ? Phase 2 of this @lindseyburnett @kahowell

kahowell: Yes, using threads per core is future work.

lindseyburnett: @san7ket yup, that's future work

lindseyburnett (approved):

PR #571, opened 2021-09-08
Subscription table API naming updates

snehagunta:
Update naming for sort parameters and fields in Json response for Subscription table api to be consistent with other apis

The final agreed upon names:

next_event_date
next_event_type
service_level

Also updated the tag for this API to be 'subscriptions'

To test, startup application locally and review update at http://localhost:8080/api-docs/index.html#/

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -391,7 +391,7 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
       tags:
-        - capacity
+        - subscriptions
   /hosts/products/{product_id}:
     description: 'Operations for hosts for a given account and product'
     parameters:
@@ -1429,11 +1429,11 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/SkuCapacitySubscription"
-        upcoming_event_date:
+        next_event_date:
           description: "The most immediate date for a subscription event."
           type: string
           format: date-time
-        upcoming_event_type:
+        next_event_type:
           description: "The most immediate subscription event, such as a subscription beginning or ending."
           $ref: "#/components/schemas/SubscriptionEventType"
         quantity:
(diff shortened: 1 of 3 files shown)


redsaz (approved):
lgtm!

cdcabrera:
These changes are what was originally agreed upon...
- `next_event_date`
- `next_event_type`
- `service_level`
...so for that aspect lgtm!

However, @snehagunta pointed out that we use `sla` for a variety of other query params on other API endpoints... it is to our benefit to consider instead making `service_level` into `sla` to help align with the other API(s) endpoints, thoughts? @kahowell @joeldmyers @redsaz @lindseyburnett

joeldmyers:
> These changes are what was originally agreed upon...
> 
> * `next_event_date`
> * `next_event_type`
> * `service_level`
>   ...so for that aspect lgtm!
> 
> However, @snehagunta pointed out that we use `sla` for a variety of other query params on other API endpoints... it is to our benefit to consider instead making `service_level` into `sla` to help align with the other API(s) endpoints, thoughts? @kahowell @joeldmyers @redsaz @lindseyburnett

That makes sense to me to change it to `sla` for consistency's sake. Agreed, otherwise lgtm.

kahowell:
+1 on bringing about consistency, but if we're going to make a change, I'd rather change `sla` to `serviceLevel` (or `service_level` depending on context) (selfishly because we use `ServiceLevel` more times in the backend than `sla`). `git grep -i sla | wc -l` -> 332 < `git grep -i servicelevel | wc -l` -> 523

Other than that, they're synonymous enough to me, that I don't mind which one wins.

cdcabrera:
The GUI would be onboard for going the other direction @kahowell and @snehagunta with `serviceLevel` or `service_level`!

As a recap in our off-to-the side discussion, GUI agrees in keeping the query param as `sla` for now, and a subsequent set of stories/tasks for the API and GUI to convert the `sla` param over to `serviceLevel` or `service_level` after the "current subscriptions" work is wrapped up

@joeldmyers

kahowell (approved):

PR #572, opened 2021-09-09
ENT-4259 Add profiles to offering and subscription sync jobs

redsaz:
This is to prevent any beans the jobs use from being pulled in when other profiles are active.
Also adds "org.candlepin.subscriptions.tally.job" to the OfferingSyncConfiguration component scan so that the bean can be properly loaded when offering-sync profile is active.


mstead (approved):
Seems to fix the issues for me. Tested the jobs that were failing as well.

PR #573, opened 2021-09-09
Refactor Sync Offering JMX implementation

snehagunta:
There are 2 possible scenarios when an offering will not be synced

1. If it is not present on the allowlist
2. If it is present on the allowlist but not present in the product service

As each of these error scenarios requires different actions i.e, 

for 1, we will either ignore the error because it is expected behavior or the add the offering to the allowlist (depends on the sku)

 for 2, we will need to figure out why a sku is missing from product service and take appropriate steps to fix it.

I am recommending we separate the exception logging for these 2 scenarios, so we can alert based on the specific scenario in splunk (if that's our preferred option).

Currently we are returning a string message for an offering that's missing in Product Service but to me it feels more appropriate to throw an exception. That requires a little more refactoring than this, so if there is agreement, I would like to make that update as well.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/product/OfferingJmxBean.java
+++ b/src/main/java/org/candlepin/subscriptions/product/OfferingJmxBean.java
@@ -60,7 +60,10 @@ public class OfferingJmxBean {
       return upstream
           .map(Offering::toString)
           .orElseGet(
-              () -> "{\"message\": \"offeringSku=\"" + sku + "\" was not found/allowlisted.\"}");
+              () ->
+                  "{\"message\": \"offeringSku=\""
+                      + sku
+                      + "\" was not found in upstream service\"}");
     } catch (Exception e) {
       log.error("Error syncing offering", e);
       throw new JmxException("Error syncing offering. See log for details.");
--- a/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
@@ -58,7 +58,8 @@ public class OfferingSyncController {
   public Optional<Offering> getUpstreamOffering(String sku) {
     if (!productAllowlist.productIdMatches(sku)) {
       LOGGER.info("sku=\"{}\" is not in allowlist. Will not retrieve offering from upstream.", sku);
-      return Optional.empty();
+      throw new RuntimeException(
+          "sku=" + sku + " is not in allowlist. Will not retrieve offering from upstream.");
     }
     LOGGER.debug("Retrieving product tree for offering sku=\"{}\"", sku);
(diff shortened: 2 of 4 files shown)


redsaz (commented):
This is worth a larger discussion with @kahowell I think since he has a better understanding of how swatch is expected to run.

Based on what I see, I'll offer this:
* Exceptions should be thrown for exceptional cases, and not as part of an expected, normal business logic flow. Since we know ahead of time if a SKU is allowed or not (thanks to the allow list), this doesn't seem worthy of an exception.
* The RuntimeException isn't handled specially. That is, if it were to get thrown, it'd be caught here in the `catch (Exception e)` block, whose purpose is really more for handling "we tried to contact the upstream service, but it wasn't there" (or some other service/connection type of error).
* No changes were made to the JavaDocs or tests to cover the updates to this contract.
* No updates were made to the OfferingWorker which also uses the method
* (This isn't really about anything from the PR) I wish Java offered the Rust equivalent of `Result<T, Error>` because that's what I wanted: It would return the actual Offering if it was a success, and if not, the reason that no offering was returned. However, Java doesn't, so the next best thing was Optional.

Given the above, if it is decided that throwing an exception for when something is not allowlisted, then that should probably be made into a Jira ticket, I imagine? It seems *juuuust* a little too much effort to be a one-off PR without a story to back it.

snehagunta:
@redsaz 

- Exceptions should be thrown for exceptional cases, and not as part of an expected, normal business logic flow. Since we know ahead of time if a SKU is allowed or not (thanks to the allow list), this doesn't seem worthy of an exception.

>  I do not disagree with you, I did not really want to throw an exception if something wasn't on the allowList. My goal is to separate the 2 scenarios instead of clubbing them into one message, because it's confusing(to me atleast) when I sync a whole bunch of offering for example and I want to know via splunk which failed to sync because of which. I can not throw an exception if something isn't on the allowlist, but then I would need to change the upstream offering method, because returning Optional.empty for both cases masks the exact reason why an offering sync.

The RuntimeException isn't handled specially. That is, if it were to get thrown, it'd be caught here in the catch (Exception e) block, whose purpose is really more for handling "we tried to contact the upstream service, but it wasn't there" (or some other service/connection type of error).

> Again I don't disagree. But I do need to do a slightly larger refactor to do the right thing. I wanted to get this question out there so we can discuss the right implementation for achieving this goal.

No changes were made to the JavaDocs or tests to cover the updates to this contract.

> I did not realize there was a JavaDoc that needed to be updated for JMX API contracts. If you point them to me, I'll be happy to make the updates based on what we decide.

No updates were made to the OfferingWorker which also uses the method

> I did not see this Class in develop and I pulled this morning. Possibly I need to refresh my local repo.

PR #575, opened 2021-09-10
Awood/rehome clients

awood:
Move all our generated clients under a single directory.  They are generally uninteresting and infrequently touched.  Let's tuck them away in a sub-directory so that the project isn't so sprawled out at the top level.  My scroll wheel finger is sore from all the movement.


kahowell:
@awood needs a rebase

PR #581, opened 2021-09-14
Fix issue with subscription table api tags 

snehagunta:
As part of a PR merged recently, I unintentionally updated subscription table api to be tagged under capacity and capacity related api to be tagged under subscription. This PR fixes that. It also updates the SkuCapacityReport examples to reflect the latest naming for next_event_date, next_event_type and service_level.


cdcabrera:
I think we got it!
Screen Shot 2021-09-14 at 1 51 53 PM

Screen Shot 2021-09-14 at 1 51 58 PM

PR #587, opened 2021-09-17
ENT-3376 Enable subscription sync for cron job

redsaz:
This will set SUBSCRIPTION_SYNC_ENABLED to true for the subscription-sync cron job. With the updates to app-interface (https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/25597) which sets it to false for QA, STAGE, and PROD, this allows only CI to run the periodic subscription sync, until we're ready.

### DIFF
--- a/templates/rhsm-subscriptions-scheduler.yml
+++ b/templates/rhsm-subscriptions-scheduler.yml
@@ -67,6 +67,8 @@ parameters:
     value: 128Mi
   - name: TOKEN_REFRESHER_MEMORY_LIMIT
     value: 150Mi
+  - name: SUBSCRIPTION_SYNC_ENABLED
+    value: true
   # TODO: this should be removed once OCM has non-standard billing_model clusters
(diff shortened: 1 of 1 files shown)


lindseyburnett (approved):

ntkathole:
@redsaz this change is causing failure in deployment 

```
error: failed to read input object (not a Template?): unable to decode "rhsm-subscriptions-scheduler.yml": v1.Template.Parameters: []v1.Parameter: v1.Parameter.Value: ReadString: expects " or n, but found t, error found in #10 byte of ...|,"value":true},{"nam|..., bigger context ...|Mi"},{"name":"SUBSCRIPTION_SYNC_ENABLED","value":true},{"name":"OPENSHIFT_ENABLED_ACCOUNT_PROMQL","v|...
error: no objects passed to apply
```
Seems 'true' should be inside quotes.

redsaz:
Thanks @ntkathole I'll check it out!

PR #605, opened 2021-09-27
ENT-4331: lburnett/data transfer tagprofile ent 4331

lindseyburnett:


awood (changes requested):

swatch-core/src/main/resources/tag_profile.yaml
```diff
@@ -66,6 +71,15 @@ tagMetrics:
       prometheusMetric: group(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) by (_id)
       prometheusMetadataMetric: subscription_labels
 
+  # RHOSAK
+  - tag: rhosak
+    metricId: redhat.com:rhosak:transfer_gb
+    uom: TRANSFER_GIGABYTES
+    queryParams:
+      product: rhosak
+      prometheusMetric: kafka_id:haproxy_server_bytes_in_out_total:rate1h_gibibytes
```
awood, line 80: We're labeling this as transfer *gigabytes* but Prometheus is returning *gibibytes*.  I'm assuming we have plans for conversion in place?  I think a comment here would be appropriate though explaining how the discrepency is handled.

awood (changes requested):

swatch-core/src/main/resources/tag_profile.yaml
lindseyburnett: We're not going to be doing any conversions.  We got the RHOSAK team to do whatever conversions they need to and send us the metrics as they want to be billed.  I could see a good argument to change some names that we're using to avoid confusion though.  eg `TRANSFER_GIGIBYTES`.  

Thoughts?
@awood @kahowell @mstead @ksynvrit

awood: Personally, I vote that we report as Gibibytes.  Switching the unit seems like it only adds unnecessary confusion and a potential landmine for the future.

awood (approved):

PR #609, opened 2021-09-27
ENT-4373 Reconcile capacities when offering synced

redsaz:
If an offering's data is changed in the course of getting synced, then
any capacities relating to that offering should be reconciled.

Because reconciling capacities could add time to an offering sync, and
because there could be many offerings to sync each day, a reconciliation
job will be enqueued into kafka per each sku.

Test overview:

- Pick an offering that has at least one or more customers
- Sync the offering, if it isn't already in the swatch DB
- Sync the subscriptions of at least one of those customers
- Change one of the details of the offering in the swatch DB
- Re-sync the offering. Because the offering was changed, the
  capacities for the customer should be reconciled (async).
- Sync the offering a third time. Because the offering was not changed,
  capacities for the customer should not be reconciled

Testing this commit with locally running containers would look like the
following example:

- When running this locally, make sure the logs from bootRun can be seen
- All of these JMX bean commands will use the following bash function
  (copy and paste into bash terminal):

```bash
curl_jmx() {
  curl 'http://localhost:8080/actuator/hawtio/jolokia/?maxDepth=7&maxCollectionSize=50000&ignoreErrors=true&canonicalNaming=false' \
    -H 'Accept: application/json' \
    --silent \
    --data-raw "$1"
}
```

- Sync the offering using the OfferingJmxBean, let it complete:
```bash
curl_jmx '{"type":"exec","mbean":"org.candlepin.subscriptions.product:name=offeringJmxBean,type=OfferingJmxBean","operation":"syncOffering(java.lang.String)","arguments":["RH00006"]}'
```
- Sync the subscriptions of that org (depending on the org, this may
  take a few seconds to a few minutes, and the operation works in pages,
  so could continue even after the command returns)
```bash
curl_jmx '{"type":"exec","mbean":"org.candlepin.subscriptions.subscription:name=subscriptionJmxBean,type=SubscriptionJmxBean","operation":"syncSubscriptionsForOrg(java.lang.String)","arguments":["ASK_FOR_ORGID"]}'
```
- Edit the entry of that offering in the DB, so that when it gets synced
  again, it will change:
```bash
psql -U rhsm-subscriptions -d rhsm-subscriptions  -h localhost -p 5432 -c "UPDATE offering SET role = 'bogus' WHERE sku = 'RH00006';"
```
- Sync the offering a second time. The logs should show the offering was
  updated and the capacities reconciled. The output should have:
  syncResult=FETCHED_AND_SYNCED
```bash
curl_jmx '{"type":"exec","mbean":"org.candlepin.subscriptions.product:name=offeringJmxBean,type=OfferingJmxBean","operation":"syncOffering(java.lang.String)","arguments":["RH00006"]}'
```
- Sync the offering a third time. The logs should show the offering
  matches what was previously stored, so was not updated nor were the
  capacities reconciled. The output should have:
  syncResult=SKIPPED_MATCHING
```bash
curl_jmx '{"type":"exec","mbean":"org.candlepin.subscriptions.product:name=offeringJmxBean,type=OfferingJmxBean","operation":"syncOffering(java.lang.String)","arguments":["RH00006"]}'
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
@@ -26,6 +26,7 @@ import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import org.candlepin.subscriptions.capacity.CapacityReconciliationController;
 import org.candlepin.subscriptions.capacity.files.ProductWhitelist;
 import org.candlepin.subscriptions.db.OfferingRepository;
 import org.candlepin.subscriptions.db.model.Offering;
@@ -44,11 +45,12 @@ public class OfferingSyncController {
   private static final Logger LOGGER = LoggerFactory.getLogger(OfferingSyncController.class);
 
   private static final String SYNC_LOG_TEMPLATE =
-      "{} for offeringSku=\"{}\" in offeringSyncTimeMills={}.";
+      "{} for offeringSku=\"{}\" in offeringSyncTimeMillis={}.";
 
   private final OfferingRepository offeringRepository;
   private final ProductWhitelist productAllowlist;
   private final ProductService productService;
+  private final CapacityReconciliationController capacityReconciliationController;
   private final Timer syncTimer;
   private final Timer enqueueAllTimer;
   private final KafkaTemplate<String, OfferingSyncTask> offeringSyncKafkaTemplate;
@@ -59,12 +61,14 @@ public class OfferingSyncController {
       OfferingRepository offeringRepository,
       ProductWhitelist productAllowlist,
       ProductService productService,
+      CapacityReconciliationController capacityReconciliationController,
       MeterRegistry meterRegistry,
       KafkaTemplate<String, OfferingSyncTask> offeringSyncKafkaTemplate,
       @Qualifier("offeringSyncTasks") TaskQueueProperties taskQueueProperties) {
     this.offeringRepository = offeringRepository;
     this.productAllowlist = productAllowlist;
     this.productService = productService;
+    this.capacityReconciliationController = capacityReconciliationController;
     this.syncTimer = meterRegistry.timer("swatch_offering_sync");
     this.enqueueAllTimer = meterRegistry.timer("swatch_offering_sync_enqueue_all");
     this.offeringSyncKafkaTemplate = offeringSyncKafkaTemplate;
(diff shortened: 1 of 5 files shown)


src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
```diff
@@ -129,11 +133,11 @@ private SyncResult syncOffering(Offering newState) {
     }
 
     // Update to the new entry or create it.
-    offeringRepository.save(newState);
+    offeringRepository.saveAndFlush(newState);
```
lindseyburnett, line 136: Why are we switching to a save and flush here?

redsaz: I was still new with Spring Data at the time this was written, I did not realize that `save` did not necessarily mean that it gets committed to the DB. At this point in the sync call, it is expected that the offering should be updated within the DB.

If my understanding of `save` and `saveAndFlush` are not correct, I can switch it back.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -172,8 +172,7 @@ void syncSubscriptions(String orgId, int offset, int limit) {
     boolean hasMore = subscriptions.size() >= pageSize;
     subscriptions.forEach(this::syncSubscription);
     if (hasMore) {
-      offset = offset + limit;
-      enqueueSubscriptionSync(orgId, offset, limit);
+      enqueueSubscriptionSync(orgId, offset + limit, limit);
```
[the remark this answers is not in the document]
redsaz: This was changed so that the log statement a couple lines down would become correct. It was reporting that the next page of subscriptions were synced, when in fact it was the previous page that got synced.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
redsaz: It may be worth mentioning this isn't wrapped in a `@Transactional` annotation, which I figured meant that I needed to flush it manually like this rather than rely on the auto-magicks that happen in such cases.

lindseyburnett: Got it, makes sense to me!

lindseyburnett (approved):

PR #610, opened 2021-09-28
Read Clowder configuration using JSON Path

awood:
For certain configuration keys, the Clowder JSON provides a list instead
of a map.  Additionally, the list can be in an indeterminate order.  For
example,

```
"endpoints": [
  {
    "app": "rhsm-clowdapp",
    "hostname": "rhsm-clowdapp-api.rhsm.svc",
    "name": "api",
    "port": 8000
  },
  {
    "app": "rhsm-clowdapp",
    "hostname": "rhsm-clowdapp-metrics-worker.rhsm.svc",
    "name": "metrics-worker",
    "port": 8000
  },
  {
    "app": "rbac",
    "hostname": "rbac-service.rhsm.svc",
    "name": "service",
    "port": 8000
  }
]
```

The previous method of locating Clowder configuration values simply
flattened the JSON, allowing us to use an expression like
`endpoints[0].hostname`.  But, if the order of the list is variable, we
can't specify a static index and expect it to work.

This commit takes properties prefixed with `clowder.` and evaluates them
as JSON path expressions.  This approach allows us to us expressions
like `endpoints[?(@.app == 'rbac')]` which filters the list to find
items that have the `app` key equal to `rbac`.

See https://goessner.net/articles/JsonPath/ for the JSON Path syntax.

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonEnvironmentPostProcessor.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonEnvironmentPostProcessor.java
@@ -44,7 +44,7 @@ public class ClowderJsonEnvironmentPostProcessor implements EnvironmentPostProce
   public static final String JSON_RESOURCE_LOCATION =
       "rhsm-subscriptions.clowder.json-resource-location";
 
-  public static final String CLOWDER_STRICT_LOADING = "rhsm-subscriptions.clowder.strictLoading";
+  public static final String CLOWDER_STRICT_LOADING = "rhsm-subscriptions.clowder.strict-loading";
 
   // At a minimum this needs to run after the ConfigDataEnvironmentPostProcessor so that we can read
   // JSON_RESOURCE_LOCATION out of the config files
--- a/swatch-core/src/main/resources/swatch-core/application.yaml
+++ b/swatch-core/src/main/resources/swatch-core/application.yaml
@@ -125,4 +125,4 @@ rhsm-subscriptions:
       kafka-group-id: offering-worker
   clowder:
     json-resource-location: file:/cdapp/cdappconfig.json
-    strictLoading: true
+    strict-loading: true


swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonEnvironmentPostProcessor.java
```diff
@@ -20,97 +20,37 @@
...
-  private static final String SERVLET_ENVIRONMENT_CLASS =
-      "org.springframework.web.context.support.StandardServletEnvironment";
-
-  private static final Set<String> SERVLET_ENVIRONMENT_PROPERTY_SOURCES =
-      new LinkedHashSet<>(
-          Arrays.asList(
-              StandardServletEnvironment.JNDI_PROPERTY_SOURCE_NAME,
-              StandardServletEnvironment.SERVLET_CONTEXT_PROPERTY_SOURCE_NAME,
-              StandardServletEnvironment.SERVLET_CONFIG_PROPERTY_SOURCE_NAME));
+  public static final String CLOWDER_STRICT_LOADING = "rhsm-subscriptions.clowder.strictLoading";
```
kahowell, line 47: Nitpick: Spring best practice is lower-case kebab format according to [the tip in Spring docs](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.typesafe-configuration-properties.relaxed-binding) (note that #624 should be adjusted if you change it).
```suggestion
  public static final String CLOWDER_STRICT_LOADING = "rhsm-subscriptions.clowder.strict-loading";
```

swatch-core/src/main/resources/swatch-core/application.yaml
```diff
@@ -125,3 +125,4 @@ rhsm-subscriptions:
       kafka-group-id: offering-worker
   clowder:
     json-resource-location: file:/cdapp/cdappconfig.json
+    strictLoading: true
```
kahowell, line 128: See other comment about lower kebab case...

```suggestion
    strict-loading: true
```

PR #619, opened 2021-09-30
ENT-4260 Provide more data to HBI

ksynvrit:
This pull request is to address [ENT-4260 ](https://issues.redhat.com/browse/ENT-4260).

Add logic to filter through the IP address and map them to their Network Interface. So that rhsm-conduit as a reporter will provide HBI the following fields:

- Release
- IPv4 addresses
- IPv6 addresses
- Interfaces/NICs
Also, I have added a field for the distro/OS name as well.


swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/ConduitFacts.java
```diff
@@ -76,4 +88,15 @@ public Integer getCpuCores() {
   public Long getMemory() {
     return super.getMemory();
   }
+
+  @Override
```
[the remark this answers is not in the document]
ksynvrit: This was to resolve a issue I had with Sonar qube, since I add a additional field in the ConduitFacts. The check failed as it wants subclass that have different fields than the parent class, to override the equal method.
So this is why I have that there, and some unit test were failing due to them expecting ConsumerInventory, but they got a different argument instead.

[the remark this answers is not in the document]
lindseyburnett: @ksynvrit gotcha....I think the equals should compare network interfaces as well

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/InventoryService.java
```diff
@@ -168,6 +170,16 @@ public OrgInventory getInventoryForOrgConsumers(List<ConduitFacts> conduitFactsF
     return rhsmFactMap;
   }
 
+  private SystemProfileOperatingSystem buildOS(String name) {
```
lindseyburnett, line 173: Super minor, but could we rename this param to `operatingSystemName`?

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/InventoryService.java
```diff
@@ -168,6 +170,16 @@ public OrgInventory getInventoryForOrgConsumers(List<ConduitFacts> conduitFactsF
...
+    var operatingSystems = new SystemProfileOperatingSystem();
```
lindseyburnett, line 174:
```suggestion
   if(!operatingSystemName.toLowerCase().contains("red hat enterprise linux")){
      return null;
   }
   var operatingSystems = new SystemProfileOperatingSystem();
   operatingSystems.setName(SystemProfileOperatingSystem.NameEnum.RHEL);
   return operatingSystems;
  }
```

Also, if there's any potential for operatingSystemName to be null, we'd want to null check that before `.toLowerCase()`

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/ConduitFacts.java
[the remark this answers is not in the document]
ksynvrit: Yeah I tried before but it was cause some unit test to fail because of the the different arguments. So I left this as a work around. If you really think it would should do it, maybe we can tackle it together?

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
```diff
@@ -65,9 +59,13 @@
   private static final BigDecimal KIBIBYTES_PER_GIBIBYTE = BigDecimal.valueOf(1048576);
   private static final BigDecimal BYTES_PER_KIBIBYTE = BigDecimal.valueOf(1024);
   private static final String COMMA_REGEX = ",\\s*";
+  private static final String PERIOD_REGEX = "\\.";
   private static final String NON_HYPHEN_REGEX = "[0-9a-fA-F]{8}([0-9a-fA-F]{4}){3}[0-9a-fA-F]{12}";
   private static final String UUID_REGEX = "[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}";
 
+  public static final String OS_DISTRIBUTION_NAME = "distribution.name";
+  public static final String DISTRIBUTION_VERSION = "distribution.version";
```
kahowell, line 67: nitpick: for consistency:
```suggestion
  public static final String OS_DISTRIBUTION_VERSION = "distribution.version";
```

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
```diff
@@ -65,9 +59,13 @@
...
+  public static final String OS_BIOS_VERSION = "bios.version";
```
kahowell, line 68: Actually, this is not an OS fact, instead `bios.version` is actually redundant with `dmi.bios.version`. I think we should drop `bios.version` handling
```suggestion
```

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
```diff
@@ -215,6 +213,10 @@ private void extractHardwareFacts(Map<String, String> rhsmFacts, ConduitFacts fa
       }
     }
 
+    facts.setOsName(rhsmFacts.get(OS_DISTRIBUTION_NAME));
+    facts.setOsVersion(rhsmFacts.get(DISTRIBUTION_VERSION));
+    facts.setOsKernelVersion(rhsmFacts.get(OS_BIOS_VERSION));
```
kahowell, line 218: See above comment
```suggestion
```

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
```diff
@@ -330,6 +333,70 @@ protected void extractIpAddresses(Map<String, String> rhsmFacts, ConduitFacts fa
...
+      }
+    }
+    // creates a lo interface if ips exist for it, but no mac was given
+    checkLoopbackIPs(nicSet, rhsmFacts);
+    return nicSet;
+  }
+
+  private void mapInterfaceIps(
+      NetworkInterface networkInterface, Map<String, String> facts, String suffix) {
+    String regex = MAC_PREFIX + networkInterface.getName() + suffix;
```
kahowell, line 357: Nitpick: this isn't a regex; I'd prefer a different variable name, such as `prefix`.

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/ConduitFacts.java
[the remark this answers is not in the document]
kahowell: I tried this locally too, I see why it happens (`null != new ArrayList<>()`), and actually I think the correct thing here to do is to leave the networkinterfaces as null if we have 0 NIC entries. Leaving a suggestion for where to do this.

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
```diff
@@ -280,6 +282,7 @@ private void extractNetworkFacts(Map<String, String> rhsmFacts, ConduitFacts fac
       facts.setFqdn(fqdn);
     }
 
+    facts.setNetworkInterfaces(populateNICs(rhsmFacts));
```
kahowell, line 285: We should not set `networkInterfaces` if `populateNics(rhsmFacts)` returns an empty list.
```suggestion
    List<NetworkInterface> networkInterfaces = populateNICs(rhsmFacts);
    if (!networkInterfaces.isEmpty()) {
      facts.setNetworkInterfaces(networkInterfaces);
    }
```

This will address the test failures you were seeing with comparing `networkInterfaces` in `equals`.

PR #620, opened 2021-10-01
ENT-4406 Speed up per-org subscription syncs

redsaz:
The following changes are made to improve performance on per-org
subscription syncs:

- Use SubscriptionCapacityRepository's
  findByKeyOwnerIdAndKeySubscriptionIdIn(
  String ownerId, List<String> subscriptionIds) method instead of
  findByKeySubscriptionId(String subscriptionId), since the former
  can take advantage of the (owner_id, product_id, subscription_id)
  primary key
- Subs expired longer than 2 months or future dated subs more than 2
  months from now will not be synced.
- Fetch more subs-per-page (again)

To test locally:

```bash
    curl 'http://localhost:8080/actuator/hawtio/jolokia/?maxDepth=7&maxCollectionSize=50000&ignoreErrors=true&canonicalNaming=false' \
        -H 'Accept: application/json' \
        --data-raw '{"type":"exec","mbean":"org.candlepin.subscriptions.subscription:name=subscriptionJmxBean,type=SubscriptionJmxBean","operation":"syncSubscriptionsForOrg(java.lang.String)","arguments":["$ORG_ID"]}'
```

A log message should appear that shows how many subs were fetched and
how many were synced.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
@@ -26,6 +26,7 @@ import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
@@ -94,7 +95,10 @@ public class CapacityReconciliationController {
 
     Collection<SubscriptionCapacity> newCapacities = mapSubscriptionToCapacities(subscription);
     reconcileSubscriptionCapacities(
-        newCapacities, subscription.getSubscriptionId(), subscription.getSku());
+        newCapacities,
+        subscription.getOwnerId(),
+        subscription.getSubscriptionId(),
+        subscription.getSku());
   }
 
   @Transactional
@@ -128,11 +132,17 @@ public class CapacityReconciliationController {
   }
 
   private void reconcileSubscriptionCapacities(
-      Collection<SubscriptionCapacity> newCapacities, String subscriptionId, String sku) {
+      Collection<SubscriptionCapacity> newCapacities,
+      String ownerId,
+      String subscriptionId,
+      String sku) {
 
     Collection<SubscriptionCapacity> toSave = new ArrayList<>();
     Map<SubscriptionCapacityKey, SubscriptionCapacity> existingCapacityMap =
-        subscriptionCapacityRepository.findByKeySubscriptionId(subscriptionId).stream()
+        subscriptionCapacityRepository
+            .findByKeyOwnerIdAndKeySubscriptionIdIn(
+                ownerId, Collections.singletonList(subscriptionId))
+            .stream()
             .collect(Collectors.toMap(SubscriptionCapacity::getKey, Function.identity()));
 
     if (productWhitelist.productIdMatches(sku)) {
--- a/src/main/java/org/candlepin/subscriptions/subscription/StubSearchApi.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/StubSearchApi.java
@@ -20,9 +20,11 @@
  */
 package org.candlepin.subscriptions.subscription;
 
+import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.List;
 import org.candlepin.subscriptions.subscription.api.model.Subscription;
+import org.candlepin.subscriptions.subscription.api.model.SubscriptionProduct;
 import org.candlepin.subscriptions.subscription.api.resources.SearchApi;
 
 /** Stub version of the SearchApi for the Subscription service for local testing. */
@@ -46,6 +48,11 @@ public class StubSearchApi extends SearchApi {
   }
 
   private Subscription createData() {
-    return new Subscription().subscriptionNumber("2253591");
+    var now = OffsetDateTime.now();
+    return new Subscription()
+        .subscriptionNumber("2253591")
+        .effectiveStartDate(now.minusYears(10).toEpochSecond() * 1000L)
+        .effectiveEndDate(now.plusYears(10).toEpochSecond() * 1000L)
+        .subscriptionProducts(List.of(new SubscriptionProduct().sku("sku")));
   }
 }
--- a/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
@@ -27,6 +27,7 @@ import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.transaction.Transactional;
 import lombok.extern.slf4j.Slf4j;
 import org.candlepin.subscriptions.capacity.CapacityReconciliationController;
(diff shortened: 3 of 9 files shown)


src/main/java/org/candlepin/subscriptions/capacity/CapacityReconciliationController.java
```diff
@@ -128,11 +132,17 @@ public void enqueueReconcileCapacityForOffering(String sku) {
...
+      String ownerId,
+      String subscriptionId,
+      String sku) {
 
     Collection<SubscriptionCapacity> toSave = new ArrayList<>();
     Map<SubscriptionCapacityKey, SubscriptionCapacity> existingCapacityMap =
-        subscriptionCapacityRepository.findByKeySubscriptionId(subscriptionId).stream()
+        subscriptionCapacityRepository
+            .findByKeyOwnerIdAndKeySubscriptionIdIn(
+                ownerId, Collections.singletonList(subscriptionId))
```
redsaz, line 144: Switching off of findByKeySubscriptionId and on to findByKeyOwnerIdAndKeySubscriptionIdIn is what provides the majority of our speedup, as it makes use of the compound primary key (owner_id, product_id, subscription_id). 

Performance testing on a particular org improved time-to-sync-100-subs from 39.9 seconds to 2.9 seconds, an over 10x speed up.

For that same test org, Performance tests showed time-to-sync the entire org went from 36m 47s to 2m 52s, again an over 10x speed up.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -49,7 +50,7 @@
 @Component
 @Slf4j
 public class SubscriptionSyncController {
-  private static final int SUBS_PAGE_SIZE = 100;
+  private static final int SUBS_PAGE_SIZE = 1000;
```
redsaz, line 53: Previously we did larger-than-100-subs-per-page, but backed off to 100, as the time to complete 2000 subs would take over an hour. 1 hour is how long kafka will wait for a job to complete before declaring it stale, and would push the job back onto the queue.

Thanks to the query change used in CapacityReconcilliationController, the bottleneck is no longer in the database, but in the time to fetch subs from upstream. So we can fetch A LOT more subs and still not run into the 1 hour kafka message limit.

However, there's diminishing returns in fetching more than 1000 subs at a time, so we cap to that. Switching from 100-subs-per-page to 1000, for the test-org with lots of subs, improved fetch time up to 3x.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -158,32 +159,46 @@ public void syncSubscription(String subscriptionId) {
...
         syncDuration.toMillis());
   }
 
+  private boolean shouldSyncSub(Subscription sub) {
+    // Reject subs expired long ago, or subs that won't be active quite yet.
+    OffsetDateTime now = clock.now();
+    long earliestAllowedFutureStartDate = now.plusMonths(2).toEpochSecond() * 1000;
+    long latestAllowedExpiredEndDate = now.minusMonths(2).toEpochSecond() * 1000;
+    return sub.getEffectiveStartDate() < earliestAllowedFutureStartDate
+        && sub.getEffectiveEndDate() > latestAllowedExpiredEndDate;
```
redsaz, line 199: By limiting to only refreshing subscriptions that have only recently expired or are near-future-dated, this improved time-to-sync on my local machine with the test-org with lots of subs by 2x.

kahowell: 1. Can you please update the StubSearchApi to populate these fields? Otherwise, I get NPE when using the stub.
2. If we are 100% sure that the API won't return `null`, we should probably mark the properties as required in the spec. This will prevent intellij (and possibly other tools, but not sonar this time?) from freaking out over possible NPE.
image

3. Can you please put the cutoffs in configuration?
 a. Add Duration properties to `SubscriptionServiceProperties` for both.
 b. Add entries to `application.yaml` for both (with defaults).
 c. Add template params to `templates/rhsm-subscriptions-capacity-ingress.yml` along with corresponding pod env vars (similar to `SUBSCRIPTION_URL`).

kahowell (changes requested):
Great work! Have a few small requests before merging.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
redsaz: 1. Will do!
2. I had to double-check... Turns out effectiveStartDate and effectiveEndDate *CAN* be null. I will investigate this further with the subprod team, as I don't see how that is possible... In every instance, at least one subscription product had a non-null activeStartDate and activeEndDate. In the meantime, I will allow such subs to "slip past" this filter for now, as that is how it would've happened before this code was added.
3. I debated adding this, and figured if somebody wanted it, it'd come up in the PR. And that's exactly what happened, so woot! I'll get it added.

redsaz: Revised answer:
1. It is possible for effecdtiveStartDate and effectiveEndDate to be null, but they *SHOULDN'T*. That is, if they come back null from the upstream server, then that indicates bad data and needs to be fixed. I added logging that will tell us when this happens so we can identify the subs and ask for them to be corrected upstream. When we encounter such a sub, we will skip syncing it.
2. Added unit tests to cover the null case of these dates, and to skip them if so.
3. Cutoffs are now configurable:
    a. I used Period instead of Duration, since Period is date-scale and allows for years, months, and days, whereas Duration is time-scale and allows for seconds, minutes, hours, days. Since subscription dates are dates (the fact that it is represented upstream as an exact timestamp is a by-product of the tech used, as I understand it), Period seems a better fit.
    b. added
    c. added

src/main/java/org/candlepin/subscriptions/subscription/StubSearchApi.java
```diff
@@ -46,6 +47,10 @@ public Subscription getSubscriptionById(String id) throws ApiException {
   }
 
   private Subscription createData() {
-    return new Subscription().subscriptionNumber("2253591");
+    var now = OffsetDateTime.now();
+    return new Subscription()
+        .subscriptionNumber("2253591")
```
kahowell, line 52:
```suggestion
        .subscriptionNumber("2253591")
        .subscriptionProducts(List.of(new SubscriptionProduct().sku("sku")))
```

src/main/resources/application.yaml
```diff
@@ -75,14 +75,16 @@ rhsm-subscriptions:
...
-    backOffInitialInterval: ${SUBSCRIPTION_BACKOFF_INITIAL_INTERVAL:1s}
-    maxRetryAttempts: ${SUBSCRIPTION_MAX_RETRY_ATTEMPTS:4}
-    pageSize: ${SUBSCRIPTION_PAGE_SIZE:500}
+    keystore-password: ${SUBSCRIPTION_KEYSTORE_PASSWORD:changeit}
+    max-connections: ${SUBSCRIPTION_MAX_CONNECTIONS:100}
+    back-off-initial-interval: ${SUBSCRIPTION_BACKOFF_INITIAL_INTERVAL:1s}
+    max-retry-attempts: ${SUBSCRIPTION_MAX_RETRY_ATTEMPTS:4}
+    page-size: ${SUBSCRIPTION_PAGE_SIZE:1000}
+    ignore-expired-older-than: ${SUBSCRIPTION_IGNORE_EXPIRED_OLDER_THAN:P2M}
+    ignore-starting-later-than: ${SUBSCRIPTION_IGNORE_STARTING_LATER_THAN:P2M}
```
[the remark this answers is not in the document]
redsaz: Ah, good to know, I was not aware. Fixed!

src/main/java/org/candlepin/subscriptions/subscription/StubSearchApi.java
redsaz: Added

kahowell (approved):

PR #637, opened 2021-10-20
ENT-4460: Add script to reach hawtio via oc-proxy and nginx

kahowell:
https://issues.redhat.com/browse/ENT-4460

Essentially this uses nginx to force a couple headers and "serve" hawtio on the path turnpike would.

The script emits a message like:

```
Proxying pod marketplace-worker-241-chqb4 port 8080 at path /app/rhsm-marketplace/actuator/hawtio so that it can be reached via http://localhost:8085/app/rhsm-marketplace/actuator/hawtio
```

which shows the URL hawtio can be reached at.

### DIFF
--- a/bin/hawtio-proxy.sh
+++ b/bin/hawtio-proxy.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+# This script does all the proxy magic we need to access hawtio on a cluster
+NGINX_PORT=${NGINX_PORT:-8085}
+FORWARD_PORT=${FORWARD_PORT:-9095}
+pod=$1
+if [ $# -lt 1 ]; then
+  echo "Usage: $0 <pod>"
+  exit 1
+fi
+if [ -z "$(which jq)" ]; then
+  echo "jq not found (hint: sudo dnf install -y jq)"
+  exit 1
+fi
+nginx_config=$(mktemp)
+hawtio_base_path=/actuator/hawtio
+pod_json=$(oc get pod/$pod -o json)
+hawtio_path_override="$(echo "$pod_json" | jq -r '.spec.containers[].env[] | select(.name=="HAWTIO_BASE_PATH") | .value')"
+if [ -n "$hawtio_path_override" ]; then
+  hawtio_base_path="$hawtio_path_override"
+fi
+container_port=8080
+web_port=$(echo "$pod_json" | jq -r '.spec.containers[].ports[] | select(.name=="web") | .containerPort')
+if [ -n "$web_port" ]; then
+  container_port=$web_port
+fi
+oc port-forward pod/$pod $FORWARD_PORT:$container_port&
+port_forward_pid=$!
+auth = $(echo -n '{"identity":{"associate":{"email":"test@example.com"},"auth_type":"saml-auth","type":"Associate"}}' | base64 -w0)
+cat <<EOF > $nginx_config
+events {}
+
+http {
+    server {
+        listen       $NGINX_PORT;
+
+        location $hawtio_base_path {
+            proxy_pass http://localhost:$FORWARD_PORT/actuator/hawtio;
+            proxy_set_header origin "example.cloud.redhat.com";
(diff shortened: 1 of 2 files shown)


bin/hawtio-proxy.sh
```diff
@@ -0,0 +1,51 @@
...
+# This script does all the proxy magic we need to access hawtio on a cluster
+NGINX_PORT=${NGINX_PORT:-8085}
+FORWARD_PORT=${FORWARD_PORT:-9095}
+pod=$1
+if [ $# -lt 1 ]; then
+  echo "Usage: $0 <pod>"
+  exit 1
+fi
+if [ -z "$(which jq)" ]; then
+  echo "jq not found (hint: sudo dnf install -y jq)"
```
awood, line 11: Do you want to bomb out here?  As it is, the script will keep running and then fail when `jq` can't be found on line 16.

bin/hawtio-proxy.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+nginx_config=$(mktemp)
+hawtio_base_path=/actuator/hawtio
+pod_json=$(oc get pod/$pod -o json)
+hawtio_path_override=$(echo "$pod_json" | jq -r '.spec.containers[].env[] | select(.name=="HAWTIO_BASE_PATH") | .value')
+if [ -n "$hawtio_path_override" ]; then
+  hawtio_base_path=$hawtio_path_override
```
awood, line 18: We might want to double quote `hawtio_path_override` here and on line 16.  I'm not sure if there's the potential for any special shell characters in there or not.  I would err on the side of quoting it myself.

bin/hawtio-proxy.sh
```diff
@@ -0,0 +1,51 @@
...
+events {}
+
+http {
+    server {
+        listen       $NGINX_PORT;
+
+        location $hawtio_base_path {
+            proxy_pass http://localhost:$FORWARD_PORT/actuator/hawtio;
+            proxy_set_header origin "example.cloud.redhat.com";
+            proxy_set_header x-rh-identity "eyJpZGVudGl0eSI6eyJhc3NvY2lhdGUiOnsiZW1haWwiOiJ0ZXN0QGV4YW1wbGUuY29tIn0sImF1dGhfdHlwZSI6InNhbWwtYXV0aCIsInR5cGUiOiAiQXNzb2NpYXRlIn19Cg==";
```
awood, line 37: I would maybe just assign this to a variable:

$auth = $(echo -n '{"identity":{"associate":{"email":"test@example.com"},"auth_type":"saml-auth","type": "Associate"}}' | base64 -w 0)

That was people can a) see what it is and b) easily modify it if they need to.

bin/hawtio-proxy.sh
kahowell: Done

awood:
> @awood I tweaked the script per your feedback, please take another look when you get a chance.

Looks good, but I noticed that `.gitignore` has "bin" in it.  I downloaded this PR as a patch and applied it and was kind of surprised that `git status` didn't show it as a new file!

@lindseyburnett `bin` is added under a section marked "VS Code".  I have `bin/test` and `bin/main` directories that looks like VSCode might have created when I was using it for minikube dev work.  In my case looks like an [issue](https://github.com/redhat-developer/vscode-java/issues/634) in Language Support for Java by Red Hat might be responsible.  Can we just ignore those directories?  Otherwise, we'll need to carve out exceptions for the bin scripts `hawtio-proxy.sh` and `insert-mock-hosts`

This will do the needful if you want to apply it @kahowell 

```diff
diff --git a/.gitignore b/.gitignore
index df6d59da..c0095f1d 100644
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,9 @@ config/
 #VS Code
 .project
 .settings
-bin
+/**/bin/*
+!bin/hawtio-proxy.sh
+!bin/insert-mock-hosts
 .vscode
 .editorconfig
```

PR #654, opened 2021-11-01
ENT-4463 Marketplace Fixes for RHOSAK

mstead:
There were quite a few things addressed in this PR in order to send the usage to marketplace.

* Removed the ProductProfileRegistry from the MarketplaceSubscriptionIdProvider
so that product config can all be done in the tag profile. Also updated the
offering lookup query to be based on Offering name instead of role.
* Added rhosak tag to eligible-swatch-product-ids in the marketplace config so that
it will be included in metrics that are sent to marketplace.
* Updated the product name mappings in the tag profile config to reflect the values
that will come from the product service to be ready for turning on subscription
table functionality. Also added a RHOSAK mapping.
* Updated the offering product names in DB for OCP and OSD, as well as added RHOSAK
* Set Offering name from PRODUCT_NAME attribute instead of the DESCRIPTION.
* I also created a small python script to import Events from a CSV file. It is very cumbersome to use the Event JMX API directly and this makes it a little easier to view the event time line without sifting through JSON.

### Testing
The following shows how to test this PR against an account with actual cluster instances running, and allows verifying that data will be sent to Marketplace.

https://docs.engineering.redhat.com/display/ENT/Testing+RHOSAK+metering+in+a+development+environment

### DIFF
--- a/bin/import_events.py
+++ b/bin/import_events.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+import sys
+import argparse
+import csv
+import requests
+
+
+try:
+    import jsonpickle
+except ImportError as e:
+    print('Please install jsonpickle: python -m pip install --user "jsonpickle"')
+    sys.exit(1)
+
+
+class Measurement:
+    def __init__(self, uom, value):
+        self.uom = uom
+        self.value = value
+
+
+class Event:
+    def __init__(self, event_type, account, instance, timestamp, expiration, role, sla, uom, value, service_type):
+        self.event_source = "prometheus"
+        self.event_type = "snapshot_" + event_type
+        self.account_number = account
+        self.instance_id = instance
+        self.timestamp = timestamp
+        self.expiration = expiration
+        self.display_name = instance
+        self.measurements = [ Measurement(uom, value)]
+        self.role = role
+        self.sla = sla
+        self.service_type = service_type
+
+
+def import_events(url, csv_path, dry_run):
+    events = parse_csv(csv_path)
+    print("Found {num} records".format(num=len(events)))
+    return post_events(url, events, dry_run=dry_run)
+
+
+def parse_csv(csv_path):
+    with open(csv_path, mode='r') as csv_file:
+        reader = csv.DictReader(csv_file)
+        events = []
+        for row in reader:
+            events.append(Event(**row))
+        return events
+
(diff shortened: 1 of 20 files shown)


bin/import_events.py
```diff
@@ -0,0 +1,90 @@
...
+    return post_events(url, events, dry_run=dry_run)
+
+
+def parse_csv(csv_path):
+
+    with open(csv_path, mode='r') as csv_file:
+        reader = csv.DictReader(csv_file)
+        events = []
+        for row in reader:
+            events.append(create_event(row['event_type'], row['account'], row['instance'], row['timestamp'],
```
awood, line 20: Maybe just create an event object?  Similar to example 1 in https://stackoverflow.com/a/2466380/6124862

mstead: DONE - Thanks!

mstead: Forgot to clean  up...

mstead: Now it is done! :)

awood (approved):
I think the Python looks good, but I would ask for a second review on the RHOSAK specific stuff since I'm not really plugged into that design right now.

src/main/java/org/candlepin/subscriptions/product/UpstreamProductData.java
```diff
@@ -61,8 +61,7 @@
     SERVICE_TYPE,
     PRODUCT_FAMILY,
     USAGE,
-    /** Name of Offering comes from opProd description field, not the PRODUCT_NAME attribute. */
-    X_DESCRIPTION,
+    PRODUCT_NAME,
```
kahowell, line 64: In order to not break in-flight subscription table changes, please add a `description` column to the DB, and populate `description` appropriately here. Then update `SubscriptionCapacityView` to use `o.description` instead of `o.product_name` in its query. (This will ensure the display name of the subscription remains description while still allowing us to use `PRODUCT_NAME` as we wish here).

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/product/UpstreamProductData.java
mstead: DONE

kahowell (approved):

PR #656, opened 2021-11-02
ENT-4383: Implementation of InstanceAPI

ksynvrit:
This PR is to address [ENT-4383](https://issues.redhat.com/browse/ENT-4383)

### Testing Step for Instance
Profiles use for Testing
```
DEV_MODE=true;
USER_USE_STUB=true;
PROM_URL=http://localhost:8082/api/metrics/v1/telemeter/api/v1
```

Get connected to the testing telemeter instance:
Uncomment 'Eval' in application-openshift-metering-worker.yaml so that the Kafka cluster metrics get picked up by the query.
```
min_over_time(#{metric.queryParams[prometheusMetadataMetric]}{product="#{metric.queryParams[product]}", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None|Eval"}[1h])
```

1. Gather measurements from Telemeter:
```
curl 'http://localhost:8080/actuator/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.metering.jmx:name=meteringJmxBean,type=MeteringJmxBean","operation":"performCustomMeteringForAccount(java.lang.String,java.lang.String,java.lang.String,java.lang.Integer)","arguments":["6043640", "rhosak", "2021-10-30T00:00:00Z", 15760]}'
```
2. Start TallyJMX:
```
curl 'http://localhost:8080/actuator/hawtio/jolokia' \
  -d '{
    "type":"exec",
    "mbean":"org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean",
    "operation":"tallyAccountByHourly(java.lang.String,java.lang.String,java.lang.String)",
    "arguments":["6043640", "2021-10-15T00:00Z", "2021-10-31T00:00Z"]
  }'
```
3. Use this curl to test Instance API:
```
curl -X 'GET' \
  'http://localhost:8080/api/rhsm-subscriptions/v1/instances/products/rhosak?beginning=2021-10-15T00%3A00Z&ending=2021-10-31T00%3A00Z' \
  -H 'accept: application/json' \
  -H 'x-rh-identity: ewogICAgImlkZW50aXR5IjogewogICAgICAgICJhY2NvdW50X251bWJlciI6ICI2MDQzNjQwIiwKICAgICAgICAidHlwZSI6ICJVc2VyIiwKICAgICAgICAidXNlciIgOiB7CiAgICAgICAgICAgICJpc19vcmdfYWRtaW4iOiB0cnVlCiAgICAgICAgfSwKICAgICAgICAiaW50ZXJuYWwiIDogewogICAgICAgICAgICAib3JnX2lkIjogIjYwNDM2NDAiCiAgICAgICAgfQogICAgfQp9'
```
This won't work if you try to Authorize with the default account ID as I'm using a specific account number for RHOSAK.

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -690,6 +690,83 @@ paths:
           $ref: '#/components/responses/InternalServerError'
       tags:
         - ingress
+  /instances/products/{product_id}:
+    parameters:
+      - name: product_id
+        description: The ID for the product we wish to query
+        in: path
+        required: true
+        schema:
+          $ref: '#/components/schemas/ProductId'
+      - name: offset
+        in: query
+        schema:
+          type: integer
+        description: "The number of items to skip before starting to collect the result set"
+      - name: limit
+        in: query
+        schema:
+          type: integer
+          minimum: 1
+          maximum: 100
+        description: "The numbers of items to return"
+    get:
+      operationId: getInstancesByProduct
+      parameters:
+        - name: sla
+          in: query
+          schema:
+            $ref: '#/components/schemas/ServiceLevelType'
+          description: "Include only hosts for the specified service level."
+        - name: usage
+          in: query
+          schema:
+            $ref: '#/components/schemas/UsageType'
+          description: "Include only hosts for the specified usage level."
+        - name: display_name_contains
+          description: Include only hosts containing the specified display name. Passing an empty string behaves the same way as passing null
+          schema:
+            type: string
+            #Default to empty string to prevent complex jpql statement later
+            default: ''
+          in: query
+        - name: beginning
+          in: query
+          required: false
+          schema:
+            type: string
+            format: date-time
+          description: "Defines the start of the report period. Dates should be provided in ISO 8601
+                     format but the only accepted offset is UTC. E.g. 2017-07-21T17:32:28Z.
+                     Only applicable for OpenShift-dedicated-metrics and OpenShift-metrics products"
+        - name: ending
+          in: query
+          required: false
+          schema:
+            type: string
+            format: date-time
+          description: "Defines the end of the report period. Dates should be provided in ISO 8601
+                     format but the only accepted offset is UTC. E.g. 2017-07-21T17:32:28Z.
+                     Only applicable for OpenShift-dedicated-metrics and OpenShift-metrics products"
+        - name: sort
+          in: query
+          schema:
+            $ref: "#/components/schemas/InstanceReportSort"
+          description: "What property to sort by (optional)"
+        - name: dir
+          in: query
+          schema:
+            $ref: "#/components/schemas/SortDirection"
+          description: "Which direction to sort by (default: asc)"
+      tags:
+        - instances
+      responses:
+        '200':
+          description: The request for instances was successful.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/InstanceResponse'
   /subscriptions/products/{product_id}:
     description: "Operations for total capacity by SKU for all of the owner's active subscriptions for given Swatch product ID."
     parameters:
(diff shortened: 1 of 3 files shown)


api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -1450,6 +1527,106 @@ components:
...
+        count: 10
+        product: RHEL
+        service_level: Premium
+        usage: Production
+        measurements:
+          - Instance-Hours
+          - Storage-Gigabytes
+          - Transfer-Gigabytes
+    InstanceResponse:
+      title: Root Type for DefaultDataResponse
```
lindseyburnett, line 1558: Is this title accurate?  Not seeing any other references to `DefaultDataResponse`

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -1450,6 +1527,106 @@ components:
         cloud_provider:
           description: If applicable, the cloud provider for this host.
           type: string
+    InstanceMeta:
+      title: Root Type for InstanceMeta
+      description: ''
```
lindseyburnett, line 1532: We should fill out description

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -1450,6 +1527,106 @@ components:
...
+        product: RHEL
+        service_level: Premium
+        usage: Production
+        measurements:
+          - Instance-Hours
+          - Storage-Gigabytes
+          - Transfer-Gigabytes
+    InstanceResponse:
+      title: Root Type for DefaultDataResponse
+      description: ''
```
lindseyburnett, line 1559: We should fill out description

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -1450,6 +1527,106 @@ components:
...
+        - last_seen
+        - Cores
+        - Sockets
+        - Core-seconds
+        - Instance-hours
+        - Storage-gibibytes
+        - Transfer-gibibytes
+    InstanceData:
+      title: Root Type for Instance
+      description: ''
```
lindseyburnett, line 1603: We should fill out description

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -1622,4 +1799,4 @@ components:
             - subscription_type
 
 security:
-  - 3ScaleIdentity: []
+  - 3ScaleIdentity: []
```
lindseyburnett, line 1802: File should end in empty line

src/main/resources/application-openshift-metering-worker.yaml
```diff
@@ -11,6 +11,7 @@ rhsm-subscriptions:
             #{metric.queryParams[prometheusMetric]}
             * on(_id) group_right
             min_over_time(#{metric.queryParams[prometheusMetadataMetric]}{product="#{metric.queryParams[product]}", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None"}[1h])
+          #   min_over_time(#{metric.queryParams[prometheusMetadataMetric]}{product="#{metric.queryParams[product]}", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None|Eval"}[1h])
```
lindseyburnett, line 14: I think we should delete this so we don't forget to remove it later

src/test/java/org/candlepin/subscriptions/resource/InstancesResourceTest.java
```diff
@@ -0,0 +1,106 @@
...
+            null,
+            ServiceLevelType.PREMIUM,
+            UsageType.PRODUCTION,
+            null,
+            null,
+            null,
+            InstanceReportSort.DISPLAY_NAME,
+            null);
+
+    // Reassigned values for measurements as logic cannot be mocked correctly.
```
lindseyburnett, line 102: Can you elaborate on what can't be mocked correctly?  It feels weird to take the "actual" value, and then overwrite it before comparing it to "expected".

lindseyburnett (changes requested):

src/test/java/org/candlepin/subscriptions/resource/InstancesResourceTest.java
ksynvrit: The `asTallyHostViewApiInstance` method is used to populate measurements, If I remove the line that creates a empty list the method would return a list of nulls `[null,null,null]`for each of the uoms. 
I can't populate the expected with a list of nulls. So I opted to use a work around.

src/main/resources/application-openshift-metering-worker.yaml
ksynvrit: Agreed. I put it there for convenience of testing the InstanceAPI for rhosak.

src/test/java/org/candlepin/subscriptions/resource/InstancesResourceTest.java
lindseyburnett: I'm going to look into this a little more locally.  If we want/expect an empty list rather than a list of nulls (which makes sense to me), my gut is telling me to look into the logic returning the list of nulls and maybe update it to return an empty list instead of hacking the unit tests.

Thoughts? @ksynvrit @kahowell

src/main/resources/application-openshift-metering-worker.yaml
lindseyburnett: Instead of checking in commented-out properties for convenience of testing, you can create a `config/application.yaml` locally in the project and override whatever properties you'd like.  Files in this directory are omitted via the .gitignore file.

```yaml
rhsm-subscriptions:
  metering:
    prometheus:
      metric:
        queryTemplates:
          default: >-
            #{metric.queryParams[prometheusMetric]}
            * on(_id) group_right
            min_over_time(#{metric.queryParams[prometheusMetadataMetric]}{product="#{metric.queryParams[product]}", ebs_account="#{runtime[account]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None|Eval"}[1h])
```

lindseyburnett (changes requested):

src/test/java/org/candlepin/subscriptions/resource/InstancesResourceTest.java
kahowell: > I can't populate the expected with a list of nulls.

Why not? I would expect nulls here actually.

ksynvrit: I think if a measurement isn't there it should be null. Like in the example output in [ENT-4383](https://issues.redhat.com/browse/ENT-4383). The way it's implemented now. I believe it's consistent with the example in the card. I don't want to adjust the implementation of the logic to return an empty list, if it's only relevant for the unit test. If there's a case where we would want to return a empty list, than I' all for add the logic to return an empty list.

src/test/java/org/candlepin/subscriptions/resource/InstancesResourceTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    Mockito.when(
+            repository.findAllBy(
+                any(), any(), any(), any(), any(), anyInt(), anyInt(), any(), any(), any()))
+        .thenReturn(new PageImpl<>(List.of(host)));
+    var data = new InstanceData();
+    data.setId(host.getInstanceId());
+    data.setDisplayName(host.getDisplayName());
+    data.setLastSeen(host.getLastSeen());
+
```
lindseyburnett, line 77:
```suggestion

    List<Double> expectedMeasurements = new ArrayList<>();
    for (int i = 0; i < 3; i++) {
      expectedMeasurements.add(null);
    }

    data.setMeasurements(expectedMeasurements);
```

lindseyburnett (changes requested):

src/test/java/org/candlepin/subscriptions/resource/InstancesResourceTest.java
ksynvrit: > > I can't populate the expected with a list of nulls.
> 
> Why not? I would expect nulls here actually.
I'm working on adjusting it so excepted get a list of nulls. It's was just I could directly add a list of nulls from the setter method.

lindseyburnett: I agree with not changing the implementation to return an empty list, based on the additional information.

Since this is the case, please do not override the returned data before asserting equals.  Update the expected `data` to include a list of three null values.  I added a simple example of how to achieve this using github's suggestion feature at line 77.

PR #659, opened 2021-11-04
ENT-4052: Validate/fix capacity ingress clowder deployment

lindseyburnett:
The environment ephemeral-41 has an rhsm-capacity-ingress pod running successfully that was done using the steps added to the clowder.md file as part of this PR.

### DIFF
--- a/deploy/clowder.md
+++ b/deploy/clowder.md
@@ -177,8 +177,7 @@ but here are some essentials:
   $ podman push quay.io/awood/swatch-system-conduit
   ```
 
-* When you deploy with bonfire during development, specify the image and
-  image tag you want to use like so:
+* When you deploy with bonfire during development, you'll want to specify the image and image tag you want to use like so:
 
   ```
   `bonfire deploy rhsm-subscriptions -n NAMESPACE --no-remove-resources=all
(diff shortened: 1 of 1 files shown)


deploy/clowder.md
```diff
@@ -219,3 +218,19 @@ but here are some essentials:
...
+```bash
+curl https://gitlab.cee.redhat.com/rhsm/swatch-product-allowlist/-/raw/main/templates/capacity-allowlist.yml | oc process -f - | oc apply -f -
+```
+## bonfire "deploy" command and namespace reservation
+If you use `bonfire deploy` without already having a namespace reserved, it will reserve the namespace for you **BUT** if the app doesn't start up in the default amount of time, bonfire will take down/give up the namespace it reserved to begin with.  To get around this, you can manually reserve the namespace, then pass `-n <NAMESPACE>` as an argument when running `bonfire deploy`.
+
+# TL;DR Quickstart Steps
+1. Start bonfire virtual environment
+2. Reserve a namespace
+3. Apply the `capacity-allowlist` configmap
```
awood, line 235: Should we add a step to import all the other needed stuff?  The `tls`, `pinhead`, and `ingress` secrets for example.  I have a .gitignored file that has all these but we don't have a great way yet to share that stuff around and I had to create the file by combing through Vault to get the secrets.

awood (changes requested):

deploy/clowder.md
lindseyburnett: We automated the population of `tls`,`pinhead`, and `ingress` as part of [ENT-4459](https://issues.redhat.com/browse/ENT-4459).

awood (approved):

PR #660, opened 2021-11-04
ENT-4053 Add activeDeadlineSeconds to jobs, add sidecars

kahowell:
https://issues.redhat.com/browse/ENT-4053

activeDeadlineSeconds ensures they are terminated if they hang, and for metrics-cron, it ensures that the sidecar doesn't keep the job from finishing.

Note that activating sidecars requires a change the ClowdEnvironment. I added a note to clowder.md about that.

Note that offering-sync needs the allowlist populated (see #659).

### DIFF
--- a/deploy/rhsm-eph-clowdenv.yaml
+++ b/deploy/rhsm-eph-clowdenv.yaml
@@ -1,3 +1,5 @@
+# Derived from https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/ephemeral-cluster-clowdenvironment.yaml
+# but with the token-refresher sidecar enabled
 apiVersion: v1
 kind: Template
 metadata:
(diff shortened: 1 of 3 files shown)


deploy/rhsm-eph-clowdenv.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: v1
```
awood, line 1: Would you mind adding something like this that kind of documents why this file exists
```suggestion
# Derived from https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/ephemeral-cluster-clowdenvironment.yaml
# but with the token-refresher sidecar enabled
apiVersion: v1
```

awood (changes requested):

deploy/rhsm-eph-clowdenv.yaml
kahowell: :+1:

awood (approved):

PR #664, opened 2021-11-08
ENT-4530: Splunk HEC MVP

lindseyburnett:
Relates to: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/29062

Testing:

1.  Pull branch
2.  Build image and push to quay
3.  Deploy to ephemeral environments using bonfire
```
bonfire deploy rhsm-subscriptions \
--no-remove-resources=all \
-i quay.io/lburnett/rhsm=splunk_hec \
-p rhsm-subscriptions/IMAGE=quay.io/lburnett/rhsm \
-i quay.io/awood/swatch-system-conduit=latest \
-p rhsm-subscriptions/CONDUIT_IMAGE=quay.io/awood/swatch-system-conduit \
--timeout 3600
```
4.  Take note of ephemeral environment it deployed to and go see that the logs are getting sent to splunk

https://splunk.corp.redhat.com/en-US/app/search/search
`index=rh_rhsm namespace="ephemeral*"`

### DIFF
--- a/swatch-core/build.gradle
+++ b/swatch-core/build.gradle
@@ -6,6 +6,7 @@ plugins {
 dependencies {
 
     implementation "com.splunk.logging:splunk-library-javalogging:1.11.0"
+    implementation "org.codehaus.janino:janino:3.1.6"
 
     implementation project(":api") // for util methods in ResourceUtils
     implementation project(":clients:rbac-client") // for security integration w/ rbac
--- a/swatch-core/build.gradle
+++ b/swatch-core/build.gradle
@@ -4,10 +4,6 @@ plugins {
 }
 
 dependencies {
-
-    implementation "com.splunk.logging:splunk-library-javalogging:1.11.0"
-    implementation "org.codehaus.janino:janino:3.1.6"
-
     implementation project(":api") // for util methods in ResourceUtils
     implementation project(":clients:rbac-client") // for security integration w/ rbac
     implementation "org.hibernate.validator:hibernate-validator"
@@ -23,6 +19,8 @@ dependencies {
     implementation "io.micrometer:micrometer-core"
     implementation "io.hawt:hawtio-springboot"
     implementation "com.jayway.jsonpath:json-path"
+    implementation "com.splunk.logging:splunk-library-javalogging"
+    implementation "org.codehaus.janino:janino"
     implementation("io.confluent:kafka-avro-serializer") {
         exclude group: "org.apache.kafka"
         // A transitive dependency, org.apache.zookeeper:zookeeper, includes log4j as well as slf4j.  We use
--- a/swatch-core/build.gradle
+++ b/swatch-core/build.gradle
@@ -4,6 +4,9 @@ plugins {
 }
 
 dependencies {
+
+    implementation "com.splunk.logging:splunk-library-javalogging:1.11.0"
+
     implementation project(":api") // for util methods in ResourceUtils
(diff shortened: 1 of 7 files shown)


swatch-core/src/main/java/org/candlepin/subscriptions/logback/RhsmSplunkHecEventHeaderSerializer.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class RhsmSplunkHecEventHeaderSerializer implements EventHeaderSerializer {
+  @Override
+  public Map<String, Object> serializeEventHeader(
+      HttpEventCollectorEventInfo eventInfo, Map<String, Object> metadata) {
+
+    var fields = (Map<String, Object>) metadata.getOrDefault("fields", new HashMap<>());
+
+    // TODO this better
+    fields.put("namespace", System.getenv("SPLUNKMETA_namespace"));
```
lindseyburnett, line 17: I took a stab at trying to make this configurable via a properties bean or something, but I think it's invoked at an awkward part of the lifecycle.  This "works", but ideally I'd like this not to be so buried.

swatch-core/src/main/resources/logback-spring.xml
```diff
@@ -39,4 +39,34 @@
         <appender-ref ref="ConsoleAppender" />
         <appender-ref ref="SplunkAppender" />
     </root>
+
+    <if condition='property("ENABLE_SPLUNK_HEC").equals("true")'>
```
lindseyburnett, line 43: Another sort of buried env variable that I would like to be more obvious eventually.

swatch-core/build.gradle
```diff
@@ -4,6 +4,10 @@ plugins {
 }
 
 dependencies {
+
+    implementation "com.splunk.logging:splunk-library-javalogging:1.11.0"
+    implementation "org.codehaus.janino:janino:3.1.6"
```
lindseyburnett, line 9: The janino library is used for conditional processing in the logback xml file.

Reference: http://logback.qos.ch/manual/configuration.html

awood: These should be defined in `build.gradle`'s `dependencyManagement` section.  Then we can reference them here like

```
implementation "com.splunk.logging:splunk-library-javalogging"
implementation "org.codehaus.janino:janino"
```

This is to make sure that the entire project uses exactly the same versions

swatch-core/src/main/resources/logback-spring.xml
```diff
@@ -39,4 +39,34 @@
...
+        <then>
+            <appender name="http" class="com.splunk.logging.HttpEventCollectorLogbackAppender">
+                <url>${SPLUNK_HEC_URL}</url>
```
awood, line 46: Are these getting pulled from the environment?  Looking at the [docs](https://logback.qos.ch/manual/configuration.html#variableSubstitution) it looks like it pulls from Java system properties instead?  Same question on the `<if condition='property` bit.

One thought is that we could use the `<property file=` syntax and load the properties from a config map that we write.  That would actually cut down on a lot of the redundancy in the clowder config since instead of redefining the environment variables over and over again, we could just mount in the same properties file into each one.

lindseyburnett: "During substitution, properties are looked up in the local scope first, in the context scope second, in the system properties scope third, and in the OS environment fourth and last."  http://logback.qos.ch/manual/configuration.html#scopes

PR #669, opened 2021-11-10
Clowder Quality of Life

awood:


kahowell (changes requested):
For the port naming, to be consistent with clowder, I'd like to go ahead and rename the port to "metrics": For each template, there are a few places this needs to happen:

1. `prometheus.io/port` annotation
2. (may as well) in the probes, use "metrics" instead of 9000.
3. In the podspec, use the name "metrics" instead of "metrics-port"
4. In the Service, use "metrics" instead of 9000 and "metrics" instead of "metrics-port".

For the API deployment specifically, we should do similarly for "8000". It should be called "public" to be consistent with clowder. I'm thinking we just drop 8080 from port definitions in all other templates (though I'd be fine with renaming them to "public" and changing to 8000 instead).

Also, capacity-ingress is a bit... odd, it should only have 8443 and 9000 (doesn't need 8000 or 8080). With planned changes (offering sync & subscription sync enablement) 8443 doesn't really matter, so I'm not concerned w/ clowder differences for 8443.

A few other references to 8080 to address:
- `container_port` default in `bin/hawtio-proxy.sh` - may as well set to 9000 to avoid confusion
- `privatePort` in `dev-clowdenv.yaml` - may as well set to 10000 to be consistent with ephemeral environments instead
- `privatePort` in `test-clowder-config.json` - may as well set to 10000 to avoid confusion.

The paths in `deploy/rhsm-clowdapp.yaml` probes need to be updated to remove `/actuator`. Also the port needs to change from 8000 to "metrics". We should go ahead and drop the lines in ssl.properties that reference overriding the port, since it'll get set to 9000 and be consistent if we do.

`swatch-core/src/main/java/org/candlepin/subscriptions/security/JolokiaActuatorConfiguration.java` needs updating, specifically the matchers still have `/actuator` prefix.

awood:
> For each template, there are a few places this needs to happen:
> 
>     1. `prometheus.io/port` annotation

Is that going to resolve?  I didn't do it because `prometheus.io/port` is an metadata annotation.

kahowell:
> > For each template, there are a few places this needs to happen:
> > ```
> > 1. `prometheus.io/port` annotation
> > ```
> 
> Is that going to resolve? I didn't do it because `prometheus.io/port` is an metadata annotation.


Actually, our current prometheus setup doesn't even use these annotations. I dug a bit and took some notes on how on monitoring integration works here (internal only): https://docs.engineering.redhat.com/x/uCviBg, so nevermind on the prometheus annotations (we can actually drop them at some point in fact if we want).

lindseyburnett (approved):

PR #681, opened 2021-11-26
ENT-3968: Replace ProductRegistry with TagProfile

ksynvrit:
This pull request is to address [ENT-3968](https://issues.redhat.com/browse/ENT-3968).

Update tag profile yaml with product registry data.
Remove unused imports of ProductRegistry
Add missing data from the ProductRegistry Yaml to the TagProfile yaml
Remove unneeded method from TagProfile
Add Test specific data in the Test yaml
Delete ProductProfileRegistry & related unit test
Remove ProductProfileRegistrySource bean config from RegistryConfiguration


swatch-core/src/main/java/org/candlepin/subscriptions/registry/TagProfile.java
```diff
@@ -243,4 +248,70 @@ public String tagForOfferingProductName(String offeringProductName) {
...
+    if (!tagSupportsGranularity(productId.toString(), requestedGranularity)) {
+      String msg =
+          String.format(
+              "%s does not support any granularity finer than %s",
+              productId, requestedGranularity.getValue());
+      throw new IllegalStateException(msg);
+    }
+  }
+
+  public Map<Integer, Set<String>> getEngProductIdToSwatchProductIdsMap() {
```
mstead, line 270: Let's create this mapping when the object is initialized.  Take a look at `handleTagMapping` and build it there and allow access via a getter.

swatch-core/src/main/java/org/candlepin/subscriptions/registry/TagProfile.java
```diff
@@ -243,4 +248,70 @@ public String tagForOfferingProductName(String offeringProductName) {
...
+      if (engProductIdToSwatchProductIdsMap.containsKey(engId)) {
+        throw new IllegalStateException("Duplicate engineering product ID found: " + engId);
+      }
+      engProductIdToSwatchProductIdsMap.put(engId, tag.getTags());
+    }
+    return engProductIdToSwatchProductIdsMap;
+  }
+
+  public Map<String, Set<String>> getRoleToSwatchProductIdsMap() {
+    Map<String, Set<String>> roleToProductsMap = new HashMap<>();
```
mstead, line 286: Unless I'm mistaken, this is already being done in `handleTagMapping`. Should be able to just allow access to the mapping via  a getter.

swatch-core/src/main/java/org/candlepin/subscriptions/registry/TagProfile.java
```diff
@@ -243,4 +248,70 @@ public String tagForOfferingProductName(String offeringProductName) {
...
+
+      if (roleToProductsMap.containsKey(role)) {
+        throw new IllegalStateException("Duplicate engineering product role found: " + role);
+      }
+      roleToProductsMap.put(role, tag.getTags());
+    }
+    return roleToProductsMap;
+  }
+
+  public Map<String, String> getArchToSwatchProductIdsMap() {
```
mstead, line 300: Let's create this mapping when the object is initialized.  Take a look at `handleTagMapping` and build it there and allow access via a getter.

mstead (changes requested):
Overall looks pretty good. Just a few suggestions.

swatch-core/src/main/java/org/candlepin/subscriptions/registry/TagProfile.java
ksynvrit: Hey Micheal, I had this method so that I could return a map of <Integer,String> datatypes for Capacity processing, While the `tagToEngProductsLookup` is a <String, Set<String>> map. I changed the other methods based on your suggestions and added a new lookup for arch in the `handleTagMapping`, But I want to leave this methods as is. As creating this mapping in the lookup would be somewhat redundant as it would hold the same information as  `tagToEngProductsLookup` only that the keys are Integers.

swatch-core/src/main/java/org/candlepin/subscriptions/registry/TagProfile.java
```diff
@@ -243,4 +252,45 @@ public String tagForOfferingProductName(String offeringProductName) {
...
+      engProductIdToSwatchProductIdsMap.put(engId, tag.getTags());
+    }
+    return engProductIdToSwatchProductIdsMap;
+  }
+
+  public Map<String, Set<String>> getRoleToSwatchProductIdsMap() {
+    return roleToTagLookup;
+  }
+
+  public Map<String, String> getArchToSwatchProductIdsMap() {
```
mstead, line 293: This method can removed and the `@Getter` annotation can be used on the field. Probably best if we rename the field to match the getter method that you are removing so that it makes more sense when accessed.

swatch-core/src/main/java/org/candlepin/subscriptions/registry/TagProfile.java
```diff
@@ -243,4 +252,45 @@ public String tagForOfferingProductName(String offeringProductName) {
...
+      if (engProductIdToSwatchProductIdsMap.containsKey(engId)) {
+        throw new IllegalStateException("Duplicate engineering product ID found: " + engId);
+      }
+      engProductIdToSwatchProductIdsMap.put(engId, tag.getTags());
+    }
+    return engProductIdToSwatchProductIdsMap;
+  }
+
+  public Map<String, Set<String>> getRoleToSwatchProductIdsMap() {
+    return roleToTagLookup;
```
mstead, line 290: This method can removed and the `@Getter` annotation can be used on the field. Probably best if we rename the field to match the getter method that you are removing so that it makes more sense when accessed.

swatch-core/src/main/java/org/candlepin/subscriptions/registry/TagProfile.java
```diff
@@ -111,6 +114,8 @@ private void handleTagMapping(TagMapping mapping) {
       roleToTagLookup
           .computeIfAbsent(mapping.getValue(), k -> new HashSet<>())
           .addAll(mapping.getTags());
+    } else if ("arch".equals(mapping.getValueType())) {
+      archToTagLookup.put(mapping.getValue(), mapping.getTags().stream().findFirst().orElseThrow());
```
mstead, line 118: It feels like the mapping here should be Map<String, Set<String>>. I realize that the code that uses this grabs the first tag it finds for the arch, but if we ever need the entire list for something we would need to change this. Perhaps it would be better to have the code that uses this mapping decide whether it wants the first item.

I'd also like to see a test around this code.

mstead (changes requested):
A couple more change requests with your latest changes.

mstead (approved):

PR #682, opened 2021-11-29
Reduce log level for AccessDenied errors.

barnabycourt:
AccessDenied can happen whenever a user attempts access when an organization has
not opted in. They get a warning back with the reason. As this is not an
application error we should reduce the log level so that real errors can be found
in the logs more easily.


mstead:
I would prefer WARN level for this error given the subject matter of the message.

PR #724, opened 2022-01-13
Update import_events.py with updated jolokia URL path and port

lindseyburnett:
Prereq: swatch running with jmx endpoints enabled

On develop branch, running the following doesn't work and gets a 401:

```shell
cd bin
./import_events.py --file events.csv
```

Pull this branch, then run it again and it should succeed


mstead:
ACK - Waiting for PR checks before merging (event though this is just a utility).

PR #735, opened 2022-01-25
Downgrading spring boot back to 2.5.6 because our implementation of /api-docs is broken

lindseyburnett:
http://localhost:8000/api-docs was displaying a 404, unable to find openapi.yaml.  Tracked it back to the upgrade of our spring boot library version.  Reverting for now so we can get our swagger ui documentation back while we find out how to correct it in the 2.6.X version.

Caused by dependabot PR: https://github.com/RedHatInsights/rhsm-subscriptions/pull/709

JIRA ticket for resolving https://issues.redhat.com/browse/ENT-4670

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ ext {
     swagger_annotations_version = "1.6.4"
     swagger_ui_version = "4.2.1"
     resteasy_version = "3.6.3.Final"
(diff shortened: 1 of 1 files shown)


kahowell (changes requested):
Please let's not do this. I figured out a quick fix; see related PR.

PR #736, opened 2022-01-25
ENT-4670 Change path for openapi.yaml in swagger-ui

kahowell:
https://issues.redhat.com/browse/ENT-4670

Rather than serving the openapi docs in multiple places just to make our
old scheme of hosting them at `/api-docs/openapi.yaml`, it's easier to
just point at their location served from our API root.

Closes #735

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resteasy/ResteasyConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/resteasy/ResteasyConfiguration.java
@@ -53,11 +53,4 @@ public class ResteasyConfiguration implements WebMvcConfigurer {
     registry.addViewController("/api-docs").setViewName("redirect:/api-docs/index.html");
     registry.addViewController("/api-docs/").setViewName("redirect:/api-docs/index.html");
   }
-
-  @Override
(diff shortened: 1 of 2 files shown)


lindseyburnett (approved):

src/main/java/org/candlepin/subscriptions/resteasy/ResteasyConfiguration.java
```diff
@@ -53,11 +52,4 @@ public void addViewControllers(ViewControllerRegistry registry) {
     registry.addViewController("/api-docs").setViewName("redirect:/api-docs/index.html");
     registry.addViewController("/api-docs/").setViewName("redirect:/api-docs/index.html");
   }
```
awood, line 54: I think this whole method can go too.

lindseyburnett: I remove this method, but when I tested it it resulted in a 404 unless you use the full URL path to index.html

awood: Okay, pay me no mind then!

PR #744, opened 2022-02-03
ENT-4598 Add major and minor version attributes to operating system details of HBI messages

kflahert:
https://issues.redhat.com/browse/ENT-4598

Reads the distribution.version fact and sets the major and minor version numbers of the the operation system  for use in HBI.

### DIFF
--- a/swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/InventoryService.java
+++ b/swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/InventoryService.java
@@ -120,7 +120,7 @@ public abstract class InventoryService {
   private SystemProfile createSystemProfile(ConduitFacts facts) {
     SystemProfile systemProfile = new SystemProfile();
     if (facts.getOsName() != null) {
-      systemProfile.setOperatingSystem(operatingSystemName(facts.getOsName()));
+      systemProfile.setOperatingSystem(operatingSystem(facts));
     }
     systemProfile.setOsRelease(facts.getOsVersion());
     systemProfile.setArch(facts.getArchitecture());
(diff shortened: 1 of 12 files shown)


swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/InventoryService.java
```diff
@@ -120,7 +120,7 @@ protected CreateHostIn createHost(ConduitFacts facts, OffsetDateTime syncTimesta
   private SystemProfile createSystemProfile(ConduitFacts facts) {
     SystemProfile systemProfile = new SystemProfile();
     if (facts.getOsName() != null) {
-      systemProfile.setOperatingSystem(operatingSystemName(facts.getOsName()));
+      systemProfile.setOperatingSystem(operatingSystem(facts));
     }
     systemProfile.setOsRelease(facts.getOsVersion());
```
ksynvrit, line 125: @kflahert @kahowell I'm wondering if we should remove the OS release since os_release is deprecated, and the Major and minor would be giving the same information.

ksynvrit (changes requested):
@kflahert Looks good, just want to make sure about the os_release with @kahowell

ksynvrit (approved):

swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/inventory/InventoryService.java
kahowell: Let's leave it be for now; seems puptoo populates both and I couldn't find any plans to drop it. @kruai, please let us know if you feel otherwise.

kruai: @kahowell @ksynvrit We do plan to drop `os_release`, but we can't until everyone stops using it. The source of truth for OS information is the `operating_system` field.

PR #750, opened 2022-02-08
ENT-4431: add total_capacity sort param to for subscription/products api

kflahert:
https://issues.redhat.com/browse/ENT-4431

Allows for sorting by the total capacity of virtual/physical sockets and cores.

Test Steps:

```
- curl -X 'GET' \
  'http://localhost:8000/api/rhsm-subscriptions/v1/subscriptions/products/OpenShift%20Container%20Platform?sort=total_capacity&dir=asc' \
  -H 'accept: application/vnd.api+json'
```

### DIFF
--- a/api/rhsm-subscriptions-api-spec.yaml
+++ b/api/rhsm-subscriptions-api-spec.yaml
@@ -1769,6 +1769,8 @@ components:
         - next_event_date
         - next_event_type
         - quantity
+        - total_capacity
+        - product_name
 
     SkuCapacityReport:
       properties:
--- a/src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
@@ -273,6 +273,12 @@ public class SubscriptionTableController {
             case NEXT_EVENT_TYPE:
               diff = left.getNextEventType().compareTo(right.getNextEventType());
               break;
+            case TOTAL_CAPACITY:
+              diff = left.getTotalCapacity().compareTo(right.getTotalCapacity());
+              break;
+            case PRODUCT_NAME:
+              diff = left.getProductName().compareTo(right.getProductName());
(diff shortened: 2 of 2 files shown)


src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
```diff
@@ -269,6 +269,9 @@ private static void sortCapacities(
             case NEXT_EVENT_TYPE:
               diff = left.getNextEventType().compareTo(right.getNextEventType());
               break;
+            case TOTAL_CAPACITY:
```
kflahert, line 272: Will this enum be automatically generated in the pipeline or do I need to commit it somewhere? Locally I ran 'openApiGenerate' to test.

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -1766,6 +1766,7 @@ components:
         - next_event_date
         - next_event_type
         - quantity
+        - total_capacity
```
kahowell, line 1769: [ENT-4431](https://issues.redhat.com/browse/ENT-4431) also mentions product name as an option, let's go ahead and tackle that too while we're updating.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
kahowell: It's fine. We have the general build task hooked up to openApiGenerate, so that it's sufficient to simply do a `./gradlew assemble` (or any other task that invokes a build).

api/rhsm-subscriptions-api-spec.yaml
kflahert: Oh I misunderstood that originally. Will update.

kahowell (approved):

PR #751, opened 2022-02-08
Add a Spring Actuator endpoint to print certificate information

awood:


barnabycourt:
@awood Could this be test generate a certificate if one isn't present in order to avoid checking certs (even test certs) into the repository? We have run into a couple of times on candlepin were tests randomly started failing because test certs that were checked in like this expired.

PR #764, opened 2022-02-21
Awood/cert endpoint

awood:


kahowell (changes requested):
I would like to see the URL logging restored (lines such as "RH Marketplace service URL: {}"):
- [x] RHSM URL
- [x] cloudigrade URL
- [x] RH Marketplace URL
- [ ] RBAC URL
- [ ] RHIT User service URL
- [ ] RHIT Product service URL
- [ ] RHIT Subscription service URL
- [ ] Prometheus service URL

Also, it'd be nice to get this enabled for the non-conduit services too... though I'm happy to scope that to a card in the backlog, unless you see a quick way to do that?

awood:
> I would like to see the URL logging restored (lines such as "RH Marketplace service URL: {}"):

Done!

> Also, it'd be nice to get this enabled for the non-conduit services too... though I'm happy to scope that to a card in the backlog, unless you see a quick way to do that?

I'm not sure I understand.  These logging messages should get printed for anything that's using the `*ApiFactory`.

kahowell (approved):

kahowell:
> > Also, it'd be nice to get this enabled for the non-conduit services too... though I'm happy to scope that to a card in the backlog, unless you see a quick way to do that?
> 
> I'm not sure I understand. These logging messages should get printed for anything that's using the *ApiFactory.

I was referring to the cert info endpoint being available in the other deployments. FYI, there was a spotless issue (I force pushed w/ fix), but otherwise looks good. @awood do you want to squash before merge?

PR #746, opened 2022-02-04
[DRAFT] lburnett/billing producer ent 4719

lindseyburnett:
WIP

PR #767, opened 2022-02-24
ENT-4711 Rename marketplace-worker for clarity

kahowell:
https://issues.redhat.com/browse/ENT-4711

In order to be consistent with the WIP in #746, I renamed the deployment to swatch-billing-provider-rh-marketplace. In many other places I renamed to rhmarketplace or rh-marketplace or rhMarketplace depending on the context. This will also require some changes in the deployment, so there will be some app-interface MRs to go along with this.


mstead (approved):
Looks good and deploys Ok locally.

PR #782, opened 2022-03-03
Update Subscriptions for Billing Provider

ksynvrit:
This PR is to address https://issues.redhat.com/browse/ENT-4685

Add new liquidbase change file
Add new file in changelog.xml
Rename Subscription marketSubscriptionId to billingProviderId
Create logic for populating billingProvider column
Refactor query for renamed field for billingProviderId


src/main/java/org/candlepin/subscriptions/subscription/SubscriptionDtoUtil.java
```diff
@@ -72,4 +72,16 @@ public static String extractRhMarketplaceId(Subscription subscription) {
...
   }
+
+  public static String populateBillingProvider(Subscription subscription) {
+    boolean hasRef =
+        subscription.getExternalReferences() != null
+            && subscription.getExternalReferences().containsKey(IBMMARKETPLACE);
+    String billingProvider = null;
+
+    if (hasRef) {
+      billingProvider = "RH_Marketplace";
```
kahowell, line 83: This should be "red hat" per the enum that we implemented in #778

kahowell (changes requested):

kahowell (approved):

PR #783, opened 2022-03-03
Incorporate swatch-producer-aws as subproject

kahowell:
I did a few things also to ensure we get versions managed properly.

Namely, I introduced a couple of "conventions" gradle plugins, which apply common config to projects.

I moved our dependency management to primarily dependencies.gradle, which was necessary to give dependabot a consistent place to look.

The spring dependency-managent gradle plugin is also still used, but is now limited to certain projects (see projects using the
`swatch.spring-boot-dependencies-conventions` plugin.

I also cleaned up the `dependencyJson` task a bit, it had been broken.


kahowell:
09695046 to 4640b942: Fix git properties application to buildinfo to restore NPE safety (done as fixup)

lindseyburnett (approved):

PR #786, opened 2022-03-04
Query the Cloudigrade user API for user existence before further calls

awood:
[ENT-3400](https://issues.redhat.com/browse/ENT-3400)

Querying the Cloudigrade concurrent API with for a user that does not
exist results in a non-200 return code.  We interpret this is a failure
and retry the call and log an error.  This commit queries the internal
Cloudigrade user API to check for the existence of a user prior to
making any subsequent calls and therefore avoids entering error handling
code.


awood:
TODO:

- [ ] Make sure the Cloudigrade PSK secret is available in all environments.  I believe it should be -- it is in the ephemeral environment I created as "cloudigrade-psk" but I can't list secrets for prod
- [ ] Convert our existing Cloudigrade calls to use the PSK?  This would be a pretty simple change.
- [ ] Run an end-to-end test in the ephemeral environments

awood:
Testing:

* Login to the ephemeral environment
* Port forward a cloudigrade-api pod from `ephemeral-hv80yo`.  (Run `oc get pods` if necessary to get pod name) E.g.
  `oc -n ephemeral-hv80yo port-forward pod/cloudigrade-api-677868856c-q4rhq 9999:8000`
* Get the ephemeral pre-shared key for Cloudigrade and base64 decode the result.  The PSK looks like a uuid.  Also I can just   
  send it to you privately.
  `oc get secret/cloudigrade-psk -o yaml`
* Set up your deployment with the following properties
  ```
  rhsm-subscriptions.cloudigrade.url=https://localhost:9999/api/cloudigrade/v2
  rhsm-subscriptions.cloudigrade.presharedKey=PSK_HERE
  rhsm-subscriptions.cloudigrade.internal.url=http://localhost:9999/internal/api/cloudigrade/v1
  rhsm-subscriptions.cloudigrade.internal.presharedKey=PSK_HERE
  rhsm-subscriptions.cloudigrade-enabled=true
  logging.
  ```
* Note that the cloudigrade user service **is not https**.
* Run a tally for user over Jolokia with the JMX bean
  ```
  curl 'http://localhost:9000/hawtio/jolokia' -H 'Content-Type: text/json' -d '{"type":"exec",   "mbean":"org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean", "operation":"tallyAccount(java.lang.String)", "arguments":["1234567"]}'
  ```

kflahert (approved):

PR #792, opened 2022-03-08
Update gradle to 7.4

kahowell:
First I ran

```
./gradlew wrapper --gradle-distribution-sha256-sum=8cc27038d5dbd815759851ba53e70cf62e481b87494cc97cfd97982ada5ba634 --gradle-version=7.4
```

and then dealt with the fallout. This was mostly just replacing removed dependency scopes with their replacements.

I ran the wrapper command once more so that the scripts would be updated. (See https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper)

Finally, I ran `./gradlew help --warning-mode=all` and addressed one remaining deprecation from that.

Co-authored-by: Barnaby Court <bcourt@redhat.com>


awood (approved):
This works for me. Maybe get one more approver since this change has the potential to break workflows?

kahowell:
@mstead @lindseyburnett @kflahert @ksynvrit can I get another +1 (or let me know if you hit issues?)

ksynvrit (approved):
I'm able to run everything fine on my end, only the Quarkus component failed to build for me, but I think it do to me just running it locally without the proper setup?

kahowell:
a94b081..1e8e236: rebase
94ba773: bump memory per @lindseyburnett's suggestion after troubleshooting w/ @ksynvrit

PR #801, opened 2022-03-09
podman_run.sh: s/-t/--time/

kahowell:
Current build_deploy.sh fails with

```
Error: unknown shorthand flag: 't' in -t
```

Hoping that changing to `--time` will address, and if not, we'll at least get the podman version output for reference.


kahowell:
Self-merging since this is more-or-less trivial.

PR #822, opened 2022-03-22
ENT-4698 Add admin API endpoint to resend specific tally snapshots

awood:
Exposing administrator REST endpoints. Unlike our other REST endpoints, this resource is meant to be deployed across all pods. With that being the case, we need to account for the contingency where a bean we depend on is not available (due to the Spring profile the pod is deployed with or for other architectural reasons). To account for this possibility, dependencies are injected as typed ObjectProviders. The provider then calls the `getIfAvailable` method with a dummy default provider.  The default provider implements or extends the dependency and has dummy implementations of the methods that
throw UnsupportedOperationExceptions.

This architecture has the potential to become very cumbersome if this resource begins to pull in a lot of dependencies. However, due to the current situation where we are using Spring deployment profiles to determine runtime functionality, any universal admin endpoint needs to be able to cope with missing beans.

---
## Testing (using [HTTPie](https://httpie.io/))

These instructions assume that you are either running locally (`./gradlew clean :bootRun --args="--spring.profiles.active=api,kafka-queue,worker,capacity-ingress"`) or that you are using `oc port-forward` to forward your `worker` pod to localhost and using the `bin/hawtio-proxy.sh` to get a hawtio console for the `api` pod.
```
$ bin/insert-mock-hosts --account account123 --num-physical 25 --product RHEL --hbi
# Run a tally for account123 using the Hawtio web console or via Jolokia REST request as seen below
$ http -v :9000/hawtio/jolokia type=exec mbean=org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean operation="tallyAccount(java.lang.String)" arguments:='["account123"]'
$ psql -h localhost -U rhsm-subscriptions rhsm-subscriptions -c "select id from tally_snapshots limit 1;"
$ http -v POST :8000/api/rhsm-subscriptions/v1/internal/tally/resend  x-rh-identity:$(echo -n '{"identity":{"account_number":"account123","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w 0) uuids:="[\"536e564f-19b7-476c-bf1d-882a97f385a1\"]"
```

But replace that UUID with a UUID from the `psql` command.

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/validator/UuidValidatorTest.java
+++ b/src/test/java/org/candlepin/subscriptions/validator/UuidValidatorTest.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.validator;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.util.List;
+import java.util.Set;
+import javax.validation.ConstraintViolation;
+import javax.validation.ConstraintViolationException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.context.TestConfiguration;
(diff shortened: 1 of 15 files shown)


src/test/java/org/candlepin/subscriptions/validator/UuidValidatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @ParameterizedTest
+  @ValueSource(
+      strings = {
+        "hellowor-ldhe-llow-worl-dhelloworldh",
+        "b682126b88d840b7ada6212f0e65e30a",
+        "b682126b88d840b7ada6212f",
+        "192.168.2.8.4",
+        "redhat.com"
```
lindseyburnett, line 85: what about something in a UUID format but has a bunch of special characters that aren't alpha numerical?

kahowell (changes requested):
This is a bit cumbersome. How about:

1. Separate API spec for tally "service" (what we call worker profile). (We would use separate packages to keep them from clashing for now). - to keep things simple, instead of adding another subproject we can just add another openApiGenerate task into `$buildDir/tally-api` and then add that to `sourceSets.main.java.srcDirs`
2. Resource class for the tally service endpoints only imported in work profile.

Additionally, can you please rename retally to resend? I don't want someone to think this causes recalculation.

awood:
> This is a bit cumbersome. How about:
> 
>     1. Separate API spec for tally "service" (what we call worker profile). (We would use separate packages to keep them from clashing for now). - to keep things simple, instead of adding another subproject we can just add another openApiGenerate task into `$buildDir/tally-api` and then add that to `sourceSets.main.java.srcDirs`
> 
>     2. Resource class for the tally service endpoints only imported in work profile.
> 
> 
> Additionally, can you please rename retally to resend? I don't want someone to think this causes recalculation.

Updated now.

kahowell (changes requested):

kahowell (approved):

PR #834, opened 2022-03-25
Update Tally Api with Billing Provider

ksynvrit:
https://issues.redhat.com/browse/ENT-4687

Update ReportCriteria with BillingProvider field.
Add new method for sanitizing billing provider in ResourceUtils.
Update yaml for Tally metadata to include billing provider.
Add implementation for billing provider for TallyResource.
Rework Unit tests to include billing provider as a parameter.
Update Tally db model for billing provider & repository.

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/db/TallySnapshotRepository.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/db/TallySnapshotRepository.java
@@ -24,10 +24,7 @@ import java.time.OffsetDateTime;
 import java.util.Collection;
 import java.util.UUID;
 import java.util.stream.Stream;
-import org.candlepin.subscriptions.db.model.Granularity;
-import org.candlepin.subscriptions.db.model.ServiceLevel;
-import org.candlepin.subscriptions.db.model.TallySnapshot;
-import org.candlepin.subscriptions.db.model.Usage;
+import org.candlepin.subscriptions.db.model.*;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.JpaRepository;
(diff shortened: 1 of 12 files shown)


kflahert (approved):

swatch-core/src/main/java/org/candlepin/subscriptions/db/TallySnapshotRepository.java
```diff
@@ -41,12 +38,13 @@
   // method
   @SuppressWarnings({"linelength", "java:S107"})
   Page<TallySnapshot>
-      findByAccountNumberAndProductIdAndGranularityAndServiceLevelAndUsageAndSnapshotDateBetweenOrderBySnapshotDate(
+      findByAccountNumberAndProductIdAndGranularityAndServiceLevelAndUsageAndBillingProviderAndSnapshotDateBetweenOrderBySnapshotDate(
```
lindseyburnett, line 41: @kahowell thoughts on maybe using Specifications or QueryDSL after a certain threshold of parameters because my brain does not like reading method names this long

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/BillingProvider.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public BillingProviderType asOpenApiEnum() {
+    return openApiEnum;
+  }
+
+  /** JPA converter for BillingProvider */
+  @Converter(autoApply = true)
+  public static class EnumConverter implements AttributeConverter<BillingProvider, String> {
```
lindseyburnett, line 69: Wondering if there is something we can do to make this a little more DRY.  This will be the third time that we're declaring an enum that implements `StringValueEnum`.

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/Usage.java
swatch-core/src/main/java/org/candlepin/subscriptions/db/model/ServiceLevel.java

cc: @kahowell

swatch-core/src/main/java/org/candlepin/subscriptions/db/TallySnapshotRepository.java
ksynvrit: Same, I discussed this with @kahowell as i noticed that this will get longer with billingProductId. I'm going to tackle this in my next PR for ENT-4843

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/BillingProvider.java
kahowell: That's a good thought. Happy to either timebox it to an hour or create a tech-debt card for the backlog for this. We already worked on getting this DRY, (That's how StringValueEnum came about in the first place); if there are improvements to be had, it's probably in pulling more things into that interface.

PR #835, opened 2022-03-25
Improve dependabot compatibility

kahowell:
By specifying repos in a central location - dependencies.gradle, dependabot knows to consult additional maven repos.

I also found that dependabot is a bit conservative with updating versions in properties; specifically, if those versions are used by more than one dep, it skips those deps. So it kind of defeats the purpose of having a versions object.

I found there is a good workaround though: using bom artifacts instead.

I checked the results against this branch by updating the dependabot-script `update-script.rb` with the following logic:

```rb
dependencies.each { |d|
  checker = Dependabot::Gradle::UpdateChecker.new(
    dependency: d,
    dependency_files: files,
    credentials: credentials,
  )

  if not checker.up_to_date?
    if checker.latest_resolvable_version.nil?
      puts "#{d.name}:#{d.version} not up to date, but unable to update; TODO figure out why"
      binding.irb
    else
      puts "#{d.name}: #{d.version} to #{checker.latest_resolvable_version}"
    end
  end
}

exit
```

This showed better results on the branch:

```
com.adarshr:gradle-test-logger-plugin: 3.0.0 to 3.2.0
com.commercehub.gradle.plugin:gradle-avro-plugin: 0.9.1 to 0.99.99
com.github.davidmc24.gradle.plugin:gradle-avro-plugin:1.3.0 not up to date, but unable to update; TODO figure out why
io.quarkus:gradle-application-plugin: 2.7.1.Final to 2.7.5.Final
com.redhat.cloud.common:clowder-quarkus-config-source: 0.5.3 to 1.0.0
io.quarkus.platform:quarkus-bom: 2.7.1.Final to 2.7.5.Final
org.jboss.resteasy:resteasy-bom: 3.6.3.Final to 6.0.0.Final
org.jboss.resteasy:resteasy-spring-boot-starter: 3.9.1.Final to 5.0.0.Final
com.github.tomakehurst:wiremock-jre8:2.32.0 not up to date, but unable to update; TODO figure out why
```

For the two com.github ones, I fired up a ruby debugger and noticed that they're trying to resolve against GitHub, rather than maven repos. I *suspect* this isn't an issue with dependabot the service, because we've had wiremock updated via dependabot in the past successfully, but at least this is a step in the right direction.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -170,8 +170,8 @@ project(":api") {
         implementation "javax.validation:validation-api"
         implementation "org.jboss.spec.javax.ws.rs:jboss-jaxrs-api_2.1_spec"
         implementation "io.swagger:swagger-annotations"
-        implementation "com.google.code.findbugs:jsr305:3.0.2"
-        implementation "org.openapitools:jackson-databind-nullable:0.2.2"
+        implementation libraries["jsr305"]
+        implementation libraries["jackson-databind-nullable"]
     }
 
     sourceSets.main.java.srcDirs = ["${buildDir}/generated/src/gen/java"]
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,18 +1,20 @@
 // By keeping dependencies in this file, they get picked up by dependabot reliably
 // inspired by mockito's gradle structure, which dependabot uses as a test case
-ext {
-    libraries = [:]
-    plugins = [:]
-    versions = [:]
+
+// Repeating repositories here allows dependabot to use them to check for updates
+buildscript {
+  repositories {
+      mavenCentral()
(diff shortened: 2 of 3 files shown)


kahowell:
8666d54 to f0c30aa:
* remove resteasy-bom; instead declare each dep independently. It's not ideal because we should upgrade all resteasy deps at once, but at least this way we'll be notified. Using resteasy-bom was problematic as it was downgrading versions of all kinds of other deps. Ideally, we'll update resteasy-related things altogether in another PR at some point.
* remove places where testcontainers was being set explicitly; it's versioned in the quarkus-bom :smile:

lindseyburnett (approved):

PR #867, opened 2022-04-06
Awood/merge internal apis

awood:


awood:
TODO:

- [ ] Need to update the Openshift configs to handle the map value binding
- [ ] Need to update the secrets in Vault

PR #874, opened 2022-04-08
Remove Gradle renaming of our OpenAPI spec file

awood:
The consoledot guidelines want the API spec to be served under
"openapi.yaml" which we can do by controlling the name of the endpoint
that serves the file in the specification itself.  There's no need to
rename the file with Gradle.

### DIFF
--- a/swatch-system-conduit/build.gradle
+++ b/swatch-system-conduit/build.gradle
@@ -71,6 +71,6 @@ jacocoTestReport {
     }
 }
 
-bootJar {
+springBoot {
     mainClass = "org.candlepin.subscriptions.SystemConduitApplication"
 }
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,7 @@
+import com.github.davidmc24.gradle.plugin.avro.GenerateAvroJavaTask
 import groovy.json.JsonOutput
+import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
+import org.openapitools.generator.gradle.plugin.tasks.ValidateTask
 
 buildscript {
     // dependabot will parse dependencies.gradle by detecting the `apply from` below
@@ -122,7 +125,7 @@ ext {
     sub_sync_config_file = "${internal_spec_dir}/internal-subscriptions-sync-api-config.json"
 }
 
-tasks.register("buildApiTally", org.openapitools.generator.gradle.plugin.tasks.GenerateTask) {
+tasks.register("buildApiTally", GenerateTask) {
     generatorName = "jaxrs-spec"
     inputSpec = tally_api_spec_file
     configFile = tally_config_file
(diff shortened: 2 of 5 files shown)


clients/build.gradle
```diff
@@ -104,22 +107,19 @@ configure(subprojects) {
     apply plugin: "swatch.java-conventions"
     apply plugin: "org.openapi.generator"
 
-    openApiValidate {
+    tasks.register("apiValidate", ValidateTask) {
         inputSpec = api_spec_path
     }
```
kahowell, line 112: I don't actually want to do this, because it means that we end up with extraneous tasks defined, see `./gradlew :clients:rhsm-client:tasks --all`. Additionally, gradle syntax and plugin implementation make this unclear, but actually, `openApiValidate` task is already declared via `tasks.register`. See [generator plugin source](https://github.com/OpenAPITools/openapi-generator/blob/d17316e8d98be750fd87af528174b6584578d394/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/OpenApiGeneratorPlugin.kt#L82-L88).

`openApiValidate { /*...*/ }` defines [an extension instance](https://github.com/OpenAPITools/openapi-generator/blob/d17316e8d98be750fd87af528174b6584578d394/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/extensions/OpenApiGeneratorValidateExtension.kt) (essentially just a pojo), that the plugin then consults when running `tasks.register`.

Additionally we lose some things when defining custom tasks (namely task `group` and `description`, and any defaults applied in the extension class or plugin class).

swatch-system-conduit/build.gradle
```diff
@@ -9,7 +12,7 @@ ext {
     config_file = "${projectDir}/system-conduit-api-config.json"
 }
 
-openApiGenerate {
+tasks.register("apiGenerate", GenerateTask) {
```
kahowell, line 15: See previous comment about extensions and custom tasks.

swatch-system-conduit/build.gradle
```diff
@@ -21,12 +24,12 @@ openApiGenerate {
     ]
 }
 
-openApiValidate {
+tasks.register("apiValidate", ValidateTask) {
```
kahowell, line 27: See previous comment about extensions and custom tasks.

kahowell (changes requested):

clients/build.gradle
awood: Yeah, I know it's not ideal, but it leaves us in the position of having 2 different tasks that do the same thing.  Some projects have `openApiValidate` while the main project has `apiValidate` since we can't define `openApiValidate` twice.  Let me experiment and see if I can pull `tasks.named('openApiValidate')` and add a depends on to our two validation tasks for the 2 internal API specs.

awood: So I tried

```
tasks.named("openApiGenerate") {
    dependsOn(provider {
        tasks.findAll { task -> task.name.startsWith('buildApi') }
    })
}
```

In the main build file and Gradle dies saying "In plugin 'org.openapi.generator' type 'org.openapitools.generator.gradle.plugin.tasks.GenerateTask' property 'inputSpec' doesn't have a configured value."  So it looks like either we move everything to tasks named "generateApi" or we have an inconsistency where the generation task has a different name for the main project.  Maybe it's not a big deal since it's all dependent on `compilreJava` anyway.  I just call the generation tasks directly a lot so keeping consistency in naming what a goal here.

kahowell: @awood see #881

kahowell (approved):

PR #881, opened 2022-04-11
Move back to openApiGenerate extension where possible

kahowell:
With these changes, `./gradlew openApiGenerate` triggers all the codegen tasks as desired.


awood:
I just cherry picked your commit onto #874.  It made the history nicer and was easier for me to resolve the conflicts that way.

PR #891, opened 2022-04-13
Express defaults directly in property value

awood:
The reason to define an property that's meant to have the same name as
an environment variable is for when we would otherwise have to nest
expressions (e.g. we can't nest expressions like
${CLOUDIGRADE_INTERNAL_HOST:$CLOUDIGRADE_HOST}). That situation doesn't
apply to these properties, so we can define the defaults inline.

### DIFF
--- a/src/main/resources/application-worker.yaml
+++ b/src/main/resources/application-worker.yaml
@@ -2,10 +2,6 @@ CLOUDIGRADE_HOST: ${clowder.endpoints[?(@.app == 'cloudigrade')].hostname:localh
 CLOUDIGRADE_PORT: ${clowder.endpoints[?(@.app == 'cloudigrade')].port:8080}
 CLOUDIGRADE_INTERNAL_HOST: ${CLOUDIGRADE_HOST}
 CLOUDIGRADE_INTERNAL_PORT: ${CLOUDIGRADE_PORT}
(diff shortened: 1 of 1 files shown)


kahowell (changes requested):
This pushes our env vars into two places, it hurts readability IMO, and it's not obvious when you should define the env var up front vs. inline. Any reasons to do this other than preference?

awood:
> This pushes our env vars into two places, it hurts readability IMO, and it's not obvious when you should define the env var up front vs. inline. Any reasons to do this other than preference?

The ones defined up-top are the outliers.  The only reason we do them that way is because we can't nest references.  Everything else we have follows the

```
my-config-value: ${SOME_ENV_VAR:thedefault}`
```

convention that this PR is attempting to apply to the `TALLY_SUMMARY_*` vars.

kahowell (approved):
@awood fair enough that we're not doing it that way currently (been staring at `swatch-producer-aws` a lot). I actually prefer them separated out at top at least in the components in quarkus... There it lends itself to an organization like:
* defaults
* dev profile defaults
* test profile defaults
* implementation details & aliasing

I'll put it on our agenda, would like to get a feel for whether folks like that way or not anyways, and if so, will maybe backlog a card to reorganize env vars and properties a bit.

PR #897, opened 2022-04-14
Fix broken instances API

mstead:
Need to allow searching by null BillingProvider because we
do not to apply a default when adding the new field.

The resource was updated to not use the 'sanitize' method, since
it does not make use of the _ALL or "" types.


mstead:
This is missing a unit test, but I wanted to get this fix out as I won't be online on Monday.

kahowell:
LGTM and previous CI test run passed.

PR #928, opened 2022-04-26
ENT-4695: Adjust usage latency to meet AWS SLA

mstead:
https://issues.redhat.com/browse/ENT-4695

Usage metering now occurs every hour on the half hour
and is delayed by only 30 minutes relative to when the job
is run. This means that if the metering job runs at 2:30PM,
metrics will be gathered for 1pm (the 1pm - 2pm range).

**NOTE:** When the metrics query is run, it is run against the
2pm - 2:59:59.99999pm range since the query does a 1h look back
and the data actually represents usage for 1pm.

The hourly tally schedule has been updated to run with a
1h delay, relative to when the job is run. This give the
metrics tasks 30m to complete and to have events populated.

## Testing
The metering job should be offset by only 1h based on when the job was run. It does not matter at what point in the current hour the job is run, the result will be the same. For example, if you run the job at 2:15pm, metrics should be gathered for 2pm-2:59:59.9999pm. This would be the same for 2:45pm.

**REMEBER: The resulting metric would be for 1pm due to the PromQL query look back.**

Manually run the MetricsJob and verify the resulting metering range in the logs.
```
# Current time 14:47UTC
PROM_URL="http://localhost:8888/api/metrics/v1/telemeter/api/v1" SPRING_PROFILES_ACTIVE=metering-job,kafka-queue ./gradlew :bootRun
```
```
2022-04-26 14:47:31,695 [thread=restartedMain] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.task.PrometheusMetricsTaskManager] - Queuing rhosak metric updates for range: 2022-04-26T14:00Z -> 2022-04-26T14:59:59.999999
```

Run the hourly job at any point. The tally range should be offset by only 1h.
```
# Current time: 14:49
$ SPRING_PROFILES_ACTIVE=capture-hourly-snapshots,kafka-queue ./gradlew :bootRun
```
```
2022-04-26 14:49:47,061 [thread=restartedMain] [INFO ] [org.candlepin.subscriptions.tally.job.CaptureSnapshotsTaskManager] - Queuing hourly snapshot production for accountNumber 941133 between 2022-04-26T12:00:00Z and 2022-04-26T13:00:00Z
```

**NOTE**: With this configuration, metrics will be gathered every hour on the 1/2 hour, and the tally will occur at the top of the next hour.

Metrics for 1pm are gathered @ 2:30pm and these hourly metrics are tallied at 3pm.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
@@ -103,7 +103,7 @@ public class ApplicationProperties {
    * Offsets the range to look at metrics to account for delay in prometheus having metrics
    * available
    */
-  private Duration prometheusLatencyDuration = Duration.ofHours(4L);
+  private Duration prometheusLatencyDuration = Duration.ofHours(0L);
 
   /**
    * Amount of time from current timestamp to start looking for metrics during a tally, independent
--- a/src/main/java/org/candlepin/subscriptions/metering/jmx/MeteringJmxBean.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/jmx/MeteringJmxBean.java
@@ -108,11 +108,8 @@ public class MeteringJmxBean {
     Object principal = ResourceUtils.getPrincipal();
     log.info("Metering for all accounts triggered via JMX by {}", principal);
 
-    OffsetDateTime end = getDate(null);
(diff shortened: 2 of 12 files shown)


kflahert (approved):
Will merge once checks are done.

PR #931, opened 2022-04-27
ENT-4693: Add filter for logging user requests to admin APIs

kflahert:
https://issues.redhat.com/browse/ENT-4693

Test Steps:
- Start app with default profiles
- Run against an internal endpoint and check server logs for message with path and user
`curl -X 'PUT'   'http://localhost:8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/123' -H 'x-rh-swatch-psk: dummy'`

- Run non internal endpoint and check that same log message does not come up
`curl -X 'GET'   'http://localhost:8000/api/rhsm-subscriptions/v1/subscriptions/products/OpenShift%20Container%20Platform'   -H 'accept: application/vnd.api+json'   -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoib3JnMTIzIn19fQo='`

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/security/ApiSecurityConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/security/ApiSecurityConfiguration.java
@@ -148,6 +148,10 @@ public class ApiSecurityConfiguration extends WebSecurityConfigurerAdapter {
     return new MdcFilter();
   }
 
+  public LogPrincipalFilter logPrincipalFilter() {
+    return new LogPrincipalFilter();
+  }
+
   @Override
   protected void configure(HttpSecurity http) throws Exception {
     String apiPath =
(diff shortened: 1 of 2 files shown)


lindseyburnett:
Is this something that we're going to need in the new swatch-producer-aws component as well?  If so, we need to figure out the non-Spring equivalent of doing this.  Using MDC wasn't super straight forward when I added the splunk logger, so we opted to skip the security stack eval MDC property.

cc: @kflahert @kahowell

kflahert:
@lindseyburnett I think this can be accomplished more easily in Quarkus than the MdcFilter. This JAX-RS ContainerRequestFilter I think will work for what we need: https://quarkus.io/guides/security-customization#jaxrs-security-context

lindseyburnett (approved):

PR #942, opened 2022-04-28
Add SWATCH_SELF_PSK to swatch-producer-aws clowdapp

kahowell:
In order to fix pod restarts due to:

```
Failed to load config value of type class java.lang.String for: SWATCH_SELF_PSK
```

Reduced initial probe delays to a more reasonable value for quicker deployment.

Updated bonfire config in clowder README. Bonfire now checks app-interface for the components and expects the same "app name" as the saas-file name in app-interface, so had to update accordingly.

Testing
------

Regenerate your bonfire config per the updated README.

```
bonfire deploy -C swatch-producer-aws -f rhsm -i quay.io/cloudservices/swatch-producer-aws=9897ac3
```
*latest doesn't work as the image tag - it'll pull an old version - above is at the time of writing the latest hash.

Observe that it deploys correctly to ephemeral. Optionally, go look at the logs in the ephemeral namespace as well.

### DIFF
--- a/deploy/clowder.md
+++ b/deploy/clowder.md
@@ -17,7 +17,7 @@
   bonfire config write-default
 
   cat <<BONFIRE >>  ~/.config/bonfire/config.yaml
-  - name: rhsm-subscriptions
+  - name: rhsm
     components:
       - name: capacity-allowlist
         host: gitlab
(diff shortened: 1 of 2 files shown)


deploy/clowder.md
```diff
@@ -17,7 +17,7 @@
   bonfire config write-default
 
   cat <<BONFIRE >>  ~/.config/bonfire/config.yaml
-  - name: rhsm-subscriptions
```
kflahert, line 20: Just want to confirm, is this name change intentional?

kflahert: nvm read your comment

kflahert (approved):

PR #958, opened 2022-05-03
Rename swatch-producer-aws clowdapp deployment service

kahowell:
The k8s deployment name is generated by clowder as `$clowdapp-$deploymentName`, so it was being set to `swatch-producer-aws-swatch-producer-aws`; With this change it'll get set instead to `swatch-producer-aws-service`, which is much less redundant.


lindseyburnett (changes requested):
I'd love to get away from the word "worker"

PR #973, opened 2022-05-05
ENT-4912: Protect subscription sync against failures due to invalid org/acct id

lindseyburnett:
I traced back the remaining kafka exceptions in stage to an ApiException from the subscription service.  This was happening because there's an orgId in stage - `lsjel`.  This happens when subscription sync cron job runs, which is why it's buried in a kafka exception.

# Reproducing on develop

The easiest/fastest way to produce locally though is to use the `syncSubscriptionsForOrg` jolokia endpoint.

Set environment variables to point to the stage subscription service.  Keystore and password can be found in vault.  You might want to turn on some http logging.  You can do this with `LOGGING_LEVEL_ORG_APACHE_HTTP_WIRE=debug`.
  
```
SUBSCRIPTION_URL=https://subscription.stage.api.redhat.com/svcrest/subscription/v5
SUBSCRIPTION_KEYSTORE=<redacted>
SUBSCRIPTION_KEYSTORE_PASSWORD=<redacted>
SPRING_PROFILES_ACTIVE =api,worker,capacity-ingress
```

Start the service `./gradlew :bootRun`

Kick off a syncSubscriptionsForOrg request
```
curl 'http://localhost:9000/hawtio/jolokia/?maxDepth=7&maxCollectionSize=50000&ignoreErrors=true&canonicalNaming=false' \
  -H 'Accept: application/json, text/javascript, */*; q=0.01' \
  --data-raw '{"type":"exec","mbean":"org.candlepin.subscriptions.subscription:name=subscriptionJmxBean,type=SubscriptionJmxBean","operation":"syncSubscriptionsForOrg(java.lang.String)","arguments":["lsjel"]}' \
  --compressed
```

If you have http logging on, you'll notice in the console some 500 errors and that it's retrying the request a few times.  Eventually when the function exits, you'll get this response from your curl request:

```
{"request":{"mbean":"org.candlepin.subscriptions.subscription:name=subscriptionJmxBean,type=SubscriptionJmxBean","arguments":["lsjel"],"type":"exec","operation":"syncSubscriptionsForOrg(java.lang.String)"},"error_type":"org.candlepin.subscriptions.exception.ExternalServiceException","error":"org.candlepin.subscriptions.exception.ExternalServiceException : Error during attempt to request subscription info","status":500}
```
For anyone who is curious, this is the example of the response body that the subscription service returns:
```
{"msgName":"com.redhat.services.util.rest.ExceptionMessage","type":["java.lang.NumberFormatException"],"message":["For input string: \"lsjel\""],"serialized":null}
```

# "The Fix"
There isn't a need to keep retrying the rest request if we know it's caused by an invalid orgId.  I updated the subscription sync retry template to /not/ do that.

Also added a log.error to print out the response body when an ApiException happens.

# "TODO"
- [ ] locally reproduce the nasty kafka stack trace that i fished up out of the stage console logs 

```
... 11 common frames omitted
Caused by: org.candlepin.subscriptions.subscription.ApiException: <HTML><HEAD><TITLE>Error</TITLE></HEAD><BODY>
An error occurred while processing your request.<p>
Reference&#32;&#35;30&#46;5efe3017&#46;1651156484&#46;323bf55c
</BODY></HTML>
```

### DIFF
--- a/deploy/rhsm-clowdapp.yaml
+++ b/deploy/rhsm-clowdapp.yaml
@@ -99,10 +99,14 @@ parameters:
     value: https://subscription.qa.api.redhat.com/svcrest/subscription/v5
   - name: SUBSCRIPTION_MAX_CONNECTIONS
     value: '100'
-  - name: SUBSCRIPTION_BACKOFF_INITIAL_INTERVAL
-    value: 1s
   - name: SUBSCRIPTION_MAX_RETRY_ATTEMPTS
     value: '4'
+  - name: SUBSCRIPTION_BACKOFF_INITIAL_INTERVAL
+    value: 1s
+  - name: SUBSCRIPTION_BACK_OFF_MAX_INTERVAL
+    value: 64s
+  - name: SUBSCRIPTION_BACK_OFF_MULTIPLIER
+    value: '2'
   - name: SUBSCRIPTION_PAGE_SIZE
     value: '500'
   - name: USER_HOST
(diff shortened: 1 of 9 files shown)


mstead (commented):
I took another look at this PR this morning. I think the approach you have taken with the custom exception is how the framework expects us to bail on retries, so I tossed my thought on using a boolean out the window.

One thought I had to get rid of the sonar linting issue, you could have the new exception class extend SubscriptionsException and wrap another SubscriptionsException. This way, the unretryable exception can be used for things other than just external service exceptions should we choose. Should then probably rename it to simply UnretryableException.

```java
package org.candlepin.subscriptions.exception;

public class UnretryableException extends SubscriptionsException {

  public UnretryableException(SubscriptionsException e) {
   // NOTE: Will need a new constructor for SubscriptionsException
    super(e);
  }
}
```

```java
            if (e.getResponseBody().contains("NumberFormatException")) {
              throw new UnretryableException(new ExternalServiceException(
                  ErrorCode.REQUEST_PROCESSING_ERROR,
                  ERROR_DURING_ATTEMPT_TO_REQUEST_SUBSCRIPTION_INFO_MSG,
                  e));
            }
```

lindseyburnett:
> 

Opting for approach #1

PR #991, opened 2022-05-13
Optimize force sync subscriptions

lindseyburnett:
The problem on develop is hard to reproduce locally if you just have the default three values in the offering table.  I was able to reproduce slowness by grabbing the subscriptions/offering tables from s3, and importing that data locally.

A lot of time was being spent checking if the sku from the service exists in our offering table (https://github.com/RedHatInsights/rhsm-subscriptions/blob/develop/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java#L138)

Made some changes where we query the offering table up front for the list of skus, .filter the subscriptions (coming from the service), and then perform sync after filtering out unknowns.

Will probably have to make similar changes so that the rest of subscription sync (ie other than the force api) still works.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
@@ -361,7 +361,7 @@ public class SubscriptionSyncController {
           boolean doesContain = knownOfferings.contains(sku);
 
           if (!doesContain) {
-            continue;
+            return;
           }
 
           syncSubscription(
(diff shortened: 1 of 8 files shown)


src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -135,15 +136,6 @@ public void syncSubscription(
       return;
     }
 
-    if (!offeringRepository.existsById(sku)) {
```
lindseyburnett, line 138: Ripping this out probably breaks the subscription sync job

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -343,18 +335,47 @@ public void deleteSubscription(String subscriptionId) {
...
         getActiveSubscriptionsForOrg(orgId)
             .collect(
                 Collectors.toMap(
                     org.candlepin.subscriptions.db.model.Subscription::getSubscriptionId,
                     sub -> sub));
+    log.info("Stop getting active subscriptions for org");
+
+    log.info("Start syncing subscriptions", subscriptionMap);
+    subscriptions.stream()
+        .filter(this::shouldSyncSub)
```
kflahert, line 350: shouldSyncSub() is called again in 'syncSubscription()' so I think we should get rid of it here.

kahowell:
Closing for now

PR #1024, opened 2022-05-23
ENT-5005: Ensure billingAccountId is set before it's inserted to the db as a key

lindseyburnett:


swatch-core/src/main/java/org/candlepin/subscriptions/tally/UsageCalculation.java
```diff
@@ -41,66 +50,16 @@ public class UsageCalculation {
...
-          && Objects.equals(usage, that.usage)
-          && Objects.equals(billingProvider, that.billingProvider)
-          && Objects.equals(billingAccountId, that.billingAccountId);
-    }
-
-    @Override
-    public int hashCode() {
-      return Objects.hash(productId, sla, usage);
-    }
+    @NonNull private final String productId;
```
lindseyburnett, line 58: https://projectlombok.org/features/NonNull

swatch-core/src/main/java/org/candlepin/subscriptions/tally/UsageCalculation.java
```diff
@@ -41,66 +50,16 @@ public class UsageCalculation {
    *
    * <p>Note that already data is scoped to an account, so account is not included in the key.
```
lindseyburnett, line 51: Looking for suggestions on how to make this javadoc less confusing since we now include `billingAccountId`

kahowell (approved):

PR #1027, opened 2022-05-24
DRAFT: ENT-5019 Update host api to filter on proper fields for billing provider info

lindseyburnett:
Both the Host and HostTallyBucket tables include billingAccountId and billingProvider.  When querying the database, we want to be comparing the key of the HostTallyBucket instead.


src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -360,10 +360,7 @@ private void addBucketsFromEvent(Host host, Event event, Optional<TagMetaData> s
         Optional.ofNullable(event.getBillingProvider())
             .map(Event.BillingProvider::toString)
             .map(BillingProvider::fromString)
-            .orElse(
-                serviceTypeMeta
-                    .map(TagMetaData::getDefaultProvider)
```
kflahert, line 365: Why was this was removed?

kahowell: @kflahert, we don't need per-product defaults for the billing provider. It was extra unneeded complexity that was adding noise when we were debugging/tracing how things work.

kflahert (approved):

PR #1032, opened 2022-05-25
DRAFT: ENT-5009 Subtasks - Add liquibase changelog and hibernate entities, stub out tracking of cumulative totals

lindseyburnett:


swatch-core/src/main/java/org/candlepin/subscriptions/db/model/BillableUsageRemittanceEntity.java
```diff
@@ -0,0 +1,91 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+@Entity
+@Table(name = "billable_usage_remittance")
+@IdClass(BillableUsageRemittanceEntityPK.class)
+public class BillableUsageRemittanceEntity {
```
mstead, line 42: Let's add JavaDoc here describing what this is and how it is going to be used?

mstead, line 42: Let's add JavaDoc here describing what this is and how it is going to be used?

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/BillableUsageRemittanceEntityPK.java
```diff
@@ -0,0 +1,71 @@
...
+package org.candlepin.subscriptions.db.model;
+
+import java.io.Serializable;
+import java.time.OffsetDateTime;
+import javax.persistence.Column;
+import javax.persistence.Id;
+import lombok.Data;
+
+@Data
+public class BillableUsageRemittanceEntityPK implements Serializable {
```
mstead, line 30: Instead of duplicating all these key fields in the BillableUsageRemittanceEntity, we can make this @Embeddable and define it as `@EmbeddedId private  BillableUsageRemittanceEntityPK key` in  BillableUsageRemittanceEntity.

mstead, line 30: Instead of duplicating all these key fields in the BillableUsageRemittanceEntity, we can make this @Embeddable and define it as `@EmbeddedId private  BillableUsageRemittanceEntityPK key` in  BillableUsageRemittanceEntity.

mstead (changes requested):
Overall I like where this is going. I noted a couple things that caught my eye.

lindseyburnett:
closing and passing off to @mstead

PR #1054, opened 2022-05-31
ENT-5013: Update RHM Payload Mapper for unconfigured rhMarketplaceMetricId

ksynvrit:
Add logic to ignore Usage metric for undefined rhmMetricId.
Update Unit test for ignoring undefined rhmMetricId.


ksynvrit:
I thought about adding logic to the `RhMarketplaceProducer` class, but left it as is because, it will just place empty usage to the rejected counter. Let me know, if more should be done. Thanks in advance.

InsightsDroid:
11bb30a1-6af0-436b-9b86-9f4f93a732ed
789d934d-c607-4d4d-8f4d-59987f466e8e

PR #1055, opened 2022-05-31
Add admin endpoint to terminate subscriptions

awood:
https://issues.redhat.com/browse/ENT-4932

PAYG subscriptions terminated more than an hour from the current time
will log a warning.


InsightsDroid:
1aa2e30b-483d-453c-b609-ee930b84e0a2
db0c9041-ef22-4360-91ca-ce18cc0c04af

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -343,6 +345,32 @@ public void deleteSubscription(String subscriptionId) {
...
+
+    var subscription = subscriptionResult.get();
+    subscription.setEndDate(terminationDate);
+
+    OffsetDateTime now = OffsetDateTime.now();
+    // The calculation returns a whole number, representing the number of complete units
+    // between the two temporals. For example, the amount in hours between the times 11:30 and
+    // 12:29 will zero hours as it is one minute short of an hour.
+    var delta = ChronoUnit.HOURS.between(terminationDate, now);
+    if (tagProfile.isProductPAYGEligible(subscription.getSku()) && delta > 0) {
```
kahowell, line 363: This `if` statement doesn't work as intended. Reason being that subscription SKU is like `MW01882` where `isProductPAYGEligible` deals with swatch product tags like `rhosak`. So we need to do a couple of things here:
1. Lookup the offering based on the SKU (using `offeringRepository`).
2. Lookup the swatch product tag via offering - the data is in tag profile, but it doesn't look like we made a method for this yet.

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -343,6 +345,39 @@ public void deleteSubscription(String subscriptionId) {
...
+    // end date.  We want validation to occur before we start mutating data.
+    subscription.setEndDate(terminationDate);
+
+    OffsetDateTime now = OffsetDateTime.now();
+    // The calculation returns a whole number, representing the number of complete units
+    // between the two temporals. For example, the amount in hours between the times 11:30 and
+    // 12:29 will zero hours as it is one minute short of an hour.
+    var delta = ChronoUnit.HOURS.between(terminationDate, now);
+    var productTag = tagProfile.tagForOfferingProductName(offering.getProductName());
+    if (tagProfile.isProductPAYGEligible(productTag) && delta > 0) {
```
kahowell, line 370: I found that delta is negative if termination date is in the past. `Math.abs(delta)` helps

kahowell (changes requested):

kahowell:
Confirmed via:

```
DEVTEST_EVENT_EDITING_ENABLED=true ./gradlew :bootRun&
http POST :8000/api/rhsm-subscriptions/v1/internal/subscriptions/terminate/12351253 x-rh-swatch-psk:dummy timestamp==2022-06-22T00:00:00Z
```

PR #1061, opened 2022-06-01
Remove calls to deprecated Spring Data method getById

awood:
Spring Data deprecated getById because it didn't do what people thought
it did.  Rather than fetching the entity, getById returned a JPA proxy
object.  The method has been replaced with the more accurately named
getReferenceById.  In the case of the code this commit patches, it
appears that the original author wanted the semantics associated with
findById and got tripped up by the confusing naming.

See https://github.com/spring-projects/spring-data-jpa/issues/2232


InsightsDroid:
29f13cb7-3fe3-4c8c-a442-7c506219ac06
fe718069-ae68-41c4-8c04-06a2c9d2eb28

kahowell (approved):

PR #1064, opened 2022-06-02
ENT-5016: Add Monthly Storage metric to rhosak tag metrics

ksynvrit:
Update Tag profile with new configuration for rhosak
Add new Unit Test to verify that configuration is working.
Update event.yaml & tally_snapshot.yaml for new UOM

PR #1065, opened 2022-06-02
Adjust awsDimension to actual values

kahowell:
What we had in tag_profile was placeholders. Notably, the actual values are the same, but without the prefixes.

Also removed the `awsDimension` from redhat.com:rhosak:storage_gb, as the changes in #1064 set up the metric we will use for awsDimension `storage_gb`.


InsightsDroid:
1a059d0e-c2d8-42eb-b394-4de3c88e25c9
ba2f1b00-a06a-4ce8-a9d1-20eb995523ee

ksynvrit (approved):

PR #1096, opened 2022-06-07
Add conduit sync API spec

haynesce:
https://issues.redhat.com/browse/ENT-5027


barnabycourt:
ok to test

swatch-system-conduit/build.gradle
```diff
@@ -18,6 +18,7 @@ openApiGenerate {
         interfaceOnly: "true",
         generatePom: "false",
         dateLibrary: "java8",
+        useTags: "true"
```
lindseyburnett, line 21: What's this setting do?

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,312 @@
...
+    variables:
+      PATH_PREFIX:
+        default: api
+      APP_NAME:
+        default: rhsm-subscriptions
+  - url: https://{HOSTNAME}/{PATH_PREFIX}/{APP_NAME}/v1
+    variables:
+      HOSTNAME:
+        enum:
+          - ci.cloud.redhat.com
```
lindseyburnett, line 18: @kahowell should we still include CI and QA in this enumeration?

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,312 @@
...
+          $ref: "../../../../spec/error-responses.yaml#/$defs/BadRequest"
+        '403':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/Forbidden"
+        '404':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/ResourceNotFound"
+        '500':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/InternalServerError"
+
+  /internal/organizations/remove/org/{orgs}:
+    description: "Initiate removing a given list of Org IDs from database sync list"
```
barnabycourt, line 114: Similar, singular vs. plural confusion.

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,312 @@
...
+          $ref: "../../../../spec/error-responses.yaml#/$defs/BadRequest"
+        '403':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/Forbidden"
+        '404':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/ResourceNotFound"
+        '500':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/InternalServerError"
+
+  /internal/organizations/inventory/org/{org_id}:
+    description: "Initiate viewing conduit representation of an org's systems from RHSM"
```
barnabycourt, line 173: Description is confusing. This appears to be returning the list for an org.

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,312 @@
...
+    parameters:
+      - name: org_id
+        in: path
+        required: true
+        schema:
+          type: string
+    get:
```
barnabycourt, line 180: The inventory list should be paged with some reasonable maximum number returned in a given call. 
You can see https://github.com/RedHatInsights/rhsm-subscriptions/blob/b64cc53272ed8f7bea37e19158bf9c50c7663f52/api/rhsm-subscriptions-api-spec.yaml#L675 & the offset & limit usage for details. 
The API should specify what values will be used by default & what the maximum values that can be specified for limit would be.

barnabycourt (commented):
Overall this is a great first pass. I'd like to see some more detail in the descriptions & documentation portion as well as fixing a few of the areas where there appear to be contradictions. 

The biggest change is that the endpoints that are returning lists need to have paging information (to return a limited number at a time) and documentation on the maximum number of items returned per page as well as the default number of items returned per page.

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
barnabycourt: Is there a way to avoid specifying hostnames which can change in the API at all?

swatch-system-conduit/build.gradle
kflahert: It makes it so that the tag attributes in your spec are used as class names when you generate them.

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
kahowell: Yeah, we let's just drop `servers` from the spec altogether (it's optional anyways).

kahowell (changes requested):
We should make `limit` `required: true` and add a `maximum` as discussed this morning. Otherwise looks good!

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,314 @@
...
+          content:
+            application/vnd.api+json:
+              schema:
+                $ref: "#/components/schemas/DefaultResponse"
+        '400':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/BadRequest"
+        '403':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/Forbidden"
+        '404':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/ResourceNotFound"
```
awood, line 50: Would this call actually return a 404, since it's not a call to a specific org but a general global operation?

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,314 @@
...
+          description: "The request for adding organizations to the sync list has completed successfully."
+          content:
+            application/vnd.api+json:
+              schema:
+                $ref: "#/components/schemas/DefaultResponse"
+        '400':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/BadRequest"
+        '403':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/Forbidden"
+        '404':
```
awood, line 78: Not sure this one would actually return a 404 either since its adding to a global list.

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,314 @@
...
+          $ref: "../../../../spec/error-responses.yaml#/$defs/ResourceNotFound"
+        '500':
+          $ref: "../../../../spec/error-responses.yaml#/$defs/InternalServerError"
+  /internal/organizations-sync-list:
+    description: "Initiate altering database sync list by either adding or removing a given list of Org IDs"
+    parameters:
+      - name: org_ids
```
awood, line 56: We should define this as an array based query parameter.  Something like this should do the trick

```
- name: org_ids
  in: query
  required: true
  style: form
  explode: false
  schema:
    type: array
    items:
      type: string
```
See https://swagger.io/docs/specification/describing-parameters/#query-parameters and https://swagger.io/docs/specification/serialization/

swatch-system-conduit/src/main/spec/internal-organizations-sync-api-spec.yaml
```diff
@@ -0,0 +1,314 @@
...
+      tags:
+        - internalOrganizations
+      responses:
+        '202':
+          description: "The request for syncing all organizations is successfully running."
+          content:
+            application/vnd.api+json:
+              schema:
+                $ref: "#/components/schemas/DefaultResponse"
+        '400':
```
awood, line 45: I don't think we can throw a BadRequest here since it's just a completely empty POST

kahowell (approved):

awood (approved):

PR #1127, opened 2022-06-13
ENT-5094: Retry billable-usage-producer transient errors

kahowell:
https://issues.redhat.com/browse/ENT-5094

With warning messages rather than verbose error messages with stack traces. Note that this will retry indefinitely...

Note that it was already retrying 10 times by default, but without backoff, and I verified that it failed even with retries a couple
of times in stage environment. By retrying indefinitely (for transient errors only), we'll eventually work past transient issues.
kafka topic lag will detect if we ever get into a situation where this causes issues.

Testing
-------

Run as

```
SPRING_PROFILES_ACTIVE=worker ./gradlew :bootRun
```

Simulate a transient error by replacing the logic in `TallySummaryMessageConsumer#receive` with:

```
throw new ObjectOptimisticLockingFailureException("test", "test");
```

Then use the kafka producer in intellij to queue up a message to `platform.rhsm-subscriptions.tally` with value `{}`.

You can also simulate other errors if desired to see that the retry remains unchanged for other exception types.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/billing/BillingProducerConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/billing/BillingProducerConfiguration.java
@@ -23,22 +23,28 @@ package org.candlepin.subscriptions.tally.billing;
 import static org.candlepin.subscriptions.task.queue.kafka.KafkaTaskProducerConfiguration.getConfigProps;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


InsightsDroid:
16ab6b6b-0459-4575-ad9d-a25113ee62db
d06ed884-ffb3-4440-bf46-18027cba6f26

mstead:
Feel free to merge after tests if I don't get to it before you.

PR #1153, opened 2022-06-21
Change response code for missing account from 204 to 404

kahowell:
This was causing issues w/ swatch-producer-aws crashlooping.

I also renamed the error code constant to make it clear that the code is to be used when the account does not exist according to RH IT services.


lindseyburnett (approved):

kahowell:
c4924b3 to 9ed5b37: spotlessApply

PR #1157, opened 2022-06-21
Filter expired subscriptions from IT subs queries

kahowell:
In order to reduce load of data to subscriptions we actually care about.


kahowell:
Closing, this causes problems with subscriptions expiring

kahowell:
s/expiring/terminations/

PR #1165, opened 2022-06-22
Fix ID parsing for UMB messages

kahowell:
There were a few things wrong with the mocked message:

1. `Identifiers` can have additional elements.
2. `Subscription number` is written in an `Identifier` element.

This updates the mocked message and code appropriately.

Testing
-------

Testing via `mocked-subscription-message.xml`/unit tests.


InsightsDroid:
20bbf174-03ad-4d12-a93f-3ed25eb99624

PR #1166, opened 2022-06-22
DRAFT: Add billing provider/account to response payload of subscriptions api

lindseyburnett:
Part of ENT-5123.  Got this far before the story was moved back to the backlog.  billing_provider and billing_account_id will always return null from this API for now, but at least they're there and exist for the UI that is expecting them.

Need to figure out what data to get in here in order to test via curl command

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
@@ -293,6 +293,15 @@ public class SubscriptionTableController {
       skuCapacity.setNextEventDate(nearestEventDate);
       skuCapacity.setNextEventType(SubscriptionEventType.END);
     }
+
+    //TODO
+    subscription.getBillingAccountId();
+    subscription.getBillingProvider();
+    subscription.getBillingProviderId();
+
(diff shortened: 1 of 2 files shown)


InsightsDroid:
Test results are available in Ibutsu:
https://url.corp.redhat.com/ibutsu-runs-ef5ce580-1f24-495a-ba57-0d49ec8e7281

src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
```diff
@@ -293,6 +293,11 @@ public void addOnDemandSubscriptionInformation(
       skuCapacity.setNextEventDate(nearestEventDate);
       skuCapacity.setNextEventType(SubscriptionEventType.END);
     }
+
+    //Populate as part of ENT-5123
+    skuCapacity.setBillingAccountId(null);
+    skuCapacity.setBillingProvider(null);
```
lindseyburnett, line 299: what do we set here?  billing_provider enum, or billing_provider_id

lindseyburnett: billing_provider

kahowell:
Closing for now

PR #1184, opened 2022-06-28
Create a Splunk HEC error handler that counts errors

awood:
Keeping a count of HEC errors in the Spring metrics actuator will allow
us to use OpenShift's monitoring capabilities to alert us when Splunk
communication fails.

This commit also changes the logging configuration slightly.  Before,
the root logger was set to INFO which resulted in INFO messages from all
third party libraries being logged.  Generally we're not interested in
these, so I changed the root level to WARN and then set the log level
for all classes under org.candlepin to INFO.

---
# Testing
* `ENABLE_SPLUNK_HEC=true;SPLUNK_HEC_URL=http://localhost:12345 ./gradlew :bootRun`
* On startup, you'll see a message printed to `stderr` reading `2022-06-28 19:14:30.654 [OkHttp http://localhost:12345/...] HttpEventCollectorError exception for batch of 10 log events. Total errors/events: 2/20. Message: ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:12345`
* If you go to `localhost:9000/metrics`, you'll see a Prometheus metric `splunk_hec_failure_total`.  The value will be `1.0`.
* Perform some operations with the application.  Or just insert a logging statement in a resource method and call the method repeatedly.  You can just apply the patch below and run the following command 

`http :8000/api/rhsm-subscriptions/v1/tally/products/RHEL granularity==monthly beginning==2021-03-02T00:00:00.000Z ending==2021-03-03T23:59:59.999Z   x-rh-identity:$(echo -n '{"identity":{"account_number":"account123","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w0)`

```patch
diff --git a/src/main/java/org/candlepin/subscriptions/resource/TallyResource.java b/src/main/java/org/candlepin/subscriptions/resource/TallyResource.java
index aa0f8c03..ca42163a 100644
--- a/src/main/java/org/candlepin/subscriptions/resource/TallyResource.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/TallyResource.java
@@ -398,6 +398,7 @@ public class TallyResource implements TallyApi {
             null,
             offset,
             limit);
+    log.error("I am error");
 
     Page<org.candlepin.subscriptions.db.model.TallySnapshot> snapshotPage =
         repository.findSnapshot( 
```
* Every 10 log messages (Splunk HEC batches the message reporting), another error message will appear on `stderr` and the counter will increment by 1.


InsightsDroid:
Test results are available in Ibutsu:
https://url.corp.redhat.com/ibutsu-runs-1c664e00-7312-4cf6-a5f0-a8901dea6ada

kahowell (approved):

PR #1262, opened 2022-07-27
SWATCH-257: Tolerate missing accountnumber

kartikshahc:
https://issues.redhat.com/browse/SWATCH-257

1. **Description**:

- Currently, tolerate-missing-account-number has default value as false. It is set in stage as true and in prod it is set as false. This is defined in app-interface.

App-interface PR: ```https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44020```

2. **Test Steps**:

- The flag is defined in swatch-system-conduit/src/main/resources/application.yaml

- Change the Stub: StubRhsmApi

- Comment out the account number to test the above flag locally.

 For false it will throw error and for true it will tolerate the missing account number.

-Run: `RHSM_USE_STUB=true DEV_MODE=true ./gradlew :swatch-system-conduit:bootRun`

Another terminal execute:
`curl -X 'POST'  'http://localhost:8000/api/rhsm-subscriptions/v1/internal/rpc/syncOrg' -H 'accept: application/vnd.api+json' -H 'x-rh-swatch-psk: dummy' -H 'Content-Type: application/json' -d '{ "org_id": "org123" }'`

### DIFF
--- a/deploy/rhsm-clowdapp.yaml
+++ b/deploy/rhsm-clowdapp.yaml
@@ -4,6 +4,8 @@ kind: Template
 metadata:
   name: rhsm
 parameters:
+  - name: TOLERATE_MISSING_ACCOUNT_NUMBER
+    value: 'false'
   - name: SUBSCRIPTIONS_HAWTIO_BASE_PATH
     value: /app/rhsm-subscriptions/hawtio
   - name: RH_MARKETPLACE_HAWTIO_BASE_PATH
@@ -1188,6 +1190,8 @@ objects:
                 secretKeyRef:
                   name: splunk-hec-external
                   key: token
+            - name: TOLERATE_MISSING_ACCOUNT_NUMBER
+              value: ${TOLERATE_MISSING_ACCOUNT_NUMBER}
             - name: SPLUNK_SOURCE
               value: ${SPLUNK_SOURCE}
             - name: SPLUNK_SOURCE_TYPE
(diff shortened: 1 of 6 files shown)


lindseyburnett (changes requested):
In order for the new `TOLERATE_MISSING_ACCOUNT_NUMBER` to be available to override the default in the application.yaml file, the OpenShift/ClowdApp templates need to be updated to have it added as a parameter.

See this commit for an example of the how to update a ClowdApp template: https://github.com/RedHatInsights/rhsm-subscriptions/commit/7f03cf5708d5609e13340d9dc817b2bbea0adc81#diff-a51abd82d91ba26dc24c7aa58fa54eb8bd506abe079eb3b01f991fe460d39b97.

Here's the file/section you'll want to modify for that: https://github.com/RedHatInsights/rhsm-subscriptions/blob/develop/deploy/rhsm-clowdapp.yaml#L1145 

For the standard OpenShift template change, you'll want to make the same kind of change, but in this file: https://github.com/RedHatInsights/rhsm-subscriptions/blob/develop/templates/rhsm-conduit.yaml

lindseyburnett (approved):

lindseyburnett:
MR to toggle TOLERATE_MISSING_ACCOUNT_NUMBER to 'true' during ephemeral environment deployments

https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44181

PR #1287, opened 2022-08-01
SWATCH-157 Verify Org ID Exists in Sync List - API Implementation

chaynesred:
This implements the REST API for checking if an Org ID exists in the sync list in accordance with API spec.

https://issues.redhat.com/browse/SWATCH-157

Tested using test file and manually using endpoint.

### DIFF
--- a/swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/admin/InternalOrganizationSyncResourceTest.java
+++ b/swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/admin/InternalOrganizationSyncResourceTest.java
@@ -24,6 +24,8 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
+import static org.springframework.test.util.AssertionErrors.assertFalse;
+import static org.springframework.test.util.AssertionErrors.assertTrue;
 
 import org.candlepin.subscriptions.conduit.InventoryController;
 import org.candlepin.subscriptions.conduit.job.OrgSyncTaskManager;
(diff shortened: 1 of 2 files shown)


swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/admin/InternalOrganizationSyncResourceTest.java
```diff
@@ -68,4 +70,20 @@ void syncOrgShouldThrowInternalServerError() throws Exception {
   void syncAllOrgsShouldReturnSuccess() {
     assertEquals("Success", resource.syncFullOrgList().getStatus());
   }
+
+  @Test
+  void hasOrgInSyncListShouldReturnTrue() {
+    when(repo.existsById("123")).thenReturn(true);
+    assertTrue(
+        "Org ID expected to exist but does not",
```
kartikshahc, line 78: Shouldn't the assertTrue message be "Org Id exists" instead of "does not"?

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/admin/InternalOrganizationSyncResourceTest.java
```diff
@@ -68,4 +70,20 @@ void syncOrgShouldThrowInternalServerError() throws Exception {
...
+        resource.hasOrgInSyncList("123").getExistsInList());
+  }
+
+  @Test
+  void hasOrgInSyncListShouldReturnFalse() {
+    when(repo.existsById("123")).thenReturn(false);
+    assertFalse(
+        "Org ID expected to not exist but does",
```
kartikshahc, line 86: In this case shouldn't it be "Org Id doesn't exist"?

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/admin/InternalOrganizationSyncResourceTest.java
chaynesred: This message is displayed only if the assertion fails, so if it does not exist then this message is displayed.

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/admin/InternalOrganizationSyncResourceTest.java
chaynesred: Same as above comment but vice versa

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/admin/InternalOrganizationSyncResourceTest.java
kartikshahc: Usually the assertion shouldn't fail which means in this case it should be always true. May be, I am missing something here.

kflahert: @kartikshahc the 'message' field only gets displayed when an assertion failed. It shouldn't fail but if something in the code changes and now it does fail this message will explain why the test failed.

kartikshahc: Ah ok! Makes sense will resolve these comments in that case.

kartikshahc (approved):

PR #1308, opened 2022-08-03
Limit trigger recursion to (hopefully) avoid deadlock

kahowell:
NOTE: I was unable to reproduce. Hoping to get this into stage, and see if it helps.


kahowell:
f372b8b to 289c133: add `if exists` in places I missed it.

lindseyburnett (approved):

PR #1313, opened 2022-08-04
Remove token-refresher secrets since Clowder handles sidecars

lindseyburnett:


kahowell (approved):

kahowell:
Was incorporated into #1318

PR #1318, opened 2022-08-05
Fix EE deployment issues

kahowell:
Should fix failure to deploy in EE.


kahowell:
Failing test seems totally unrelated to the change: https://ci.ext.devshift.net/job/RedHatInsights-rhsm-subscriptions-pr-check/1106/testReport/tests.rbac/test_rbac/test_verify_rbac_with_swatch_group_permission/

For the sake of progress, merging and we can dig into the test failure with subsequent test runs on other PRs.

PR #1321, opened 2022-08-05
SWATCH-241 Set orgId on TallySummary when mapping snapshots

mstead:
https://issues.redhat.com/browse/SWATCH-241

* Add org_id field to TallySummary definition.
* Set orgId on TallySummary when snapshots are mapped to the summary.

**Testing Notes**
At the moment we are not really able to manually test that the orgId will be set on the TallySummary since the orgId is currently coming from the snapshots. The best we can do at the moment is to ensure that the setting of the orgId is unit tested appropriately.

**TallySnapshot.orgId will be populated by the following:**
- https://issues.redhat.com/browse/SWATCH-97
- https://issues.redhat.com/browse/SWATCH-459

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/tally/TallySummaryMapperTest.java
+++ b/src/test/java/org/candlepin/subscriptions/tally/TallySummaryMapperTest.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


src/test/java/org/candlepin/subscriptions/tally/TallySummaryMapperTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.candlepin.subscriptions.db.model.HardwareMeasurementType;
+import org.candlepin.subscriptions.db.model.ServiceLevel;
+import org.candlepin.subscriptions.db.model.TallyMeasurementKey;
+import org.candlepin.subscriptions.db.model.TallySnapshot;
+import org.candlepin.subscriptions.db.model.Usage;
+import org.candlepin.subscriptions.json.Measurement.Uom;
+import org.candlepin.subscriptions.json.TallySummary;
+import org.junit.jupiter.api.Test;
+
+public class TallySummaryMapperTest {
```
awood, line 42:
```suggestion
class TallySummaryMapperTest {
```

awood (approved):

PR #1340, opened 2022-08-16
Add warning message if ownerId/orgId is null in account calculations

lindseyburnett:


kartikshahc:
I have incorporated these changes in a different way in SWATCH-309 branch since when the org is not available we will use account number but, with the correct header.

PR #1345, opened 2022-08-18
SWATCH-494: Use Join fetch on Host child entities

kahowell:
https://issues.redhat.com/browse/SWATCH-494

This should reduce roundtrips to the DB for host updates.


lindseyburnett (approved):

kahowell:
This does nothing actually given the eager fetches already present, as mentioned by @lindseyburnett in a chat conversation. Closing.

PR #1351, opened 2022-08-19
Add missing DB_CHANGELOG_CLEANUP_SQL env var

kahowell:
Missed in #1350. Doh!


kahowell:
Self-merging since trivial.

PR #1355, opened 2022-08-19
Various tally performance/stability fixes

kahowell:
1. Set postgresql statement_timeout to 5min

Except for liquibase updates, where it is set to undefined (no timeout).

2. Remove usages of `clear()` in JPA collections

They were causing unnecessary recreates and issues with trying to apply version column.

3. Avoid extraneous selects w/ host_tally_bucket changes
    
By providing a version column, hibernate doesn't need to check the DB to see if the record exists before inserting. I'm not entirely sure why it can't make this determination without this change...

4. Make tally queries use join fetches, reducing DB queries


mstead (approved):

mstead:
Approved pending PR check.

mstead (changes requested):
Something is up with the PR check. Approved pending passing checks.

mstead (approved):

mstead:
Approved pending PR checks.

kahowell:
NOTE: I want to re-evaluate whether the optimistic locking change is still helping or necessary.

mstead (approved):
@kahowell The latest commits look good to me and make sense. General testing of tally appears to behaving correctly. At this point I'd be Ok with testing this out again.

PR #1370, opened 2022-08-26
Use swatch-producer-aws pattern to remove gradle builds in dockerfiles

barnabycourt:
The overall goal of this PR is to reduce the number of general gradle
builds that are done when the pr_check.sh script is run.

This adds an override to the .gitignore because the .dockerignore is a
symlink to .gitignore and otherwise the docker COPY command can not be used
to add any finished jars from the build directories.


kahowell (approved):

kahowell:
Test failures unrelated.

kahowell (approved):

PR #1371, opened 2022-08-26
Make our PR check deployment slimmer

kahowell:
1. Remove HBI deps
2. Add rbac to the API deployment (we were getting it for free w/ HBI deployment).


kahowell:
Interestingly, this didn't seem to impact our deployment time, but it did of course have an effect on quota. (I manually invoked the same bonfire commands to deploy to EE and test via iqe).

kahowell:
Needs xjoin because iqe tests depend on HBI queries that require HBI to function. This is feeling like a fool's errand.

PR #1387, opened 2022-08-30
Fix and prevent null values for instance_count and sockets

lindseyburnett:
Related to: https://issues.redhat.com/browse/SWATCH-468

On the main branch first

* Shove stage data into your local instance.  i posted a helper script in slack.  hit me up if you can't find it
* Run the purgesnapshot job and observe error

```bash
SPRING_PROFILES_ACTIVE="purge-snapshots,kafka-queue" DEV_MODE=true ./gradlew :bootRun
```

```
Caused by: java.lang.IllegalArgumentException: Can not set int field org.candlepin.subscriptions.db.model.HardwareMeasurement.instanceCount to null value
```

Switch branches, then retest.  Job should complete without error

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurement.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurement.java
@@ -23,7 +23,6 @@ package org.candlepin.subscriptions.db.model;
 import java.io.Serializable;
 import javax.persistence.Column;
 import javax.persistence.Embeddable;
-import javax.validation.constraints.NotNull;
 import lombok.Getter;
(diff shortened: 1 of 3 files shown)


swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurement.java
```diff
@@ -31,9 +32,13 @@
 @Getter
 @Setter
 public class HardwareMeasurement implements Serializable {
+
+  @NotNull
```
kahowell, line 36: `@NotNull` on `int` a bit redundant, since Java will default to `0` but I don't mind.

kahowell (approved):

PR #1388, opened 2022-08-30
Remove probes from all scheduled jobs

mstead:


barnabycourt:
Probe removed because it was causing deployment errors when the pod started up, did it's work, & completed before a probe returned successfully. The window during which the probe would work was too small (~ 1.5 seconds out of 25 seconds total runtime)

PR #1397, opened 2022-09-01
Simplify sasl handling

kahowell:


swatch-core/src/main/java/org/candlepin/subscriptions/clowder/KafkaJaasBeanPostProcessor.java
```diff
@@ -20,13 +20,12 @@
...
 import org.springframework.beans.BeansException;
-import org.springframework.beans.InvalidPropertyException;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.boot.autoconfigure.kafka.KafkaProperties;
 import org.springframework.core.Ordered;
 import org.springframework.core.env.Environment;
+import org.springframework.util.StringUtils;
 
 /**
  * A bean post-processor to correctly configure JAAS for Kafka. Clowder gives us a value indicating
```
awood, line 31: The Javadoc needs updating as a result of these changes.  Also please javadoc the methods and maybe drop a line comment or two about how this is all meant to to work in the various scenarios.

awood (changes requested):

awood (approved):

PR #1406, opened 2022-09-06
SWATCH-511: add rhacs skus to product allow list

kflahert:
https://issues.redhat.com/browse/SWATCH-511

Test Steps:

- Deploy to EE
- Check config map capacity-allowlist for new entries


kahowell (changes requested):
"RHACS" is not a SKU

PR #1409, opened 2022-09-07
Use IQE plugin image for tests

ntkathole:
Currently we are using [quay.io/cloudservices/iqe-tests:latest](http://quay.io/cloudservices/iqe-tests:latest) image which comes up with all iqe plugins bundle. Rather than using common image, let's use plugin specific image quay.io/cloudservices/iqe-tests:rhsm-subscriptions to avoid issues due to other plugins.


kahowell (approved):

kahowell:
Merging given only ci.ext checks exercise this change, and they passed.

PR #1410, opened 2022-09-07
SWATCH-488: set account_config pk to org_id

kflahert:
https://issues.redhat.com/browse/SWATCH-488

### DIFF
--- a/src/main/resources/liquibase/202209121018-update-primary-key-of-account-config.xml
+++ b/src/main/resources/liquibase/202209121018-update-primary-key-of-account-config.xml
@@ -1,0 +1,28 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+
+    <changeSet id="202209121018-1" author="kflahert" dbms="postgresql">
+        <comment>Set placeholder values for missing org_ids in stage</comment>
+        <update tableName="account_config">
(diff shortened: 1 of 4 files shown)


kahowell (changes requested):
I was looking at our stage environment, and realized this won't cleanly apply there, because there are records from test data which are missing org_id. I'm thinking the most pragmatic thing to do is to add another changeset before the account_id which inserts org_id as `PLACEHOLDER_$accountNumber`; we'll want to very careful with this, and ensure that there are no real accounts in prod that would be affected (e.g. because of a transient issue with the org id populator or services it uses) before we roll out to production.

src/main/resources/liquibase/202209121018-update-primary-key-of-account-config.xml
```diff
@@ -0,0 +1,28 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+
+    <changeSet id="202209121018-1" author="kflahert">
```
kahowell, line 8: Unit tests fail because this query won't work on hsqldb.

In order to skip this on hsqldb:

```suggestion
    <changeSet id="202209121018-1" author="kflahert" dbms="postgresql">
```

kahowell (changes requested):
Unit tests needing updates:
* AccountConfigRepositoryTest. saveAndUpdate()
* AccountConfigRepositoryTest. testDelete()
* AccountConfigRepositoryTest. testFindAccountsWithEnabledSync()
* AccountConfigRepositoryTest. testIsReportingEnabled()
* AccountConfigRepositoryTest. testOptInCount()
* OptInControllerTest. testOptInViaAccountNumberDoesNotUseApiIfOptInExists()

kahowell (approved):

PR #1419, opened 2022-09-09
SWATCH-545: Fix cloudigrade auth errors

awood:


src/main/java/org/candlepin/subscriptions/tally/CloudigradeAccountUsageCollector.java
```diff
@@ -99,30 +101,29 @@ public void enrichUsageWithCloudigradeData(
...
+
+    if (ObjectUtils.isEmpty(orgId)) {
+      /* As of this writing (9 Sep 2022), Cloudigrade does still support the
+      X-RH-CLOUDIGRADE-ACCOUNT-NUMBER header, so this statement is not true in the narrow sense.
+      However, sending both the account number header and the X-RH-CLOUDIGRADE-ORG-ID is a
+      recipe for confusion if one of them is empty (See SWATCH-545).  OpenAPI does not make it
+      easy to declare mutually exclusive parameters, so I've elected to just ban account number
+      since we are transitioning away from it anyway.
+      */
+      throw new IllegalStateException("Cloudigrade requires an org ID to authenticate.");
```
awood, line 116: I'd like a second opinion on this.  I can reconfigure the API client to keep sending account number through, but getting the OpenAPI specification right is going to be a little tricky.

kahowell: I'm generally okay with this, but looks like we will get unnecessary retries from https://github.com/RedHatInsights/rhsm-subscriptions/blob/dd51397d99b94b53a2572f2d893ad807684abbc1/src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java#L167-L182... We should adjust the exception handling here a bit so that it doesn't retry when org ID is missing.

kahowell: thinking about it, maybe easier to `log.warn` and `return null` instead of throwing.

kahowell (changes requested):
Need to adjust the handling of missing orgId so that we don't get superfluous retries.

kahowell (approved):

PR #1336, opened 2022-08-15
SWATCH-309 Use orgId instead of accountNumber

kartikshahc:

PR #1466, opened 2022-09-28
SWATCH-566 instead null check AccountUsageCalculation we will look up…

kartikshahc:
… org id from AccountConfigRepository.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
@@ -117,7 +117,7 @@ public class TallySnapshotController {
               context -> usageCollector.collect(this.applicableProducts, accountLookup)));
       String orgId = accountRepo.findOrgByAccountNumber(account);
       if (props.isCloudigradeEnabled()) {
-        if (StringUtils.hasLength(orgId)) {
+        if (StringUtils.hasText(orgId)) {
           attemptCloudigradeEnrichment(account, accountCalcs, orgId);
         } else {
           log.warn(
--- a/src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
@@ -22,11 +22,9 @@ package org.candlepin.subscriptions.tally;
 
 import io.micrometer.core.annotation.Timed;
 import java.time.OffsetDateTime;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
 import org.candlepin.subscriptions.ApplicationProperties;
 import org.candlepin.subscriptions.db.AccountConfigRepository;
 import org.candlepin.subscriptions.db.model.Granularity;
@@ -114,9 +112,11 @@ public class TallySnapshotController {
       accountCalcs.putAll(
           retryTemplate.execute(
               context -> usageCollector.collect(this.applicableProducts, accountLookup)));
-      if (props.isCloudigradeEnabled() && null != accountCalcs.get(account)) {
-        String orgId = accountCalcs.get(account).getOwner();
+      String orgId = accountRepo.findOrgByAccountNumber(account);
+      if (props.isCloudigradeEnabled() && StringUtils.isNotEmpty(orgId)) {
         attemptCloudigradeEnrichment(account, accountCalcs, orgId);
+      } else {
+        log.info("orgId={} not found or cloudigrade is not enabled", orgId);
       }
     } catch (Exception e) {
       log.error("Could not collect existing usage snapshots for account {}", account, e);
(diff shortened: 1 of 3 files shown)


kartikshahc:
https://issues.redhat.com/browse/SWATCH-566:

If we are testing with cloudigrade, we have to follow the same steps mentioned in: https://github.com/RedHatInsights/rhsm-subscriptions/pull/1336

Test locally to see the org id is pulled from account config do the following steps:
1. ran this branch locally with `CLOUDIGRADE_ENABLED=true SPRING_PROFILES_ACTIVE="worker,api,kafka-queue" DEV_MODE=true ./gradlew :bootRun`
2. inserted mock data into hbi with `python bin/insert-mock-hosts --hbi --num-physical 2`
3. opted in account123/org123 with JMX endpoint
4. kicked off a tally for configured accounts via curl `curl 'http://localhost:9000/hawtio/jolokia/?maxDepth=7&maxCollectionSize=50000&ignoreErrors=true&canonicalNaming=false' -H 'Content-Type: text/json' --data-raw '{"type":"exec","mbean":"org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean","operation":"tallyConfiguredAccounts()"}' --compressed`
5. If the org id isn't present in account config table then it will log following message: orgId={} not found or cloudigrade is not enabled
6. If the org id is available in account config table then it will call cloudigrade service.

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
```diff
@@ -22,11 +22,9 @@
 
 import io.micrometer.core.annotation.Timed;
 import java.time.OffsetDateTime;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
awood, line 25: Please don't use wildcard imports for classes (static method wildcard imports are okay).  Let me know if you need help configuring your editor to avoid these.

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
```diff
@@ -22,11 +22,9 @@
...
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
```
awood, line 27: Please use Spring's StringUtils `hasLength` method.  That class is what we use throughout most of the rest of the codebase

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
```diff
@@ -114,9 +112,11 @@ public void produceSnapshotsForAccount(String account) {
       accountCalcs.putAll(
           retryTemplate.execute(
               context -> usageCollector.collect(this.applicableProducts, accountLookup)));
-      if (props.isCloudigradeEnabled() && null != accountCalcs.get(account)) {
-        String orgId = accountCalcs.get(account).getOwner();
+      String orgId = accountRepo.findOrgByAccountNumber(account);
+      if (props.isCloudigradeEnabled() && StringUtils.isNotEmpty(orgId)) {
         attemptCloudigradeEnrichment(account, accountCalcs, orgId);
+      } else {
+        log.info("orgId={} not found or cloudigrade is not enabled", orgId);
```
awood, line 119: This log message will get printed *a lot* in non-cloudigrade enabled environments.  Let's change the code to

if (props.isCloudigradeEnabled()) {
  if (StringUtils.hasLength(orgId)) {
    attemptCloudigradeEnrichment(...);
  } else {
    log.warn("Org Id {} not found for account {} during cloudigrade enrichment")
  }
}

and bump up the level to warn because if we can't find the org ID for the account, something is out of whack

awood (changes requested):

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
kartikshahc: @awood Sure I can replace it. I actually used it so, we don't have to trim the whitespace. But, I assume orgId will never be incorrect value " " (empty whitespace)? Correct?

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
kartikshahc: Thanks for the suggestion I have changed that setting.

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
kartikshahc: Done

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
awood: You can use `StringUtils.hasText` to check for that.  Here's the relevant source.

```java
	public static boolean hasText(@Nullable String str) {
		return (str != null && !str.isEmpty() && containsText(str));
	}

	private static boolean containsText(CharSequence str) {
		int strLen = str.length();
		for (int i = 0; i < strLen; i++) {
			if (!Character.isWhitespace(str.charAt(i))) {
				return true;
			}
		}
		return false;
	}
```

kartikshahc: Addressed code review changes.

PR #1474, opened 2022-09-30
Update Metric based Tally With OrgId

ksynvrit:
This PR is to address https://issues.redhat.com/browse/SWATCH-459

Added one liner to populate OwnerId for TallySnapshot. Refactor AccountService slightly
Added OrgId to event.csv and insert events.py

1. Start Application with following profiles -> worker, api
2. Insert events below with EventJMX
```
{"event_source": "prometheus", "event_type": "snapshot_redhat.com:rhosak:transfer_gb", "account_number": "account123", "org_id": "orgId123", "instance_id": "c5mu16smf1c22rn8e730", "timestamp": "2021-10-31T19:00:00Z", "expiration": "2021-11-01T00:00:00Z", "display_name": "c5mu16smf1c22rn8e730", "measurements": [{"uom": "Transfer-gibibytes", "value": "1.0"}], "role": "rhosak", "sla": "Premium", "service_type": "Kafka Cluster", "billing_provider": "red hat", "billing_account_id": "dummyId"}
```
3. Run Tally Hourly with the following parameter in the TallyJMX bean.
```
account123,
 2021-10-01T00:00Z,
 2021-11-01T00:00Z 
```
4. Observe in database that the tally_snapshot now have OwnerId populated.


src/test/java/org/candlepin/subscriptions/tally/MetricUsageCollectorTest.java
```diff
@@ -30,12 +30,7 @@
 import static org.mockito.Mockito.when;
 
 import java.time.OffsetDateTime;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
awood, line 33: Please omit this from the commit since the rest of the class doesn't actually have any changes

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -87,6 +87,12 @@ public CollectionResult collect(String serviceType, String accountNumber, DateRa
             .findById(new AccountServiceInventoryId(accountNumber, serviceType))
             .orElse(new AccountServiceInventory(accountNumber, serviceType));
 
+    // SWATCH-261 This logic should be implemented much cleaner
+    accountServiceInventory.getServiceInstances().values().stream()
```
awood, line 91: Why'd this jump locations out of curiosity?  Are we referencing the orgId in some of the method calls below or something?

awood (changes requested):

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
ksynvrit: I moved it , so it easier to find in the clean up in the other cards, and I was using it but I came to the decision to just use the one liner

src/test/java/org/candlepin/subscriptions/tally/MetricUsageCollectorTest.java
ksynvrit: Will Do

PR #1478, opened 2022-10-03
SWATCH-102: Set database ssl properties from clowder config

kflahert:
https://issues.redhat.com/browse/SWATCH-102

To test need to setup local database with SSL and add a mock clowder config similar to below:

```
{
	"database": {
		"sslMode": "require",
		"rdsCa": "-----BEGIN CERTIFICATE-----placeholderCertData-----END CERTIFICATE-----\n"
	}
}
```


mstead (commented):
Functionally seems to work correctly. One though/question...

Do we want to pin the PostProcessor class to 'rds' naming? Perhaps make it configurable, but default to 'rds'?

PR #1479, opened 2022-10-03
Fix kafka producer properties building

kahowell:
We weren't passing enough properties along into the producer config which meant producers were being instantiated without proper auth config for managed kafka.


mstead (approved):
Looks good.

PR #1481, opened 2022-10-03
Fix inventory host ingress references

kahowell:
This appears to be the only topic we missed when having our code reference actual topic names from clowder config.


kahowell:
fe1784b to 251e47c: fix one more failing test

PR #1503, opened 2022-10-11
SWATCH 408: Tally hypervisors as hypervisors, not virtual systems

awood:
I've got a few points to discuss with the reviewer:

* How should QE plan on testing this?
* Do we need to worry about migrating existing HostTallyBuckets?

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/collector/RHELProductUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/collector/RHELProductUsageCollector.java
@@ -76,7 +76,7 @@ public class RHELProductUsageCollector implements ProductUsageCollector {
       prodCalc.addPhysical(appliedCores, appliedSockets, 1);
       return Optional.of(
           createBucket(
-              prodCalc, true, appliedCores, appliedSockets, HardwareMeasurementType.PHYSICAL));
+              prodCalc, false, appliedCores, appliedSockets, HardwareMeasurementType.PHYSICAL));
     }
 
     // nothing applied to calculation so no bucket to return.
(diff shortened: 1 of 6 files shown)


src/main/java/org/candlepin/subscriptions/tally/collector/RHELProductUsageCollector.java
```diff
@@ -93,15 +94,26 @@ public Optional<HostTallyBucket> collectForHypervisor(
...
           account);
     }
 
     prodCalc.addHypervisor(appliedCores, appliedSockets, 1);
-    return Optional.of(
-        createBucket(
-            prodCalc, false, appliedCores, appliedSockets, HardwareMeasurementType.VIRTUAL));
+    HardwareMeasurementType hardwareType = translateHardwareType(hypervisorFacts.getHardwareType());
+    return Optional.of(createBucket(prodCalc, true, appliedCores, appliedSockets, hardwareType));
+  }
```
kahowell, line 105: Actually, this is where I expected we would see HardwareMeasurementType.HYPERVISOR, something like

```
createBucket(prodCalc, false, appliedCores, appliedSockets, HardwareMeasurementType.HYPERVISOR));
```

kahowell:
>    How should QE plan on testing this?

QE should be update any existing tests for Hypervisors. I expect we'd have two types of hypervisors tested:
* RHEL-based hypervisors, where I'd expect two contributions (PHYSICAL, and HYPERVISOR), and
* non-RHEL hypervisors where I'd expect a single contribution to HYPERVISOR.

Mocked data in HBI should be sufficient to test this.

>    Do we need to worry about migrating existing HostTallyBuckets?

No, we'll leave existing HostTallyBuckets alone (they will be updated on next tally anyways).

kahowell (changes requested):

PR #1518, opened 2022-10-14
move db definition to swatch-tally clowdapp

kflahert:


lindseyburnett (changes requested):
Could you migrate the deploy/clowder.md information to a section in the toplevel README.md, then we can get rid of the whole deploy folder later (i think? i don't think we need the clowdenv file that lives in there)

lindseyburnett (approved):

PR #1525, opened 2022-10-17
Workaround some migration issues w/ HardwareMeasurement

kahowell:
Because the `hardware_measurements` to `tally_measurements` migrations were very large, they can't reasonably be done via liquibase without blocking CronJobs from running. Previous PRs (#1520 and follow-ons) have been written to perform the migrations async in the background. In order to unblock the current release, we want to have the readers of the data read from either `hardware_measurements` or `tally_measurements` (whichever has data). (Once the migrations are done, we'll remove this workaround).

1. Drop instances from the changelog, because it's not actually used by the UI anyways.
2. Temporarily comment out the problematic changelog.
3. Restore HardwareMeasurement JPA entity.
4. Add logic to read cores/sockets, etc. from HardwareMeasurement when apparently missing from `tally_measurements` table.

Testing
=======

Insert some data to mimic a couple of scenarios:

```shell

insert into tally_snapshots(id, product_id, account_number, granularity, owner_id, snapshot_date, unit_of_measure, sla)
VALUES (gen_random_uuid(), 'RHEL', 'account123', 'DAILY', 'org123', '2022-10-16', null, 'Premium');
insert into tally_snapshots(id, product_id, account_number, granularity, owner_id, snapshot_date, unit_of_measure, sla)
VALUES (gen_random_uuid(), 'RHEL', 'account123', 'DAILY', 'org123', '2022-10-17', null, 'Premium');
insert into hardware_measurements(snapshot_id, measurement_type, cores, instance_count, sockets)
select id, 'TOTAL', 2, 3, 4
from tally_snapshots;
insert into tally_measurements(snapshot_id, measurement_type, uom, value)
select snapshot_id, measurement_type, 'CORES', 10
from hardware_measurements
limit 1;
insert into tally_measurements(snapshot_id, measurement_type, uom, value)
select snapshot_id, measurement_type, 'SOCKETS', 20
from hardware_measurements
limit 1;

EOF
```

Note that the above creates a couple of snapshots and then provides `tally_measurements` rows for only one of them, but provides `hardware_measurements` for both.

Ensure opt-in:

```shell
  x-rh-identity:$(echo -n '{"identity":{"account_number":"account123","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w0)
```

Tally:

```shell
http :8000/api/rhsm-subscriptions/v1/tally/products/RHEL \
  sla==Premium \
  granularity==DAILY \
  beginning=='2022-10-16T00:00Z' \
  ending=='2022-10-18T00:00Z' \
  x-rh-identity:$(echo -n '{"identity":{"account_number":"account123","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w0)
```

Notice that 2022-10-16 has data coming from tally_measurements (10 cores, 20 sockets), while 2022-10-17 has data coming from hardware_measurements (2 cores, 4 sockets).

### DIFF
--- a/src/main/resources/liquibase/202208251616-migrate-cores-sockets-data.xml
+++ b/src/main/resources/liquibase/202208251616-migrate-cores-sockets-data.xml
@@ -46,14 +46,4 @@
       on conflict(snapshot_id, measurement_type, uom) do update set value=excluded.value;
     </sql>
   </changeSet>
-  <changeSet id="202208251616-5" author="khowell" dbms="postgresql">
-    <comment>Copy data from hardware_measurements.instance_count to tally_measurements.</comment>
-    <sql>
-      insert into tally_measurements(snapshot_id, measurement_type, uom, value)
-      select snapshot_id, measurement_type, 'INSTANCES', instance_count
-      from hardware_measurements
-      where instance_count is not null
-      on conflict(snapshot_id, measurement_type, uom) do update set value=excluded.value;
-    </sql>
-  </changeSet>
 </databaseChangeLog>
--- a/src/main/resources/liquibase/changelog.xml
+++ b/src/main/resources/liquibase/changelog.xml
@@ -73,7 +73,11 @@
     <include file="liquibase/202208301033-default-null-hardware-measurements.xml"/>
     <include file="liquibase/202208301423-set-tally-snapshot-owner-id.xml"/>
     <include file="liquibase/202209121018-update-primary-key-of-account-config.xml"/>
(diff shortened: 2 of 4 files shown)


kahowell:
d118dfd to 5e3fd8b: update comment with proper psql invocation

mstead (approved):

kahowell:
Only failing test also fails on dep update, so merging

PR #1549, opened 2022-10-21
Mark orgId as optional in aws usage context lookup

kahowell:
Was causing:

```
SUBSCRIPTIONS1002: Client request failed validation. - getAwsUsageContext.orgId: must not be null
```

and

```
SWATCHAWS1004: Error looking up AWS usage context
```


kahowell:
Trivial fix, self-merging

PR #1553, opened 2022-10-21
SWATCH-588: Allow capacity reports to be filtered by hypervisor/non-hypervisor

awood:
If the report category is set to physical, virtual, or cloud, the capacity report will return only non-hypervisor capacity.  If the category is set to hypervisor, the hypervisor capacity will be returned. If no category is given, both types of capacity are returned.

See SWATCH-588


kflahert (approved):

swatch-core/src/main/java/org/candlepin/subscriptions/db/SubscriptionCapacityViewRepository.java
```diff
@@ -110,8 +122,49 @@ private SearchCriteria searchCriteriaMatchingUomOfSockets() {
...
+          SearchCriteria.builder()
+              .key(SubscriptionCapacityView_.VIRTUAL_SOCKETS)
+              .operation(SearchOperation.GREATER_THAN)
+              .value(0)
+              .build(),
+          SearchCriteria.builder()
+              .key(SubscriptionCapacityView_.VIRTUAL_CORES)
+              .operation(SearchOperation.GREATER_THAN)
+              .value(0)
+              .build());
```
kahowell, line 157: The semantics of this are `and`; instead we want `or`. This is most clearly seen in the generated query (captured from debug logging and reformatted):

```
http :8000/api/rhsm-subscriptions/v1/subscriptions/products/RHEL x-rh-identity:$(echo -n '{"identity":{
    "type":"User",
    "user":{"is_org_admin":true},
    "internal":{"org_id":"org123"}
  }}' | base64 -w0) granularity==DAILY beginning==2022-10-01T00:00Z ending==2022-10-05T00:00Z category==hypervisor
```

```sql
select subscripti0_.owner_id            as owner_id1_0_,
       -- more columns, removed for brevity
       subscripti0_.virtual_sockets     as virtual17_0_
from (SELECT sc.subscription_id,
             -- more columns, removed for brevity
             o.description as product_name
      FROM subscription_capacity sc
               JOIN subscription s on sc.subscription_id = s.subscription_id
          AND s.end_date > CURRENT_TIMESTAMP
               JOIN offering o on sc.sku = o.sku) subscripti0_
where subscripti0_.owner_id=?
  and subscripti0_.product_id=?
  and subscripti0_.virtual_sockets > 0
  and subscripti0_.virtual_cores > 0
  and subscripti0_.begin_date<=?
  and subscripti0_.end_date>=?
```

Because our generic searches don't have boolean capabilities (and I think it'd be overkill), it'll probably be necessary to push this into `org.candlepin.subscriptions.db.SubscriptionCapacityViewSpecification#mapSingleSearchCriteriaToPredicate` (maybe it's cleaner to just add a category field to that class?)

kahowell (changes requested):

swatch-core/src/main/java/org/candlepin/subscriptions/db/SubscriptionCapacityViewRepository.java
awood: > Because our generic searches don't have boolean capabilities (and I think it'd be overkill), it'll probably be necessary to push this into `org.candlepin.subscriptions.db.SubscriptionCapacityViewSpecification#mapSingleSearchCriteriaToPredicate` (maybe it's cleaner to just add a category field to that class?)

They do now! :grin: As of #1575

swatch-core/src/main/java/org/candlepin/subscriptions/db/SubscriptionCapacityViewRepository.java
```diff
@@ -140,8 +152,32 @@ static Specification<SubscriptionCapacityView> matchingUomForSockets() {
             builder.isTrue(root.get(SubscriptionCapacityView_.hasUnlimitedUsage)));
   }
 
+  static Specification<SubscriptionCapacityView> matchesCategories(
+      Set<SubscriptionReportCategory> subsReportCategories) {
+    return (root, query, builder) -> {
+      var predicate = builder.conjunction();
+      if (subsReportCategories.contains(SubscriptionReportCategory.NON_HYPERVISOR)) {
+        Expression<Boolean> nonhypervisorExp =
+            builder.or(
```
kahowell, line 161:
```suggestion
            builder.and(
```

kahowell (changes requested):

kahowell (approved):

PR #1568, opened 2022-10-26
SWATCH-575: Apply MAC address validation on HbiNetworkInterface object creation

awood:
See SWATCH-575


awood:
TODO: It would be good to get a unit test on this

awood:
> Tests are failing because we were actually depending on loose matching in a few places.

LOL.  Fixed now.

kahowell (approved):

PR #1575, opened 2022-10-28
Refactor Spring Data Specification usage to be less abstract

awood:


lindseyburnett (approved):

mstead (commented):
Being new to the Spring Data Specification, and recently working with it, IMO this PR establishes a pattern that I like and is pretty simple to follow.

Since @lindseyburnett has more experience in this area, I'll hand the official review and sign-off over to her.

PR #1589, opened 2022-11-03
Fix logging to display exception

kflahert:


src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
```diff
@@ -251,7 +251,7 @@ private String ensureOptIn(String orgId) {
       log.warn("Error while attempting to automatically opt-in orgId={}", orgId);
       // Keep the logs clean unless specified.
       if (log.isDebugEnabled()) {
-        log.debug("Opt-in error for orgId={}.", orgId, e);
+        log.debug(String.format("Opt-in error for orgId=%s.", orgId), e);
       }
```
awood, line 255: Would you mind just deleting the surrounding `if`?  It's not necessary since the `debug` method won't print unless debug is enabled and there's no extra cost since `orgId` is already a String and we don't need to run toString on it.  In fact, you could just change it to
```java
log.debug("Opt-in error for orgId=" + orgId, e);
```

I'm curious why the SLF4J placeholder syntax didn't work for you though.  Would you mind writing up an explanation and adding that to the commit message?

awood (changes requested):
Looks fine, but I'm curious why you had to make this change.

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusMeteringController.java
kflahert: @awood Updated this. Here's the commit message too:

This code was using method `public void debug(String format, Object arg1, Object arg2) `which
treated the exception as another argument for the string. Instead we want to use this Logger method
`public void debug(String msg, Throwable t);`

awood: Approved, but I'm curious why [Logback's LoggingEvent](https://github.com/qos-ch/logback/blob/d0dcec81cae593dcf4f73911f445c0a4055a946d/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java#L160) didn't successfully extract the throwable.  That's a mystery for another day.

awood (approved):

PR #1590, opened 2022-11-03
Add BurstBalance to DB stats to grafana dashboard & set default environment to production

barnabycourt:
Update the db dashboards to include a panel for the database burst balance. 
Also set default environment to production

You can export the dashboard json by:

`oc create --dry-run=client -f dashboards/grafana-dashboard-subscription-watch.configmap.yaml -o json | jq -r '.data["subscription-watch.json"]' > dashboard.json
`
and then import into stage grafana. Observe that all graphs are functional again.

### DIFF
--- a/dashboards/grafana-dashboard-subscription-watch.configmap.yaml
+++ b/dashboards/grafana-dashboard-subscription-watch.configmap.yaml
@@ -27,8 +27,7 @@ data:
       "editable": true,
       "fiscalYearStartMonth": 0,
       "graphTooltip": 0,
-      "id": 540,
-      "iteration": 1662728700464,
+      "iteration": 1667394379897,
       "links": [],
       "liveNow": false,
       "panels": [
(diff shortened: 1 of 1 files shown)


dashboards/grafana-dashboard-subscription-watch.configmap.yaml
```diff
@@ -3220,9 +3342,9 @@ data:
           },
           {
             "current": {
-              "selected": false,
-              "text": "AWS insights-stage",
-              "value": "AWS insights-stage"
+              "selected": true,
+              "text": "AWS insights-prod",
```
lindseyburnett, line 3346: `stage` vs `prod`

lindseyburnett (changes requested):

dashboards/grafana-dashboard-subscription-watch.configmap.yaml
barnabycourt: Not sure I follow, This would change the default selection to be production instead of stage. Is the request here to not do that or will it break something?

lindseyburnett (approved):

barnabycourt:
Commit comment & PR description updated to include information about changing the default environment to prod as well.

PR #1591, opened 2022-11-04
[hotfix] Sync allowlist with internal repo

kahowell:
There were some SKUs present in our internal repo not yet in this repo's copy.


kahowell:
Self-merged since allowlist contents are already in the internal allowlist copy.

PR #1592, opened 2022-11-04
Sync allowlist with internal repo

kahowell:
There were a few SKUs added to the internal repo not yet present in this repo.


kahowell:
Going to self-merge since this is a trivial config change and the contents are already in the internal allowlist copy.

PR #1595, opened 2022-11-07
Add pod level memory widget for the tally services

barnabycourt:
The divide by 1000000 in the equation is to convert to Mi used by app-interface specification.

### DIFF
--- a/dashboards/grafana-dashboard-subscription-watch.configmap.yaml
+++ b/dashboards/grafana-dashboard-subscription-watch.configmap.yaml
@@ -27,7 +27,7 @@ data:
       "editable": true,
       "fiscalYearStartMonth": 0,
       "graphTooltip": 1,
-      "iteration": 1667586950301,
+      "iteration": 1667877595996,
       "links": [],
       "liveNow": false,
       "panels": [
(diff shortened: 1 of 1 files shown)


kahowell (changes requested):
Instead of the divide, it's better to set `Unit` to `Data` -> `bytes(IEC)`

barnabycourt:
I'm happy to consider bytes(IEC) but It doesn't seem to quite match. For example, 
Openshift Console shows: 2.53 GB
Grafana using divide by 1000000: 2556
Grafana using bytes(IEC): 2.38 GB
Grafana using bytes(SI): 2.56 GB

It looks like bytes(SI) is closer. I'd rather over-estimate on our dashboard than under-estimate.  @kahowell what do you think?

kahowell:
@barnabycourt let's take a look at this tomorrow some time. It looks like IEC on my end, unless you click in to the generic prometheus query UI.

I also prefer IEC as we've been specifying our memory request/limits in IEC units.

barnabycourt:
I updated to use IEC memory unit and set the max value of the graph to the current quote limit per conversation with @kahowell

kahowell (approved):

PR #1611, opened 2022-11-09
Reorder account_services liquibase change

kahowell:
In order to fix an issue deploying where an error like:

```
Caused by: liquibase.exception.DatabaseException: ERROR: update or delete on table "account_services" violates foreign key constraint "fk_hosts_account_services" on table "hosts"
Detail: Key (account_number, service_type)=(6356567, HBI_HOST) is still referenced from table "hosts". [Failed SQL: (0) DELETE FROM public.account_services WHERE org_id is null]
```

happens because there are host rows that reference account_services rows missing org_id in stage specifically.


mstead (approved):
:+1:

PR #1624, opened 2022-11-10
SWATCH-631: Migrate Data for OrgId To TallySnapshot Table

ksynvrit:
This is address https://issues.redhat.com/browse/SWATCH-631

Add new migrator for OrgId to populate Tally Snapshot Add new endpoint to reference New data migration.
Adjusted query to filter properly by accountNumber

Test Steps:
1. Either clear org_id in Tally snapshot or insert test data below.
```
INSERT INTO public.tally_snapshots (id, product_id, account_number, granularity, org_id, snapshot_date, unit_of_measure, sla, usage, billing_provider, billing_account_id) VALUES ('c5310125-46bd-4991-94f8-dccadab62245', 'rhosak', 'account123', 'DAILY', '', '2021-10-31 00:00:00.000000 +00:00', null, 'Premium', '_ANY', 'red hat', 'dummyId');
INSERT INTO public.tally_snapshots (id, product_id, account_number, granularity, org_id, snapshot_date, unit_of_measure, sla, usage, billing_provider, billing_account_id) VALUES ('8d81e216-0c6a-432a-a4b4-a96a43186fd4', 'rhosak', 'account123', 'DAILY', '', '2021-10-31 00:00:00.000000 +00:00', null, '_ANY', 'Production', 'red hat', 'dummyId');
```
2. If you use the test data make sure to insert script in account config table
```
INSERT INTO public.account_config (account_number, opt_in_type, created, updated, org_id) VALUES ('account123', 'PROMETHEUS', '2022-10-07 16:13:04.231491 +00:00', '2022-10-07 16:13:04.231491 +00:00', 'orgId123');
INSERT INTO public.account_config (account_number, opt_in_type, created, updated, org_id) VALUES ('account321', 'PROMETHEUS', '2022-11-10 15:36:02.522000 +00:00', '2022-11-10 15:36:10.849000 +00:00', 'orgId321');
```
3. Now Boot run application and trigger the migrateOrgIdToTallySnapshot(String, int) method in Hawtio, pass null and 0, and all blank org_id in tally snapshot table will be filled with the correct values based on account config.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/jmx/TallyJmxBean.java
+++ b/src/main/java/org/candlepin/subscriptions/jmx/TallyJmxBean.java
@@ -25,6 +25,7 @@ import javax.validation.constraints.NotNull;
 import org.candlepin.subscriptions.resource.ResourceUtils;
 import org.candlepin.subscriptions.tally.admin.DataMigrationRunner;
 import org.candlepin.subscriptions.tally.admin.HardwareMeasurementMigration;
+import org.candlepin.subscriptions.tally.admin.OrgIdToTallySnapshotMigration;
 import org.candlepin.subscriptions.tally.job.CaptureSnapshotsTaskManager;
 import org.candlepin.subscriptions.util.DateRange;
 import org.candlepin.subscriptions.validator.ParameterDuration;
@@ -138,10 +139,30 @@ public class TallyJmxBean {
       description = "Offset to start from (may be null to start from beginning)")
   @ManagedOperationParameter(name = "batchSize", description = "Batch size")
   public void migrateHardwareMeasurements(String snapshotId, int batchSize) {
+    if (batchSize < 1) {
+      throw new IllegalArgumentException("Batch size cannot be zero");
+    }
     log.info(
         "hardware_measurements migration triggered over JMX by {}", ResourceUtils.getPrincipal());
     dataMigrationRunner.migrate(
         HardwareMeasurementMigration.class,
+        StringUtils.hasText(snapshotId) ? snapshotId : null,
+        batchSize);
+  }
+
+  @ManagedOperation(description = "Trigger tally_snapshots Org_id migration")
+  @ManagedOperationParameter(
+      name = "snapshotId",
+      description = "Offset to start from (may be null to start from beginning)")
+  @ManagedOperationParameter(name = "batchSize", description = "Batch size")
+  public void migrateOrgIdToTallySnapshot(String snapshotId, int batchSize) {
+    if (batchSize < 1) {
+      throw new IllegalArgumentException("Batch size cannot be zero");
+    }
+    log.info(
+        "Org_id tally_snapshot migration triggered over JMX by {}", ResourceUtils.getPrincipal());
+    dataMigrationRunner.migrate(
+        OrgIdToTallySnapshotMigration.class,
         StringUtils.hasText(snapshotId) ? snapshotId : null,
         batchSize);
   }
--- a/src/main/java/org/candlepin/subscriptions/tally/admin/OrgIdToTallySnapshotMigration.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/admin/OrgIdToTallySnapshotMigration.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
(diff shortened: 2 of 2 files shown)


barnabycourt:
@ksynvrit Do you have an estimate of the number of affected records which this PR will be updating in production?

ksynvrit:
> @ksynvrit Do you have an estimate of the number of affected records which this PR will be updating in production?

Not at the top of my head, but I check with gabi and this is the result I got in prod

    http -b https://gabi-subscription-watch-prod.apps.crcp01ue1.o9m8.p1.openshiftapps.com/query Authorization:"Bearer $(oc whoami -t)" query="select count(*) from tally_snapshots where org_id is null"
{
    "error": "",
    "result": [
        [
            "count"
        ],
        [
            "625145"
        ]
    ]
}

awood:
> 3\. Now Boot run application and trigger the migrateOrgIdToTallySnapshot(String, int) method in Hawtio, pass null and 0, and all blank org_id in tally snapshot table will be filled with the correct values based on account config.

A batch size of 0 does not work.  This tripped me up for a little while.  We should actually do some validation in the JMX call to disallow batch sizes of 0.  Could you please add that both for your new method and the old HardwareMeasurement method?

ksynvrit:
> > 3. Now Boot run application and trigger the migrateOrgIdToTallySnapshot(String, int) method in Hawtio, pass null and 0, and all blank org_id in tally snapshot table will be filled with the correct values based on account config.
> 
> A batch size of 0 does not work. This tripped me up for a little while. We should actually do some validation in the JMX call to disallow batch sizes of 0. Could you please add that both for your new method and the old HardwareMeasurement method?



> > 3. Now Boot run application and trigger the migrateOrgIdToTallySnapshot(String, int) method in Hawtio, pass null and 0, and all blank org_id in tally snapshot table will be filled with the correct values based on account config.
> 
> A batch size of 0 does not work. This tripped me up for a little while. We should actually do some validation in the JMX call to disallow batch sizes of 0. Could you please add that both for your new method and the old HardwareMeasurement method?

I have added validation steps to both methods, let me know if your fine with the changes

src/main/java/org/candlepin/subscriptions/tally/admin/OrgIdToTallySnapshotMigration.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.util.StringUtils;
+
+@Slf4j
+public class OrgIdToTallySnapshotMigration extends DataMigration {
+
+  public static final SqlRowSetResultSetExtractor SQL_ROW_SET_RESULT_SET_EXTRACTOR =
+      new SqlRowSetResultSetExtractor();
+
+  public static final String UPDATE_SQL =
+      "update tally_snapshots set org_id=? where id =? and org_id=''";
```
awood, line 41: This doesn't work with null orgIds; only empty string orgIds.  If you insert the records below, the migration does not work.  It will find the records thanks to the `!StringUtils.hasText(tallyOrg)` check but the insert won't do anything because both condition aren't true.

```sql
INSERT INTO public.tally_snapshots (id, product_id, account_number, granularity, org_id, snapshot_date,
unit_of_measure, sla, usage, billing_provider, billing_account_id) VALUES ('c5310000-46bd-4991-94f8-dccadab62245',
'rhosak', 'account123', 'DAILY', null, '2021-10-31 00:00:00.000000 +00:00', null, 'Premium', '_ANY', 'red hat',
'dummyId');
INSERT INTO public.tally_snapshots (id, product_id, account_number, granularity, org_id, snapshot_date,
unit_of_measure, sla, usage, billing_provider, billing_account_id) VALUES ('8d810000-0c6a-432a-a4b4-a96a43186fd4',
'rhosak', 'account123', 'DAILY', null, '2021-10-31 00:00:00.000000 +00:00', null, '_ANY', 'Production', 'red hat',
'dummyId');
```

We know the orgIds are null and not empty string using Gabi:

```bash
% http -b https://gabi-subscription-watch-stage.apps.crcs02ue1.urby.p1.openshiftapps.com/query query="select count(id) from tally_snapshots where org_id =''" 'Authorization:Bearer $TOKEN'
{
    "error": "",
    "result": [
        [
            "count"
        ],
        [
            "0"
        ]
    ]
}

% http -b https://gabi-subscription-watch-stage.apps.crcs02ue1.urby.p1.openshiftapps.com/query query="select count(id) from tally_snapshots where org_id is null" 'Authorization:Bearer $TOKEN'
{
    "error": "",
    "result": [
        [
            "count"
        ],
        [
            "3248"
        ]
    ]
}
```

I would just remove the `and org_id=''` clause.  We already know we want to update that record; there's no reason to second-guess it here.

awood (changes requested):

src/main/java/org/candlepin/subscriptions/jmx/TallyJmxBean.java
```diff
@@ -138,11 +139,33 @@ public void tallyAllAccountsByHourly(String start, String end) throws IllegalArg
       description = "Offset to start from (may be null to start from beginning)")
   @ManagedOperationParameter(name = "batchSize", description = "Batch size")
   public void migrateHardwareMeasurements(String snapshotId, int batchSize) {
+    if (batchSize == 0) {
+      log.info("Batch size cannot be zero");
```
awood, line 143: You need to throw an exception otherwise it will be non-obvious to the user that nothing happened.  Also might as well make the check `if (batchSize < 1)` since there are crazy people with their negative numbers out there.

awood (changes requested):

src/main/java/org/candlepin/subscriptions/jmx/TallyJmxBean.java
awood: IllegalArgumentException should be fine

ksynvrit: Will do

awood (approved):

PR #1634, opened 2022-11-15
Change templates apiVersion to template.openshift.io/v1

kahowell:
Otherwise we face issues with newer `oc` clients.


kahowell:
Self-merging since it's a trivial change, and CI passed.

PR #1661, opened 2022-11-28
SWATCH-582: Add New TagMetric Field For OSD

ksynvrit:
This change is for phase 1 for https://issues.redhat.com/browse/SWATCH-582

Adding new tag metric for a field for BIllingFactor, which is going to be use to calculate the units that are a multiple of what we collect.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/rhmarketplace/RhMarketplacePayloadMapper.java
+++ b/src/main/java/org/candlepin/subscriptions/rhmarketplace/RhMarketplacePayloadMapper.java
@@ -173,8 +173,9 @@ public class RhMarketplacePayloadMapper {
     // We need a longer term process to get that information onto the SKU/product
     // definition
     // itself so that we are not hard coding this type of value in our code. This will do
-    // for now.
+    // for now. Wll be removed eventually in Phase 3 of SWATCH-582 subtasks
     if (OPENSHIFT_DEDICATED_4_CPU_HOUR.equalsIgnoreCase(rhmMarketplaceMetricId)
+        && (billableUsage.getBillingFactor() == 1.0 || billableUsage.getBillingFactor() == null)
         && !Objects.isNull(value)
         && Uom.CORES.equals(uom)) {
       value = value / 4;
--- a/src/main/resources/liquibase/202212011334-add-billing-factor-column-to-remittance-table.xml
+++ b/src/main/resources/liquibase/202212011334-add-billing-factor-column-to-remittance-table.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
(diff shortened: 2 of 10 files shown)


barnabycourt:
I would suggest a test to verify the results if the billing factor does not evenly divide the unit. For example, if the billing factor is dividing CPU hours by 4. How much should be billed if 5 CPU hours of usage is found?

barnabycourt:
Also, using a non-integer multiplication as opposed to an integer based divisor seems like it could result in much larger rounding differences than we would have by doing a division but I'd be interested in the take from whoever is doing the formal review.

kahowell (changes requested):
@ksynvrit I've been mulling this over and I've come to the conclusion that we should move the billingFactor handling to BillableUsageController.calculateBillableUsage. Otherwise we get confusing and inconsistent behavior with our billable usage tracking.

Because this has interactions with already billed usage, we need to have a migration plan as well. I'm thinking we split this work into several phases/PRs/releases:

### 1. Implement billingFactor option

- [x] Via liquibase, add a billing_factor column to billable_usage_remittance, and set it to 1.0 for all existing rows, this will be used to record the billing factor applied to this particular remittance record (e.g. given a remitted value of 4.0 and a billingFactor of 0.25, the actual usage is implied to be greater than or equal to 16.0 and less than 20.0) (>= 4.0 * 4, < 5.0 * 4).
- [x] Update the billable usage message to include the applied billing factor.
- [ ] In order to support updating billing factors, when fetching recorded billable usage, derive the implied raw value by multiplying the inverse of the recorded billing factor (e.g. 1.0 / 0.25) by the remitted value. Then convert this to the configured billingFactor by multiplying by the currently configured billingFactor. The usage record should be updated with the configured billingFactor. This is where the added complexity is and definitely needs extensive testing.
- [ ] Adjust the RHM listener to continue to apply the divide by 4 only when the incoming recorded billing factor is 1.0 (null or missing should be interpreted as 1.0). (By doing this, the listener will continue to divide by 4 until we actually change the tag_profile.yaml value, and without coordinating a complex release).
- [ ] Undo any other (now redundant) changes to RHM listener and swatch-producer-aws.

### 2. Migrate OSD billingFactor to 0.25.

- Simply change the billingFactor for OSD cores to 0.25 in tag_profile.yaml - in a separate release (to be done only after the prior work lands).

### 3. Remove hardcoded division from RHM listener.

-  Remove hardcoded division from RHM listener - in a separate release (to be done only after the prior work lands).

I am thinking we use this PR to accomplish the first phase (it would be a no-op/regression test only), and then create follow-up Jira issues for (2) and (3).

kahowell:
@ksynvrit FYI, this needs a rebase.

ksynvrit:
> @ksynvrit FYI, this needs a rebase.

kahowell (approved):

PR #1667, opened 2022-11-30
Update constraints on host table

mstead:
https://issues.redhat.com/browse/SWATCH-715

* org_id: Add not-null constraint
* account_number: Remove not-null constraint

== Testing
1. Launch the application to trigger the liquibase update.
2. Ensure that `hosts.account_number` DOES NOT have the `not null` constraint.
3. Ensure that the `hosts.org_id`  HAS the `not null` constraint.


barnabycourt:
@mstead This is updating 2 million records in prod, will that cause a time constraint problem? Also, there are currently 914 hosts with a null org. Will that cause a problem?

kahowell (approved):

PR #1673, opened 2022-12-05
Add apiPath to swatch-api-nginx-proxy

kahowell:
In order to support frontend testing in EE


kahowell:
Merging, since it's a one-liner and passed ci checks.

PR #1684, opened 2022-12-08
Adjust RHACS product name

kahowell:
Actual SKU product name ended up different than initially expected.


kahowell:
Self-merging since it's a trivial config change

PR #1685, opened 2022-12-08
[hotfix] Adjust RHACS product name

kahowell:
Actual SKU product name ended up different than initially expected.


kahowell:
self-merging since it's a trivial config change

PR #1177, opened 2022-06-24
ENT-5098 Support remittance syncing and update metrics to HOURLY billing window

mstead:
https://issues.redhat.com/browse/ENT-5098

**Add endpoint to sync remittance**
The endpoint will fetch all of the latest snapshots for each
metric for an account, filter them based on the configured
billing window and billing validity, and run them through the
remittance process.

If remittance already exists for a metric, a new remittance
value will not be calculated as the metric has already been
billed.

When remittance is synced, the new remittance value is calculated
and no billable usage will be sent as we consider this usage already
billed for.

**Configured metrics to use integer billing (monthly billing window)**

**NOTE:** rhosak - Storage-gibibytes remains as HOURLY billing window due to existing
subscription constraints.

## How To Test

1. Clean out the applicable data so we can start fresh
```
psql -U rhsm-subscriptions rhsm-subscriptions -c "delete from events; delete from tally_snapshots; delete from instance_monthly_totals; delete from instance_measurements; delete from billable_usage_remittance;"
```
2. Load events into the DB. Make sure that it includes Events for the metrics that we are switching over (Reach out for sql script to load events if you need some. I pulled the data from stage and imported it.)

```
psql -U rhsm-subscriptions rhsm-subscriptions < ~/event_data.sql
```

3. Check out the develop branch and deploy the application
```
SPRING_PROFILES_ACTIVE="worker,api,kafka-queue" DEV_MODE=true ./gradlew :bootRun
```
4. Opt in all accounts from the events.
```
for var in 6 6043640 6266656 6305667;  do curl 'http://localhost:9000/hawtio/jolokia/' -H 'Content-Type: application/json' -d "{\"type\":\"exec\",\"mbean\":\"org.candlepin.subscriptions.jmx:name=optInJmxBean,type=OptInJmxBean\",\"operation\":\"createOrUpdateOptInConfig(java.lang.String,java.lang.String,boolean,boolean,boolean)\",\"arguments\":[\"$var\", \"org"$var"\", true, true, true]}"; done;
```

5. Do an hourly tally for all accounts.
```
curl 'http://localhost:9000/hawtio/jolokia/' -H 'Content-Type: application/json' -d '{"type":"exec","mbean":"org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean","operation":"tallyAllAccountsByHourly(java.lang.String,java.lang.String)","arguments":["2022-06-01T00:00Z", "2022-07-01T00:00Z"]}'
```
[...]

PR #1694, opened 2022-12-13
SWATCH-729: Migrate osd billling factor

ksynvrit:
This is to address  [SWATCH-729 ](https://issues.redhat.com/browse/SWATCH-729) 

This is to allow usage to calculate tag-metric for osd specifically as billing factor will be used to remove the hard coded division logic in the RHM Producer. Which is currently dividing the cores uom values for osd by 4. Billing Factor is configured for osd to be 0.25 which should give the same output and make the hard coded logic in the RHM Marketplace Mapper redundant. 

This is a followup to https://github.com/RedHatInsights/rhsm-subscriptions/pull/1661

# Testing Steps
Run workflow with 'main' and than using the data saved in the DB rerun tally with this branch. The remittance table should have billing factor updated for `osd` and the calculation should be the same for cores value.
```
select * from billable_usage_remittance where product_id = 'OpenShift-dedicated-metrics';
```
Run the service:

```shell
USER_USE_STUB=true SUBSCRIPTION_USE_STUB=true CG_USE_STUB=true CLOUDIGRADE_ENABLED=false DEV_MODE=true ./gradlew clean :bootRun
```

Load events into the DB:

```shell
bin/import-events.py --file bin/events.csv
```

Opt in all accounts from the events:

```shell
http :9000/hawtio/jolokia \
  type=exec \
  mbean=org.candlepin.subscriptions.jmx:name=optInJmxBean,type=OptInJmxBean \
  operation='createOrUpdateOptInConfig(java.lang.String,java.lang.String,boolean,boolean,boolean)' \
  arguments:='["account123","org123",true,true,true]'
```

Do an hourly tally for all accounts:

```shell
http :9000/hawtio/jolokia \
  type=exec \
  mbean=org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean \
  operation='tallyAllAccountsByHourly(java.lang.String,java.lang.String)' \
  arguments:='["2021-10-18T00:00Z","2021-10-19T00:00Z"]'
```
Make sure that the remittance table values gets updated with the new billing factor value, if you run into issues I used https://github.com/RedHatInsights/rhsm-subscriptions/pull/1177 as a reference to check the remittance table.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageController.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageController.java
@@ -34,8 +34,11 @@ import org.candlepin.subscriptions.db.model.ServiceLevel;
 import org.candlepin.subscriptions.db.model.TallyMeasurementKey;
 import org.candlepin.subscriptions.db.model.Usage;
 import org.candlepin.subscriptions.json.BillableUsage;
+import org.candlepin.subscriptions.json.Measurement;
 import org.candlepin.subscriptions.json.Measurement.Uom;
 import org.candlepin.subscriptions.registry.BillingWindow;
+import org.candlepin.subscriptions.registry.TagMetric;
+import org.candlepin.subscriptions.registry.TagProfile;
 import org.candlepin.subscriptions.util.ApplicationClock;
 import org.springframework.stereotype.Component;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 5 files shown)


src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageController.java
```diff
@@ -90,16 +96,36 @@ public BillableUsageRemittanceEntity getLatestRemittance(BillableUsage billableU
...
+    double tagFactor =
+        tagMetricOptional
+            .map(TagMetric::getBillingFactor)
+            .orElse(1.0); // get configured billingFactor in tag_profile yaml
+    double billableValue;
+    double remittedValue;
+    var currentRemittedValue = remittance.getRemittedValue();
+
+    if (tagFactor != 1.0) {
+      var prevBilled = currentRemittedValue / remittance.getBillingFactor(); // previously billed
```
awood, line 112: What is the `remittance.getBillingFactor()` going to be here?  Will it be the same factor as what's in the tag profile or the reciprocal?

I.e.: is this line doing something inadvertent like `currentRemittedValue / 0.25` which would actually multiply it by four.

awood (changes requested):

src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageController.java
awood: Okay, thanks.

awood (approved):

PR #1730, opened 2023-01-03
SWATCH-358 On OperationalProduct message listener call syncoffering

kartikshahc:
https://issues.redhat.com/browse/SWATCH-358
1. Run:
```USER_USE_STUB=true SUBSCRIPTION_USE_STUB=true CG_USE_STUB=true CLOUDIGRADE_ENABLED=false DEV_MODE=true ./gradlew clean :bootRun```

2. Execute the JMX endpoint in offeringJmxBean> syncProductFromXml(String) with the content present in one of the file: 

>src/test/resources/mocked-product-message.xml
or
src/test/resources/mocked-svc-product-message.xml

3. Result in log:
```2023-01-04 21:00:56,921 [thread=org.springframework.jms.JmsListenerEndpointContainer#0-1] [INFO ] [org.candlepin.subscriptions.product.OfferingWorker] - Received UMB message for productSku=RH0180191```

5. Note: The rest of the controller syncUmbProduct code will be done as part of SWATCH-395

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/product/OfferingWorker.java
+++ b/src/main/java/org/candlepin/subscriptions/product/OfferingWorker.java
@@ -20,13 +20,19 @@
  */
 package org.candlepin.subscriptions.product;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.dataformat.xml.XmlMapper;
 import lombok.extern.slf4j.Slf4j;
 import org.candlepin.subscriptions.task.TaskQueueProperties;
+import org.candlepin.subscriptions.umb.CanonicalMessage;
+import org.candlepin.subscriptions.umb.UmbOperationalProduct;
+import org.candlepin.subscriptions.umb.UmbProperties;
 import org.candlepin.subscriptions.util.KafkaConsumerRegistry;
 import org.candlepin.subscriptions.util.SeekableKafkaConsumer;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Profile;
+import org.springframework.jms.annotation.JmsListener;
 import org.springframework.kafka.annotation.KafkaListener;
 import org.springframework.stereotype.Service;
 
@@ -36,15 +42,20 @@ import org.springframework.stereotype.Service;
 public class OfferingWorker extends SeekableKafkaConsumer {
 
   private final OfferingSyncController controller;
+  private final UmbProperties umbProperties;
+  private final XmlMapper xmlMapper;
 
   @Autowired
   protected OfferingWorker(
       @Qualifier("offeringSyncTasks") TaskQueueProperties taskQueueProperties,
       KafkaConsumerRegistry kafkaConsumerRegistry,
-      OfferingSyncController controller) {
+      OfferingSyncController controller,
+      UmbProperties umbProperties) {
     super(taskQueueProperties, kafkaConsumerRegistry);
 
     this.controller = controller;
+    this.umbProperties = umbProperties;
+    this.xmlMapper = CanonicalMessage.createMapper();
   }
 
   @KafkaListener(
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/main/java/org/candlepin/subscriptions/product/OfferingWorker.java [hunk 1]
```diff
@@ -57,4 +68,18 @@ public void receive(OfferingSyncTask task) {
 
     controller.syncOffering(sku);
   }
+
+  @JmsListener(destination = "#{@umbProperties.productTopic}")
+  public void receive(String productMessageXml) throws JsonProcessingException {
+    log.debug("Received message from UMB offering product{}", productMessageXml);
+    if (!umbProperties.isProcessingEnabled()) {
+      return;
```
mstead, line 76: Feels like there should be some logging that happens here. Would also be nice to have a test case around this logic.

src/main/java/org/candlepin/subscriptions/subscription/ActiveMQUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Service;
+import org.springframework.util.StringUtils;
+
+@Service
+@Slf4j
+@Profile("capacity-ingress")
+public class ActiveMQUtils {
```
mstead, line 36: This feels more like an ActiveMQ service configuration class than a 'utils' class and could use a rename IMO.

src/main/java/org/candlepin/subscriptions/subscription/ActiveMQUtils.java
```diff
@@ -0,0 +1,73 @@
...
+    ActiveMQSslConnectionFactory factory = new ActiveMQSslConnectionFactory();
+    factory.setExceptionListener(e -> log.error("Exception thrown in ActiveMQ connection", e));
+    if (StringUtils.hasText(activeMQProperties.getBrokerUrl())) {
+      factory.setBrokerURL(activeMQProperties.getBrokerUrl());
+      if (!umbProperties.providesTruststore() || !umbProperties.usesClientAuth()) {
+        log.warn("UMB config requires keystore and truststore - not provided or not valid.");
+      }
+    } else {
+      // default to an embedded broker
+      factory.setBrokerURL("vm://localhost?broker.persistent=false");
```
mstead, line 61: We should log that we are defaulting to an embedded broker so that it is clear when we are running the app.

mstead (changes requested):

src/main/java/org/candlepin/subscriptions/subscription/ActiveMQUtils.java
lindseyburnett: agreed

src/main/java/org/candlepin/subscriptions/product/OfferingWorker.java
kartikshahc: Added test for the calling method and changed this method to private

src/main/java/org/candlepin/subscriptions/subscription/ActiveMQUtils.java
kartikshahc: Done

src/main/java/org/candlepin/subscriptions/umb/UmbProperties.java
```diff
@@ -56,11 +58,15 @@ public String getConsumerTopic(String topic) {
     if (!topic.startsWith("VirtualTopic")) {
       return topic;
     }
-    String subscriptionId = String.format("swatch-%s-%s", namespace, topic.replace(".", "_"));
-    return String.format("Consumer.%s.%s.%s", serviceAccountName, subscriptionId, topic);
+    String id = String.format("swatch-%s-%s", namespace, topic.replace(".", "_"));
+    return String.format("Consumer.%s.%s.%s", serviceAccountName, id, topic);
```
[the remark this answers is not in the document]
kartikshahc: It was suppose to be subscriptionId since this is the Active MQ subscription and not the RHSM subscription. Added unit test as well.

src/main/java/org/candlepin/subscriptions/product/OfferingWorker.java
mstead: Please add a debug message for when UMB processing is not enabled. Could be helpful in cases where messages are being skipped because of this setting.

mstead (changes requested):
One small change request. Other than that, the changes look good!

src/main/java/org/candlepin/subscriptions/product/OfferingWorker.java
kartikshahc: Done

mstead (approved):

PR #1746, opened 2023-01-09
Add ENABLE_PAYG_SUBSCRIPTION_FORCE_SYNC to RHM deployment, default true

kahowell:


kahowell:
Self-merging since this is a trivial & urgent template param & env var change

PR #1758, opened 2023-01-11
Change RHACS awsDimension to vCPU_Hour

kahowell:


kahowell:
Self-merging since it's a trivial config change w/ passing tests.

PR #1762, opened 2023-01-13
SWATCH-782: Add a configurable threshold for HBI tallies

kahowell:
https://issues.redhat.com/browse/SWATCH-782

`TALLY_MAX_HBI_ACCOUNT_SIZE` when overridden limits the size of accounts that will be processed. Any account going over this limit will be logged with a warning, and skipped.

Note that I intentionally set the default to `Integer.MAX_VALUE`, as this makes it easy to "disable" the check by just not overriding it, and I plan to override with a reasonable value for prod specifically via template parameter.

Testing
-------

Set up some fake HBI data:

```shell
bin/insert-mock-hosts --num-physical=2 --org=org123 --hbi
```

Run the service with a low threshold:

```shell
TALLY_MAX_HBI_ACCOUNT_SIZE=1 ./gradlew :bootRun
```

Trigger a tally:

```shell
http :9000/hawtio/jolokia \
  type=exec \
  arguments:='["org123"]'
```

Observe logging that shows the org was skipped:

```
2023-01-13 15:16:19,185 [thread=rhsm-subscriptions-task-processor-0-C-1] [WARN ] [org.candlepin.subscriptions.tally.TallySnapshotController] - Skipping tally for orgId=org123 because it has activeHbiSystems=2 (greater than threshold=1)
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
+++ b/src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
@@ -128,4 +128,7 @@ public class ApplicationProperties {
 
   /** If enabled, will allow synchronous operations when requested. */
   private boolean enableSynchronousOperations = false;
+
+  /** Sets a hard limit on the size of accounts that HBI-based tally will attempt to process. */
+  private int tallyMaxHbiAccountSize;
 }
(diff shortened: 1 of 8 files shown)


mstead (changes requested):
When the account is over the threshold, the retry logic in the controller kicks in when the exception occurs resulting in many useless retries. We should fix this.

mstead (approved):

PR #1763, opened 2023-01-13
Add MACHINE_POOL parameter to swatch-tally

kahowell:
Default to empty string, this doesn't set tolerations at all (see https://github.com/RedHatInsights/clowder/blob/a85b276baf1b70f31d8fab1f325dcc5e64cc2617/controllers/cloud.redhat.com/providers/job/impl.go#L80)


kahowell:
Self-merging since this is a trivial change to allow changing machine pool.

PR #1776, opened 2023-01-19
SWATCH-687 Perform nightly tallies via DB

mstead:
## Detail
Perform the nightly tally for an org by querying the DB instead
of holding all hosts and their facts in memory to perform the
calculations.

Since everything we are tallying for an org has already been persisted
as 'buckets', we can perform our calculations via the query using
the bucket data. The query is using the cores and sockets on the bucket itself
because these fields are normailzed before being stored, unlike
the instance measurements.

An environment variable was added to allow enabling the old
legacy tally functionality should we identify any issues and would like
to disable the new tally logic.

This new tally logic should address the issue outlined in bug [SWATCH-49 Virtual satellites are not considered in tally](https://issues.redhat.com/browse/SWATCH-49) and will likely cause some discrepancy in IQE tests. Any virtual non-RHEL instance will now be properly tallied as VIRTUAL in both the DB based tally AND the legacy tally.

## References:
https://issues.redhat.com/browse/SWATCH-687

## How To Test
Use the steps below as a guide to testing, but please try and find ways to break this outside of what I describe. Using the insert-mock-hosts script is a good way to generate data for testing against a large number of hosts per org, BUT, it isn't the greatest at creating the various combos of data that we need to pull in from HBI. I'd suggest finding an org in stage and pulling a smallish subset of data from the inventory DB to test. Ideally the data should have some RHEL and non-RHEL hosts to tally.

**NOTES:**
* Be sure to replace `YOUR_TEST_ORG_ID`,  in the commands below, with the org_id that will have data in the inventory DB.
* Load some host records into the inventory DB for an org. This inventory data will be what we are basing our tally tests off of.

### Scenario 1: Existing Tally Data and re-tally occurs via DB Tally

Start with a fresh swatch DB, deploy the **main** branch.
```
dropdb -U rhsm-subscriptions rhsm-subscriptions && createdb -U rhsm-subscriptions rhsm-subscriptions  &&  DEV_MODE=true ./gradlew clean :bootRun
```

Run a nightly tally on the test org.
```
http :9000/hawtio/jolokia   type=exec   mbean='org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean'   operation='tallyOrg(java.lang.String)'   arguments:='["YOUR_TEST_ORG_ID"]'
```

Once complete, capture the key parts of the tally as follows.
```
[...]

PR #1777, opened 2023-01-19
SWATCH-686: Minimize nightly memory usage by streaming system updates

kahowell:
https://issues.redhat.com/browse/SWATCH-686

By ordering records from HBI and swatch in a consistent order, and then collating the results, we
can handle updates with minimal memory usage.

I integrated with changes from #1776; this update strategy is applied only when
`legacyNightlyTallyEnabled=false`.

In order to make the logic easier to follow (and hopefully easier to re-use once we start processing
HBI events), I split the changes up as follows:

1. a new class `InventorySwatchDataCollator` that handles collation of streams of system information
   from HBI and swatch.
2. methods added to InventoryAccountUsageCollector that handle reconciliation of state between HBI
   and swatch
3. a coordinator method in InventoryAccountUsageCollector that invokes `InventorySwatchDataCollator`
   and flushes at configured intervals.

I had to set auto-commit to false on the HBI connection in order to allow streaming.

Testing
=======

Set up an HBI org with 50,000 physical systems:

```shell
bin/insert-mock-hosts --org=org456 --hbi --clear --num-physical=50000
```

Add 3 hypervisors with 100,000 systems each:

```shell
for i in $(seq 3); do
  bin/insert-mock-hosts --org=org456 --hbi --num-hypervisors=1 --num-guests=100000
done
```

Next run the following testing steps (repeat for `main` and this branch for comparison)...

Reset hosts and related tables (critical between test runs):

```shell
truncate table hosts cascade;
EOF
```

Run the application:

```shell
LOGGING_LEVEL_ORG_CANDLEPIN_TALLY=DEBUG \
  ./gradlew :bootRun \
  --args='--logging.level.org.hibernate.SQL=DEBUG'
```

Connect to the application via `jconsole`:

```shell
jconsole
```

Select `org.candlepin.subscriptions.BootApplication`

This will allow you to see the memory usage as the tally progresses.

Trigger a tally for org456:

```shell
http :9000/hawtio/jolokia \
  type=exec \
  arguments:='["org456"]'
```

Trigger a retally:

```shell
http :9000/hawtio/jolokia \
  type=exec \
  arguments:='["org456"]'
```

Observe:

1. The memory usage for this branch is minimal (~300 Mb, and if you force GC, it stays even lower)
2. The logged sql queries are minimal. (On a re-tally there are no inserts, and only the initial selects)

Also make the following call to check the overall runtime once the tally completes (or use a clock):

Notice the overall runtime is often faster (feel free to adjust the shape of the HBI data and see
how it performs, though I expect we'll work with performance engineers to validate as well).

```shell
http :9000/metrics | grep spring_kafka_listener_seconds_sum | grep rhsm-subscriptions-task-processor
```

### DIFF
--- a/bin/insert-mock-hosts
+++ b/bin/insert-mock-hosts
@@ -82,7 +82,8 @@ def generate_host(inventory_id=None, insights_id=None, account_number=None, org_
         }
 
     _generate_insert('hosts', **host_fields)
-    _generate_measurements(host_id, uom, 1.0)
+    if not is_hbi:
+        _generate_measurements(host_id, uom, 1.0)
     if not is_hbi and not skip_buckets:
         _generate_buckets(host_id, product, sla, usage, as_hypervisor=is_hypervisor, cores=cores, sockets=sockets, measurement_type=measurement_type,
         billing_provider=billing_provider, billing_account_id=billing_account_id)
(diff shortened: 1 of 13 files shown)


mstead (changes requested):
The code looks good to me, performs pretty well and definately reduces memory usage. Pretty clever. I tested this in as many ways as I could think of, both from quantity and shape of data.

One suggestion would be to add a disclaimer around the queries to spell out the importance of the result sorting defined by the queries.

mstead:
> The code looks good to me, performs pretty well and definately reduces memory usage. Pretty clever. I tested this in as many ways as I could think of, both from quantity and shape of data.
> 
> One suggestion would be to add a disclaimer around the queries to spell out the importance of the result sorting defined by the queries.

Thanks for adding the comment. Merge away when ready!

mstead (approved):

PR #1794, opened 2023-01-25
SWATCH-749: Add sockets as a UOM for RHEL product ids

awood:
See https://issues.redhat.com/browse/SWATCH-749
---
# Testing

1. Before applying this patch, `gunzip` and import [swatch-749-test-data.sql.gz](https://github.com/RedHatInsights/rhsm-subscriptions/files/10501352/swatch-749-test-data.sql.gz) .  Let me know if you have problems with the import.  I believe this should work but it's my first time trying an export with a new method.
2. Run
    ```shell
    http :8000/api/rhsm-subscriptions/v1/instances/products/RHEL%20for%20IBM%20z \
    beginning==2022-01-01T00:00:00.000Z ending==2023-01-10T23:59:59.999Z dir==desc sort==last_seen uom==sockets \
    x-rh-identity:$(echo -n '{"identity":{"account_number":"account123","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w0)
    ```
3. Observe that no measurements are present on any of the returned elements.
4. Apply this patch.
5. Rerun the `http` command and observe that the measurements are present.


kflahert (approved):

mirekdlugosz:
@awood I know that was not mentioned in SWATCH-749, but should we also add sorting by sockets as part of this task?

[GUI has disabled sorting on that column](https://github.com/RedHatInsights/curiosity-frontend/pull/1038/files#diff-5fab37fd02bbf0985221e56a1b4fda23c59238b441dd18f7d6e38637a3974d1dR226) but I believe this is temporary, while sorting doesn't work. In general, users should be able to sort by that column.

I can create another jira for sorting if that's easier.

awood:
> I can create another jira for sorting if that's easier.

I'd like to have a card for this so that it can be tracked explicitly.

mirekdlugosz:
@awood Sure, opened [SWATCH-820](https://issues.redhat.com/browse/SWATCH-820)

PR #1798, opened 2023-01-26
SWATCH-787: Add product-denylist.txt entries

barnabycourt:
Add product denylist entries. 

These entries are unused at this point but it is a pre-req for future work so should only require a regression test of the swatch-subscription-sync service


kahowell (approved):
LGTM, feel free to merge once CI passes

PR #1802, opened 2023-01-27
Added sku SER0730 - Self-Support: Red Hat OpenShift Platform Plus wit…

rspagnola:
What:
Added sku SER0730 - Self-Support: Red Hat OpenShift Platform Plus with Red Hat OpenShift Data Foundation Advanced, Self-Support (2 Cores or 4vCPUs, NFR, Partner Only)

Why:
Customer needs to apply subscription against this sku

Tickets:
https://access.redhat.com/support/cases/#/case/03409895


barnabycourt:
Closing per discussion with product management as this SKU will not be counted correctly as it currently structured.

PR #1830, opened 2023-02-08
SWATCH-829 Endpoint that swatch contracts can use to map a SKU to the swatch product tags

kartikshahc:
https://issues.redhat.com/browse/SWATCH-829

- **Description**: This is a new API that will allow any service to lookup the swatch product(s) associated with a given SKU. (This lookup will use the offering information already stored in the database) and map the`product_name` to a swatch `product_tag` via info in `tag_profile.yaml`. If the offering does not exist then return 404. If it does exist, then return an empty list if there are no tags found for that particular offering.

- **Test Steps**:
```
INSERT INTO public.offering (sku, product_name, product_family, cores, sockets, hypervisor_cores, hypervisor_sockets, role, sla, usage, description, has_unlimited_usage, derived_sku) VALUES ('MW01484', 'OpenShift Dedicated', null, 0, 0, 0, 0, 'osd', 'Premium', 'Production', 'Red Hat OpenShift Dedicated on Customer Cloud Subscription (Hourly)', null, null);

INSERT INTO public.offering (sku, product_name, product_family, cores, sockets, hypervisor_cores, hypervisor_sockets, role, sla, usage, description, has_unlimited_usage, derived_sku) VALUES ('RH0180191', 'RHEL Server', 'Red Hat Enterprise Linux', null, 2, null, null, 'Red Hat Enterprise Linux Server', 'Standard', 'Production', 'Red Hat Enterprise Linux Server, Standard (1-2 sockets) (Up to 4 guests) with Smart Management', false, null);
```
  
  - Run: `USER_USE_STUB=true SUBSCRIPTION_USE_STUB=true CG_USE_STUB=true RHSM_USE_STUB=true  CLOUDIGRADE_ENABLED=false DEV_MODE=true ./gradlew clean :bootRun`
  
1.  No matching Offering Product Names and sku
   - Run: `curl -X 'GET'   http://localhost:8000/api/rhsm-subscriptions/v1/internal/offerings/123/product_tags   -H 'accept: application/json'   -H 'x-rh-swatch-psk: placeholder'`
   -  Response: {"errors":[{"status":"404","code":"SUBSCRIPTIONS3003","title":"Sku 123 not found in Offering"}]}
 2. No matching Offering Product Name
  - Run: `curl -X 'GET'   http://localhost:8000/api/rhsm-subscriptions/v1/internal/offerings/RH0180191/product_tags   -H 'accept: application/json'   -H 'x-rh-swatch-psk: placeholder'`
  -  Response: {}
 3. Matching sku and Offering Product Name
  - Run: `curl -X 'GET'   http://localhost:8000/api/rhsm-subscriptions/v1/internal/offerings/MW01484/product_tags   -H 'accept: application/json'   -H 'x-rh-swatch-psk: placeholder'`
  - Response: {"data":["OpenShift-dedicated-metrics"]}

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
@@ -37,6 +37,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.persistence.EntityNotFoundException;
 import javax.transaction.Transactional;
+import javax.ws.rs.core.Response;
 import lombok.extern.slf4j.Slf4j;
 import org.candlepin.subscriptions.capacity.CapacityReconciliationController;
 import org.candlepin.subscriptions.capacity.files.ProductAllowlist;
@@ -48,6 +49,8 @@ import org.candlepin.subscriptions.db.model.ReportCriteria;
 import org.candlepin.subscriptions.db.model.ServiceLevel;
 import org.candlepin.subscriptions.db.model.Subscription_;
 import org.candlepin.subscriptions.db.model.Usage;
+import org.candlepin.subscriptions.exception.ErrorCode;
+import org.candlepin.subscriptions.exception.MissingOfferingException;
 import org.candlepin.subscriptions.registry.TagProfile;
 import org.candlepin.subscriptions.subscription.api.model.Subscription;
 import org.candlepin.subscriptions.tally.UsageCalculation.Key;
@@ -598,7 +601,9 @@ public class SubscriptionSyncController {
   /**
    * This will allow any service to lookup the swatch product(s) associated with a given SKU. (This
    * lookup will use the offering information already stored in the database) and map the
-   * `product_name` to a swatch `product_tag` via info in `tag_profile.yaml`
+   * `product_name` to a swatch `product_tag` via info in `tag_profile.yaml` If the offering does
+   * not exist then return 404. If it does exist, then return an empty list if there are no tags
+   * found for that particular offering.
    *
    * @param sku
    * @return OfferingProductTags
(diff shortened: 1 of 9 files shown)


src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
```diff
@@ -592,4 +594,22 @@ public List<org.candlepin.subscriptions.db.model.Subscription> findSubscriptions
...
+   * lookup will use the offering information already stored in the database) and map the
+   * `product_name` to a swatch `product_tag` via info in `tag_profile.yaml`
+   *
+   * @param sku
+   * @return OfferingProductTags
+   */
+  public OfferingProductTags findProductTags(String sku) {
+    OfferingProductTags productTags = new OfferingProductTags();
+    var productTag = offeringRepository.findProductNameBySku(sku);
+    if (productTag.isPresent()
```
kflahert, line 609: @kartikshahc Thinking this should return a 404 if the SKU isn't present in offering table.

kflahert (changes requested):

src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
kartikshahc: Not sure either. 4xx may add to splunk and this could be a valid case. Do you want me to log a warning instead of returning 404?

kflahert: I think 404 is valid in this case and on the client side we can log a warning or error if needed.  I don't think this should be an issue in splunk but paging @kahowell if that is a concern for this.

kartikshahc: Done

kflahert (approved):

PR #1854, opened 2023-02-15
SWATCH-869 Remove cast on instance_id in tally_instance_view

mstead:
https://issues.redhat.com/browse/SWATCH-869

Our instance view was casting host.instance_id to a UUID, but,
host.instance_id on the hosts table isn't always a proper UUID(
i.e OpenShift Clusters).

When querying against the view, if the instance_id stored in the
host table is not a proper UUID, it would cause an error at the DB level.

This patch removes the cast on instance_id and updates the field to a
String in the view class.

**NOTE** This PR also fixes a couple issues in the insert-mock-hosts script. The fixes are in separate commits.

## Testing
**Check out this PR's branch so that the fixes are applied to the insert-mock-hosts script** and insert a couple mock hosts into the tally DB.
```
bin/insert-mock-hosts --org=123456 --num-physical=2
```

**CHECK OUT THE MAIN BRANCH**

Determine the host IDs so that we can update the instance_id of one of them to not be a UUID (script generates a UUID as the instance_id)
```
$ psql -U rhsm-subscriptions rhsm-subscriptions -c "select id, instance_id from hosts where org_id='123456'"
                  id                  |             instance_id              
--------------------------------------+--------------------------------------
 d0a96421-f5f6-436c-b7e9-d8eaa514afa4 | 7239bc70-1300-4903-844f-369fd833933b
 831f224c-ce01-426f-9698-bc8becfcc1a0 | 2b6cb6c5-f5e5-4244-a6e9-de0f0764c75f
(2 rows)
```

Update the instance_id of one record with something that is not a UUID. This example uses a cluster id from OpenShift metrics.
```
$ psql -U rhsm-subscriptions rhsm-subscriptions -c "update hosts set instance_id='c5ff3f4ctr234ihcgr8g' where id='d0a96421-f5f6-436c-b7e9-d8eaa514afa4'"
```
A simple query will reproduce the bug.
```
$ psql -U rhsm-subscriptions rhsm-subscriptions -c "select * from tally_instance_view where org_id='123456'"
ERROR:  invalid input syntax for type uuid: "c5ff3f4ctr234ihcgr8g"
```

Deploy the application
```
DEV_MODE=true ./gradlew clean :bootRun
```

Reproduce via the Instances API
```
$ curl -X 'GET'   'http://localhost:8000/api/rhsm-subscriptions/v1/instances/products/RHEL?uom=cores&sort=category&dir=asc' -H 'accept: application/json'   -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMTIzNDU2In19fQ=='
{"errors":[{"status":"500","code":"SUBSCRIPTIONS1000","title":"An internal server error has occurred. Check the server logs for further details.","detail":"could not extract ResultSet; SQL [n/a]; nested exception is org.hibernate.exception.DataException: could not extract ResultSet"}]}
```

Examining the stack trace you should see:
```
Caused by: org.postgresql.util.PSQLException: ERROR: invalid input syntax for type uuid: "c5ff3f4ctr234ihcgr8g"
```

Check out this PR's branch and deploy the app.
```
DEV_MODE=true ./gradlew clean :bootRun
```

Run the query again and results should be returned.
```
$ psql -U rhsm-subscriptions rhsm-subscriptions -c "select * from tally_instance_view where org_id='123456'"
```

Test the API again as well and a proper response should be returned.
```
curl -X 'GET'   'http://localhost:8000/api/rhsm-subscriptions/v1/instances/products/RHEL?uom=cores&sort=category&dir=asc' -H 'accept: application/json'   -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImFjY291bnQxMjMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsiaXNfb3JnX2FkbWluIjp0cnVlfSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMTIzNDU2In19fQ=='
```

### DIFF
--- a/src/main/resources/liquibase/202302151544-remove-uuid-cast-on-instance-id-in-tally-instance-view.xml
+++ b/src/main/resources/liquibase/202302151544-remove-uuid-cast-on-instance-id-in-tally-instance-view.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+  <changeSet id="202302151544-1" author="mstead">
+    <dropView viewName="tally_instance_view"/>
+  </changeSet>
+
+  <changeSet id="202302151544-2" author="mstead">
+    <comment>Create view of hosts and host_tally_buckets for use in Instance API</comment>
+    <createView
+      replaceIfExists="true"
+      viewName="tally_instance_view">
+      select
+          distinct org_id,
+          h.id,
+          h.instance_id as instance_id,
+          h.display_name,
+          h.billing_provider as host_billing_provider,
+          h.billing_account_id as host_billing_account_id,
+          b.billing_provider as bucket_billing_provider,
+          b.billing_account_id as bucket_billing_account_id,
+          h.last_seen,
+          coalesce(h.num_of_guests, 0) AS num_of_guests,
+          b.product_id,
+          b.sla,
+          b.usage,
+          coalesce(b.measurement_type, 'EMPTY') AS measurement_type,
+          m.uom,
+          m.value,
+          b.sockets,
+          b.cores,
+          h.subscription_manager_id
+      from hosts h
+          join host_tally_buckets b on h.id = b.host_id
+          left outer join instance_measurements m on h.id = m.instance_id;
+    </createView>
+  </changeSet>
+
+</databaseChangeLog>
+<!-- vim: set expandtab sts=4 sw=4 ai: -->
--- a/bin/insert-mock-hosts
+++ b/bin/insert-mock-hosts
@@ -82,7 +82,8 @@ def generate_host(inventory_id=None, insights_id=None, account_number=None, org_
         }
 
     _generate_insert('hosts', **host_fields)
-    _generate_measurements(host_id, uom, 1.0)
+    if not is_hbi:
+        _generate_measurements(host_id, uom, 1.0)
     if not is_hbi and not skip_buckets:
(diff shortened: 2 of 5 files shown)


src/main/resources/liquibase/202302151544-remove-uuid-cast-on-instance-id-in-tally-instance-view.xml
```diff
@@ -0,0 +1,43 @@
...
+                      sla,
+        usage,
+        COALESCE(b.measurement_type, 'EMPTY') AS measurement_type,
+        m.uom,
+        m.value,
+        b.sockets,
+        b.cores,
+        h.subscription_manager_id
+      from hosts h
+        join host_tally_buckets b on h.id = b.host_id left outer join instance_measurements m on h.id = m.instance_id;
```
awood, line 38: Can you tidy this up?  It's kind of hard to read like this.  The indentation is inconsistent, the case is inconsistent (`coalesce` vs `coalesce`), some columns have the table prefix `h.` while others don't, and some columns are aliased when they don't need to be (`h.instance id as instance_id`).  I think you could just drop this text into DataGrip and run the formatter and it would fix a lot of the problems.

awood (changes requested):

awood:
Awesome test steps.  Everything passed.  I'd just like the view SQL to get tidied up.

mstead:
@awood Thanks for the review. Sorry 'bout the force push.

mstead:
Actually, hold up. I didn't get all the query changes you requested.

src/main/resources/liquibase/202302151544-remove-uuid-cast-on-instance-id-in-tally-instance-view.xml
mstead: DONE

awood (approved):

PR #1870, opened 2023-02-21
SWATCH-853: Tally - increase timing slack with conduit

kahowell:
https://issues.redhat.com/browse/SWATCH-853

Assumptions about conduit updates being processed before tally were insufficient. Accordingly, we're adding some timing slack.

Per PM input:
* Updated stale threshold to 30 hours (instead of 24 hours).
* Updated the schedule for nightly tallies to 6 am UTC, so HBI has longer to process updates before we tally.

With this change, systems that haven't checked in can take up to 30 hours before being filtered from tallies.

I also changed the stale threshold to use midnight UTC as its reference point, so that the stale threshold doesn't get offset by the tally schedule. A 30 hour threshold should mean that a system checked in more than 30 hours prior to the last sync, rather than 30 hours from the time of the tally.

Testing
=======

This is essentially just configuration changes.

Observe the logging of the updated default of 30 hours when deployed to ephemeral environment (`/retest` and look at the tally startup logs, or alternatively manually deploy to ephemeral and then view the logs).

Also observe the updated cronjob schedule for `swatch-tally-tally` in the environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ RHSM_RBAC_USE_STUB=true ./gradlew bootRun
 * `PATH_PREFIX`: path prefix in the URLs (default: api)
 * `INVENTORY_USE_STUB`: Use stubbed inventory REST API
 * `INVENTORY_API_KEY`: API key for inventory service
-* `INVENTORY_HOST_LAST_SYNC_THRESHOLD`: reject hosts that haven't checked in since this duration (e.g. 24h)
+* `HOST_LAST_SYNC_THRESHOLD`: reject hosts that haven't checked in since this duration (e.g. 24h)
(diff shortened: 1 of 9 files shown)


kahowell:
a1cf08b to 46c9542: rebase against `main`

kahowell:
a1cf08b to 46c9542: fix botched rebase (accidentally removed some template params)

kahowell:
cc8345be to b589f968: see `git diff cc8345be..b589f968`, essentially, this preserves the defaulting inside of ApplicationProperties, since unit tests depend on it.

kflahert (approved):

kflahert (approved):

PR #1898, opened 2023-03-01
SWATCH-965: Add MW02291 to allowlist

kahowell:
RHACS is using an additional SKU. Will also create a hotfix version so we can get this update to prod faster.


kahowell:
Planning to self-merge once tests pass since this is a one-line allowlist change previously discussed.

kflahert (approved):

PR #1899, opened 2023-03-01
SWATCH-965: [HOTFIX] Add MW02291 to allowlist

kahowell:
RHACS is using an additional SKU.


kahowell:
Will merge once #1898 lands

kflahert (approved):

PR #1902, opened 2023-03-02
SWATCH-1004: Update swatch-contracts package structure

lindseyburnett:


kahowell (changes requested):
This needs some further package changes:
* [x] build.gradle needs `apiPackage`, `modelPackage`, `invokerPackage` updates to `com.redhat.swatch.contract.openapi` prefix.

It also needs some visibility modifier fixes:
* [x] ContractMapper should be public
* [ ] ContractService.createContract should be public
* [ ] ContractService.updateContract should be public
* [ ] ContractService.deleteContract should be public
* [ ] BaseUnitTest should be public

One more:
* [ ] ContractMapperTest has a stale import for ContractMapper. This needs to be removed.

There might be a few more, but these were what I found.

kahowell (approved):

PR #1909, opened 2023-03-03
Remove db from swatch-db-changelog-cleanup-db clowdapp

kahowell:


kahowell:
Self-merging since it passed CI and is a mostly trivial change.

PR #1923, opened 2023-03-07
SWATCH-953: Use MapStruct @AfterMapping to populate contractUuid

lindseyburnett:


swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractMapper.java
```diff
@@ -32,4 +40,30 @@ public interface ContractMapper {
...
+      final Contract contractDto) {
+
+    if (Objects.requireNonNullElse(contractDto.getMetrics(), new ArrayList<>()).isEmpty()) {
+      contractEntity.metrics(new HashSet<>());
+    } else {
+      var metrics =
+          contractDto.getMetrics().stream()
+              .map(
+                  (x -> {
+                    var builder = ContractMetricEntity.builder();
```
kartikshahc, line 56: We should revisit this logic later and make it simpler. We should already be getting the ContractMetric detail from dtoToContractEntity except UUID. For now this should be fine.

kartikshahc (approved):

PR #1952, opened 2023-03-14
SWATCH-1036 Adding SKU MW01624 & MW01624F3

rspagnola:
Adding SKU MW01624 - Standard: Red Hat OpenShift Platform Plus (Bare Metal Node), Standard (1-2 sockets)


barnabycourt:
@rspagnola should this include MW01624F3 as well?

barnabycourt (approved):

PR #1968, opened 2023-03-17
Read all DB config values from clowder config

kahowell:


kahowell:
Note this is scoped to just the tally clowdapp. We can follow-up and do similar in other clowdapps as well later on.

PR #1971, opened 2023-03-17
SWATCH-1039: Fix issues with hypervisor & non-hypervisor keys

kahowell:
https://issues.redhat.com/browse/SWATCH-1039

Note that I'm not 100% sure these fixes will address 1039, but there is a class of bugs these should address, and it is likely that SWATCH-1039 is one of these.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
@@ -512,6 +512,7 @@ public class InventoryAccountUsageCollector {
 
   private void reconcileHypervisorData(
       NormalizedFacts normalizedFacts, Host system, OrgHostsData orgHostsData, Set<Key> usageKeys) {
+    Set<HostBucketKey> seenBucketKeys = new HashSet<>();
     if (system.getHypervisorUuid() != null
         && orgHostsData.hasHypervisorUuid(system.getHypervisorUuid())) {
       // system is a guest w/ known hypervisor, we should add its buckets to hypervisor-guest data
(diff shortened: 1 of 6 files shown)


kahowell:
I changed the title and description, as I changed this PR to a subtask, and intend to follow-up with another subtask to validate whether this fixes the root issue in prod or not.

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -541,12 +541,12 @@ private void reconcileHypervisorData(
                   system.addBucket(bucket);
                   seenBucketKeys.add(bucket.getKey());
                 });
-        // remove any buckets for guests no longer present
-        system
-            .getBuckets()
-            .removeIf(b -> b.getKey().getAsHypervisor() && !seenBucketKeys.contains(b.getKey()));
       }
     }
+    // remove any buckets for guests no longer present
```
awood, line 546: Would you mind just adding a note here about what `asHypervisor` is meant to represent here?  I've been looking at this code for a few days and my understanding is still murky.  In 2 weeks time, I'll never remember :smile:

awood (changes requested):

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
kahowell: Done, feel free to reopen if you'd like me to clarify further.

kartikshahc:
@kahowell After our final test together yesterday the changes seem good to me.

mstead (approved):
I've tested this last night and again this morning and these changes address the issues.

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
awood: Perfect.  Now we just need to change the column name so that it makes sense. :laughing:

PR #1975, opened 2023-03-21
Enable swatch-contracts test APIs in stage

lindseyburnett:
Just to unblock QE testing in stage.  There's more involved work for SWATCH-882


swatch-contracts/src/main/resources/application.properties
```diff
@@ -10,7 +10,7 @@ DATABASE_PASSWORD: ${clowder.database.password:rhsm-subscriptions}
 
 SWATCH_TEST_APIS_ENABLED=true
 # following line to be removed in SWATCH-882
```
kahowell, line 12:
```suggestion
```

kahowell: stale comment

kahowell (changes requested):

kahowell (approved):

PR #2010, opened 2023-03-29
SWATCH-1011: Implement rate limiting for RHSM API interactions

kahowell:
https://issues.redhat.com/browse/SWATCH-1011

This introduces the following template params/env vars:

* `RHSM_API_LIMIT_FOR_PERIOD` - how many RHSM API operations allowed in the rate limit window.
* `RHSM_API_LIMIT_REFRESH_PERIOD` - how often the rate limit should be refreshed.
* `RHSM_API_LIMIT_TIMEOUT_DURATION` - the max time a pending RHSM API interaction should wait before timing out.

These are set by default to values that enforce an effective limit of 150 requests/second:

* limit: 150
* refresh period: 100ms
* timeout duration: 30m

Timeout chosen as 30m, as this corresponds to our kafka max poll interval (if it takes longer than that the kafka message is going to be rejected anyways)

Testing
=======

Run the application with some overrides to test the rate limiting:

```shell
DEV_MODE=true \
  RHSM_USE_STUB=true \
  RHSM_MAX_ATTEMPTS=1 \
  RHSM_API_LIMIT_REFRESH_PERIOD=60s \
  RHSM_API_LIMIT_FOR_PERIOD=1 \
  RHSM_API_LIMIT_TIMEOUT_DURATION=10s \
  ./gradlew :swatch-system-conduit:bootRun
```

Check the metrics for the rate limiter:

```shell
http :9000/metrics/ | grep resilience4j | grep rhsmApi
```

Observe there is a single "permission available" - indicating the rate limit is not yet exhausted.

Next, invoke conduit operations:

```shell
http :8000/api/rhsm-subscriptions/v1/internal/rpc/syncOrg \
  org_id=org123 \
  x-rh-swatch-psk:placeholder
```

Next, consult the metrics and see that there are no remaining permissions available.

Make the above request again, and see that it times out after 10s and responds with HTTP 429.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -51,6 +51,7 @@ libraries["quarkus-qpid-jms-bom"] = "io.quarkus.platform:quarkus-qpid-jms-bom:2.
 libraries["quarkus-bom"] = "io.quarkus.platform:quarkus-bom:2.16.5.Final"
 libraries["quarkus-logging-logback"] = "io.quarkiverse.logging.logback:quarkus-logging-logback:0.13.0"
 libraries["quarkus-logging-splunk"] = "io.quarkiverse.logging.splunk:quarkus-logging-splunk:2.5.1"
+libraries["resilience4j-spring-boot2"] = "io.github.resilience4j:resilience4j-spring-boot2:2.0.2"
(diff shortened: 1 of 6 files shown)


kartikshahc (approved):
LGTM

PR #2017, opened 2023-04-03
SWATCH-743:Add ui resources to trigger ui tests in post stage deploy

san7ket:


deploy/stage-internal-test-job.yaml
```diff
@@ -71,6 +71,16 @@ objects:
...
+          - name: DYNACONF_main__use_beta
+            value: "yes"
+          - name: DYNACONF_main__use_browser
+            value: "chrome"
+          - name: DYNACONF_browsers__chrome__webdriver
+            value: "Remote"
+          - name: DYNACONF_browsers__chrome__command_executor
+            value: "http://10.0.148.118:4444/wd/hub"
+          - name: DYNACONF_browsers__chrome__webdriver_options__desired_capabilities__chromeOptions__args
+            value: "['--headless','--window-size=1600,900']"
```
mirekdlugosz, line 83: FYI: graph component is known to skip days when forced in small area. This resolution is the same as we use on stage right now, and we observe test failures every now and then due to this. Currently there's two tests that are failing for this reason pretty reliably.

I think we _should_ keep the same screen resolution here as we have in standard stage job.

@diegomaranhao FYI. If we change screen resolution in `swatch-ci` settings file, we should change it here, too.

mirekdlugosz (approved):
LGTM :+1:

ntkathole (approved):

PR #2038, opened 2023-04-10
Remove the Selenium container for using grid

san7ket:
Removes the selenium Container as the iqe plugin is set to use the selenium grid in place


ntkathole (approved):
LGTM

PR #2063, opened 2023-04-17
SWATCH-907: Adds sanity job to detect clowder outages

san7ket:


smoke_test.sh
```diff
@@ -0,0 +1,29 @@
...
+# this test job is used as a 'canary job' to monitor for an ephemeral env outage
+
+set -ex
+
+IMAGE="quay.io/cloudservices/rhsm-subscriptions"
+APP_NAME="rhsm"  # name of app-sre "application" folder this component lives in
+COMPONENT_NAME="rhsm"  # name of app-sre "resourceTemplate" in deploy.yaml for this component
+IQE_PLUGINS="rhsm-subscriptions"
+IQE_MARKER_EXPRESSION=""
+IQE_FILTER_EXPRESSION="test_validate_tally_on_physical_rhel_sockets"
```
ntkathole, line 12: I think purpose of this should not just be detecting outage in clowder but also to check if all tests are working fine, so that it will help devs to compare main vs code branch. This job definition should be similar to pr_check.

san7ket: yea, We can do that, but then we need to reduce the frequency of the job. I guess that's still ok.

ntkathole (approved):

PR #2078, opened 2023-04-21
SWATCH-1164: Use org_id as the message key for tally summary messages

kahowell:
https://issues.redhat.com/browse/SWATCH-1164

This ensures that the messages for a given org_id are processed by the same consumer in order, reducing the risk of concurrency bugs in billable usage production.

Testing
=======

Start the app with debug logging turned on:

```shell
LOGGING_LEVEL_ORG_CANDLEPIN_SUBSCRIPTIONS_TALLY_BILLING=DEBUG \
  SUBSCRIPTION_USE_STUB=true \
  USER_USE_STUB=true \
  DEV_MODE=true \
  ./gradlew :bootRun
```

Import some events into the DB:

```shell
bin/import-events.py --file bin/events.csv
```

Run the hourly tally for that the test events:

```shell
http :9000/hawtio/jolokia \
  type=exec \
  mbean=org.candlepin.subscriptions.jmx:name=tallyJmxBean,type=TallyJmxBean \
  operation='tallyOrgByHourly(java.lang.String,java.lang.String,java.lang.String)' \
  arguments:='["org123", "2021-10-01T00:00Z", "2021-10-30T00:00Z"]'
```

Look for log messages indicating the key used:

```text
2023-04-21 21:20:29,372 [thread=swatch-producer-billing-0-C-1] [DEBUG] [org.candlepin.subscriptions.tally.billing.TallySummaryMessageConsumer] - Tally Summary received w/ key=org123. Producing billable usage.
```

Per [official kafka documentation](https://kafka.apache.org/intro):

> Events with the same event key (e.g., a customer or vehicle ID) are written to the same partition,
> and Kafka guarantees that any consumer of a given topic-partition will always read that partition's events in exactly the same order as they were written.

Therefore, by design, this will ensure that an org's tally summaries will always be emitted to the same partition.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/billing/TallySummaryMessageConsumer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/billing/TallySummaryMessageConsumer.java
@@ -33,6 +33,9 @@ import org.candlepin.subscriptions.util.SeekableKafkaConsumer;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.kafka.annotation.KafkaListener;
+import org.springframework.kafka.support.KafkaHeaders;
+import org.springframework.messaging.handler.annotation.Header;
+import org.springframework.messaging.handler.annotation.Payload;
 import org.springframework.retry.support.RetryTemplate;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
@@ -71,8 +74,10 @@ public class TallySummaryMessageConsumer extends SeekableKafkaConsumer {
       topics = "#{__listener.topic}",
       containerFactory = "billingProducerKafkaTallySummaryListenerContainerFactory")
   @Transactional
-  public void receive(TallySummary tallySummary) {
-    log.debug("Tally Summary received. Producing billable usage.}");
+  public void receive(
+      @Payload TallySummary tallySummary,
+      @Header(name = KafkaHeaders.RECEIVED_MESSAGE_KEY, required = false) String kafkaMessageKey) {
+    log.debug("Tally Summary received w/ key={}. Producing billable usage.", kafkaMessageKey);
 
     billableUsageMapper
         .fromTallySummary(tallySummary)
--- a/src/main/java/org/candlepin/subscriptions/tally/SnapshotSummaryProducer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/SnapshotSummaryProducer.java
@@ -70,7 +70,8 @@ public class SnapshotSummaryProducer {
                     summary -> {
                       if (validateTallySummary(summary)) {
                         kafkaRetryTemplate.execute(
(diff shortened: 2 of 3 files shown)


Aurobinda55:
@kahowell I have created one rhacs event and synced the hourly tally. Then, I checked the log for the particular message but couldn't find it. Anything I am missing here?

src/main/java/org/candlepin/subscriptions/tally/billing/TallySummaryMessageConsumer.java
```diff
@@ -71,8 +74,10 @@ public TallySummaryMessageConsumer(
       topics = "#{__listener.topic}",
       containerFactory = "billingProducerKafkaTallySummaryListenerContainerFactory")
   @Transactional
-  public void receive(TallySummary tallySummary) {
-    log.debug("Tally Summary received. Producing billable usage.}");
+  public void receive(
+      @Payload TallySummary tallySummary,
+      @Header(name = KafkaHeaders.RECEIVED_MESSAGE_KEY, required = false) String kafkaMessageKey) {
+    log.debug("Tally Summary received w/ key={}. Producing billable usage.", kafkaMessageKey);
```
lindseyburnett, line 80: By not requiring the key, that just means that it wouldn't guarantee the message is processed by the same consumer, right?  Other than not reducing the risk of concurrency bugs, is there any other drawback to that?  I'm assuming we're doing required=false for now so we don't get rejected messages....but would we want to toggle that to required=true in a future release?

lindseyburnett:
Confirmed test steps
```
2023-04-25 18:02:46,826 [thread=swatch-producer-billing-0-C-1] [DEBUG] [org.candlepin.subscriptions.tally.billing.TallySummaryMessageConsumer] - Tally Summary received w/ key=org123. Producing billable usage.
```

src/main/java/org/candlepin/subscriptions/tally/billing/TallySummaryMessageConsumer.java
kahowell: `required = false` just allows `null` to be passed.

Per the javadocs:

> Default is true, leading to an exception if the header is missing. Switch this to false if you prefer a null value in case of a header missing.

With this, if we have some pending messages during rollout, then they won't cause exceptions.

Aurobinda55:
- Deployed EE with swatch-tally/LOGGING_LEVEL=DEBUG 
- Created a rhacs event 
- Synced hourly tally 
- Checked tally service and got the blow log message 
`2023-04-26 06:45:31,498 [thread=swatch-producer-billing-0-C-1] [DEBUG] [org.candlepin.subscriptions.tally.billing.TallySummaryMessageConsumer] - Tally Summary received w/ key=3340851. Producing billable usage.`

kahowell:
@Aurobinda55 are you happy with the PR? If so, can you please approve?

Aurobinda55 (approved):
happy path looks good and further testing will do tomorrow

PR #2103, opened 2023-05-01
SWATCH-1188: Remove metering JMX endpoint and replace with REST endpoint

awood:
See https://issues.redhat.com/browse/SWATCH-1188

# Testing

1.
   ```bash
   ./gradlew :bootRun --args="--spring.profiles.active=kafka-queue,openshift-metering-worker --rhsm-subscriptions.enableSynchronousOperations=true --rhsm-subscriptions.auth.swatchPsks.self=secret"
   ```
1. Run metering for a specific org.
   ```bash
   http POST :8000/api/rhsm-subscriptions/v1/internal/metering/OpenShift-metrics orgId==org123 rangeInMinutes==60 x-rh-swatch-synchronous-request:true x-rh-swatch-psk:secret
   ```
1. You will see log messages indicating the metering has started.  However, unless you are connected to a Prometheus instance, you'll see "connection refused" messages.
1. Run metering for all orgs.
   ```bash
   http POST :8000/api/rhsm-subscriptions/v1/internal/metering/OpenShift-metrics rangeInMinutes==60 x-rh-swatch-synchronous-request:true x-rh-swatch-psk:secret
   ```
   This will likewise fail, but you'll see a message reading "Metering OpenShift-metrics for all accounts in the past 60 minutes" above the "Connection Refused" stacktrace.


ntkathole:
@Aurobinda55 PTAL ^

Aurobinda55:
Deployed an EE  and checked the internal api for metering with different parameters like endDate,rangeInMinutes,product_id and orgId and got the response as mentioned in the test steps above. Modification is done on smqe_tools to use the new internal api and modified the test that is failing in the current PR check job.  

@awood  I can see a Step value = 3600 by default in the log, can we use it in the metering call? It was in the JMX call before?

kahowell (approved):

Aurobinda55 (approved):
Verified on EE and looks good

PR #2107, opened 2023-05-02
"Delete" a contract by updating the end date to current timestamp

lindseyburnett:
Jira issue: [SWATCH-XXXX](https://issues.redhat.com/browse/SWATCH-XXXX)

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.


swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -170,17 +170,35 @@ public List<Contract> getContracts(
...
+   * Look up a contract via uuid and update the record to have the given end date.
+   *
+   * @param uuid
+   * @param endDate
+   */
+  @Transactional
+  Optional<ContractEntity> updateContractEndDate(String uuid, OffsetDateTime endDate) {
+    var contract = contractRepository.findContract(UUID.fromString(uuid));
+
+    if (Objects.nonNull(contract)) {
```
kartikshahc, line 194: I assuming we should change the end_date only for active contract. In order to check that should you check if the end_date is null?

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -170,17 +170,35 @@ public List<Contract> getContracts(
...
+      log.debug("Setting end date to {} for contract {}", endDate, uuid);
+      contract.setEndDate(endDate);
```
kartikshahc, line 196: In addition please set the last_updated to the same value.

kartikshahc (changes requested):

kartikshahc:
@lindseyburnett Couple of issues that I found were we don't update the last_updated timestamp when we update the end_date. Also I was assuming we can update the end date only for the active contract. However with the current code I am able to update end_date for both active and non-active contracts.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
lindseyburnett: Opting to not do a null check first to allow for flexibility and "weird" testing that QE wants to do

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -170,17 +170,36 @@ public List<Contract> getContracts(
...
+      contract.setLastUpdated(OffsetDateTime.now());
```
kartikshahc, line 197: I was hoping to set the same endDate here since the enddate and lastDate gets updated at same time.

kartikshahc:
I have left a comment for the update timestamp. I know this is a test API so we are okay with some flexibility but, just in case you decide to incorporate the change. Otherwise LGTM.

kartikshahc (approved):

himmAllRight:
I have updated our tests and added coverage for this:
https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/422

The helper function updates in smqe-tools have also been released now, so we should be all good to go 🙂 .

himmAllRight (approved):

PR #2108, opened 2023-05-02
Remove quarkus-qpid-jms-bom from deps

kahowell:
## Description

When we pivoted to smallrye messaging, we neglected to remove, and this dep is now stale/unreferenced.

## Testing

Automated tests should be sufficient for this one.


kahowell:
Self-merging this one since it's trivial and passed CI.

PR #2112, opened 2023-05-03
Slim down image by using the base image that just contains runtime

lindseyburnett:


kahowell:
FYI, looks like the runtime images don't include s2i:

```java
/bin/bash: /usr/local/s2i/run: No such file or directory
```

I suspect the default command for these is likely fine:

```
/opt/jboss/container/java/run/run-java.sh
```

Accordingly, we could probably just drop `command` field from affected k8s objects.

PR #2137, opened 2023-05-15
SWATCH-1261 Add MCT2741 to the deny-list

barnabycourt:


swatch-subscription-sync/deploy/clowdapp.yaml
```diff
@@ -1241,6 +1241,7 @@ objects:
         MCT2537F5
         MCT2537RN
         MCT2660
+        MCT2741
```
ntkathole, line 1244: should we add all other variants ? or do we have logic to handle it already ?

barnabycourt: We don't, do you have a list of the variants handy?

barnabycourt: nevermind, found it

lindseyburnett (approved):

PR #2141, opened 2023-05-16
Separate building, testing and linting into separate stages

lindseyburnett:
Quality of life change.  It's currently a pain to dig into a bunch of logs when the `Subscription PR Check` jenkins job fails due on the **Build/Test/Lint** step

Check the "Details" link on the continuous-integration/jenkins/pr-merge in the Checks section to the changes.

Before
image

After
image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,12 +31,25 @@ pipeline {
                 input 'ok to test?'
             }
         }
-        stage('Build/Test/Lint') {
+
+        stage('Assemble'){
             steps {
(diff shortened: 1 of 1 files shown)


barnabycourt:
Should any of these stages be running in parallel?

lindseyburnett:
> Should any of these stages be running in parallel?

They weren't before, `./gradlew build` was executing sequentially in this order.  I think we'll always have to assemble before test, but lint could theoretically happen in parallel if we wanted to look into how to make the jenkins steps execute in parallel

PR #2164, opened 2023-05-24
SWATCH-1274 - Add trial SKUs to deny list

barnabycourt:


barnabycourt:
Closing per nack from PM

PR #2174, opened 2023-05-26
SWATCH-1270: Make denylist consider suffixes

kahowell:
Jira issue: [SWATCH-1270](https://issues.redhat.com/browse/SWATCH-1270)

Description
===========

Make denylist consider suffixes. The algorithm for denylist now first normalizes the SKU by removing suffixes. Removing suffixes tries all defined suffixes in length descending order, which avoids subtle issues when a suffix coincidentally has another suffix in it (e.g. `F3RN` has `F3` and `RN`).

Testing
=======

Setup
-----

1. Create a denylist:

```shell
echo 'ASKEW' > /tmp/denylist
```

2. Start the service w/ the configured denylist:

```shell
DEV_MODE=true \
  PRODUCT_DENYLIST_RESOURCE_LOCATION=file:/tmp/denylist \
  ./gradlew :bootRun
```

Steps
-----

Attempt to sync the offering, but with a suffix:

```shell
http :9000/hawtio/jolokia \
  type=exec \
  mbean='org.candlepin.subscriptions.product:name=offeringJmxBean,type=OfferingJmxBean' \
  operation='syncOffering(java.lang.String)' \
  arguments:='["ASKEW"]'
```

Verification
------------

Observe the result is `SKIPPED_DENYLISTED`. You can try other suffixes or no suffix as desired.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/files/ProductDenylist.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/files/ProductDenylist.java
@@ -22,23 +22,33 @@ package org.candlepin.subscriptions.capacity.files;
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
 import java.util.Set;
+import java.util.stream.Stream;
 import javax.annotation.PostConstruct;
+import lombok.extern.slf4j.Slf4j;
 import org.candlepin.subscriptions.ApplicationProperties;
 import org.candlepin.subscriptions.files.PerLineFileSource;
 import org.candlepin.subscriptions.util.ApplicationClock;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.context.ResourceLoaderAware;
 import org.springframework.core.io.ResourceLoader;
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/subscriptions/capacity/files/ProductDenylist.java
```diff
@@ -22,23 +22,33 @@
...
 import org.springframework.stereotype.Component;
 import org.springframework.util.StringUtils;
 
-/** List of products to be considered for capacity calculations. */
+/** List of products to be skipped for capacity calculations. */
+@Slf4j
 @Component
 public class ProductDenylist implements ResourceLoaderAware {
+  private static final List<String> suffixes =
+      Stream.of("F2", "F3", "F3RN", "F4", "F5", "HR", "MO", "RN", "S")
```
mstead, line 44: This seems like something that should be configurable. How likely is this to change?

kahowell: Highly unlikely to change, and we'd want to do as part of a release anyways, so don't care to make configurable.

mstead (approved):

PR #2175, opened 2023-05-26
SWATCH-885: Remove (almost) all references to cloudigrade

kahowell:
Jira issue: [SWATCH-885](https://issues.redhat.com/browse/SWATCH-885)

Description
===========

I kept the liquibase changelog references, and I kept the enum value in HardwareMeasurementType, so that accessing old data doesn't cause issues.

Testing
=======

See that the following shows only HardwareMeasurementType and liquibase:

```shell
git grep -li cloudigrade
```


ntkathole (approved):
Looks good to me 👍  We will also need cleanup in app-interface

mstead (approved):
This looks good to me. Gave the test updates a good look over and seems fine.

PR #2180, opened 2023-05-30
SWATCH-1202: Implement subscription_measurements table

awood:
Jira issue: [SWATCH-1202](https://issues.redhat.com/browse/SWATCH-1202)

## Description
Records in subscription_capacity had several issues: denormalized data (sku, SLA, usage), limited metric storage (only cores and sockets), and the necessity of a mostly redundant view.  The new subscription_measurement table (and associated model and business logic), creates referential integrity between itself and the subscription table; allows for a completely flexible handling of any metric and measurement type combination; and does not require any additional views.

## Testing

A note on migrations: Right now, Stage has 1,338,000 records in subscription_capacity while Prod has 2,915,057.  I was testing this on exports from the S3 buckets with around 600,000 records.  Performance was good: a full run of the migrations took 45 seconds or so, but I need to do further testing on a larger record set.  The primary issue of concern is with having a migration that will take longer than the grace period OpenShift gives an init container.

### Setup

**You probably don't want to annihilate your database for this.**  Do this instead:
1. Create a copy of your existing database
   ```
   createdb -h localhost -U postgres -O rhsm-subscriptions -T rhsm-subscriptions migration
   ```
1. Run a migration against this new database
   ```
   DATABASE_DATABASE=migration ./gradlew :bootRun --args="--spring.profiles.active=liquibase-only"   
   ```

### Steps

1. Import database.  See the private comment on https://issues.redhat.com/browse/SWATCH-1202 for a link to the file.
   ```
   createdb -h localhost -U postgres swatch-1202
   zcat swatch-1202-db.dump.gz | psql -h localhost -U rhsm-subscriptions swatch-1202
   ```
1. Deploy main
   ```
   DATABASE_DATABASE=swatch-1202 ./gradlew :bootRun
   ```
1. Opt in an org
   ```
   http -v :9000/hawtio/jolokia type=exec mbean=org.candlepin.subscriptions.jmx:name=optInJmxBean,type=OptInJmxBean operation="ensureOptInForOrgId(java.lang.String,boolean,boolean,boolean)" arguments:='["13269666", true, true, true]' origin:foobar:8000
   ```
1. Fetch capacity for a few things
   ```
   http :8000/api/rhsm-subscriptions/v1/capacity/products/RHEL/Sockets granularity==daily beginning==2023-01-31T11:00:00.000Z ending==2023-02-01T11:59:59.999Z x-rh-identity:$(echo -n '{"identity":{"account_number":"","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"13269666"}}}' | base64 -w0)
   http :8000/api/rhsm-subscriptions/v1/capacity/products/RHEL/Sockets granularity==monthly beginning==2022-01-31T11:00:00.000Z ending==2022-05-01T11:59:59.999Z x-rh-identity:$(echo -n '{"identity":{"account_number":"","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"13269666"}}}' | base64 -w0)
   ```
1. Switch to branch `awood/SWATCH-1202-subscriptions-measurements`
1. `DATABASE_DATABASE=swatch-1202 ./gradlew :bootRun` which will also run the
   migration.
1. Rerun those calls to the capacity resource.
1. Do note, however, that the results **are not** going to be exactly the same numbers.  Because of duplication in stage data and the overall lack of referential integrity between `subscription_capacity` and `subscription`, the migration makes a best effort but it just isn't possible to migrate every `subscription_capacity` record since many of them are orphans.


src/main/resources/liquibase/202305031403-create-subscription-measurements-table.xml
```diff
@@ -0,0 +1,229 @@
...
+      columnNames="product_id, subscription_id, start_date"/>
+    <addForeignKeyConstraint constraintName="subs_product_ids_fk"
+      baseTableName="subscription_product_ids"
+      referencedTableName="subscription"
+      baseColumnNames="subscription_id, start_date"
+      referencedColumnNames="subscription_id, start_date"/>
+  </changeSet>
+
+  <!-- NB: All the changesets dealing with functions and triggers are scoped to postgresql.  We
+  don't need them during unit tests -->
```
kahowell, line 60: I dislike the added complexity of the triggers/functions. Given we do a nightly sync of subscriptions, I'd rather remove them and let any inconsistencies be addressed by the next subscription sync.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/resource/CapacityResource.java
```diff
@@ -315,15 +326,36 @@ protected List<CapacitySnapshotByMetricId> getCapacitiesByMetricId(
...
+    var criteria =
+        DbReportCriteria.builder()
+            .orgId(orgId)
+            .productId(productId.toString())
+            .serviceLevel(sla)
+            .usage(usage)
+            .beginning(reportBegin)
+            .ending(reportEnd)
+            .build();
+    return !subscriptionRepository.findUnlimited(criteria).isEmpty();
```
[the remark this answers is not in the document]
kahowell: Even better I realized when I got further down you can create an `exists` variant.

src/main/resources/liquibase/202305031403-create-subscription-measurements-table.xml
lindseyburnett: +1

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/Subscription.java
```diff
@@ -73,8 +77,96 @@ public class Subscription {
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof Subscription sub)) {
+      return false;
+    }
+
+    var ourProduceIds =
```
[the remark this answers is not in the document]
lindseyburnett: 🥕

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/Subscription.java
```diff
@@ -73,8 +77,96 @@ public class Subscription {
...
+  @Column(name = "has_unlimited_usage")
+  private Boolean hasUnlimitedUsage;
+
+  public Boolean getHasUnlimitedUsage() {
+    return hasUnlimitedUsage;
+  }
+
+  @Override
+  // Implemented to avoid fetching child entities -- especially lazy loaded ones
+  public String toString() {
```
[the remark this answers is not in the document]
lindseyburnett: I generally use `@ToString.Exclude`

awood:
Running on my machine with a fresh import of stage data from `subscription` (500K records) and `subscription_capacity` (1.3 million records), the migration took about 75 seconds.

src/main/resources/liquibase/202305031403-create-subscription-measurements-table.xml
kahowell: After some discussion, we're leaning towards keeping them (but they won't be long lived)

src/main/resources/liquibase/changelog.xml
```diff
@@ -101,6 +101,9 @@
     <include file="liquibase/202212161446-add-derived-sku-to-offering.xml"/>
     <include file="liquibase/202301311609-update-tally-instance-view-default-num-of-guest-add-subscription_manager_id.xml"/>
     <include file="liquibase/202302141154-add-basilisk-product.xml"/>
-    <include file="/liquibase/202302151544-remove-uuid-cast-on-instance-id-in-tally-instance-view.xml"/>
+    <include file="liquibase/202302151544-remove-uuid-cast-on-instance-id-in-tally-instance-view.xml"/>
+    <include file="liquibase/202305031403-create-subscription-measurements-table.xml"/>
+    <!-- Uncomment once subscription-measurements has been verified -->
```
lindseyburnett, line 106: Should we make a card for this so it doesn't get forgotten?

kahowell (approved):

lindseyburnett (approved):

awood:
Draft status temporarily while I track down a bug in how capacities are being reported.  My test steps caught this, so they were definitely worth writing up.

kahowell:
Replacing @diegomaranhao with @TrayvonMcKnight as QE reviewer

TrayvonMcKnight (approved):
This looks good to me.

PR #2189, opened 2023-05-31
SWATCH-1216: Issue single DELETE query when purging Event records to prevent OOM killing pods

lindseyburnett:
Jira issue: [SWATCH-1216](https://issues.redhat.com/browse/SWATCH-1216)

## Description
Calls to DELETE /internal/metering/events were OOM killing pods.

By default, a derived delete query will return instances matching the where clause prior to deleting the entities one at a time.  This was too memory intensive.  See [this slack thread](https://redhat-internal.slack.com/archives/C01F7QFNATC/p1685035237143889) for details about how the cause was chased down.  Since we're not concerned with invoking any lifecycle callbacks, we can override the behavior with @Query to delete all matching records in bulk. 

https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.modifying-queries.derived-delete

## Testing
If you want to test/monitor the memory consumption, I suggest downloading data from s3 data dumps from stage.  The steps I'm listing here are to manually observe that prior to these changes, a SELECT statement was loading EventRecord entities and deleting them individually.  Then after the changes you can observe the delete happens with a single DELETE statement.

You need to change some logging levels in order to see the difference.  Add this to the bottom of `src/main/resources/application-openshift-metering-worker.yaml`
```yaml
logging:
  level:
    org:
      hibernate:
        SQL: DEBUG
spring:
  jpa:
    properties:
      hibernate:
        format_sql: true
```

### On `main` branch

`DEV_MODE=true SPRING_PROFILES_ACTIVE=kafka-queue,openshift-metering-worker ./gradlew :bootRun`

`curl -X DELETE -s "Content-Type: application/json" "localhost:8000/api/rhsm-subscriptions/v1/internal/metering/events" -H "x-rh-swatch-psk: placeholder"`

Notice in the logs:
```
    select
        eventrecor0_.id as id1_3_,
        eventrecor0_.account_number as account_2_3_,
        eventrecor0_.data as data3_3_,
        eventrecor0_.event_source as event_so4_3_,
        eventrecor0_.event_type as event_ty5_3_,
        eventrecor0_.instance_id as instance6_3_,
        eventrecor0_.org_id as org_id7_3_,
        eventrecor0_.timestamp as timestam8_3_ 
    from
        events eventrecor0_ 
    where
        eventrecor0_.timestamp<?
```

### Switch to `lburnett/delete-events-oom` branch

Restart the application and run the same curl command

Notice in the logs this time:
```
    delete 
    from
        events 
    where
        timestamp<=?
```


swatch-core/src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
```diff
@@ -82,9 +83,11 @@ Stream<EventRecord> findByOrgIdAndTimestampGreaterThanEqualAndTimestampLessThanO
   /**
    * Delete old event records given a cutoff date
    *
-   * @param cutoffDate Dates BEFORE this timestamp get deleted
+   * @param cutoffDate Dates before this timestamp get deleted
    */
-  void deleteEventRecordsByTimestampBefore(OffsetDateTime cutoffDate);
+  @Modifying
+  @Query("DELETE FROM EventRecord e WHERE e.timestamp<:cutoffDate")
+  void deleteInBulkEventRecordsByTimestampBefore(OffsetDateTime cutoffDate);
```
awood, line 90: Would you mind adding to the javadoc why this method exists?  Just something like what you have in the PR description.

```
By default, a derived delete query will return instances matching the where clause prior to deleting the entities one at a time. This was too memory intensive.  Since we're not concerned with invoking any lifecycle callbacks, we can override the behavior with @query to delete all matching records in bulk.  See the <a href="https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.modifying-queries.derived-delete">Spring Data documentation</a> on deletion.
```

awood (changes requested):

swatch-core/src/main/java/org/candlepin/subscriptions/db/EventRecordRepository.java
lindseyburnett: i can do that

awood (approved):

TrayvonMcKnight:
In the /tmp/rhsm-subscriptions.log file I only see timestamps.

TrayvonMcKnight:
In the rhsm-subscriptions.log file, I only see time stamps instead of a SELECT statements.

`2023-06-02T19:26:55.786+0000 [thread=http-nio-8000-exec-1] [level=INFO] [category=org.candlepin.subscriptions.security.LogPrincipalFilter] self - Internal API: /api/rhsm-subscriptions/v1/internal/metering/events requested by user: self
2023-06-02T19:26:55.813+0000 [thread=http-nio-8000-exec-1] [level=INFO] [category=org.candlepin.subscriptions.metering.api.admin.InternalMeteringResource] self - Purging event records older than 2023-03-04T00:00Z
2023-06-02T19:26:55.832+0000 [thread=http-nio-8000-exec-1] [level=INFO] [category=org.candlepin.subscriptions.metering.api.admin.InternalMeteringResource] self - Event record purge completed successfully
`

lindseyburnett:
Looks like your logging level needs to be changed to DEBUG

> In the rhsm-subscriptions.log file, I only see time stamps instead of a SELECT statements.
> 
> 
> 
> `2023-06-02T19:26:55.786+0000 [thread=http-nio-8000-exec-1] [level=INFO] [category=org.candlepin.subscriptions.security.LogPrincipalFilter] self - Internal API: /api/rhsm-subscriptions/v1/internal/metering/events requested by user: self
> 
> 2023-06-02T19:26:55.813+0000 [thread=http-nio-8000-exec-1] [level=INFO] [category=org.candlepin.subscriptions.metering.api.admin.InternalMeteringResource] self - Purging event records older than 2023-03-04T00:00Z
> 
> 2023-06-02T19:26:55.832+0000 [thread=http-nio-8000-exec-1] [level=INFO] [category=org.candlepin.subscriptions.metering.api.admin.InternalMeteringResource] self - Event record purge completed successfully
> 
> `
> 
>

TrayvonMcKnight:
Ack, I had the DEBUG statement nested.
[OCM_pod_testing.txt](https://github.com/RedHatInsights/rhsm-subscriptions/files/11653417/OCM_pod_testing.txt)

This looks good to me!

kartikshahc (approved):

PR #2260, opened 2023-06-25
SWATCH-1107 - IQE_RP_ARGS is boolean value now

ntkathole:
Jira issue: [SWATCH-1107](https://issues.redhat.com/browse/SWATCH-1107)

## Description

- As per https://gitlab.cee.redhat.com/insights-qe/iqe-core/-/merge_requests/1524 change, IQE_RP_ARGS is boolean value  instead of string. 
- Also making IQE_PARALLEL_ENABLED as false as currently we try to find if any test exists that can run in parallel even if we do not have any test running in parallel on pr_check. This will save around 1 min on every PR.


ntkathole:
@san7ket Please review.

lindseyburnett (approved):

PR #2277, opened 2023-06-29
add snapshot_id to billable_usage_remittance table

kflahert:
Add the snapshot_id to billable_usage_remittances


mstead (changes requested):
Overall this looks good to me.

One thing that needs to be addressed however, is that we need to re-align this with the rollback changeset.

PR #2317, opened 2023-07-14
SWATCH-1520: Invalidate the SKU existence cache properly when offerings are saved

awood:
Jira issue: [SWATCH-1520](https://issues.redhat.com/browse/SWATCH-1520)

## Description

During a subscription sync, the following events can occur:

* A check for a SKU is performed.  The SKU is not found
* A value of "false" for that SKU is written to the cache that stores the existence of SKUs
* We save an offering with that SKU
* Another subscription with the same SKU comes along and we make another call to OfferingRepository.existsById
* The cache returns false instead of true because we did not update the offering existence cache after saving the offering

This patch removes a SKU from the SKU existence cache after a save or saveAndFlush operation.  For saveAll or saveAllAndFlush operations, the entire cache is invalidated because there is no way to loop through the iterable of saved entities to extract the SKU from each one.

## Testing

Testing is covered by the unit tests.

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/db/OfferingRepositoryCachingTest.java
+++ b/src/test/java/org/candlepin/subscriptions/db/OfferingRepositoryCachingTest.java
@@ -22,8 +22,7 @@ package org.candlepin.subscriptions.db;
 
 import static java.util.Optional.empty;
 import static java.util.Optional.ofNullable;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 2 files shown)


barnabycourt:
@awood Since we have multiple instances of the service running at any point in time. Won't the cache eviction only affect one of the pods with the current cache implementation?

awood:
> @awood Since we have multiple instances of the service running at any point in time. Won't the cache eviction only affect one of the pods with the current cache implementation?

This is an excellent point.  I'm going to close this PR and revert the cache stuff until we can figure out a way to handle it in a unified (across pods) manner.

PR #2339, opened 2023-07-19
SWATCH-1519: On running total tallies, report the past as has_data and skip future

awood:
Jira issue: [SWATCH-1519](https://issues.redhat.com/browse/SWATCH-1519)

## Description

On running total tallies, report the past as has_data and skip future

Right now on calls to /tally/products/{product_id} we are marking days that don't have an associated measurement with has_data=false.  This choice doesn't really make sense with a running total since even if we don't have a measurement, we do still have data.  Therefore, for days in the past, we will set has_data=true and set the value equal to the running total. However, we should keep marking days in the future days with has_data=true.

This PR also makes modifications to how we are handling time in our tests.  I modified the fixedClock that we were using to allow tests to modify the underlying instant so that tests have more flexibility around when they consider "now" to be.  This change touched a lot of files, so the easiest way to review this PR is commit by commit.  Both commits compile, pass tests, and pass spotless.

### Setup
1. Unzip and import [swatch-1519.zip](https://github.com/RedHatInsights/rhsm-subscriptions/files/12099880/swatch-1519.zip) (you'll see some insertions error, but just ignore them.  They're just an artifact of my export process)
   ```shell
   zcat swatch-1519.zip | psql -h localhost -U rhsm-subscriptions rhsm-subscriptions
   ```

### Steps
1. `git checkout origin/main`
1. `./gradlew :bootRun`
1. Ask for a tally report
   ```shell
    http :8000/api/rhsm-subscriptions/v1/tally/products/rosa/Instance-hours \
    granularity==Daily \
    beginning==2023-07-01T00:00:00.000Z \
    ending==2023-07-31T23:59:59.999Z \
    use_running_totals_format==true \
    billing_category==prepaid \
    x-rh-identity:$(echo -n '{"identity":{"account_number":"account123","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w 0)
   ```
1. You'll see 31 entries in the report and days without specific measurements will be marked with `has_data: false`.

### Verification
1. `git checkout -t origin/awood/SWATCH-1519-running-total-correction`
1. `./gradlew :bootRun`
1. Ask for a tally report
   ```shell
    http :8000/api/rhsm-subscriptions/v1/tally/products/rosa/Instance-hours \
    granularity==Daily \
    beginning==2023-07-01T00:00:00.000Z \
    ending==2023-07-31T23:59:59.999Z \
    use_running_totals_format==true \
    billing_category==prepaid \
    x-rh-identity:$(echo -n '{"identity":{"account_number":"account123","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w 0)
   ```
1. Since `2023-07-31` is currently in the future, you should see records up to whatever today's date is with `has_data` set to true.  You should see `has_data` marked as false for each day that's in the future.  Additionally, days in the future will have a value of zero.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/filler/TallyReportDataPointAdapter.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/filler/TallyReportDataPointAdapter.java
@@ -22,9 +22,16 @@ package org.candlepin.subscriptions.tally.filler;
 
 import java.time.OffsetDateTime;
 import java.util.Objects;
+import org.candlepin.subscriptions.util.ApplicationClock;
 import org.candlepin.subscriptions.utilization.api.model.TallyReportDataPoint;
 
 public class TallyReportDataPointAdapter implements ReportFillerAdapter<TallyReportDataPoint> {
+
+  private final ApplicationClock clock;
+
+  public TallyReportDataPointAdapter(ApplicationClock clock) {
+    this.clock = clock;
+  }
 
   @Override
   public boolean itemIsLarger(TallyReportDataPoint oldData, TallyReportDataPoint newData) {
@@ -35,12 +42,15 @@ public class TallyReportDataPointAdapter implements ReportFillerAdapter<TallyRep
   public TallyReportDataPoint createDefaultItem(
       OffsetDateTime itemDate, TallyReportDataPoint previous, boolean useRunningTotalFormat) {
     var point = new TallyReportDataPoint().date(itemDate).hasData(false).value(0);
-    if (useRunningTotalFormat) {
-      // Guard against previous being null and against a previous object with a value field equal
-      // to null
-      var prevValue = (previous == null) ? 0 : Objects.requireNonNullElse(previous.getValue(), 0);
-      point.value(prevValue);
+    // itemDate is already adjusted in ReportFiller to be at the period start.  ReportFiller is also
+    // responsible for ticking through every period contained in the requested range, so we should
+    // not see any time gaps in the finished report.
+    if (itemDate.isBefore(clock.now()) && useRunningTotalFormat && previous != null) {
+      point
+          .hasData(previous.getValue() != null)
+          .value(Objects.requireNonNullElse(previous.getValue(), 0));
     }
+
     return point;
   }
 
--- a/src/test/java/org/candlepin/subscriptions/resource/TallyResourceTest.java
+++ b/src/test/java/org/candlepin/subscriptions/resource/TallyResourceTest.java
@@ -40,7 +40,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
-import org.candlepin.subscriptions.FixedClockConfiguration;
 import org.candlepin.subscriptions.db.AccountConfigRepository;
 import org.candlepin.subscriptions.db.BillableUsageRemittanceRepository;
 import org.candlepin.subscriptions.db.TallySnapshotRepository;
@@ -56,6 +55,7 @@ import org.candlepin.subscriptions.json.Measurement.Uom;
 import org.candlepin.subscriptions.resteasy.PageLinkCreator;
 import org.candlepin.subscriptions.security.RoleProvider;
 import org.candlepin.subscriptions.security.WithMockRedHatPrincipal;
+import org.candlepin.subscriptions.test.TestClockConfiguration;
 import org.candlepin.subscriptions.util.ApplicationClock;
 import org.candlepin.subscriptions.util.SnapshotTimeAdjuster;
 import org.candlepin.subscriptions.utilization.api.model.*;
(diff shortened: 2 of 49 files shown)


src/test/java/org/candlepin/subscriptions/resource/TallyResourceTest.java
```diff
@@ -67,24 +71,31 @@
...
   @MockBean AccountConfigRepository accountConfigRepository;
   @Autowired TallyResource resource;
 
+  @Autowired ApplicationClock applicationClock;
+
   @BeforeEach
   public void setupTests() {
     when(accountConfigRepository.existsByOrgId("owner123456")).thenReturn(true);
+    var testClock = (TestClock) applicationClock.getClock();
+    testClock.setInstant(MID_MONTH_INSTANT);
```
kahowell, line 98: Doesn't this pollute state for other tests? I don't mind having an adjustable clock, but then injecting it to be shared seems problematic.

src/main/java/org/candlepin/subscriptions/tally/filler/TallyReportDataPointAdapter.java
```diff
@@ -34,7 +34,8 @@ public boolean itemIsLarger(TallyReportDataPoint oldData, TallyReportDataPoint n
   @Override
   public TallyReportDataPoint createDefaultItem(
       OffsetDateTime itemDate, TallyReportDataPoint previous, boolean useRunningTotalFormat) {
-    var point = new TallyReportDataPoint().date(itemDate).hasData(false).value(0);
+    // Mark the datapoint as having data since this is a running total
+    var point = new TallyReportDataPoint().date(itemDate).hasData(true).value(0);
```
kahowell, line 38: We need `hasData=false` when `useRunningTotalFormat=false`:
```suggestion
    var point = new TallyReportDataPoint().date(itemDate).hasData(useRunningTotalFormat).value(0);
```

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/tally/filler/TallyReportDataPointAdapter.java
kartikshahc: I didn't see any other problem than ^
I tested for `use_running_totals_format==true` and it returns `has_data: true` and no future date
But when I tested `use_running_totals_format==false` and it returns `has_data: true` instead of `useRunningTotalFormat` with future dates

kartikshahc (changes requested):

src/test/java/org/candlepin/subscriptions/resource/TallyResourceTest.java
awood: I thought that the application context was brought up on a per-test class basis.  Your point is well taken though.  I will see if I can make some adjustments here.

ntkathole:
Since we are returning ﻿﻿records up to whatever today's date is, I think this may change frontend to not display future dates for rosa offering. Are we ok with this change in frontend ? 
Tagging @cdcabrera for visibility

kartikshahc (approved):

src/test/java/org/candlepin/subscriptions/resource/TallyResourceTest.java
kahowell: See https://docs.spring.io/spring-framework/reference/testing/testcontext-framework/ctx-management/caching.html

kartikshahc (changes requested):
After this morning discussion with David turning the PR review to request changes

awood:
> @awood Please update the API spec for /tally/products/{product_id}/{metric_id}: to describe the expectations on what would be provided in the results based on the changes here, use_running_totals_format & what the reporting period is set to. The expectation of the results should be discoverable from the API spec.

Done.

awood:
@kahowell @kartikshahc @ntkathole  Should be updated and ready for re-review

ntkathole:
> @kahowell @kartikshahc @ntkathole Should be updated and ready for re-review

Looks good to me! I leave the qe ack to @diegomaranhao

src/main/java/org/candlepin/subscriptions/tally/filler/TallyReportDataPointAdapter.java
```diff
@@ -35,12 +42,13 @@ public boolean itemIsLarger(TallyReportDataPoint oldData, TallyReportDataPoint n
...
-      var prevValue = (previous == null) ? 0 : Objects.requireNonNullElse(previous.getValue(), 0);
-      point.value(prevValue);
+    // itemDate is already adjusted in ReportFiller to be at the period start
+    if (itemDate.isBefore(clock.now()) && useRunningTotalFormat && previous != null) {
+      point
+          .hasData(previous.getValue() != null)
+          .value(Objects.requireNonNullElse(previous.getValue(), 0));
     }
+
     return point;
```
kahowell, line 52: How does this behave if we have a gap in the data, especially if a customer turns off systems for the weekend, and then comes back, do we fill in the gaps until `now()` or not? If there's not a gap, I would appreciate some code comments explaining how this behaves in that scenario.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/tally/filler/TallyReportDataPointAdapter.java
awood: There shouldn't be a gap.  The `testTallyReportDataFillerMarksHasDataFalseAfterNowWithRunningTotal` unit test covers this I believe:

```java
List<TallySnapshot> snapshots =  List.of(1, 2, 8).stream()...
```

And then ensures there are records for every day requested.  The `ReportFiller.fillWithRange()` method is what handles making sure we tick through every period in the range requested.  I can add a comment to that effect.

kartikshahc (approved):
LGTM

kahowell (approved):

PR #2355, opened 2023-07-25
SWATCH-1375 Unify the values of metricId and UOM 

kartikshahc:
Jira issue: [SWATCH-1375](https://issues.redhat.com/browse/SWATCH-1375)

## Description

SWATCH-1375 Unify the values of metricId and UOM in tag profile/configuration

## Testing

There are no testing steps since the goal is to update metricId with uom values. IMO as long as the unit tests passes I assume it should be good.


kartikshahc:
Do not merge until SWATCH-1392 is deployed. Or this PR and SWATCH-1392 has to go in prod at same time

PR #2368, opened 2023-07-28
SWATCH-1554 Reduce subscription sync memory use

lindseyburnett:
Jira issue: [SWATCH-1554](https://issues.redhat.com/browse/SWATCH-1554)

## Description

Old Flow:
- Create a List of all the Subscription dtos for an org from the Subscription Service
- Create a Map of all the Subscrpition entities for an org from the database
- Loop through the dtos, cross referencing the entities map.  Sync or delete corresponding Subscription entity.

New Flow:
- Create a Map of all the Subscription dtos for an org from the Subscription Service
- Stream Subscription entities from the database.  Added a @QueryHint to  `Stream<Subscription> findByOrgId(String orgId)` to fetch in batches.
- Loop over Stream<Subscription> in batches, allowing us to periodically flush `subscriptionRepository` and clear the entity manager to release already processed Subscription entities hanging out in memory.

Misc. Changes: 
- Stop using @EntityGraph when invoking findByOfferingSku, which was the source of `HHH90003004: firstResult/maxResults specified with collection fetch; applying in memory` warning when syncing an offering during capacity reconcilliation.

- Switch `QueryHints` for `HibernateHints` because `QueryHints` is marked as deprecated as of Hibernate 6.

Co-authored-by: Kevin Howell <khowell@redhat.com>

## Testing

Unit tests & IQE/Regression testing mainly.

If you want to play yourself, it's easiest to grab some data off of stage to load up locally.  I used a combination of the **jconsole** profiler as well as the one built into IntelliJ to compare memory usage.

The closest way I've found to replicating the hardware limitations in stage is by assembling the jar and passing some JVM arguments, rather than starting the application with :bootRun like we usually do.  When we started experiencing the crash looping due to OOM, the max memory was set to 800m.  At the time of writing this, it's at 2G.

```bash
./gradlew :assemble

DEV_MODE=true SPRING_PROFILES_ACTIVE=kafka-queue,capacity-ingress java -Xmx800m -XX:ActiveProcessorCount=1 -jar /home/lburnett/code/rhsm-subscriptions/build/libs/rhsm-subscriptions-1.1.0-snapshot.202308031816.uncommitted+lburnett.subscription.sync.memory.2eebc5c.jar
```

### Steps

1.  Get a bunch of data in your database
2. Start application with `capacity-ingress,kafka-queue` profiles.  Don't use stubs for the product service.
3. Kick of subscription sync for an org
```
http PUT :8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/**orgId** x-rh-swatch-psk:placeholder
```
OR
Kick off a subscription sync for all orgs
```
http PUT :8000/api/rhsm-subscriptions/v1/internal/rpc/subscriptions/sync x-rh-swatch-psk:placeholder
```

### Verification
Here are some screenshots of resource usage from before and after the changes.  These were taken with no JVM limitations, and were captured while syncing all orgs (via the http command above).

**main branch**
image

**this branch**

Note the memory doesn't spike as high, the overall process takes significantly less time, and the CPU/Memory usage are more uniform.  The slow increase in heap usage towards the end is when the service was just sitting idle in the background because I didn't disconnect the profile immediately after it finishing.

image

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/inventory/db/InventoryRepository.java
+++ b/src/main/java/org/candlepin/subscriptions/inventory/db/InventoryRepository.java
@@ -20,8 +20,8 @@
  */
 package org.candlepin.subscriptions.inventory.db;
 
-import static org.hibernate.jpa.QueryHints.HINT_FETCH_SIZE;
-import static org.hibernate.jpa.QueryHints.HINT_READONLY;
+import static org.hibernate.jpa.HibernateHints.HINT_FETCH_SIZE;
+import static org.hibernate.jpa.HibernateHints.HINT_READ_ONLY;
 
 import jakarta.persistence.QueryHint;
 import java.util.Collection;
@@ -42,7 +42,7 @@ public interface InventoryRepository extends Repository<InventoryHost, UUID> {
   @QueryHints(
       value = {
         @QueryHint(name = HINT_FETCH_SIZE, value = "1024"),
-        @QueryHint(name = HINT_READONLY, value = "true")
+        @QueryHint(name = HINT_READ_ONLY, value = "true")
       })
   Stream<InventoryHostFacts> streamFacts(
       @Param("orgId") String orgId, @Param("culledOffsetDays") Integer culledOffsetDays);
@@ -109,7 +109,7 @@ public interface InventoryRepository extends Repository<InventoryHost, UUID> {
   @QueryHints(
       value = {
         @QueryHint(name = HINT_FETCH_SIZE, value = "1024"),
-        @QueryHint(name = HINT_READONLY, value = "true")
+        @QueryHint(name = HINT_READ_ONLY, value = "true")
       })
   Stream<String> streamActiveSubscriptionManagerIds(
       @Param("orgId") String orgId, @Param("culledOffsetDays") Integer culledOffsetDays);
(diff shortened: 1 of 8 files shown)


ntkathole (approved):
QE will test once merge into https://github.com/RedHatInsights/rhsm-subscriptions/pull/2393

src/main/java/org/candlepin/subscriptions/inventory/db/InventoryRepository.java
```diff
@@ -20,8 +20,8 @@
  */
 package org.candlepin.subscriptions.inventory.db;
 
-import static org.hibernate.jpa.QueryHints.HINT_FETCH_SIZE;
-import static org.hibernate.jpa.QueryHints.HINT_READONLY;
+import static org.hibernate.jpa.HibernateHints.HINT_FETCH_SIZE;
```
kartikshahc, line 23: In the QueryHints documentation they have proposed to use AvailableHints. I suggest to use that however AvailableHints use HibernateHints internally so, either ways it is fine.
Reference: https://docs.jboss.org/hibernate/orm/6.0/javadocs/org/hibernate/annotations/QueryHints.html

kartikshahc (approved):

lindseyburnett:
Per discussion during muitos amigues meeting - merging this so it can be tested in stage environment due to limitations with subscription sync IQE tests against ephemeral.

PR #2383, opened 2023-08-07
SWATCH-1557: Add order by clauses on SubscriptionRepository to avoid Hibernate issue

awood:
[HHH-17040](https://hibernate.atlassian.net/browse/HHH-17040) is a Hibernate issue that can cause a single root entity to be returned in duplicate depending on the row order of the result set.  We can side-step the issue by adding an "order by" clause based on the composite primary key we use for Subscription.  That "order by" will ensure all rows of the same Subscription are grouped together.

Jira issue: [SWATCH-1557](https://issues.redhat.com/browse/SWATCH-1557)

## Description
This is actually a follow on PR based on some additional research and investigation.  We augmented the `Collectors.toMap` call in `SubscriptionSyncController` to handle multiple copies of the same `Subscription` being returned.  However, as I was looking into this bug, I noticed that when we run into the issues with duplicate objects, the `productId` collection is incorrect.  Basically, we end up with 25 copies of the same `Subscription` all with 1 `productId` on the `Offering` instead of 1 `Subscription` with 25 different `productIds` on the `Offering`.  So while modifying the `Collectors.toMap` does at least prevent an exception from being thrown, it's also giving us a false sense of security.  

Currently, the `DuplicateKeyException` we were seeing was popping up in `reconcileSubscriptionsWithSubscriptionService` but I have concerns that `subscriptionRepository.findActiveSubscription` called in `syncSubscriptions` might be behaving improperly.  That subscription gets passed on to the `CapacityReconcilliationController` who, in turn, references the `productId` collection on the `Offering`.

## Testing
Unfortunately, these queries need to be exercised via the unit tests.

### Steps
1.  Run some of the unit tests that will exercise the modified queries like so:
   ```
   SPRING_APPLICATION_JSON='{"spring":{"jpa":{"show-sql":"true","properties":{"hibernate":{"format_sql":"true"}}}}}' ./gradlew -i clean :test --tests="SubscriptionRepositoryTest.findsAllSubscriptionsForAGivenSku"
   ```
2.  I recommend the above test along with `findsUnlimitedSubscriptions`, `testSubscriptionIsActive`, and `canMatchOfferings` (the last to verify that we can still specify sort order if we want).
### Verification
1.  Observe that the generate queries have `order by s1_0.subscription_id, s1_0.start_date` on the end.

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/db/SubscriptionRepository.java
@@ -54,23 +54,33 @@ public interface SubscriptionRepository
     extends JpaRepository<Subscription, Subscription.SubscriptionCompoundId>,
         JpaSpecificationExecutor<Subscription> {
 
+  // Added an order by clause to avoid Hibernate issue HHH-17040
   @Query(
       """
         SELECT s FROM Subscription s
         WHERE (s.endDate IS NULL OR s.endDate > CURRENT_TIMESTAMP)
           AND s.subscriptionId = :subscriptionId
+        ORDER BY s.subscriptionId, s.startDate
       """)
   @EntityGraph(value = "graph.SubscriptionSync")
   Optional<Subscription> findActiveSubscription(@Param("subscriptionId") String subscriptionId);
 
   @EntityGraph(value = "graph.SubscriptionSync")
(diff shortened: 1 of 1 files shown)


ntkathole (approved):
LGTM! Could not verify the exact root cause, but verified that capacity sync is working fine.

lindseyburnett (approved):

PR #2399, opened 2023-08-14
Add gradle build to RHTAP pipelines

kahowell:
Needed to perform the gradle build step prior to doing a container build.


kahowell:
Self-merging since this affects WIP RHTAP setup only.

PR #2411, opened 2023-08-17
SWATCH-1244: Refactor aws integration to use swatch-product-configuration

Sgitario:
Jira Issue: https://issues.redhat.com/browse/SWATCH-1244

# Description

Replace the tag_profile functionality by using swatch-product-configuration.

Note that I'm getting the metric just once and use it upwards to get the aws dimension from it. 

# Testing

I haven't added any new tests because the existing coverage already ensures that nothing is broken (the existing swatch-product-configuration covers what tag_profile did).


Sgitario:
I think the test failure is unrelated to these changes.

swatch-producer-aws/src/main/java/com/redhat/swatch/processors/BillableUsageProcessor.java
```diff
@@ -92,7 +91,8 @@ public void process(BillableUsage billableUsage) {
       MDC.put("account_id", billableUsage.getAccountNumber());
     }
 
-    if (!isApplicable(billableUsage)) {
+    Optional<Metric> metric = lookupMetric(billableUsage);
```
kahowell, line 94: nitpick: method name should reflect that this also rejects non-AWS usage. Maybe `validateUsageAndLookupMetric` ? Alternatively, we could keep them separated. Something like:

```java
var isValid = isApplicable(billableUsage);
Optional<Metric> metric = Optional.empty();
if (isValid) {
  metric =  lookupMetric(billableUsage);
}
if (!isValid || metric.isEmpty()) {
```
Either would be fine to me

kahowell (changes requested):

swatch-producer-aws/src/main/java/com/redhat/swatch/processors/BillableUsageProcessor.java
Sgitario: Method name updated to `validateUsageAndLookupMetric`

Sgitario:
PR updated
fyi @kahowell

PR #2422, opened 2023-08-22
Replace tests needing prometheus service to integration tests

Sgitario:
Instead of mocking the Prometheus service in the following tests:
- PrometheusMeteringControllerTest
- PrometheusServiceTest

We can configure Wiremock to mock the Prometheus Query API.  The main issue that these changes address is that at the moment, the Prometheus API, the API client (these two are auto-generated by OpenAPI), and the Prometheus service are tightly coupled with each other, so changing one, it would require changing the rest. Moreover, the tests will be more readable and also we can improve them to also mock large content of data. 

Note that we can only use this approach in tests that also use the SpringBootTest extension.

This is related to https://issues.redhat.com/browse/SWATCH-1574 where it was hard to try to test a scenario with a large amount of data in the current tests.

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusQueryWiremockExtension.java
+++ b/src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusQueryWiremockExtension.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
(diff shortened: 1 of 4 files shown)


src/test/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusQueryWiremockExtension.java
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.jupiter.api.extension.*;
+
+/**
+ * Utility class to initialize and stub a mocked version of the prometheus server using WireMock.
+ *
+ * <p>The test classes need to configure the prometheus url to use `localhost:9000`. Example:
+ * {@code @SpringBootTest(properties =
+ * "rhsm-subscriptions.metering.prometheus.client.url=http://localhost:9000")}
+ */
+public class PrometheusQueryWiremockExtension
```
Sgitario, line 42: We can improve this extension by adding many features like mocking more Prometheus scenarios or allow to inject the PrometheusQueyWiremock directly into a test class field. 
However, I wanted to keep things simpler for now. Specially, if you prefer these tests being purely an unit test instead of an integration test.

Sgitario:
> I like this. One thing that could be done also is to set up a gradle task to specifically mock prometheus (see e.g. [swatch-producer-aws](https://issues.redhat.com/browse/SWATCH-producer-aws)/build.gradle wiremock task definition). This may prove useful to isolate memory usage of our component vs. wiremock, by running in separate processes.

What I dislike about the Gradle task vs the JUnit 5 extension is that the stubs for the Gradle task seem to be static (configured once per all the tests). For the JUnit 5 extension, we can configure the stub per test scenario. The advantage is of course what you said about isolating memory usage.

I've just updated the extension, so at least use the same WIREMOCK_PORT to configure the port. 
Of course, if the isolating memory usage does really matter, we can change it in a later pull request (I would like to merge this pull request as it is, to finish off SWATCH-1574) :)

PR #2424, opened 2023-08-22
Fix deps so jandex 2.x exclusion works everywhere

kahowell:
There's some sort of bug where dependencies managed by the `io.spring.dependency-management` gradle plugin don't respect exclusions. Work around by moving select dependencies out of the dependency management section.

### DIFF
--- a/buildSrc/src/main/groovy/swatch.spring-boot-conventions.gradle
+++ b/buildSrc/src/main/groovy/swatch.spring-boot-conventions.gradle
@@ -12,8 +12,9 @@ dependencies {
     implementation "org.springframework.boot:spring-boot-devtools"
     // This starter pulls in Spring Boot versions that we don't want.  The actual classes that we need are
     // baked into the starter artifact itself which is peculiar but sometimes that's how the cookie crumbles.
-    implementation("org.jboss.resteasy:resteasy-servlet-spring-boot-starter") {
+    implementation(libraries["resteasy-spring-boot-starter"]) {
         exclude group: "org.springframework.boot"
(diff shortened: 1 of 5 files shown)


awood (commented):
This does fix the issue I was seeing.  But there is a lot messages like

```
Execution optimizations have been disabled for task ':clients:rhsm-client:test' to ensure correctness due to the followin
g reasons:                                                                                                               
  - Gradle detected a problem with the following location: '/home/awood/devel/rhsm-subscriptions/clients/rhsm-client/buil
d/resources/main'. Reason: Task ':clients:rhsm-client:test' uses this output of task ':clients:rhsm-client:jandex' withou
t declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what orde
r the tasks are executed. Please refer to https://docs.gradle.org/7.6/userguide/validation_problems.html#implicit_depende
ncy for more details about this problem.                                                                                 
Gradle detected a problem with the following location: '/home/awood/devel/rhsm-subscriptions/clients/rhsm-client/build/re
sources/main'. Reason: Task ':clients:rhsm-client:test' uses this output of task ':clients:rhsm-client:jandex' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. This behaviour has been deprecated and is scheduled to be removed in Gradle 8.0. Execution optimizations are disabled to ensure correctness. See https://docs.gradle.org/7.6/userguide/validation_problems.html#implicit_dependency for more details.
```

So I don't want to see us in a boat where Gradle upgrades are blocked.

awood (approved):
What do you think about #2426?

PR #2423, opened 2023-08-22
SWATCH-1239: Remove tag profile from Billable Usage process.

kflahert:
Jira issue: [SWATCH-1239](https://issues.redhat.com/browse/SWATCH-1239)

## Description

Removes references to tag_profile and use swatch-product-configuration instead.
-Replace with swatch-product-config
-Remove referrences to billingWindow since all subscriptions are monthly once we remove rhosak

### Setup

1. Insert test data:
```
INSERT INTO public.account_config VALUES ('account123', 'API', '2023-06-30 11:51:29.323066+00', '2023-06-30 14:05:38.597068+00', 'org123');

INSERT INTO public.contracts VALUES ('b3a5e8b7-6260-48f4-a9eb-63a584763ce9', '12914277', '2023-06-01 18:20:07.717275+00', '2023-06-01 18:20:07.717275+00', NULL, 'org123', 'MW02393', 'aws', 'aws123', 'rosa', 'bcwuzc8ww10vvxfair55mliy8');

INSERT INTO public.contract_metrics VALUES ('b3a5e8b7-6260-48f4-a9eb-63a584763ce9', 'control_plane', 10);
INSERT INTO public.contract_metrics VALUES ('b3a5e8b7-6260-48f4-a9eb-63a584763ce9', 'four_vcpu_hour', 10);


INSERT INTO public.events VALUES ('ce64995e-c30e-4efc-b745-52017a63161b', NULL, '2023-06-19 08:00:00+00', '{"sla": "Premium", "role": "moa-hostedcontrolplane", "org_id": "org123", "event_id": "ce64995e-c30e-4efc-b745-52017a63161b", "timestamp": "2023-06-19T08:00:00Z", "event_type": "snapshot_redhat.com:rosa:cluster_hour", "expiration": "2023-06-19T09:00:00Z", "instance_id": "67e89df3-1c74-4613-ae58-708def00fe74", "display_name": "automation_moa-hostedcontrolplane_cluster_67e89df3-1c74-4613-ae58-708def00fe74", "event_source": "prometheus", "measurements": [{"uom": "Instance-hours", "value": 100.0}], "service_type": "rosa Instance", "billing_provider": "aws", "billing_account_id": "aws123"}', 'snapshot_redhat.com:rosa:cluster_hour', 'prometheus', '67e89df3-1c74-4613-ae58-708def00fe74', 'org123');
[...]

PR #2426, opened 2023-08-22
Remove all non-BOM entries from dependency management

awood:
Having non-BOM entries in dependency management results in transitive dependency exclusions not working properly.  We achieve the intended effect of dependency management (uniform versions of a library across all sub-projects) with the dependencies.gradle file anyway.

Attn @kahowell

### DIFF
--- a/buildSrc/src/main/groovy/swatch.spring-boot-dependencies-conventions.gradle
+++ b/buildSrc/src/main/groovy/swatch.spring-boot-dependencies-conventions.gradle
@@ -1,10 +1,13 @@
 // Plugin: swatch.spring-boot-dependencies-conventions
-// uses spring dependency-management to apply spring boot dependencies bom and other constraints
-plugins {
-    id "io.spring.dependency-management"
+// uses enforcedPlatform-management to apply spring boot dependencies bom and other constraints
+repositories {
+    mavenCentral()
 }
 
 dependencies {
+    annotationProcessor enforcedPlatform(libraries["spring-boot-dependencies"])
+    implementation enforcedPlatform(libraries["spring-boot-dependencies"])
+
     // common testing deps, junit + mockito + hamcrest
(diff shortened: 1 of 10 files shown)


lindseyburnett:
related: https://github.com/RedHatInsights/rhsm-subscriptions/pull/2423/files#diff-81ca6ede6d5ce5bc9718e90eed7b8887c0b4ae7980ca0a8789d96075db246461R10

buildSrc/src/main/groovy/swatch.spring-boot-dependencies-conventions.gradle
```diff
@@ -14,29 +14,10 @@ dependencies {
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine"
 }
 
-// Define all versions here so that all projects will use the same versions.
 dependencyManagement {
     imports {
         // See https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-dependencies to
         // get listings of the contents of this BOM
         mavenBom libraries["spring-boot-dependencies"]
```
kahowell, line 21: I'd be interested to go ahead and try moving this dep to the `dependencies` section, using `enforcedPlatform` and dropping the spring dependency management plugin; if it all works out, let's just go ahead and do that here.

kahowell (changes requested):

kahowell (changes requested):
- [ ] `platform` was intentional for the libraries (just revert d4f0801) - because we don't want to force the spring boot services to use quarkus provided versions.
- [ ] We can remove `io.spring.gradle:dependency-management-plugin` from dependencies.gradle now.

awood:
@kahowell Thanks for the review.  The issues you caught should be fixed now.  Good to know about the "platform was intentional for the libraries" bit.  I added a commit with comments explaining that.

kahowell (approved):

PR #2428, opened 2023-08-23
SWATCH-1574: Use jackson streaming API to parse the response from prometheus

Sgitario:
Jira issue: [SWATCH-1574](https://issues.redhat.com/browse/SWATCH-1574)

## Description
When using a Prometheus dataset with up to 53760 metrics, we were holding up to 129 MB in memory. 
These changes use the Jackson streaming API to process the metrics individually, so it won't ever hold this amount of data. 

I made the following decisions:
- I used the Consumer Java API instead of a more complicated API like the Future API which would require creating threads that would lead to performance penalties.
- I didn't include the scenario to process a large dataset because the file size was too large (up to 20MB). 
- I didn't modify the generation of the Prometheus client API on purpose, because this must match with what the real Prometheus returns (speaking about the QueryResult generated class). Instead, I created a new QuerySummaryResult that we'll be using in the Prometheus service. 
- Ideally, I wanted to use directly the Response input stream from the Prometheus response, however the generated Prometheus client by OpenAPI only gave us two options to get the Prometheus response: either an array of bytes or a File. I chose the File over the array of bytes to avoid holding the amount of data of the bytes in memory. 

## Testing
For testing, I modified the new file `prometheus_small.json` to add 53760 metrics and confirm that the memory heap of the running process is not altered when processing these metrics.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusService.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusService.java
@@ -20,16 +20,33 @@
  */
 package org.candlepin.subscriptions.metering.service.prometheus;
 
+import static org.candlepin.subscriptions.prometheus.model.QueryResult.JSON_PROPERTY_DATA;
+import static org.candlepin.subscriptions.prometheus.model.QueryResult.JSON_PROPERTY_ERROR;
+import static org.candlepin.subscriptions.prometheus.model.QueryResult.JSON_PROPERTY_ERROR_TYPE;
+import static org.candlepin.subscriptions.prometheus.model.QueryResult.JSON_PROPERTY_STATUS;
+import static org.candlepin.subscriptions.prometheus.model.QueryResultData.JSON_PROPERTY_RESULT;
+import static org.candlepin.subscriptions.prometheus.model.QueryResultData.JSON_PROPERTY_RESULT_TYPE;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.net.UrlEscapers;
+import java.io.File;
+import java.io.IOException;
 import java.time.OffsetDateTime;
+import java.util.function.Consumer;
+import java.util.function.Function;
 import org.candlepin.subscriptions.exception.ErrorCode;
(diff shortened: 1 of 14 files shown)


src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusService.java
```diff
@@ -90,4 +124,73 @@ private String formatErrorMessage(ApiException apie) {
...
+      }
+    } catch (IOException ex) {
+      throw new ExternalServiceException(
+          ErrorCode.REQUEST_PROCESSING_ERROR, "Error parsing the Prometheus response", ex);
+    }
+
+    return builder.build();
+  }
+
+  private void parseQueryData(
```
kahowell, line 152: For better clarity:

```suggestion
  /**
   * Parse the data property. See "data section of the query result" at <a
   * href="https://prometheus.io/docs/prometheus/latest/querying/api/">Prometheus HTTP API</a>
   */
  private void parseQueryData(
```

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusService.java
```diff
@@ -90,4 +124,73 @@ private String formatErrorMessage(ApiException apie) {
     return String.format(
         "Prometheus API Error! CODE: %s MESSAGE: %s", apie.getCode(), apie.getMessage());
   }
+
+  private QuerySummaryResult parseQueryResult(
```
kahowell, line 128: For better clarity:

```suggestion
  /**
   * Parse the response. See <a
   * href="https://prometheus.io/docs/prometheus/latest/querying/api/#format-overview">Prometheus
   * HTTP API format overview</a>
   */
  private QuerySummaryResult parseQueryResult(
```

kahowell (changes requested):
This is very well written! I found it a little hard to follow initially, but it helped to refer back to the prometheus docs as I was reviewing. We should add some code comments that link back to particular sections so that anyone else in the future can parse it more quickly. (I left some inline suggestions accordingly). Also, I'd like to see the stub file somewhere outside of `product-stub-data`.

src/main/resources/product-stub-data/prometheus-success.json
```diff
@@ -0,0 +1 @@
+{ "status": "success" }
```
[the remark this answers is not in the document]
Sgitario: Done

Sgitario:
PR updated after changes from review.

kahowell (approved):

PR #2434, opened 2023-08-24
Fix time format in prometheus wiremock stub

kahowell:
Without this, sometimes the stub fails to match when it should, e.g.

```
Stub:   2023-08-24T12:26:41.963482680Z
Actual: 2023-08-24T12:26:41.96348268Z
```


kahowell:
Self-merging:
* unit tests pass with this
* change is trivial
* other test failures are unrelated

PR #2439, opened 2023-08-25
Configure users as superusers

Sgitario:
This is required to load scripts via pqsl using the COPY command.


lindseyburnett (approved):

kahowell:
Merging, a transient failure on ci.ext check reporting prevented it from being reported as green :-)

PR #2463, opened 2023-08-31
[CI] Do not ask "ok to verify?" for some known users

Sgitario:
The idea is to register the users that more contribute to the project and not to ask "ok to verify?" to trigger the Jenkins job.

I also checked whether there was a way to check the user organization instead, but I could not find any way.

I took the list of users from the swatch-dev-team slack group and the ones with more contributions.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,8 +28,28 @@ pipeline {
                 expression { env.CHANGE_FORK }
                 not {
                   anyOf {
+                    // Kevin Howell
+                    changeRequest author: "kahowell"
+                    // Lindsey Burnett
+                    changeRequest author: "lindseyburnett"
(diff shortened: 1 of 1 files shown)


lindseyburnett:
Need to add Vanessa too.

Vanessa Busch
vbusch

Thinking we should probably add our QE friends too?  Thoughts @kahowell ?

Sgitario:
> Need to add Vanessa too.
> 
> Vanessa Busch vbusch
> 
> Thinking we should probably add our QE friends too? Thoughts @kahowell ?

Vanessa added!

For QE and more users, the idea is to maintain the list from now on.

lindseyburnett (approved):

kahowell (approved):

PR #2489, opened 2023-09-05
SWATCH-1651: Populate uri label in Spring Boot 3 request metrics

kahowell:
Jira issue: [SWATCH-1651](https://issues.redhat.com/browse/SWATCH-1651)

Description
===========

Spring Boot 3 introduced drastic changes to how the metrics are constructed.

Two classes are now necessary to carry resteasy URIs into metrics:

1. ResteasyUriExtractionFilter - to extract the URI from resteasy
2. ResteasyRequestObservationConvention - to set the URI as the `uri` attribute.

ResteasyUriExtractionFilter was renamed from MicrometerUriHackFilter to better reflect its purpose.

Testing
=======

Setup
-----

1. Deploy the service:

```shell
```

2. Ensure opted in:

```shell
  x-rh-identity:$(echo '{"identity":{"type":"User","internal":{"org_id":"org123"}}}' | base64 -w0)
```

Verification
------------

Hit an http API:

```shell
http ':8000/api/rhsm-subscriptions/v1/capacity/products/RHEL for x86' \
  granularity==Daily \
  beginning==2023-07-21T17:32:28Z \
  ending==2023-07-22T17:32:28Z \
  x-rh-identity:$(echo '{"identity":{"type":"User","internal":{"org_id":"org123"}}}' | base64 -w0)
```

Check metrics endpoint and see that the URI is populated:

```shell
http :9000/metrics | grep http_server_requests_seconds_count
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resteasy/ResteasyRequestObservationConvention.java
+++ b/src/main/java/org/candlepin/subscriptions/resteasy/ResteasyRequestObservationConvention.java
@@ -21,6 +21,7 @@
 package org.candlepin.subscriptions.resteasy;
 
 import io.micrometer.common.KeyValue;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.server.observation.DefaultServerRequestObservationConvention;
 import org.springframework.http.server.observation.ServerHttpObservationDocumentation.LowCardinalityKeyNames;
 import org.springframework.http.server.observation.ServerRequestObservationContext;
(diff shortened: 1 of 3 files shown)


Sgitario (changes requested):
I wrote a test to verify this change. Are you ok if I submit a commit with the test?
Also, I've found that when the call fails, the URI is still "UNKNOWN": 

```
http_server_requests_seconds_count{error="none",exception="none",method="PUT",outcome="SUCCESS",status="200",uri="/api/rhsm-subscriptions/v1/opt-in",} 1.0
http_server_requests_seconds_count{error="none",exception="none",method="PUT",outcome="CLIENT_ERROR",status="401",uri="UNKNOWN",} 1.0
```

Note that when the call works fine, the URI is populated, but now when there is a failure. 

I'm seeing how to address it.

kahowell:
> Let me know what you think about and I'm ok to update your branch to include these commits into this pull request.

@Sgitario I'm happy with those changes, feel free to commit them directly to this branch.

Sgitario:
> > Let me know what you think about and I'm ok to update your branch to include these commits into this pull request.
> 
> @Sgitario I'm happy with those changes, feel free to commit them directly to this branch.

Updated!

Sgitario:
PR updated to fix the sonar issue.

Sgitario (approved):

PR #2456, opened 2023-08-30
SWATCH-573/SWATCH-1575: Refactor swatch-metrics to produce events to kafka

Sgitario:
Jira issue: [SWATCH-573](https://issues.redhat.com/browse/SWATCH-573)
Jira issue: [SWATCH-1575](https://issues.redhat.com/browse/SWATCH-1575)

## Description
From these changes, the Swatch Metrics service will start sending events to a Kafka topic (see Readme) instead of writing directly into the database.

Note that SWATCH-573 and SWATCH-1575 are closely related with each other, that's why I provided a single pull request with separate commits that address both tickets. 

## Testing
1.- podman-compose up
2.- createdb -h localhost -U postgres -O rhsm-subscriptions swatch-1573
3.- DATABASE_DATABASE=swatch-1573 ./gradlew :bootRun
4.- Once the migration is finished, stop all the applications that started in step 3.
5.- Insert mock data to verify all scenarios:

Connect to the database:
`psql -h localhost -U rhsm-subscriptions swatch-1573`

Create an existing event for orgId=16790890 and timestamp from 2022-08-07T07:00:00Z.

`INSERT INTO public.events(id,timestamp,event_type,event_source,instance_id,org_id) VALUES ('b934c72e-f179-4d75-b589-4f703fb49812','2022-08-07T07:00:00Z','snapshot_rosa_instance-hours','prometheus','76633fe9-0e9e-4737-9eb1-4fcba7ad5202','16790890');`

6.- Mock the prometheus server at localhost:

Create the stub directory:
`mkdir stub`

Add a stub file. Copy the following content into the file `stub/__files/swatch-1573.json`

```json
{
    "status" : "success",
    "data" : {
      "resultType" : "matrix",
      "result" : [
        {
          "metric" : {
            "_id" : "cluster id",
            "support" : "sla",
            "usage" : "usage",
            "product" : "product",
            "billing_marketplace" : "billing_marketplace",
            "billing_marketplace_account" : "billing_marketplace_account",
            "ebs_account" : "account"
          },
          "values": [[ 1693404431, "1" ]]
        }
      ]
    }
}
```

Note that the value 1693404431 is quite important because it's the event timestamp that we'll use. 

Run wiremock:
`podman run -it --rm -p 8101:8080 --name wiremock -v $PWD/stub:/home/wiremock wiremock/wiremock:2.32.0`

Configure stubbing for prometheus:

[...]

PR #2494, opened 2023-09-06
SWATCH-1547 Refactor event processing for hourly tally

kartikshahc:
Jira issue: [SWATCH-1547](https://issues.redhat.com/browse/SWATCH-1547)

## Description

** *Assuming that all the events in the event table have been tallied, this PR will recalculate for that month when it is deployed for the first time.*
  

> Since we have a bitemporal pattern, the "timestamp" and "actual_date" means the same.  We won't change the column name but, I am using this terminology just clarify how we see data.
>   Accordingly, "record_date" refers to the date when we entered the activity into our system.
> 
>   For reference: https://martinfowler.com/articles/bitemporal-history.html#TheTwoDimensions

Previous logic, we used to check current date range and compared it with max lastseen in hosts table. But the cost management can send data past 1 hour, 1 day, 1 week or a month in the current hour hence there have been a few adjustments made to the process of retrieving events for the hourly tally and then doing calculations based on that.

There is a slight change in the prior logic, thus in the new logic we perform the following. 

- We first fetch the events that are inside the hour according to the `record_date`. Since the timestamp and record_date could differ, we have to do some extra steps and take min(timestamp) for comparison. 
- Next, we compare the max (last_seen) value from the host table to the min(timestamp). `if(minLastSeenTimestamp.isEmpty()|| !firstEventTimestampInRange.get().isAfter(minLastSeenTimestamp.get()))`
- If the min.event.timestamp <= max.host.lastseen then recalc for whole month otherwise we take the current hour range.
  - If the min.event.timestamp is in last month then we extend our search to current month -1.
- The events are taken based on event timestamp and sorted by timestamp if we opt to recalculate or not. This is due to the possibility that the events might be out of chronological sequence or possibly take place in the past. In this case we can't rely on record_date.

**1. Previous logic:**

**Case1: Tally first time with no host record but, in event table**
- When: 
    User Range: `10-10-2023 1:00pm to 10-10-2023 2:00pm`
    Event: `timestamp 10-10-2023 1:00pm`

- Outcome:    
As lastseen in hosts may be at 6am on 10/10/2023. Lastseen is before the startdate range of 1:00 PM on 10/10/2023. So there is no need for recalculation or time adjustment 10-10-2023 1:00pm to 10-10-2023 2:00pm
    Host table now have 10-10-2023 1:00pm as lastseen entry

**Case2: Retally for the same record with Host record and in event table**
- When:
    User Range: `10-10-2023 1:00pm to 10-10-2023 2:00pm`
    Event: `timestamp 10-10-2023 1:00pm`
    Host table have 10-10-2023 1:00pm as lastseen

- Outcome:    
  Since we have already seen the event once, lastseen in host is 10-10-2023 1:00pm. Since max lastseen is equal to startdate range 10-10-2023 1:00pm. So we will recalc and we will adjust from first day of month to the end.

**2. New logic with the same cases and additional scenarios of unordered data:**

**Case1: Tally first time with no host record but, in event table**
- When
Range: `10-10-2023 1:00pm to 10-10-2023 2:00pm`
Event: `timestamp 10-05-2023 10:00pm, record_date= 10-10-2023 1:30pm`
we take the first event for the given `service_type` in events table and get its `min timestamp`= 10-10-2023 10:00pm. There could be two cases here : 
  - If the last_seen is null or last_seen is equal or after `min(timestamp)` then we do calculation from 10-01-2023 to current hour
  - If the last_seen is before `min(timestamp)` then we take range from min(timestamp) to end range. Assume this is the case then host will update with `lastseen 10-05-2023 10:00pm`

**Case2: Retally for the same record with Host record and in event table**
- When
Range: `10-10-2023 1:00pm to 10-10-2023 2:00pm`
Event: `timestamp 10-05-2023 10:00pm, record_date= 10-10-2023 1:30pm
Host: `lastseen 10-05-2023 10:00pm`

- Outcome
Since the last_seen is equal `min(timestamp)` then we do calculation from 10-01-2023 to current hour

**Case3: Tally for same hour but new record in event table and with a Host record**
- When
Range: `10-10-2023 1:00pm to 10-10-2023 2:00pm`
Event has two rows: 
   - timestamp 10-05-2023 10:00pm, record_date= 10-10-2023 1:30pm
   - timestamp 10-06-2023 10:00pm, record_date= 10-10-2023 1:45pm
Host: `lastseen 10-05-2023 10:00pm`

- Outcome
 Since the last_seen is before `min(timestamp)` then we take range from min(timestamp) to end range.
Update to host table with last_seen as 10-06-2023 10:00pm

### Setup
Execute: `RHSM_SUBSCRIPTIONS_ENABLE_SYNCHRONOUS_OPERATIONS=true DEV_MODE=true PROM_URL=http://localhost:8082/api/metrics/v1/telemeter/api/v1 ./gradlew clean :bootRun`

There are four ways to create events: fetch Prometheus events; cost management events using Kafka; internal API; and manually creating `cost events` using inserts. I've chosen to manually create events for testing purposes, but I've also written down Prometheus and cost integration methods that you may use as well.
Since these are manually created cost events, I have the `"service_type": "RHEL System"`. If you prefer to have that then I have added a new product `rhel-for-x86-eus-payg` which will be used for cost data. 

**1. Prometheus event creation**
- To run locally prometheus by pointing to stage follow the steps in: [Connecting  Observatorium to fetch Prometheus ](https://docs.engineering.redhat.com/display/ENT/Connecting+to+Observatorium+to+Fetch+Prometheus+Metrics)
- Copy CLIENT_SECRET and paste it in a file `token-refresher`
- Run: `podman run --name=telemeter-stage --rm -ti -p 8082:8080     quay.io/observatorium/token-refresher:master-2021-02-05-5da9663     --oidc.client-secret=$(cat ~/Downloads/token-refresher)     --oidc.client-id=observatorium-subwatch-staging     --oidc.issuer-url=https://sso.redhat.com/auth/realms/redhat-external     --url=https://observatorium.api.stage.openshift.com`
- Apply the below patch in `application-openshift-metering-worker.yaml` since billing_model="marketplace" won't give results in prometheus in the further steps
```

diff --git a/src/main/resources/application-openshift-metering-worker.yaml b/src/main/resources/application-openshift-metering-worker.yaml
--- a/src/main/resources/application-openshift-metering-worker.yaml	(revision f53e559df23efbc7ba6b5028b1db2f26e3f29570)
+++ b/src/main/resources/application-openshift-metering-worker.yaml	(date 1692902103848)
@@ -12,7 +12,7 @@
           default: >-
             #{metric.prometheus.queryParams[metric]}
             * on(_id) group_right
-            min_over_time(#{metric.prometheus.queryParams[metadataMetric]}{product="#{metric.prometheus.queryParams[product]}", external_organization="#{runtime[orgId]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None"}[1h])
+            min_over_time(#{metric.prometheus.queryParams[metadataMetric]}{product="#{metric.prometheus.queryParams[product]}", external_organization="#{runtime[orgId]}", support=~"Premium|Standard|Self-Support|None"}[1h])
           addonSamples: >-
             #{metric.prometheus.queryParams[metric]}
             * on(_id) group_right
```

- Run: `curl -X POST -H "x-rh-swatch-synchronous-request:true"  -H "x-rh-swatch-psk: placeholder" "http://localhost:8000/api/rhsm-subscriptions/v1/internal/metering/OpenShift-dedicated-metrics?orgId=6323660&rangeInMinutes=1200"`

**2. Cost management Kafka**
Produce Kafka event using Big Data Tool for :`platform.rhsm-subscriptions.service-instance-ingress`
`{"event_id": "06893602-d1f9-414d-8bf6-a90154e844be", "event_source": "cost-management", "event_type": "Snapshot", "account_number": "10001", "org_id": "1234567", "service_type": "RHEL System", "instance_id": "i-55555556", "timestamp": "2023-08-01T07:00:00Z", "expiration": "2023-08-01T08:00:00Z", "measurements": [{"value": "2", "uom": "vCPUs"}], "cloud_provider": "AWS", "hardware_type": "Cloud", "product_ids": ["479", "70"], "role": "Red Hat Enterprise Linux Workstation", "sla": "Self-Support", "usage": "Disaster Recovery", "billing_provider": "aws", "billing_account_id": "9999999999999"}`

**Notice**: 
- Initially, the host table might have a record or not depending on previously tallied or not. We assume its empty for now.

### Steps 1

1.`INSERT INTO public.events (event_id, account_number, timestamp, data, event_type, event_source, instance_id, org_id, span_id, metering_batch_id, record_date) VALUES ('16893602-d1f7-414d-8bf6-a90154e844be', '10001', '2023-09-01 06:00:00.000000 +00:00', '{"sla": "Self-Support", "role": "Red Hat Enterprise Linux Workstation", "usage": "Disaster Recovery", "org_id": "1234567", "event_id": "16893602-d1f8-414d-8bf6-a90154e844be", "timestamp": "2023-09-01T06:00:00Z", "event_type": "Snapshot2", "expiration": "2023-08-01T08:00:00Z", "instance_id": "i-55555558", "product_ids": ["479", "70"], "event_source": "cost-management", "measurements": [{"uom": "vCPUs", "value": 2.0}], "service_type": "RHEL System", "hardware_type": "Cloud", "account_number": "10001", "cloud_provider": "AWS", "billing_provider": "aws", "billing_account_id": "9999999999999"}', 'Snapshot2', 'cost-management', 'i-55555558', '1234567', null, null, '2023-09-01 06:00:00.000000 +00:00');`

2. `http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=1234567&start=2023-09-01T06:00Z&end=2023-09-01T07:00Z" x-rh-swatch-psk:placeholder`

### Verification 1

Since no host last seen so, outcome month to end. Update monthly total

**Logs:**
```
2023-09-15T20:07:57.157+0000 [thread=rhsm-subscriptions-task-processor-0-C-1] [level=INFO] [category=org.candlepin.subscriptions.tally.MetricUsageCollector]  - We appear to be retallying; adjusting start and end from [2023-09-01T06:00:00Z : 2023-09-01T07:00:00Z] to [2023-09-01T00:00Z : 2023-09-15T20:59:59.999999999Z]
989cf938-0b37-468e-b466-d4c671732164,,10001,,1234567,i-55555558,,false,,CLOUD,,2023-09-01 06:00:00.000000 +00:00,false,false,AWS,RHEL System,i-55555558,aws,9999999999999
```

### Steps 2

1.` http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=1234567&start=2023-09-01T06:00Z&end=2023-09-01T07:00Z" x-rh-swatch-psk:placeholder`

### Verification 2

Since we retallied, outcome month to end same values in Host
**Logs:**
`2023-09-15T20:13:22.204+0000 [thread=rhsm-subscriptions-task-processor-0-C-1] [level=INFO] [category=org.candlepin.subscriptions.tally.MetricUsageCollector]  - We appear to be retallying; adjusting start and end from [2023-09-01T06:00:00Z : 2023-09-01T07:00:00Z] to [2023-09-01T00:00Z : 2023-09-15T20:59:59.999999999Z]`

### Steps 3

1.`INSERT INTO public.events (event_id, account_number, timestamp, data, event_type, event_source, instance_id, org_id, span_id, metering_batch_id, record_date) VALUES ('26893602-d1f9-414d-8bf6-a90154e844be', '10001', '2023-08-29 07:00:00.000000 +00:00', '{"sla": "Self-Support", "role": "Red Hat Enterprise Linux Workstation", "usage": "Disaster Recovery", "org_id": "1234567", "event_id": "26893602-d1f9-414d-8bf6-a90154e844be", "timestamp": "2023-08-29T07:00:00Z", "event_type": "Snapshot", "expiration": "2023-08-01T08:00:00Z", "instance_id": "i-55555558", "product_ids": ["479", "70"], "event_source": "cost-management", "measurements": [{"uom": "vCPUs", "value": 4.0}], "service_type": "RHEL System", "hardware_type": "Cloud", "account_number": "10001", "cloud_provider": "AWS", "billing_provider": "aws", "billing_account_id": "9999999999999"}', 'Snapshot', 'cost-management', 'i-55555558', '1234567', null, null, '2023-09-01 07:00:00.000000 +00:00');`

2. `http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=1234567&start=2023-09-01T07:00Z&end=2023-09-01T08:00Z" x-rh-swatch-psk:placeholder`

### Verification 3
**Logs:**
`2023-09-15T20:19:12.200+0000 [thread=rhsm-subscriptions-task-processor-0-C-1] [level=INFO] [category=org.candlepin.subscriptions.tally.MetricUsageCollector]  - We appear to be retallying; adjusting start and end from [2023-09-01T07:00:00Z : 2023-09-01T08:00:00Z] to [2023-08-01T00:00Z : 2023-09-15T20:59:59.999999999Z]`

Update monthly total
```
989cf938-0b37-468e-b466-d4c671732164,2023-09,vCPUs,2,989cf938-0b37-468e-b466-d4c671732164
989cf938-0b37-468e-b466-d4c671732164,2023-08,vCPUs,4,989cf938-0b37-468e-b466-d4c671732164
```

### Steps 4

1.`INSERT INTO public.events (event_id, account_number, timestamp, data, event_type, event_source, instance_id, org_id, span_id, metering_batch_id, record_date) VALUES ('16893602-d1f9-414d-8bf6-a90154e844bf', '10001', '2023-09-01 08:00:00.000000 +00:00', '{"sla": "Self-Support", "role": "Red Hat Enterprise Linux Workstation", "usage": "Disaster Recovery", "org_id": "1234567", "event_id": "16893602-d1f9-414d-8bf6-a90154e844bf", "timestamp": "2023-09-01T08:00:00Z", "event_type": "Snapshot2", "expiration": "2023-08-01T08:00:00Z", "instance_id": "i-55555558", "product_ids": ["479", "70"], "event_source": "cost-management", "measurements": [{"uom": "vCPUs", "value": 3.0}], "service_type": "RHEL System", "hardware_type": "Cloud", "account_number": "10001", "cloud_provider": "AWS", "billing_provider": "aws", "billing_account_id": "9999999999999"}', 'Snapshot2', 'cost-management', 'i-55555558', '1234567', null, null, '2023-09-01 08:00:00.000000 +00:00');`

2. `http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=1234567&start=2023-09-01T08:00Z&end=2023-09-01T09:00Z" x-rh-swatch-psk:placeholder`

### Verification 4
Host table:
`989cf938-0b37-468e-b466-d4c671732164,,10001,,1234567,i-55555558,,false,,CLOUD,,2023-09-01 08:00:00.000000 +00:00,false,false,AWS,RHEL System,i-55555558,aws,9999999999999`

Update monthly_total:
```
989cf938-0b37-468e-b466-d4c671732164,2023-09,vCPUs,5,989cf938-0b37-468e-b466-d4c671732164
989cf938-0b37-468e-b466-d4c671732164,2023-08,vCPUs,4,989cf938-0b37-468e-b466-d4c671732164
```


Sgitario:
Can't we use Hibernate `@UpdateTimestamp` and `@CreationTimestamp` annotations for doing this? See more in https://thorben-janssen.com/persist-creation-update-timestamps-hibernate/. 

When using these annotations, Hibernate will take care of automatically updating these timestamps and theoretically, you don't need to implement any code changes, simply keep using the `saveAll` repository method.

kartikshahc:
> Can't we use Hibernate `@UpdateTimestamp` and `@CreationTimestamp` annotations for doing this? See more in https://thorben-janssen.com/persist-creation-update-timestamps-hibernate/.
> 
> When using these annotations, Hibernate will take care of automatically updating these timestamps and theoretically, you don't need to implement any code changes, simply keep using the `saveAll` repository method.

Sounds like a good idea! I will change to the annotations and test it before I commit.

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/EventRecord.java
```diff
@@ -91,8 +92,23 @@ public EventRecord(Event event) {
...
+  For reference: https://martinfowler.com/articles/bitemporal-history.html#TheTwoDimensions
+  */
+
   private OffsetDateTime timestamp;
 
+  @CreationTimestamp
+  @Column(name = "record_date")
+  private OffsetDateTime recordDate;
+
+  @Column(name = "update_date")
```
Sgitario, line 109: If we use the `@UpdateTimestamp` annotation, we would not need to add methods for doing so (speaking about `updateLastSeenTallyEvents`).

Sgitario (changes requested):
I think we should also use the annotation `@UpdateTimestamp` to keep the new column `update_date` up to date. 
This would save up to add a new method for doing so in the event record repository and also having to keep entity data in memory which we wanted to avoid after [SWATCH-573](https://github.com/RedHatInsights/rhsm-subscriptions/pull/2456) is merged.

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/EventRecord.java
kartikshahc: As we spoke.
Regarding the update timestamp, I have purposely not used that annotation. 
Let me clarify since we don't currently interact with event table for saving during hourly tally, hence the `save all` is not included in the hourly code flow. Previously, you saw the "save all" function included in my PR to handle the flow for saving events by Prometheus or cost management, making it beneficial the next time an hourly tally is done.
Secondly, we are not updating the `update_date` when we save events. That timestamp is like a flag that is used by hourly tally.

This is a marker solely used for hourly tally. I named the column since the Jira card had it, but we can change the name to something other like `last_seen_tally` or anything similar. The issue with the name last_seen_tally is that it will once more be mistaken with the host `last_seen` field. I don't mind changing it to `last_seen_tally` if that solves your confusion.

kartikshahc: @kahowell thoughts on keeping the same name `update_date` or changing it to some other name?

kartikshahc: We removed the use of update_date.

kahowell (approved):
I'm happy with the changes. Note I haven't evaluated the test steps.

swatch-product-configuration/src/main/resources/subscription_configs/RHEL/RHEL_for_x86_eus_payg.yaml
```diff
@@ -0,0 +1,28 @@
+---
+# Cleanup this file in SWATCH-1690 with accurate and more information
+platform: RHEL
+
+id: rhel-for-x86-eus-payg
```
Sgitario, line 5: if this file in this pull request on purpose? I don't see the relation between the PR and the JIRA ticket, to this new profile.

Sgitario (commented):
One question with some unexpected changes I saw in this pull request and a comment about renaming the migration file.

src/main/resources/liquibase/changelog.xml
```diff
@@ -128,5 +128,6 @@
     <!-- Uncomment once the above changeset has landed in prod -->
     <!-- <include file="liquibase/202309011416-drop-id-event-record.xml"/> -->
     <include file="liquibase/202309081416-add-span-id-event-record.xml"/>
+    <include file="liquibase/202309141256-add-recordate-column-events.xml"/>
```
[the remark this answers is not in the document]
kartikshahc: Yup. Thanks for catching the typo! Have made the change and also removed the use of wildcard in one of the test class.

kartikshahc:
> One question with some unexpected changes I saw in this pull request and a comment about renaming the migration file.
> 
> Moreover, I think this ticket is a great candidate for an integration test that covers the native queries and the verification steps, do you want me to help with this?

We can absolutely perform an integration test, but we should address it with a different card. I may coordinate with you separately and refer to your POC integration PR. In order to begin the end-to-end test on stage, we want to merge this in main.

swatch-product-configuration/src/main/resources/subscription_configs/RHEL/RHEL_for_x86_eus_payg.yaml
kartikshahc: This product configuration is for cost. There will be some placeholders until we receive information from outside services, but SWATCH-1690 will have some more correct information. Its one step ahead so, QE can test because we want to test against actual product rather than BASILISK test product.

Sgitario: But what is the point of adding part of SWATCH-1690 here? :/

kartikshahc: We don't want to utilize BASILISK product for testing here, for that reason. You can see that there are comments with description related to SWATCH-1690 and  PLACEHOLDERS. We are not adding SWATCH-1690 here.

kartikshahc: @Sgitario you can also check my test steps include `RHEL System` as `service_type` which is being sent by COST. For that reason we have this product for testing purposes.

kartikshahc:
@himmAllRight waiting for your final approval.

swatch-product-configuration/src/main/resources/subscription_configs/RHEL/RHEL_for_x86_eus_payg.yaml
Sgitario: I still don't get why we have partial solutions for other tickets like SWATCH-1690 in unrelated pull requests. 
But if @kahowell is ok with this, go for it.

src/test/java/org/candlepin/subscriptions/tally/TallySnapshotControllerIT.java
```diff
@@ -122,7 +122,8 @@ private void givenEventAtDay(int dayAt, double value) {
   }
 
   private void whenProduceHourlySnapshotsForOrg() {
-    controller.produceHourlySnapshotsForOrg(ORG_ID, new DateRange(start, end));
+    controller.produceHourlySnapshotsForOrg(
+        ORG_ID, new DateRange(clock.startOfCurrentHour(), clock.startOfCurrentHour().plusHours(1)));
```
Sgitario, line 126: I'm trying to reason for the motivation of this change. 
Given today is 19 September 05:00
And the following events in database:
- event with timestamp at 18 September 13:00 (automatically record_date is set to today: 19 September 05:00)
- event with timestamp at 18 September 13:15 (automatically record_date is set to today: 19 September 05:00)

When we produce the hourly snapshots for org range between 18 September 13:00 <-> 14:00
Then no data is generated because the previous events record_date was set to a recent date (today) 19 September 05:00.

Therefore, we must produce the hourly snapshots for today's date range. 

I'm glad that this integration test spotted this scenario which I think is definitely a change of behavior that needs to be noted in the release notes (but I don't know if you have this kind of information when releasing) @kahowell @barnabycourt

Where I think the new behavior is correct (here we're interested in producing data from events that have been recently inserted in our database - aka the `record_date`-, instead of using the `timestamp` - which is when the event was made in a third-party system like prometheus).

My only concern is that taking into account events might take a while to be consumed from the kafka topic, users seeing an event in Prometheus at a concrete timestamp and might want to produce the tally report for this event using a date range within this concrete timestamp, will see that no data will be generated. However, I don't think this endpoint was ever meant to be used for this purpose, so it should be ok. 

FYI @himmAllRight @ntkathole

Sgitario (approved):
Approving after understanding the motivation of the integration test change. @kartikshahc many thanks!

I also added a comment saying that I dislike adding partial solutions of other tickets in this unrelated pull request (even if it's for testing purposes), but up to @kahowell

swatch-product-configuration/src/main/resources/subscription_configs/RHEL/RHEL_for_x86_eus_payg.yaml
lindseyburnett: Not that anyone asked my opinion, but I'm fine with this product config file getting added here with placeholders.  We frequently keep cards around that end up being "verify we did this thing totally" because some of it gets done as part of something.

In general, writing stories that are truly standalone and testable is definitely an area we have an opportunity to improve upon.

lindseyburnett (approved):

swatch-product-configuration/src/main/resources/subscription_configs/RHEL/RHEL_for_x86_eus_payg.yaml
Sgitario: > Not that anyone asked my opinion, but I'm fine with this product config file getting added here with placeholders. We frequently keep cards around that end up being "verify we did this thing totally" because some of it gets done as part of something.

I would not say this is a good practice, but _whoever is without sin, cast the first stone_. :D

> In general, writing stories that are truly standalone and testable is definitely an area we have an opportunity to improve upon.

Of course!

kahowell:
> I also added a comment saying that I dislike adding partial solutions of other tickets in this unrelated pull request (even if it's for testing purposes), but up to @kahowell

I'm happy with it. It doesn't break existing functionality, and it'll enable earlier testing.

lindseyburnett:
> > I also added a comment saying that I dislike adding partial solutions of other tickets in this unrelated pull request (even if it's for testing purposes), but up to @kahowell
> 
> 
> 
> I'm happy with it. It doesn't break existing functionality, and it'll enable earlier testing.

+1 for supporting shift-left efforts 🙌

kartikshahc:
> > I also added a comment saying that I dislike adding partial solutions of other tickets in this unrelated pull request (even if it's for testing purposes), but up to @kahowell
> 
> I'm happy with it. It doesn't break existing functionality, and it'll enable earlier testing.

Yup that was my original intention. Thanks for supporting the idea! @kahowell @lindseyburnett

himmAllRight (approved):
LGTM. I spun up an EE from this PR and reproduced most of these cases in it. It seems to be following the new logic from what I can tell.

One thing to note is that I don't think the retally automation runs in EE, so that's something we might want to double check (and maybe improve on the QE side) once merged. But as far as merging goes, I think we're good 👍

PR #2503, opened 2023-09-08
SWATCH-1653: Tally should categorise hosts coming from HBI with cloud provider "gcp" as CLOUD

Sgitario:
Jira issue: [SWATCH-1653](https://issues.redhat.com/browse/SWATCH-1653)

## Description
Given a host in HBI with cloud_provider=gcp, after a tally for the org, the report for `category=cloud` should increase.

## Testing
1.- podman-compose up
2.- Create hosts for orgId=16790890 in Insights database

```
./bin/insert-mock-hosts --num-gcp 1 --hbi --clear --org 16790890 --account account123
```
3.- createdb -h localhost -U postgres -O rhsm-subscriptions swatch-1653
4.- DEV_MODE=true SUBSCRIPTION_SYNC_ENABLED=true ENABLE_SYNCHRONOUS_OPERATIONS=true DATABASE_DATABASE=swatch-1653 ./gradlew :bootRun
5.- Perform tally snapshot:

```bash
http PUT ":8000/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots/16790890" \
Origin:console.redhat.com \
x-rh-swatch-synchronous-request:true \
x-rh-swatch-psk:placeholder \
x-rh-identity:$(echo -n '{"identity":{"account_number":"","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"16790890"}}}' | base64 -w 0)
```

7.- Check hardware_type for generated host:

Connect to the database:
`psql -h localhost -U rhsm-subscriptions swatch-1653`

Create an existing event for orgId=16790890

`SELECT hardware_type FROM hosts;`

It should be CLOUD. Before these changes, it was PHYSICAL.

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurementType.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurementType.java
@@ -34,25 +34,46 @@ public enum HardwareMeasurementType {
   // integration
   @Deprecated
   AWS_CLOUDIGRADE,
-  GOOGLE,
+  GOOGLE("GCP"),
   ALIBABA,
   AZURE,
   EMPTY;
+
+  private final String[] aliases;
+
+  HardwareMeasurementType() {
+    this.aliases = new String[0];
+  }
+
+  HardwareMeasurementType(String... aliases) {
+    this.aliases = aliases;
+  }
 
   public static boolean isSupportedCloudProvider(String name) {
     if (name == null || name.isEmpty()) {
       return false;
     }
 
-    try {
-      return getCloudProviderTypes().contains(HardwareMeasurementType.valueOf(name.toUpperCase()));
-    } catch (IllegalArgumentException e) {
-      // Passed an invalid type string, consider it not supported.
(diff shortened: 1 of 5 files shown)


swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HardwareMeasurementType.java
```diff
@@ -34,25 +34,46 @@ public enum HardwareMeasurementType {
   // integration
   @Deprecated
   AWS_CLOUDIGRADE,
-  GOOGLE,
+  GOOGLE("GCP"),
```
Sgitario, line 37: Since `HardwareMeasurementType` is used in some entities like `Host`, if we rename `GOOGLE` with `GCP`, we would need to migrate the existing hosts. Moreover, for events, we do receive events with `Google` (see Event.CloudProvider`.

Therefore, I decided to instead of changing GOOGLE with GCP, use the concept of aliases. So, if `google` or `gcp` is used, it will be mapped as CLOUD.

kahowell (approved):

PR #2518, opened 2023-09-14
SWATCH-1734: Support integration tests with Postgresql

Sgitario:
Jira issue: [SWATCH-1734](https://issues.redhat.com/browse/SWATCH-1734)

## Description
These changes add a `BaseIT` class that allows writing integration tests using Postgresql and Kafka. Moreover, it brings some other utilities and testing facilities like integration between Spring Boot Security Testing and Awailitily.


kahowell (approved):
:100: This is wonderful, and I especially like that the integration tests run in parallel with the rest. We probably want/need to reconcile/consolidate test setup/locations later on, but I'm happy with this now.

PR #2552, opened 2023-09-22
Provide instructions about how to build the repository

Sgitario:
## Description
Update the CONTRIBUTING.md guide to explain how to build the repository and requirements.
Moreover, I included some sections for formatting and code style.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ Preferred Package Structure
 # Build
 
 Requirements to build the repository are:
-- Gradle 7.4+
+- Gradle
 - Java 17
 - Podman with socket listening (instructions [here](#podman-installation-instructions)) / or Docker installed
 
@@ -47,7 +47,7 @@ and set _Class count to use import with '\*'_ to `999`. Do the same with _Names 
 -------------
 # Formatting
 
-Before submit pull requests, remember to format your changes using:
+Before submitting pull requests, remember to format your changes using:
 
 ```shell
 ./gradlew spotlessApply
@@ -116,4 +116,10 @@ $ podman info | grep -A2 'remoteSocket'
 
 remoteSocket:
   path: /path/to/podman.sock
+```
+
+Finally, you need to configure the DOCKER_HOST environment property, so Testcontainers use it:
+
+```shell
+export DOCKER_HOST=unix:///run/user/${UID}/podman/podman.sock
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,9 +22,41 @@ Preferred Package Structure
     * Class names should have "Entity" suffix to differentiate them from their DTO counterparts
 * security
   * authentication, authorization, RBAC, etc.
-
-DB migrations
 -------------
+# Build
+
+Requirements to build the repository are:
+- Gradle 7.4+
+- Java 17
+- Podman with socket listening (instructions [here](#podman-installation-instructions)) / or Docker installed
+
+Once we have all the requirements properly installed and configured, we can build the repository by doing:
+
+```shell
+./gradle clean build
+```
+
+-------------
+# Code Style
+
+- Disable wildcard imports
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -74,7 +74,7 @@ preferred development workflow.
 
 ### Prerequisites
 
-First, ensure you have podman-compose, podman and java 11 installed:
+First, ensure you have podman-compose, podman and java 17 installed:
```
Sgitario, line 77: This section also mentions docker and docker-compose. This didn't work for me and I think we should remove all the mentions to docker (since we should go only for podman). wdyt?

CONTRIBUTING.md
```diff
@@ -22,9 +22,41 @@ Preferred Package Structure
...
+- Podman with socket listening (instructions [here](#podman-installation-instructions)) / or Docker installed
+
+Once we have all the requirements properly installed and configured, we can build the repository by doing:
+
+```shell
+./gradle clean build
+```
+
+-------------
+# Code Style
```
Sgitario, line 40: Any other rules that might come into your mind?

CONTRIBUTING.md
```diff
@@ -22,9 +22,41 @@ Preferred Package Structure
     * Class names should have "Entity" suffix to differentiate them from their DTO counterparts
 * security
   * authentication, authorization, RBAC, etc.
+-------------
+# Build
+
+Requirements to build the repository are:
+- Gradle 7.4+
```
Sgitario, line 29: Is this correct?

barnabycourt: Should we even be specifying? The recommendation is to use the gradlew scripts which will import the necessary gradle version automatically. I believe it is currently Gradle 8.3

Sgitario: Good point. I'm fine with removing the version then :)

CONTRIBUTING.md
```diff
@@ -22,9 +22,41 @@ Preferred Package Structure
...
+- Disable wildcard imports
+
+To disable it in the IntelliJ IDEA editor, navigate to _Editor_ -> _Code Style_ -> _Java_ -> _Imports_
+and set _Class count to use import with '\*'_ to `999`. Do the same with _Names count to use static import with '\*'_.
 
-DB migrations
 -------------
+# Formatting
+
+Before submit pull requests, remember to format your changes using:
```
kahowell, line 50: grammar:

```suggestion
Before submitting pull requests, remember to format your changes using:
```

CONTRIBUTING.md
```diff
@@ -43,11 +75,45 @@ Liquibase Hub or Liquibase Pro.
...
+systemctl --user enable podman.socket --now
+```
+
+In systems when systemd is not running (WSL etc), instead of the above command, you need to do:
+
+```shell
+podman system service --time=0
+```
+
+Then, you can obtain the path of the socket with the following command:
```
kahowell, line 112: We need to explain what to do with the socket (`export DOCKER_HOST=unix:///path/to/podman.sock`).

README.md
kahowell: Yeah, I think that's fine. FWIW, you can probably get docker to work by manually setting `DOCKER_HOST` (the gradle setup uses `DOCKER_HOST` as-is if specified, otherwise assumes rootless podman for the current user). I think it's better to standardize the setup on podman/podman-compose though.

kahowell (changes requested):
Very useful updates! See some small feedback inline.

README.md
Sgitario: Remove all the mentions to docker

CONTRIBUTING.md
Sgitario: Updated

CONTRIBUTING.md
Sgitario: Fixed

CONTRIBUTING.md
Sgitario: Added

kahowell (approved):

lindseyburnett (approved):

PR #2557, opened 2023-09-22
SWATCH-1717 & SWATCH-1718: Create metering-rhel and metering-rhel-job profiles, and define ClowdApps

lindseyburnett:
Jira issue: [SWATCH-1718](https://issues.redhat.com/browse/SWATCH-1718)

## Description

Defining a new ClowdApp to deploy another instance of swatch-metrics, named swatch-metrics-rhel, that connects to RHEL telemeter instance instead of openshift telemeter instance.  There is no need for a separate purge events job, so I omitted it from the swatch-metrics-rhel definition.

This includes creating separate profiles.  This is necessary to allow each deployment to bind to its dedicated kafka topic, and enable the deployments to scale independently.

The part that "points" to the rhel telemeter instance is actually environment variables on the token-refresher sidecar.  These environment variables are set via vault secrets.

## Testing

What we're looking to verify here is that the `swatch-metrics-rhel-*` pods deploy, start up using the appropriate spring profile, connect to the appropriate topic, and the token refresher container uses the swatch-metrics-rhel secret.

The only other thing we can try for verification is hitting the internal endpoint for metrics sync and confirm there are no http errors in the log.  Until there is a product configured in swatch that has data available regularly available in rhel telemeter, we won't actually be able to see swatch consume anything and save it to the database.

## checkout this branch
```bash
git checkout lburnett/metering-rhel-profiles
```

## reserve namespace & enable token refresher sidecar

```bash
export NAMESPACE=$(bonfire namespace reserve -f)

```

```
oc patch env env-$NAMESPACE -p '{"spec":{"providers":{"sidecars":{"tokenRefresher":{"enabled":true}}}}}' --type=merge
```

## deploy swatch-metrics component to namespace

Note about terminology:  "swatch-metrics" is still the name of the component, as defined by our resourceTemplate entry in app-interface.  Up until this point, our clowdapps & components were 1 to 1 in regards to naming.  Now, our "swatch-metrics" component contains two clowdapps - swatch-metrics and swatch-metrics-rhel.  Eventually (out of scope), the swatch-metrics components will be named "swatch-metrics-openshift" and "swatch-metrics-rhel".

This command assumes that the PR has already built and pushed a image to quay for this githash.

```bash
VER=$(git rev-parse --short=7 HEAD); bonfire deploy -n $NAMESPACE rhsm --optional-deps-method=none \
-C swatch-metrics \
-C swatch-tally \
--no-remove-resources=all \
--no-release-on-fail \
[...]

PR #2572, opened 2023-09-27
SWATCH-1745: Make "event_source" configurable when saving to the events

Sgitario:
Jira issue: [SWATCH-1745](https://issues.redhat.com/browse/SWATCH-1745)

## Description
The "event_source" column should be configured as an env var, set in the [clowdapp](https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/swatch-metrics/deploy/clowdapp.yaml#L181) file, and the configured value saves in the "events" table appropriately

## Testing
1.- podman-compose up
2.- Mock the prometheus server at localhost:

Create the stub directory:
`mkdir -p stub/__files`

Add a stub file to return some data: `stub/__files/swatch-1745.json`

```
cat > stub/__files/swatch-1745.json <<EOF
{
    "status" : "success",
    "data" : {
      "resultType" : "matrix",
      "result" : [
        {
          "metric" : {
            "_id" : "cluster id",
            "support" : "sla",
            "usage" : "usage",
            "product" : "product",
            "billing_marketplace" : "billing_marketplace",
            "billing_marketplace_account" : "billing_marketplace_account",
            "ebs_account" : "account"
          },
          "values": [[ $(date +%s), "1" ]]
        }
      ]
    }
}
EOF
```

Add a stub file to return no data: `stub/__files/empty.json`

```
cat > stub/__files/empty.json <<EOF
{
    "status" : "success",
    "data" : {
      "resultType" : "matrix",
      "result" : []
    }
}
EOF
```

Run wiremock:
`podman run -it --rm -p 8101:8080 --name wiremock -v $PWD/stub:/home/wiremock:z wiremock/wiremock:2.32.0 --verbose`

Configure stubbing for prometheus:

- this is to return some data when asking for instance-hours:
`curl -X POST --data '{ "priority": 1, "request": { "urlPath": "/api/v1/query_range", "method": "GET", "queryParameters": {"query": {"contains":"instance_hours"}}}, "response": { "status": 200, "bodyFileName": "swatch-1745.json", "headers":{"Content-Type":"application/json"} }}' http://localhost:8101/__admin/mappings/new`

- this is to return no data otherwise:
`curl -X POST --data '{ "priority": 5, "request": { "urlPath": "/api/v1/query_range", "method": "GET" }, "response": { "status": 200, "bodyFileName": "empty.json", "headers":{"Content-Type":"application/json"} }}' http://localhost:8101/__admin/mappings/new`

3.- Start the swatch metrics app: `EVENT_SOURCE=prometheus PROM_URL="http://localhost:8101/api/v1/" SUBSCRIPTION_USE_STUB=true USER_USE_STUB=true RHSM_RBAC_USE_STUB=true SUBSCRIPTION_SYNC_ENABLED=true ENABLE_SYNCHRONOUS_OPERATIONS=true SPRING_PROFILES_ACTIVE=openshift-metering-worker,kafka-queue ./gradlew :bootRun`

4.- Trigger the metrics from prometheus:
```
http POST :8000/api/rhsm-subscriptions/v1/internal/metering/rosa?orgId=16790890 \
Origin:console.redhat.com \
x-rh-swatch-synchronous-request:true \
x-rh-swatch-psk:placeholder \
x-rh-identity:$(echo -n '{"identity":{"account_number":"","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"16790890"}}}' | base64 -w 0)
```

5.- Confirm the events have been produced in the Kafka topic "platform.rhsm-subscriptions.service-instance-ingress".

You should see at least two events in http://localhost:3030/#/cluster/default/topic/n/platform.rhsm-subscriptions.service-instance-ingress/:

```
Key 16790890
Value
event_source prometheus
event_type snapshot_rosa_instance-hours
account_number account
org_id 16790890
service_type rosa Instance
instance_id cluster id
display_name cluster id
measurements [ [object Object] ]
billing_account_id billing_marketplace_account

Key 16790890
Value
event_source prometheus
event_type cleanup-snapshot_rosa_instance-hours
org_id 16790890
timestamp 2023-08-31T06:00:00Z
```

Note that both events should use "prometheus" as event_source, proving that the EVENT_SOURCE environment property works as expected and this field is configurable.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/event/EventController.java
+++ b/src/main/java/org/candlepin/subscriptions/event/EventController.java
@@ -50,7 +50,8 @@ import org.springframework.util.StringUtils;
 @Slf4j
 public class EventController {
 
-  private static final String PROMETHEUS = "prometheus";
+  private static final Set<String> EXCLUDE_LOG_FOR_EVENT_SOURCES =
+      Set.of("prometheus", "rhelemeter");
   private final EventRecordRepository repo;
   private final ObjectMapper objectMapper;
   private final OptInController optInController;
@@ -148,7 +149,7 @@ public class EventController {
         eventJson -> {
           try {
             BaseEvent baseEvent = objectMapper.readValue(eventJson, BaseEvent.class);
-            if (!PROMETHEUS.equals(baseEvent.getEventSource())) {
+            if (!EXCLUDE_LOG_FOR_EVENT_SOURCES.contains(baseEvent.getEventSource())) {
               log.info("Event processing in batch: " + baseEvent);
             }
 
--- a/src/main/java/org/candlepin/subscriptions/event/EventController.java
+++ b/src/main/java/org/candlepin/subscriptions/event/EventController.java
@@ -20,8 +20,6 @@
  */
 package org.candlepin.subscriptions.event;
 
-import static org.candlepin.subscriptions.metering.MeteringEventFactory.EVENT_SOURCE;
-
 import com.fasterxml.jackson.databind.ObjectMapper;
 import jakarta.transaction.Transactional;
 import java.time.OffsetDateTime;
@@ -51,6 +49,8 @@ import org.springframework.util.StringUtils;
 @Service
 @Slf4j
 public class EventController {
+
+  private static final String PROMETHEUS = "prometheus";
   private final EventRecordRepository repo;
   private final ObjectMapper objectMapper;
   private final OptInController optInController;
@@ -148,7 +148,7 @@ public class EventController {
         eventJson -> {
           try {
             BaseEvent baseEvent = objectMapper.readValue(eventJson, BaseEvent.class);
-            if (!EVENT_SOURCE.equals(baseEvent.getEventSource())) {
+            if (!PROMETHEUS.equals(baseEvent.getEventSource())) {
               log.info("Event processing in batch: " + baseEvent);
             }
 
--- a/src/main/resources/application-metering-job.yaml
+++ b/src/main/resources/application-metering-job.yaml
@@ -12,6 +12,5 @@ rhsm-subscriptions:
         jobBackOffMaxInterval: ${METERING_JOB_BACK_OFF_MAX_INTERVAL:50000}
         jobBackOffInitialInterval: ${METERING_JOB_BACK_OFF_INITIAL_INTERVAL:1000}
         backOffMultiplier: ${METERING_JOB_BACK_OFF_MULTIPLIER:1.5}
-        eventSource: ${EVENT_SOURCE}
     tasks:
       topic: ${METERING_TASK_TOPIC}
--- a/src/main/resources/application-metering-job.yaml
+++ b/src/main/resources/application-metering-job.yaml
@@ -12,5 +12,6 @@ rhsm-subscriptions:
         jobBackOffMaxInterval: ${METERING_JOB_BACK_OFF_MAX_INTERVAL:50000}
         jobBackOffInitialInterval: ${METERING_JOB_BACK_OFF_INITIAL_INTERVAL:1000}
         backOffMultiplier: ${METERING_JOB_BACK_OFF_MULTIPLIER:1.5}
+        eventSource: ${EVENT_SOURCE}
     tasks:
       topic: ${METERING_TASK_TOPIC}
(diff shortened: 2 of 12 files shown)


lindseyburnett:
Confirmed test steps and confirmed that when kafka messages are processed by swatch-tally, the "event_source" column in the events table populates as anticipated.

src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -51,6 +49,8 @@
 @Service
 @Slf4j
 public class EventController {
+
+  private static final String PROMETHEUS = "prometheus";
```
lindseyburnett, line 53:
```suggestion
  private static final String PROMETHEUS = "prometheus";
  private static final String RHELEMETER= "rhelemeter";
```

src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -148,7 +148,7 @@ private ServiceInstancesResult parseServiceInstancesResult(Set<String> eventJson
         eventJson -> {
           try {
             BaseEvent baseEvent = objectMapper.readValue(eventJson, BaseEvent.class);
-            if (!EVENT_SOURCE.equals(baseEvent.getEventSource())) {
+            if (!PROMETHEUS.equals(baseEvent.getEventSource())) {
```
lindseyburnett, line 151:
```suggestion
            if (!PROMETHEUS.equals(baseEvent.getEventSource()) && !RHELEMETER.equals(baseEvent.getEventSource())) {
```

src/main/resources/application-metering-job.yaml
```diff
@@ -12,5 +12,6 @@ rhsm-subscriptions:
         jobBackOffMaxInterval: ${METERING_JOB_BACK_OFF_MAX_INTERVAL:50000}
         jobBackOffInitialInterval: ${METERING_JOB_BACK_OFF_INITIAL_INTERVAL:1000}
         jobBackOffMultiplier: ${METERING_JOB_BACK_OFF_MULTIPLIER:1.5}
+        eventSource: ${EVENT_SOURCE}
```
lindseyburnett, line 15: Since the `swatch-metrics-sync` (uses this metering-job profile) doesn't produce messages to the `platform.rhsm-subscriptions.service-instance-ingress` topic, I don't think it needs to have an event_source yet since it won't be used.  Can also remove it from the job definition in the clowdapp.yaml file (lines 488)

swatch-metrics/deploy/clowdapp.yaml
```diff
@@ -285,6 +287,8 @@ objects:
                   secretKeyRef:
                     name: ${INVENTORY_SECRET_KEY_NAME}
                     key: ${INVENTORY_SECRET_KEY_NAME_PREFIX}password
+              - name: EVENT_SOURCE
+                value: ${EVENT_SOURCE}
```
lindseyburnett, line 291: Let's set this value directly to `prometheus` instead of parameterizing it in lines 96-97.  When the swatch-metrics-rhel clowdapp becomes part of this file, we'll set it directly as "rhelemeter" for that one.

lindseyburnett (changes requested):

src/main/resources/application-metering-job.yaml
Sgitario: Removed

swatch-metrics/deploy/clowdapp.yaml
Sgitario: +1 Done

src/main/java/org/candlepin/subscriptions/event/EventController.java
Sgitario: I used a collection to exclude these two event sources from being logged.

src/main/java/org/candlepin/subscriptions/event/EventController.java
Sgitario: Code updated.

Sgitario:
@lindseyburnett PR updated after your requested changes. Feel free to take another look.

lindseyburnett (approved):
Code & local testing LGTM.

Note that testing "rhelemeter" and "prometheus" event sources side by side for QE won't be available until swatch-metrics-rhel ClowdApp lands (PR #2557)

TrayvonMcKnight (approved):
https://drive.google.com/file/d/1XR8yTGa049sMQDVGkBS5ZeUXxff7yXea/view?usp=drive_link

https://drive.google.com/file/d/1Ir7acZSZzsSMfIJ2jivITFx0O6YBwnq1/view?usp=drive_link

PR #2574, opened 2023-09-27
Swatch-1691: event kafka retry

kflahert:
Jira issue: [SWATCH-1691](https://issues.redhat.com/browse/SWATCH-1691)

## Description

In order to handle failed Event kafka messages we need to retry and then if failing continues put message on a dead letter queue to be processed later.

We only put messages with valid json that can be parsed into an Event object but not able to be persisted on the dead letter topic.

Topic creation (topic had to be renamed to all lowercase): https://github.com/RedHatInsights/platform-mq/pull/305 and https://github.com/RedHatInsights/platform-mq/pull/307
### Setup

1. Clean up the Events table.
2. Save event that will be duplicated in messages and should fail on message consumption:
`INSERT INTO public.events VALUES ('b1208963-b92a-4c59-a730-88fd5fadb9b1', NULL, '2023-05-02 00:00:00+00', '{"sla": "Premium", "role": "osd", "org_id": "146", "event_id": "b1208963-b92a-4c59-a730-88fd5fadb9b1", "timestamp": "2023-05-02T00:00:00Z", "event_type": "snapshot_redhat.com:openshift_dedicated:cluster_hour", "expiration": "2023-05-02T01:00:00Z", "instance_id": "e3a62bd1-fd00-405c-9401-f2288808588d", "display_name": "automation_osd_cluster_e3a62bd1-fd00-405c-9401-f2288808588d", "event_source": "prometheus", "measurements": [{"uom": "Instance-hours", "value": 1.0}], "service_type": "OpenShift Cluster"}', 'snapshot_redhat.com:openshift_dedicated:cluster_hour', 'prometheus', 'e3a62bd1-fd00-405c-9401-f2288808588d', '146', DEFAULT, NULL, '2023-09-26 14:57:56.160104+00');`
3. Use the Big data Tools of Intellij as producer to test the consumer topic `platform.rhsm-subscriptions.service-instance-ingress`. Payloads are as below make sure to check each Test step on which payload to use:
image

Payload144: `{"sla": "Premium", "role": "osd", "org_id": "144","timestamp": "2023-05-02T00:00:00Z", "event_type": "snapshot_redhat.com:openshift_dedicated:cluster_hour", "expiration": "2023-05-02T01:00:00Z", "instance_id": "e3a62bd1-fd00-405c-9401-f2288808588d", "display_name": "automation_osd_cluster_e3a62bd1-fd00-405c-9401-f2288808588d", "event_source": "prometheus", "measurements": [{"uom": "Instance-hours", "value": 1.0}], "service_type": "OpenShift Cluster"}`

Payload145: `{"sla": "Premium1", "role": "osd", "org_id": "145","timestamp": "2023-05-02T00:00:00Z", "event_type": "snapshot_redhat.com:openshift_dedicated:cluster_hour", "expiration": "2023-05-02T01:00:00Z", "instance_id": "e3a62bd1-fd00-405c-9401-f2288808588d", "display_name": "automation_osd_cluster_e3a62bd1-fd00-405c-9401-f2288808588d", "event_source": "prometheus", "measurements": [{"uom": "Instance-hours", "value": 1.0}], "service_type": "OpenShift Cluster"}`

Payload146: `{"sla": "Premium", "role": "osd", "org_id": "146","timestamp": "2023-05-02T00:00:00Z", "event_type": "snapshot_redhat.com:openshift_dedicated:cluster_hour", "expiration": "2023-05-02T01:00:00Z", "instance_id": "e3a62bd1-fd00-405c-9401-f2288808588d", "display_name": "automation_osd_cluster_e3a62bd1-fd00-405c-9401-f2288808588d", "event_source": "prometheus", "measurements": [{"uom": "Instance-hours", "value": 1.0}], "service_type": "OpenShift Cluster"}`

### Steps

1. **Step 1: Add new events**
In the value string add payload 144, payload 145, payload 146 one by one in order and produce.
2. Start app with: `DEV_MODE=true ./gradlew clean :bootRun`

### Verification

1. One new entry in events table should exist for org_id 144 `select * from events where org_id = 144`
2. Check logs for `[INFO ] [org.candlepin.subscriptions.tally.billing.ServiceInstanceMessageConsumer] - Events received w/ event list size=3. Consuming events.`
3. Should error out and retry
4. You should also see a log where we get an event list size=1 since that is the org_id 146 record being retried since it failed to save.
5. Use big data tool to consume messages from platform.rhsm-subscriptions.service-instance-ingress.DLT which should contain the org_id 146 event.


src/test/java/org/candlepin/subscriptions/event/EventControllerTest.java
```diff
@@ -185,12 +185,12 @@ void setup() {
...
   void testPersistServiceInstances_WhenValidPayload() {
 
-    Set<String> eventRecords = new HashSet<>();
+    List<String> eventRecords = new ArrayList<>();
     eventRecords.add(eventRecord1);
     eventRecords.add(eventRecord2);
     eventRecords.add(eventRecord3);
     eventController.persistServiceInstances(eventRecords);
-    verify(optInController, times(2)).optInByOrgId(any(), any());
+    verify(optInController, times(3)).optInByOrgId(any(), any());
```
kflahert, line 193: this is a result of changing Set to List. The duplicates were getting removed earlier in the process so opt in was called one less time.

swatch-core/src/main/java/org/candlepin/subscriptions/task/TaskQueueProperties.java
```diff
@@ -41,4 +41,8 @@ public class TaskQueueProperties {
 
   /** Batch size number of records * */
   private String maxPollRecords = "500";
+
+  private int retryAttempts = 2;
```
kartikshahc, line 45: We should retry only once. :)

swatch-core/src/main/java/org/candlepin/subscriptions/task/TaskQueueProperties.java
```diff
@@ -41,4 +41,8 @@ public class TaskQueueProperties {
...
+
+  private long retryBackOffMillis = 5000L;
```
kartikshahc, line 47: Since we will retry only once may be we can remove retryBackOffMillis?
Also I assume the default retention of a Kafka topic is one week. I am wondering if for DLT we should reduce it if not consumed?

kartikshahc (commented):
Overall looks good to me. I tested and verified the steps. As we discussed yesterday, we should look to see if we can still use Set instead of List. If DLT doesn't support it then we can think about it. Also left few small comments.

kartikshahc (changes requested):
As mentioned in ^ comment property value change for retry.

swatch-core/src/main/java/org/candlepin/subscriptions/task/TaskQueueProperties.java
kflahert: I set the backoff to 0 just in case we want to override later. One week sounds ok to me for it.

src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -128,63 +146,124 @@ public Optional<OffsetDateTime> findFirstEventTimestampInRange(
...
+    return result;
+  }
+
+  /**
+   * Deduplicate eventJsonList while preserving indexes of events in batch
+   *
+   * @param eventJsonList
+   */
+  private Map<String, Integer> mapEventsToBatchIndex(List<String> eventJsonList) {
+    var result = new HashMap<String, Integer>();
```
kflahert, line 263: @kartikshahc I don't think I could get it to work with replacing the list with a set because I wanted to skip the records that failed to parse and we need to use the index for that. This method deduplicates the data upfront instead and keeps track of what index it was at for retrying.

kartikshahc (approved):
LGTM

PR #2575, opened 2023-09-27
Add empty template for swatch-producer-azure

kahowell:
So that CI can "deploy" swatch-producer-azure.

This will fix bonfire deploy failures in recent PR checks.


kahowell:
self-merging since the test failure is unrelated, and the new template is just a placeholder

PR #2582, opened 2023-09-29
SWATCH-1592 and SWATCH-1008: Refactor swatch-metrics jobs to use curl

Sgitario:
Jira issue: [SWATCH-1008](https://issues.redhat.com/browse/SWATCH-1008)
Jira issue: [SWATCH-1592](https://issues.redhat.com/browse/SWATCH-1592)

## Description
For all cronjobs defined in clowdapps, we should migrate to using k8s cronjobs w/ curl against internal endpoints.

## Testing
There are two cronjobs to verify here:
- The cronjob `swatch-metrics-sync` (old `metering-job` profile) which uses the environment property `METERING_SCHEDULE`.
- The cronjob `swatch-metrics-rhel-sync` (old `metrics-rhel-job` profile) which uses the environment property `RHEL_METERING_SCHEDULE`.

Both jobs were scheduled to be run every 30 min.

By default, the requests to prometheus do not work. To fix it, you need to run the following command against your ephemeral namespace:

```
oc patch env env-$NAMESPACE -p '{"spec":{"providers":{"sidecars":{"tokenRefresher":{"enabled":true}}}}}' --type=merge
```

Then, for each job, you need to wait the 30 min that each one is configured for.

And then, you should see a job instance for each cronjob with:

```
% Total % Received % Xferd Average Speed Time Time Time Current
Dload Upload Total Spent Left Speed
0 0 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0 0 0 0 0 0 0 0 0 --:--:-- 0:00:01 --:--:-- 0 0 0 0 0 0 0 0 0 --:--:-- 0:00:02 --:--:-- 0 0 0 0 0 0 0 0 0 --:--:-- 0:00:02 --:--:-- 0
```

And in the logs of the service `swatch-metrics` or `swatch-metrics-rhel`:

```
2023-09-29 11:30:04,160 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.security.LogPrincipalFilter] self- Internal API: /api/rhsm-subscriptions/v1/internal/metering/sync requested by user: self
2023-09-29 11:30:04,386 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.task.PrometheusMetricsTaskManager] self- Queuing OpenShift-dedicated-metrics metric updates for range: [2023-09-29T10:00Z, 2023-09-29T11:00Z)
2023-09-29 11:30:04,407 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusService] self- Fetching metrics from prometheus: 2023-09-29T11:00Z -> 2023-09-29T11:00Z [Step: 3600]
2023-09-29 11:30:05,538 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusService] self- Fetching metrics from prometheus: 2023-09-29T11:00Z -> 2023-09-29T11:00Z [Step: 3600]
2023-09-29 11:30:07,041 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusService] self- Fetching metrics from prometheus: 2023-09-29T11:00Z -> 2023-09-29T11:00Z [Step: 3600]
```

So, the call was made and we confirmed that the service processed it.


lindseyburnett:
It looks like you took care of a blocker card for this, https://issues.redhat.com/browse/SWATCH-1592.  Is that right?  Just want to make sure that card gets updated in Jira accordingly.

Also, a note about this part of the description:

> Note that prometheus is not configured in the ephemeral namespaces, so it's expected both calls to fail.

If you enable the token-refresher sidecar in the ephemeral environment you're testing with, I would expect the calls to be successful.  There might not be any data returned, but the calls themselves shouldn't fail.

Once you've reserved a namespace in ephemeral with bonfire, you can patch the ClowdEnvironment with this command:

`oc patch env env-$NAMESPACE -p '{"spec":{"providers":{"sidecars":{"tokenRefresher":{"enabled":true}}}}}' --type=merge`

src/main/spec/internal-metering-api-spec.yaml
```diff
@@ -63,7 +76,7 @@ paths:
             type: integer
             minimum: 0
             description: "Defines the amount of time (in minutes) that will be used to calculate a metric query's start date based on a given end date.
-For example, given an end date of 2021-01-06T00:00:00Z and a rangeInMinutes of 60, the calculated start date will be: 2021-01-05T23:00:00Z.  Default (60) is store in application-metering-job.yaml"
+For example, given an end date of 2021-01-06T00:00:00Z and a rangeInMinutes of 60, the calculated start date will be: 2021-01-05T23:00:00Z.  Default (60) is store in application-openshift-metering-worker.yaml"
```
lindseyburnett, line 79: in application-openshift-metering-worker.yaml and application-metrics-rhel.yaml

swatch-metrics/deploy/clowdapp.yaml
```diff
@@ -391,140 +399,28 @@ objects:
...
-                    key: ${INVENTORY_SECRET_KEY_NAME_PREFIX}password
-              - name: PROM_URL
-                value: ${PROM_URL}
-              - name: OTEL_EXPORTER_OTLP_ENDPOINT
-                value: ${OTEL_EXPORTER_OTLP_ENDPOINT}
-              - name: OTEL_SERVICE_NAME
-                value: ${OTEL_SERVICE_NAME}
-              - name: OTEL_JAVAAGENT_ENABLED
-                value: ${OTEL_JAVAAGENT_ENABLED}
+                value: metering-job
```
lindseyburnett, line 416: Could you explain why "this env property is necessary to pass the test check".  It's just a little confusing to delete the `metering-job` and `metrics-rhel-job` profiles, but still have a need to set them.

lindseyburnett (changes requested):

src/main/spec/internal-metering-api-spec.yaml
Sgitario: updated

swatch-metrics/deploy/clowdapp.yaml
Sgitario: Updated

Sgitario: Without this property, the PR check fails because it explicitly searches for this property and this value. I've just removed it, so you can see it.

Sgitario:
> It looks like you took care of a blocker card for this, https://issues.redhat.com/browse/SWATCH-1592. Is that right? Just want to make sure that card gets updated in Jira accordingly.
> 
> Also, a note about this part of the description:
> 
> > Note that prometheus is not configured in the ephemeral namespaces, so it's expected both calls to fail.
> 
> If you enable the token-refresher sidecar in the ephemeral environment you're testing with, I would expect the calls to be successful. There might not be any data returned, but the calls shouldn't fail.
> 
> Once you've reserved a namespace in ephemeral with bonfire, you can patch the ClowdEnvironment with this command:
> 
> `oc patch env env-$NAMESPACE -p '{"spec":{"providers":{"sidecars":{"tokenRefresher":{"enabled":true}}}}}' --type=merge`

@lindseyburnett PR updated and description updated as well with the above suggestion. Thanks!

lindseyburnett (approved):

Aurobinda55:
- DEploy EE and  hit the above patch command 
- Wait for 30 minutes and check the swatch-metrics-sync and swatch-metrics-rhel-sync jobs 
- Checked swatch-metrics-service and swatch-metrics-rhel-service and found the below logs 
```
self- Internal API: /api/rhsm-subscriptions/v1/internal/metering/sync requested by user: self
2023-10-04 11:30:02,426 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.task.PrometheusMetricsTaskManager] self- Queuing OpenShift-dedicated-metrics metric updates for range: [2023-10-04T10:00Z, 2023-10-04T11:00Z)
2023-10-04 11:30:02,447 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusService] self- Fetching metrics from prometheus: 2023-10-04T11:00Z -> 2023-10-04T11:00Z [Step: 3600]
2023-10-04 11:30:03,311 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.task.PrometheusMetricsTaskManager] self- Queuing OpenShift-dedicated-metrics Cores metric updates for all configured accounts.
2023-10-04 11:30:03,311 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.task.PrometheusMetricsTaskManager] self- Done queuing updates of OpenShift-dedicated-metrics Cores metric
2023-10-04 11:30:03,312 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.metering.service.prometheus.PrometheusService] self- Fetching metrics from prometheus: 2023-10-04T11:00Z -> 2023-10-04T11:00Z [Step: 3600]
```

Aurobinda55 (approved):

PR #2587, opened 2023-09-29
Avoid retally by changing the enddate during recalc

kartikshahc:
Jira issue: [[SWATCH-1780](https://issues.redhat.com/browse/SWATCH-1780)]

## Description

This is a PR same as what we deployed for hotfix branch to stop retally process hence I have put no test steps.


kahowell (approved):

kahowell (changes requested):
Needs some test updates

src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -78,9 +79,17 @@ public Stream<Event> fetchEventsInTimeRange(
   }
 
   public Stream<Event> fetchEventsInTimeRangeByServiceType(
-      String orgId, String serviceType, OffsetDateTime begin, OffsetDateTime end) {
+      String orgId,
+      String serviceType,
+      OffsetDateTime begin,
+      OffsetDateTime end,
+      OffsetDateTime processEndDateTime) {
```
kahowell, line 86: I think `asOfDateTime` might be a better name. You should add some comments around how this is used. Something like: "asOfDateTime is used to ignore events that are recorded outside a given hourly tally's window. Those events will be processed during the next hourly tally, and filtering them out here ensures they won't be processed twice."

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/event/EventController.java
kartikshahc: sounds good. Will add comments and change the variable name

kartikshahc: @kahowell have added the comment and changed the variable name.

kahowell (approved):

ssidhaye (approved):
Tested changes in EE for past events the behavior is as expected.

PR #2588, opened 2023-10-02
SWATCH-1651 (for conduit service): Populate uri label in Spring Boot 3

Sgitario:
Jira issue: [SWATCH-1651](https://issues.redhat.com/browse/SWATCH-1651)

## Description
The same changes as https://github.com/RedHatInsights/rhsm-subscriptions/pull/2489 but applied to the conduit service.

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/security/RoleProvider.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/security/RoleProvider.java
@@ -20,9 +20,10 @@
  */
 package org.candlepin.subscriptions.security;
 
-import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Builds roles based on the provided RBAC permission list. The intent of this class is to process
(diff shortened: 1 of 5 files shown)


swatch-core/src/main/java/org/candlepin/subscriptions/security/RoleProvider.java
```diff
@@ -42,20 +43,24 @@ public RoleProvider(String rulePrefix, boolean devModeEnabled) {
...
-    if (devModeEnabled || permissions.contains(rulePrefix + ":*:*")) {
+    if (permissions.contains(rulePrefix + ":*:*")) {
       roles.add(SWATCH_ADMIN_ROLE);
     }
     if (permissions.contains(rulePrefix + ":reports:read")) {
       roles.add(SWATCH_REPORT_READER);
     }
+    if (devModeEnabled) {
+      roles.add(SWATCH_ADMIN_ROLE);
+      roles.add(ROLE_INTERNAL);
```
Sgitario, line 62: I think it makes sense to also add the internal role when using DEV_MODE=true to use the internal endpoints (I needed for the conduit service).

kahowell (approved):

PR #2598, opened 2023-10-03
SWATCH-662 remove account config reposititory in opt in

vbusch:
Jira issue: [SWATCH-662](https://issues.redhat.com/browse/SWATCH-662)

## Description
- Replace usages of AccountConfigRepository with OrgConfigRepository for opt-in

## Testing
Unit tests & Regression


vbusch:
2 Failing IQE tests that are looking for account: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/blob/master/iqe_rhsm_subscriptions_api/model/opt_in_config_data.py#L89
- [tests.opt_in.test_opt_in_prometheus.test_verify_opt_in_prometheus[rhosak]](https://ci.ext.devshift.net/job/RedHatInsights-rhsm-subscriptions-pr-check/4425/testReport/junit/tests.opt_in/test_opt_in_prometheus/test_verify_opt_in_prometheus_rhosak_/)
- [tests.opt_in.test_opt_in_prometheus.test_verify_opt_in_prometheus_with_kafka_event[OpenShift-dedicated-metrics]](https://ci.ext.devshift.net/job/RedHatInsights-rhsm-subscriptions-pr-check/4425/testReport/junit/tests.opt_in/test_opt_in_prometheus/test_verify_opt_in_prometheus_with_kafka_event_OpenShift_dedicated_metrics_/)

src/main/java/org/candlepin/subscriptions/security/OptInChecker.java
```diff
@@ -61,7 +61,7 @@ public boolean checkAccess(Authentication authentication) {
      * the OptInRequiredException in the AccessDecisionVoter.vote method and then our own
      * AbstractAccessDecisionManager capable of catching that exception and rethrowing it after all
      * the other voters had been consulted. */
-    if (!accountConfigRepository.existsByOrgId(insightsUserPrincipal.getOrgId())) {
+    if (Boolean.FALSE.equals(orgConfigRepository.existsByOrgId(insightsUserPrincipal.getOrgId()))) {
```
Sgitario, line 64: Is there any reason why `existsByOrgId` returns `Boolean` instead of the primitive `boolean`? 
Using the primitive type `boolean` would significantly make its usage easier because we don't need to check if it's null and we could simply do here:
```
if (!orgConfigRepository.existsByOrgId(insightsUserPrincipal.getOrgId())) {
```

Sgitario (commented):
Only one minor suggestion about using the primitive type `boolean` in the `existsByOrgId` method if this makes sense.

src/main/java/org/candlepin/subscriptions/security/OptInChecker.java
vbusch: Thank you.  I have added this change.

src/main/java/org/candlepin/subscriptions/security/OptInController.java
```diff
@@ -51,128 +47,69 @@
...
-    performOptIn(accountNumber, orgId, optInType);
+    return buildDto(buildMeta(orgId), buildOptInOrgDTO(orgData));
   }
 
   // Separate isolated transaction needed in order to prevent opt-in errors rolling back metrics
   // updates
   @Transactional(propagation = Propagation.REQUIRES_NEW)
   public void optInByOrgId(String orgId, OptInType optInType) {
-    if (!accountConfigRepository.existsById(orgId)) {
+    if (Boolean.FALSE.equals(orgConfigRepository.existsByOrgId(orgId))) {
```
Sgitario, line 77: Using `Boolean.FALSE.equals(` is no longer necessary because existsByOrgId now returns the primitive boolean.

Sgitario (changes requested):
There is still a place that uses `Boolean.FALSE.equals`. 
The rest looks good to me.
Thanks!

Sgitario (approved):
Thanks!

PR #2600, opened 2023-10-03
Randomize testcontainer network aliases

kahowell:
Otherwise we get strange behaviors due to kubedock behavior.

Notably,

> Kubedock flattens all networking, which basicly means that everything will run in the same namespace.
> When a network alias is present, it will create a service exposing all ports that have been exposed by the container.

I *think* this will fix recent ci test failures.


kahowell:
Seems to help, Jenkins check passed, and this change won't affect the other CI checks, so self-merging.

PR #2611, opened 2023-10-05
SWATCH-1801: Use simple solution to replace SPeL expression parser

Sgitario:
Jira issue: [SWATCH-1801](https://issues.redhat.com/browse/SWATCH-1801)

## Description
This is the simplest solution I can think of, to replace the SPeL expression parser which perfectly works for our use case and avoids the usage of reflection (so it's compatible with Quarkus native if in the future we want to use it). 

The drawback is that it lacks many features that regular expression parsers have like conditional expressions and many others. 

Another alternative to SPeL expressions and Quarkus Qute is to use [Mvel2](http://mvel.documentnode.com/) (which is supported by JBoss RedHat team). This framework works in Spring Boot and Quarkus, *but not in Quarkus native*, see [the Quarkus example](https://github.com/Sgitario/quarkus-examples/tree/main/mvel2) I created for this - use `mvn clean install` from the mvel2 directory and `mvn clean install -Dnative` to run the test using the native binaries and you can see the test will fail). 

Moreover, because of using Mvel2 or Qute (or any other expression parsers), we would need to update all the properties to accommodate the new conventions (for example, instead of using `#{metric.id}`, we need to use `@{metric.id}` in Mvel2. 

Therefore, using this very simple solution, it works in Quarkus native (because it does not use reflection), also it does not require changes and it's very well performant (because of the same, neither uses reflection nor requires adding an additional dependency).

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryBuilder.java
+++ b/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryBuilder.java
@@ -48,9 +48,6 @@ public class QueryBuilder {
           "#{metric.prometheus.queryKey}",
               descriptor -> descriptor.getMetric().getPrometheus().getQueryKey());
   private static final String KEY_VALUE_CLOSE_TAG = "]}";
-  private static final String SYSTEM_PROPERTY_OPEN_TAG = "${";
-  private static final String SYSTEM_PROPERTY_CLOSE_TAG = "}";
-  private static final String SYSTEM_PROPERTY_DEFAULT_TAG = ":";
 
   private final MetricProperties metricProperties;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryBuilder.java
```diff
@@ -61,25 +78,64 @@ public String buildAccountLookupQuery(QueryDescriptor queryDescriptor) {
...
+  private String buildQuery(String query, QueryDescriptor descriptor) {
+    // Support of system properties binding in the format of "${KEY:default value}"
+    if (query.contains(SYSTEM_PROPERTY_OPEN_TAG)) {
+      String rawKey =
+          StringUtils.substringBetween(query, SYSTEM_PROPERTY_OPEN_TAG, SYSTEM_PROPERTY_CLOSE_TAG);
+      String key = rawKey;
+      String defaultValue = null;
+      if (rawKey.contains(SYSTEM_PROPERTY_DEFAULT_TAG)) {
+        String[] keyParts = rawKey.split(SYSTEM_PROPERTY_DEFAULT_TAG);
+        key = keyParts[0];
```
awood, line 94: I would like to do some checks here.  Specifically, require `key` to begin with an `metric` prefix (and maybe need to add `runtime` as well?).  Giving this class unfettered access to the system properties and environment variables can have security implications.  I think it would be safer to just eliminate access to any property or environment variable that doesn't start with a blessed prefix.

awood (changes requested):

src/main/java/org/candlepin/subscriptions/metering/service/prometheus/promql/QueryBuilder.java
Sgitario: Let's agree first if this is the right approach or not. I'm favored of getting rid of the system properties usage altogether.

lindseyburnett (approved):

PR #2615, opened 2023-10-05
SWATCH-1771 - Update promql templates to handle unexpected duplicate instance results

lindseyburnett:
Jira issue: [SWATCH-1771](https://issues.redhat.com/browse/SWATCH-1771)

## Description
We expect a single cluster per metric.  In case of bad data/misconfiguration by the reporter, it's possible to have multiple values.  When this happens, prometheus throws an exception when trying to join.

In the event that the metric corresponds to multiple clusters/instances (identified by unique instance keys, generally _id), we'll take the max value of those to ensure one-to-many matches and avoid attempting to do a many-to-many grouping.

### Steps

Get the ORG_ID from the jira card to use for testing.

`./bin/metering-promql.py --org $ORG_ID --product rhacs`

Copy the output of "Cores PromQL" and paste it in grafana

https://telemeter-lts-dashboards.datahub.redhat.com/explore
Use the `historical` data source
Set the timeframe **2023-09-26 00:00:00** to **2023-09-27 23:59:59**

### Verification

On the main branch, you'll reproduce the issue as detailed in the card.

With this update, the query should execute successfully.  The query doesn't actually have any matching data, so "No data" being displayed it the expected outcome.

### DIFF
--- a/src/main/resources/application-metrics-rhel.yaml
+++ b/src/main/resources/application-metrics-rhel.yaml
@@ -9,11 +9,11 @@ rhsm-subscriptions:
       metric:
         queryTemplates:
           default: >-
-            #{metric.prometheus.queryParams[metric]}
+            max(#{metric.prometheus.queryParams[metric]}) by (#{metric.prometheus.queryParams[instanceKey]})
             * on(#{metric.prometheus.queryParams[instanceKey]}) group_right
             min_over_time(#{metric.prometheus.queryParams[metadataMetric]}{product="#{metric.prometheus.queryParams[product]}", external_organization="#{runtime[orgId]}", billing_model="marketplace", support=~"Premium|Standard|Self-Support|None"}[1h])
           addonSamples: >-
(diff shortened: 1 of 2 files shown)


Sgitario (approved):

Aurobinda55:
- Created a RHACS instance 
- Performed metering 
- Checked the below promql query and got successful data in response 

image
```

[{'id': 'e3056400-a021-4a40-9833-3714311cb77a',
  'instance_id': 'cluster_0908365353',
  'display_name': 'cluster_0908365353',
  'billing_provider': 'aws',
  'billing_account_id': 'testkgaonm',
  'measurements': [93.0],
  'last_seen': datetime.datetime(2023, 10, 6, 7, 0, tzinfo=tzlocal()),
  'number_of_guests': 0,
  '_labeled_measurements': {'Cores': 93.0}}]
```

Aurobinda55 (approved):

PR #2617, opened 2023-10-05
SWATCH-1765: Fix Contract capacity population for ROSA

ksynvrit:
Jira issue: [SWATCH-1765](https://issues.redhat.com/browse/SWATCH-1765)

## Description

If you observe in prod you will see that `subscription_measurements` for subscription: `13579593` there's old test data in the database has the  incorrect `MetricId` for rosa and not allowing QE to see the capacity after contracts are created.

Liquibase changes will update the affect `contracts_metrics` & `subscription_measurements` and the code changes should remove any unsupported metrics from the processing subscription. `19` records that will be affected by this change.

Gabi Query:
```
http -b https://gabi-subscription-watch-prod.apps.crcp01ue1.o9m8.p1.openshiftapps.com/query Authorization:"Bearer $(oc whoami -t)" query="select * from subscription_measurements where subscription_id = '13579593';" 
    
```

example:
```
{"result":["13579593","2023-09-25T08:25:22.797423Z","four_vcpu_hour","PHYSICAL","5"],["13579593","2023-09-25T08:25:22.797423Z","control_plane","PHYSICAL","1"]]}
```
## Testing Steps
1.Run both `bootApplication` & `Swatch-Contract` make sure one of the services are running on a different port
2. Setup up DB:
```
INSERT INTO public.offering (sku, product_name, product_family, cores, sockets, hypervisor_cores, hypervisor_sockets, role, sla, usage, description, has_unlimited_usage, derived_sku) VALUES ('MW02393', 'OpenShift Dedicated', null, null, null, null, null, null, 'Premium', null, 'Red Hat OpenShift Service on AWS Hosted Control Planes (Hourly)', false, null);

```
3. Use UI : http://localhost:XXXX/api/swatch-contracts/internal/swagger-ui/#/default/createContract and paste in data:
```
{
  "subscription_number": "sn800",
  "sku": "MW02393",
  "start_date": "2023-09-05T00:00:00Z",
  "org_id": "RosaOrg",
  "billing_provider": "aws",
  "billing_account_id": "bp1",
  "product_id": "rosa",
  "vendor_product_code": "VPC",
  "metrics": [
    {
      "metric_id": "four_vcpu_0",
      "value": 100
    },
    {
      "metric_id": "control_plane_0",
      "value": 100
    },
    {
      "metric_id": "four_vpc_hours",
      "value": 100
    }
  ]
}
```
3. Observe that `Cores` is visible in `contract_metrics` & `subscription_measurements` as it was the only metric that was correctly formatted to the metrics tag given from `bootApplication`: http://localhost:XXXX/api-docs/internal-subscription-sync.html#/internalSubscriptions/getMetrics
4. Attempt another request with the correct format for `Instance-hours` and observe it is now visible in the db.

### DIFF
--- a/src/main/resources/liquibase/202310041633-fix-placeholder-values-for-rosa.xml
+++ b/src/main/resources/liquibase/202310041633-fix-placeholder-values-for-rosa.xml
@@ -1,0 +1,23 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="202310041633-1" author="ksynvrit">
+        <comment>Change four_vcpu_hour to Cores on subscription_measurements.</comment>
+        <update tableName="subscription_measurements">
+            <column name="metric_id" value="Cores"/>
+            <where>metric_id='four_vcpu_hour'</where>
+        </update>
+    </changeSet>
+
+    <changeSet id="202310041633-2" author="ksynvrit">
+        <comment>Change control_plane to Instance-hours on subscription_measurements.</comment>
+        <update tableName="subscription_measurements">
+            <column name="metric_id" value="Instance-hours"/>
+            <where>metric_id='control_plane'</where>
+        </update>
+    </changeSet>
+
+</databaseChangeLog>
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/model/MeasurementMetricIdTransformer.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/model/MeasurementMetricIdTransformer.java
@@ -20,16 +20,22 @@
  */
 package com.redhat.swatch.contract.model;
 
+import com.redhat.swatch.clients.swatch.internal.subscription.api.model.Metric;
 import com.redhat.swatch.clients.swatch.internal.subscription.api.resources.ApiException;
 import com.redhat.swatch.clients.swatch.internal.subscription.api.resources.InternalSubscriptionsApi;
 import com.redhat.swatch.contract.exception.ContractsException;
 import com.redhat.swatch.contract.exception.ErrorCode;
 import com.redhat.swatch.contract.repository.BillingProvider;
+import com.redhat.swatch.contract.repository.ContractEntity;
 import com.redhat.swatch.contract.repository.SubscriptionEntity;
+import com.redhat.swatch.contract.repository.SubscriptionMeasurementEntity;
 import com.redhat.swatch.contract.repository.SubscriptionProductIdEntity;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
 
@@ -54,6 +60,9 @@ public class MeasurementMetricIdTransformer {
                 .map(SubscriptionProductIdEntity::getProductId)
                 .orElseThrow();
         var metrics = internalSubscriptionsApi.getMetrics(tag);
+        // the metricId currently set here is actually the aws Dimension and get translated to the
+        // metric uom after calculation
+        checkForUnsupportedMetrics(metrics, subscription);
         subscription
             .getSubscriptionMeasurements()
             .forEach(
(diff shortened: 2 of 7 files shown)


swatch-contracts/src/main/resources/db/202310091227_fix_contract_metrics_for_rosa.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+<changeSet id="202310091227-1" author="ksynvrit">
+    <comment>Change four_vcpu_hour to Cores on contract_metrics.</comment>
+    <update tableName="contract_metrics">
+        <column name="metric_id" value="Cores"/>
```
kartikshahc, line 10: We shouldn't change the values, in my opinion, since these values are of the aws dimension and not the actual metric id. The aws dimension should stay in the contract_metrics table and shouldn't be replaced. This problem has to be solved in different way because `contractEntityToSubscriptionProductIdEntity` in ContractMapper maps to incorrect values due to same name `metric_id`. So either we need to rename the column to cloud_dimension or `contractEntityToSubscriptionProductIdEntity` should have the right mapping.

swatch-contracts/src/main/resources/db/changeLog.xml
```diff
@@ -8,5 +8,6 @@
 
   <include file="db/202302030707_create_contracts_table.xml"/>
   <include file="db/202304061624_add_vendorproductcode_to_contracts_table.xml"/>
+  <include file="db/202310091227_fix_contract_metrics_for_rosa.xml"/>
```
kartikshahc, line 11: Remove this since we don't want to fix contract table as mentioned ^

swatch-contracts/src/main/java/com/redhat/swatch/contract/model/MeasurementMetricIdTransformer.java
```diff
@@ -78,4 +84,21 @@ public void translateContractMetricIdsToSubscriptionMetricIds(SubscriptionEntity
       throw new ContractsException(ErrorCode.UNHANDLED_EXCEPTION, e.getMessage());
     }
   }
+
+  private void checkForUnsupportedMetrics(List<Metric> metrics, SubscriptionEntity subscription) {
+    if (metrics.size() != subscription.getSubscriptionMeasurements().size()
```
kartikshahc, line 89: should we compare the metrics by values? This condition can still pass if the metric values are different and the size of the list is same.

src/main/resources/liquibase/202310041633-fix-placeholder-values-for-rosa.xml
```diff
@@ -0,0 +1,23 @@
...
+        <update tableName="subscription_measurements">
+            <column name="metric_id" value="Cores"/>
+            <where>metric_id='four_vcpu_hour'</where>
+        </update>
+    </changeSet>
+
+    <changeSet id="202310041633-2" author="ksynvrit">
+        <comment>Change control_plane to Instance-hours on subscription_measurements.</comment>
+        <update tableName="subscription_measurements">
+            <column name="metric_id" value="Instance-hours"/>
```
kartikshahc, line 18: It appears that the `subscription_measurements` table currently has `INSTANCE_HOURS` in capital letters and with underscore. Additionally, you should update the Enum to string conversion as we spoke, or you may create a card and add that as a comment in the capacity call method.

kartikshahc (changes requested):

swatch-contracts/src/main/java/com/redhat/swatch/contract/model/MeasurementMetricIdTransformer.java
ksynvrit: Yes, we should as the metrics in the method is a set, so all i'm doing is check if the metricId of the measurement is the right type, so it wouldn't remove the measurement as long as it has the correct metric assigned to the product.

kartikshahc: As we spoke last Friday this has to be fixed. Since if we have subscription with Sockets and Cores then Contract with Sockets and Instance-hours then it won't remove the metric id that are not matching.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -280,12 +281,27 @@ private SubscriptionEntity createSubscriptionForContract(
     var subscription = new SubscriptionEntity();
     mapper.mapContractEntityToSubscriptionEntity(subscription, contract);
     measurementMetricIdTransformer.translateContractMetricIdsToSubscriptionMetricIds(subscription);
+    if (subscription.getSubscriptionMeasurements().size() != contract.getMetrics().size()) {
```
kartikshahc, line 284: Would it bypass the below condition?
If we have subscription with Sockets and Cores then Contract with Sockets and Instance-hours with this code it won't remove the metric id that are not matching. Correct?

kartikshahc (changes requested):

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
ksynvrit: At that point in the code the unsupported metricId would be removed already from subscriptions and this is merely checking the contract metrics and making sure that it reflects same metrics as the subscription_measurements

kartikshahc (approved):
Haven't tested the steps after the code review update but overall looks good.

ksynvrit:
I want to test some more before merging as I want to make sure formatting is correct for `contract_metrics` table and the expected values.

ksynvrit:
@kahowell 
Screenshot from 2023-10-16 15-18-22

Could you confirm for me which is the expected `metricId` for contract_metrics table to finalize my changes are giving the expected results

kahowell:
> @kahowell
> 
> Could you confirm for me which is the expected `metricId` for contract_metrics table to finalize my changes are giving the expected results

I'd expect only valid `awsDimension` values:
* `control_plane_0`
* `four_vcpu_0`

```shell
git grep awsDimension:
```
shows possible values though we've only set up a product with contracts for testing for rosa thus far.

ksynvrit:
@ntkathole The pr is ready for your review

PR #2630, opened 2023-10-10
SWATCH-1795: Port ApplicationClock to a new library: swatch-common-clock

Sgitario:
Jira issue: [SWATCH-1795](https://issues.redhat.com/browse/SWATCH-1795)

## Description

This is a pull request with a massive number of changes, but most of them are related to the package rename. Basically, this pull request includes:

- Move Granularity-related methods to CombiningRollupSnapshotStrategy
- Move ApplicationClock to a swatch-common-clock module

Note that I tried to use the Jakarta CDI but it seems that Spring Boot does not work well with it, so I decided not to use it and add an auto-configurable clock configuration, so all modules that depend on swatch-core will automatically have the ApplicationClock bean. 

## Testing
No new tests is required here.


swatch-core/src/main/java/org/candlepin/subscriptions/util/ClockConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.boot.autoconfigure.AutoConfigurationExcludeFilter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.TypeExcludeFilter;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.FilterType;
+
+/** Provides access to common utility beans. */
+@Configuration
```
awood, line 33: Do we want to use the `@AutoConfiguration` annotation here?  Also if we want it to be picked up with auto-configuration, we'll need to [add the fully qualified class name](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.developing-auto-configuration.locating-auto-configuration-candidates) to `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports`.

awood (changes requested):
Looks good.  I just have a question about the auto-configuration piece.  Also, do you want to go ahead and create a commit that removes the old classes?

swatch-core/src/main/java/org/candlepin/subscriptions/util/ClockConfiguration.java
Sgitario: @Configuration and @Autoconfiguration work the same. Actually, META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports is already there and works as expected.

Sgitario:
> Looks good. I just have a question about the auto-configuration piece. Also, do you want to go ahead and create a commit that removes the old classes?

The old classes are removed. Which ones do you mean?

awood:
> The old classes are removed. Which ones do you mean?

I was looking down the icons and didn't see the one for "moved" for `ApplicationClock`.  It's there now so either I missed it or needed to refresh my browser.

swatch-core/src/main/java/org/candlepin/subscriptions/util/ClockConfiguration.java
awood: There is a slight difference

``` 
Auto-configuration classes are regular {@link Configuration @Configuration} with the exception that
{@literal Configuration#proxyBeanMethods() proxyBeanMethods} is always {@code false}.
```

but more importantly, using `@AutoConfiguration` indicates to future developers that this class isn't just a regular configuration.

awood: > Actually, META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports is already there and works as expected

I missed that in the review.  It's in `swatch-core` though.  Does it need to be in `swatch-common-clock`?

Sgitario: swatch-common-clock is framework agnostic, that's why.

awood: > [swatch-common-clock](https://issues.redhat.com/browse/SWATCH-common-clock) is framework agnostic, that's why.

Makes sense.  I would still like to annotate the configuration with `@AutoConfiguration` though.

Sgitario: Updated

Sgitario:
@awood PR updated by changing the annotation `@Configuration` with `@AutoConfiguration`.

awood (approved):

PR #2657, opened 2023-10-17
ci: run build/test phase sequentially instead of in parallel

Sgitario:
## Description
The "continuous-integration/jenkins/pr-merge" used to fail with:
```
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```

This is a random but quite often failure.

Splitting the build and test stages seems to improve the above random issue.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,7 +82,7 @@ spec:
             steps {
                 // The build task includes check, test, and assemble.  Linting happens during the check
                 // task and uses the spotless gradle plugin.
-                sh "./gradlew --no-daemon build testCodeCoverageReport"
+                sh "./gradlew --no-daemon --no-parallel build testCodeCoverageReport"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -78,7 +78,14 @@ spec:
...
         }
-        stage('Build/Test/Lint') {
+        stage('Build/Lint') {
+            steps {
+                // The build task includes check, and assemble.  Linting happens during the check
+                // task and uses the spotless gradle plugin.
+                sh "./gradlew --no-daemon build -x test"
+            }
+        }
+        stage('Test/Lint') {
```
Sgitario, line 88: An alternative is to use the flag ` --no-parallel` to avoid running the tests in parallel. Wdyt?

lindseyburnett (approved):

PR #2658, opened 2023-10-17
Add floorplan for subscriptions

kahowell:
Notably this normalizes the metric_id values so that downstream consumers of this data don't have to deal with `SOCKETS` vs. `Sockets`.

Testing
=======
You can try the query with gabi in any environment, please add `limit 10` (or similar) so as to not select too much data at once.

PR #2662, opened 2023-10-17
[HOTFIX] Add floorplan for subscriptions

kahowell:
Same as #2658, but applied to hotfix branch.

Notably this normalizes the metric_id values so that downstream consumers of this data don't have to deal with `SOCKETS` vs. `Sockets`.

Testing
=======
You can try the query with gabi in any environment, please add `limit 10` (or similar) so as to not select too much data at once.


kahowell:
Self-merging since related PR to main was approved

PR #2692, opened 2023-10-23
SWATCH-1099: deleteDataAssociatedWithOrg should delete contracts

Sgitario:
Jira issue: [SWATCH-1099](https://issues.redhat.com/browse/SWATCH-1099)

## Description
Add a DELETE endpoint to swatch-contracts that's available to the "service" role.

For QE, the new endpoint is: "DELETE: /api/swatch-contracts/internal/rpc/reset/contracts/{org_id}"

## Testing
Added tests to cover the new endpoint.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/admin/InternalTallyDataController.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/admin/InternalTallyDataController.java
@@ -36,6 +36,7 @@ import org.candlepin.subscriptions.event.EventController;
 import org.candlepin.subscriptions.json.Event;
 import org.candlepin.subscriptions.security.OptInController;
 import org.candlepin.subscriptions.tally.AccountResetService;
+import org.candlepin.subscriptions.tally.billing.ContractsController;
 import org.candlepin.subscriptions.tally.job.CaptureSnapshotsTaskManager;
 import org.candlepin.subscriptions.util.DateRange;
 import org.candlepin.subscriptions.utilization.api.model.OptInConfig;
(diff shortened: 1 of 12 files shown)


ntkathole:
@Sgitario I see the new endpoint working fine and deleting contracts for org. 
```
 # swatch-client internal-api --app swatch-contracts --method DELETE --endpoint /internal/rpc/reset/contracts/3340851
INFO:root:Starting DELETE /internal/rpc/reset/contracts/3340851 in clowder_smoke env ...
INFO:root:
request response url: http://swatch-contracts-service:8000/api/swatch-contracts/internal/rpc/reset/contracts/3340851, status_code: 200
INFO:root:Response return: {"status":"SUCCESS"}
```

But I want to confirm that as per card description `update the "reset account" logic to call the DELETE endpoint` i.e. expectation is `DELETE http://swatch-tally-service:8000/api/rhsm-subscriptions/v1/internal/rpc/tally/3340851` endpoint should also delete all contracts, which is not happening currently. From QE perspective, this also work for us with just additional request call, but just want to confirm that we are making any change in  internal/rpc/tally delete method ?

ntkathole (approved):

Sgitario:
> Ack, If this make things complex for us, I am good with this change 👍

It seems that there is already a contracts client configured for the tally service, so I'm going to update the pull request, so the deleteDataAssociatedWithOrg also calls the contracts service to delete the contracts. Are you ok with?

ntkathole:
> It seems that there is already a contracts client configured for the tally service, so I'm going to update the pull request, so the deleteDataAssociatedWithOrg also calls the contracts service to delete the contracts. Are you ok with?

Yes, please. Thanks.

Sgitario:
> It seems that there is already a contracts client configured for the tally service, so I'm going to update the pull request, so the deleteDataAssociatedWithOrg also calls the contracts service to delete the contracts. Are you ok with?

PR updated!

ntkathole (approved):
Working fine now.

DELETE /internal/rpc/tally/ now deletes contracts as well 👍

kartikshahc (approved):

PR #2707, opened 2023-10-25
Get account_number from account_config table (stop-gap)

kahowell:
Until internal stakeholders can switch to org_id. Note this uses data we collected in account_config table when available.

Depends on the recreation of the account_config table in stage (we plan to comment out the changeset that removes it from prod, so that prod will have a populated table).

Depends on: #2710.

## Testing

Queries tested via gabi, need to adjust the tally snapshot date range to 10 days to get an org_id with a mapped account_number.

### DIFF
--- a/swatch-tally/deploy/clowdapp.yaml
+++ b/swatch-tally/deploy/clowdapp.yaml
@@ -649,7 +649,8 @@ objects:
     - prefix: swatch/tally
       query: >-
         select
-        org_id,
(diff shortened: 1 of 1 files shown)


Sgitario (approved):
I checked the changes in the query and it worked fine.
Btw, after https://github.com/RedHatInsights/rhsm-subscriptions/pull/2714 is merged, these queries will be automatically validated against a local database.

PR #2708, opened 2023-10-25
[HOTFIX] Get account_number from account_config table (stop-gap)

kahowell:
Until internal stakeholders can switch to org_id.


kahowell:
self-merging since query is currently not working as needed

PR #2709, opened 2023-10-25
[HOTFIX] Temporarily increase tally floorist range to 10 days

kahowell:


kahowell:
self-merging since this is a hotfix that changes only a simple query param

PR #2714, opened 2023-10-26
SWATCH-1884: Validate the floorish queries per pull request

Sgitario:
Jira issue: [SWATCH-1884](https://issues.redhat.com/browse/SWATCH-1884)

## Description
The idea is to ensure that the queries declared in the floorplan are checked to avoid issues like [SWATCH-1882](https://issues.redhat.com/browse/SWATCH-1882).

We use yaml-path which is a command line tool powered by Jbang. Yaml-path allows to extract elements from YAML resources. Here, we're extracting the queries and then executing them individually against a Postgresql database to confirm that it's working fine.

### DIFF
--- a/.github/workflows/validate-floorplan-queries.yaml
+++ b/.github/workflows/validate-floorplan-queries.yaml
@@ -1,0 +1,42 @@
+name: "FloorPlan Queries"
+on:
+  - pull_request
+jobs:
+  validate:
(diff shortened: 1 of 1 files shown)


kahowell (approved):

lindseyburnett:
I like it!

PR #2715, opened 2023-10-26
Bump kubedock version to 0.14

Sgitario:
## Description
Kubedock release: https://github.com/joyrex2001/kubedock/releases/tag/0.14.0

Related to https://github.com/RedHatInsights/kubedock/pull/4


Sgitario:
self-merging since it's a version bump and podname change only.

PR #2738, opened 2023-11-02
SWATCH-1800: Write a from-scratch equivalent of the task message JSON

Sgitario:
Jira issue: [SWATCH-1800](https://issues.redhat.com/browse/SWATCH-1800)

## Description
We need to be able to queue up tasks (eg sync metrics hourly) for our task framework, but we don't want to have to add swatch-core as a dependency.

## Testing
I've added unit tests that cover the consumer and producer tasks using an in-memory kafka instance.


Sgitario:
Build is failing with:
```
14:19:29 ��� ERROR SpringBoot properties file ./swatch-metrics/src/main/resources/application.yaml references platform.rhsm-subscriptions.metering-tasks, but does not use clowder to configure actual topic name.
```

PR #2739, opened 2023-11-02
Change tekton triggers not use cel (temporarily)

kahowell:
## Description

As part of debugging RHTAP checks, we found cel expression parsing seems to not be working properly for unknown reasons. The downside to not using cel expressions is we have no way to limit the pipeline to run for changed components without using cel. We'll open a PR to move back to using CEL expressions separately.

## Testing

Just wait for the RHTAP pipelines to finish successfully.


kahowell:
Self-merging since pipelines ran successfully.

PR #2750, opened 2023-11-03
SWATCH-1802: Port Prometheus client to quarkus

Sgitario:
Jira issue: [SWATCH-1802](https://issues.redhat.com/browse/SWATCH-1802)

## Description

- [x] add to clients/build.gradle a project ":clients:quarkus:prometheus-client" using plugin: "swatch.quarkus-rest-client-conventions"
- [x] copy PrometheusService to swatch-metrics
- [x] copy QuerySummaryResult to swatch-metrics
- [x] add config to application.yaml (reference src/main/resources/application-openshift-metering-worker.yaml)

## Testing
I added all the tests for QueryBuilder and PrometheusService to verify these changes.

Note that there are a couple of changes in comparison to the Spring-based version:

1.- We don't need to sanitize the PromQL query: https://github.com/RedHatInsights/rhsm-subscriptions/blob/ebb01ecca09e847ee3ff1e3dbaca2f05400668a7/src/main/java/org/candlepin/subscriptions/metering/service/prometheus/PrometheusService.java#L116

The tests in PrometheusServiceTest were failing because REST Client was sanitizing the query twice.

2.- The generated Prometheus Client is missing the annotation `@JsonValue` in enumerators. 

This is not an issue for production because we don't need to serialize the enumerators, but yes for tests, so that's why I had to add this workaround: https://github.com/RedHatInsights/rhsm-subscriptions/pull/2750/files#diff-966f68c6ed3c9351670f2334b88849b6346253d05cbbd862cb3ff959df8e81ceR70


clients/quarkus/prometheus-client/prometheus-client-config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "modelPackage": "org.candlepin.subscriptions.prometheus.model",
+  "apiPackage": "org.candlepin.subscriptions.prometheus.resources",
+  "invokerPackage": "org.candlepin.subscriptions.prometheus",
+  "groupId": "org.candlepin",
```
kahowell, line 5: Please use `groupId` and package prefix of `com.redhat.swatch`; it's not actually used, but is confusing to see, see `buildSrc/src/main/groovy/swatch.quarkus-rest-client-conventions.gradle` for where its overridden.

swatch-metrics/src/main/java/com/redhat/swatch/metrics/service/prometheus/PrometheusService.java
```diff
@@ -0,0 +1,193 @@
...
+  public static final String JSON_PROPERTY_DATA = "data";
+  public static final String JSON_PROPERTY_ERROR_TYPE = "errorType";
+  public static final String JSON_PROPERTY_ERROR = "error";
+
+  @Inject @RestClient QueryApi queryApi;
+  @Inject @RestClient QueryRangeApi queryRangeApi;
+  @Inject JsonFactory factory;
+
+  public QuerySummaryResult runRangeQuery(
+      String query,
```
kahowell, line 63: We previously had to manually URL-escape the query. Is this no longer needed :-) ?

clients/quarkus/prometheus-client/prometheus-api-spec.yaml
```diff
@@ -0,0 +1,148 @@
+# This is a file maintained by the rhsm-subscriptions project and describes
```
kahowell, line 1: Rather than duplicate this file, it's preferable to simply override the path in `clients/build.gradle` like:

```groovy
    openApiGenerate {
        inputSpec = "${rootDir}/clients/prometheus-client/prometheus-api-spec.yaml"
    }
```

(Once we no longer need the spring-boot client, we'd promote `:clients:quarkus:prometheus-client` to `:clients:prometheus-client`, keep the file in its current location, and remove this override).

kahowell (changes requested):

swatch-metrics/src/main/java/com/redhat/swatch/metrics/service/prometheus/PrometheusService.java
Sgitario: No, this is one of the two things I spotted when testing the changes (see the Testing section in the description).

clients/quarkus/prometheus-client/prometheus-client-config.json
Sgitario: This configuration was actually not used. Removed.

clients/quarkus/prometheus-client/prometheus-api-spec.yaml
Sgitario: Done

Sgitario:
@kahowell PR updated after your comments. Thanks!

kahowell (approved):

PR #2775, opened 2023-11-14
SWATCH-1864: quarkus-based services: consume configuration for TLS for inter-service communication from clowder config

Sgitario:
Jira issue: [SWATCH-1864](https://issues.redhat.com/browse/SWATCH-1864)

## Description
The clowder config source already deals with the configuration of the URL, truststore path and the rest of properties for SSL. All we need is to use `${clowder.endpoints.<dependency name>-service.url}` instead of `http://${clowder.endpoints.capacity-ingress.hostname}:${clowder.endpoints.capacity-ingress.port}`. 
And also to properly declare the dependencies.

Note that this strongly depends on the ClowdEnvironment object to be updated with:

```
{
  "spec": {
    "providers": {
      "web": {
        "tls": {
          "enabled": true,
          "port": 8800
        }
      }
    }
  }
}
```

With `tls.enabled=true`, the envoy sidecar will be created for all the clowd apps. 
However, if the `tls` flag is not enabled, everything will keep working as usual but using HTTP. The clowder config source will see if the `tls` flag is enabled or not, and configure the rest client accordingly. 

## Testing
Ephemeral namespace already has the "tls.enabled=true" flag, so running the QE test suite to confirm everything works as before should be enough.

For manual deployment:

1. `bonfire namespace reserve --pool default` -> output should be the oc project name
2. `oc project <project name>`
3. `oc patch env env-<project name> -p '{"spec":{"providers":{"sidecars":{"tokenRefresher":{"enabled":true}}, "web":{"tls":{"enabled": true, "port": 8800}}}}}' --type=merge`
4. `bonfire deploy rhsm --source=appsre --namespace <project name> ...`

### DIFF
--- a/deploy/dev-clowdenv.yaml
+++ b/deploy/dev-clowdenv.yaml
@@ -26,6 +26,10 @@ objects:
         port: 8000
         privatePort: 8080
         mode: operator
+        # enable envoy sidecar to handle https connections
+        tls:
+          enabled: true
+          port: 8800
 
       # provides a prometheus metrics port on 9000
       metrics:
--- a/swatch-producer-aws/deploy/clowdapp.yaml
+++ b/swatch-producer-aws/deploy/clowdapp.yaml
@@ -41,8 +41,6 @@ parameters:
     value: '3'
   - name: SPLUNK_HEC_INCLUDE_EX
     value: 'true'
-  - name: SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT
-    value: http://swatch-subscription-sync-service:8000/api/rhsm-subscriptions/v1
   - name: MOTO_REPLICAS
     value: '0'
   - name: AWS_MARKETPLACE_ENDPOINT_URL
@@ -85,6 +83,8 @@ objects:
       prometheus: quarkus
   spec:
     envName: ${ENV_NAME}
+    dependencies:
+      - swatch-subscription-sync
 
     kafkaTopics:
       - replicas: ${{KAFKA_BILLABLE_USAGE_REPLICAS}}
@@ -167,8 +167,6 @@ objects:
               value: ${SPLUNK_HEC_RETRY_COUNT}
             - name: SPLUNK_HEC_INCLUDE_EX
               value: ${SPLUNK_HEC_INCLUDE_EX}
-            - name: SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT
-              value: ${SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT}
             - name: AWS_MANUAL_SUBMISSION_ENABLED
               value: ${AWS_MANUAL_SUBMISSION_ENABLED}
             - name: AWS_MARKETPLACE_ENDPOINT_OVERRIDE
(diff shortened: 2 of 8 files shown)


deploy/dev-clowdenv.yaml
```diff
@@ -26,6 +26,10 @@ objects:
         port: 8000
         privatePort: 8080
         mode: operator
+        # enable envoy sidecar to handle https connections
+        tls:
+          enabled: true
+          port: 8800
```
Sgitario, line 32: I don't know where this file is used, but perhaps it's good to keep this change for clarity.

swatch-producer-aws/deploy/clowdapp.yaml
```diff
@@ -85,6 +83,8 @@ objects:
       prometheus: quarkus
   spec:
     envName: ${ENV_NAME}
+    dependencies:
+      - swatch-subscription-sync
```
Sgitario, line 87: With this configuration, the clowder config file will include the configuration for this service.

swatch-producer-aws/src/main/resources/application.properties
```diff
@@ -1,7 +1,7 @@
 SERVER_PORT=${clowder.endpoints.swatch-producer-aws.port:8000}
 LOGGING_LEVEL_COM_REDHAT_SWATCH=INFO
 LOGGING_LEVEL_ROOT=INFO
-SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT=http://${clowder.endpoints.capacity-ingress.hostname}:${clowder.endpoints.capacity-ingress.port}
+SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT=${clowder.endpoints.swatch-subscription-sync-service.url}
```
Sgitario, line 4: The capacity-ingress didn't exist. The SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT value was being set in the clowderapp.yaml.
After these changes, all the Quarkus-based services will populate the configuration using the clowder config source.

Sgitario:
The way I verified that the rest clients are using the correct URL is by adding the following bean in the application:

```
-@Singleton
-@Slf4j
-public class DummyBean {
-
-    @ConfigProperty(name = "my.config")
-    String url;
-    
-    void onStart(@Observes StartupEvent ev) {
-        log.info("URL is ..." + url);
-    }
-}
```

Where `my.config` is `my.config=${XXXX_ENDPOINT}`

kahowell (approved):

swatch-producer-aws/deploy/clowdapp.yaml
```diff
@@ -41,8 +41,6 @@ parameters:
     value: '3'
   - name: SPLUNK_HEC_INCLUDE_EX
     value: 'true'
-  - name: SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT
```
lindseyburnett, line 44: Do you mind putting an MR out in app-interface to clean up `SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT`  out of the [deploy-clowder.yml](https://gitlab.cee.redhat.com/service/app-interface/-/blob/78aa92465a4e3afbf1a44e28563b42fe60b1d062/data/services/insights/rhsm/cicd/deploy-clowder.yml#L137) since it's no longer being exposed to override via env var?

It actually looks like it can be removed out of the `swatch-producer-azure` of that same file too.

Sgitario: Done: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/88459

PR #2788, opened 2023-11-16
SWATCH-1936: Remove account_number from floorist queries

lindseyburnett:
Jira issue: [SWATCH-1936](https://issues.redhat.com/browse/SWATCH-1936)

## Description

TeleSense requested that we stop reporting the "account_number" column.  By removing that, it allows us to finally drop the account_config table that we stopped using as part of [the clean up account number usage epic](https://issues.redhat.com/browse/SWATCH-657) . 

## Testing

You can use gabi to verify the sql queries.  Changes should be reflected in S3 reports the next day.


mstead (approved):
Good to Go!

PR #2803, opened 2023-11-20
SWATCH-379: Sync new vCPUs field to HBI when present

Sgitario:
Jira issue: [SWATCH-379](https://issues.redhat.com/browse/SWATCH-379)

## Description
Populate the value in the system profile:

<candlepin fact> -> <system profile field>
cpu.cpu(s) -> number_of_cpus: [link to API](https://github.com/RedHatInsights/insights-host-inventory/blob/470f2264ed7be2f5b028f880a7aae2790ca9e660/swagger/system_profile.spec.yaml#L222)
cpu.thread(s)_per_core -> threads_per_core: [link to API](https://github.com/RedHatInsights/insights-host-inventory/blob/470f2264ed7be2f5b028f880a7aae2790ca9e660/swagger/system_profile.spec.yaml#L237)

## Testing
Added test to cover this change and ensure that the values are persisted. 
I don't think we need QE here (after SWATCH-401 is done, they can verify the end-to-end change).

0.- Start Kafka

`podman-compose up`

1.- Run conduit app with stubs
`RHSM_USE_STUB=true DEV_MODE=true ./gradlew :swatch-system-conduit:bootRun`

2.- Sync organization:
`http :8000/api/rhsm-subscriptions/v1/internal/rpc/syncOrg   x-rh-swatch-psk:placeholder   org_id=org123`

3. Check the message that has been sent to Kafka in http://localhost:3030/#/cluster/default/topic/n/platform.inventory.host-ingress/ .

The stub is configured to add number of cpus to 3 and threads per core to 5. This is what you should see:

```
Key
Value
operation add_host
data
org_id org123
subscription_manager_id e7ec2d55-2a17-42dc-8e70-a23126f208fa
bios_uuid 723a1ecb-d3ce-4900-b095-d3c5911bfb1a
ip_addresses [ 192.167.11.2, 192.168.1.1, 192.168.122.1, 10.0.0.1 ]
fqdn host1.test.com
facts [ [object Object] ]
system_profile
operating_system { major: 6, minor: 3, name: RHEL }
os_release 6.3
releasever 8.0
arch x86_64
cores_per_socket 2
infrastructure_type virtual
system_memory_bytes 33543999488
number_of_sockets 2
number_of_cpus 3
threads_per_core 5
owner_id e7ec2d55-2a17-42dc-8e70-a23126f208fa
is_marketplace true
network_interfaces [ [object Object] ]
stale_timestamp 2023-11-29T15:08:11.524269078Z
reporter rhsm-conduit
platform_metadata { request_id: 645cc340-f15f-4b6d-8401-296b4fc26ebc }
```

### DIFF
--- a/swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/InventoryControllerTest.java
+++ b/swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/InventoryControllerTest.java
@@ -946,4 +946,34 @@ class InventoryControllerTest {
     verify(inventoryService).scheduleHostUpdate(cfacts);
     verify(inventoryService, times(1)).flushHostUpdates();
   }
+
+  @Test
+  void testNumberOfCpusIsMappedToConduitFacts() {
+    int expectedNumberOfCpus = 5;
+    Consumer consumer = new Consumer();
+    consumer.getFacts().put("cpu.cpu(s)", "" + expectedNumberOfCpus);
+
+    // verify when cpu.cpu(s) is set
+    ConduitFacts conduitFacts = controller.getFactsFromConsumer(consumer);
+    assertEquals(expectedNumberOfCpus, conduitFacts.getNumberOfCpus());
+
+    // verify when cpu.cpu(s) is unset
+    conduitFacts = controller.getFactsFromConsumer(new Consumer());
+    assertNull(conduitFacts.getNumberOfCpus());
+  }
+
+  @Test
+  void testThreadsPerCoreIsMappedToConduitFacts() {
+    int expectedThreadsPerCore = 3;
+    Consumer consumer = new Consumer();
+    consumer.getFacts().put("cpu.thread(s)_per_core", "" + expectedThreadsPerCore);
+
+    // verify when cpu.thread(s)_per_core is set
+    ConduitFacts conduitFacts = controller.getFactsFromConsumer(consumer);
+    assertEquals(expectedThreadsPerCore, conduitFacts.getThreadsPerCore());
+
+    // verify when cpu.thread(s)_per_core is unset
+    conduitFacts = controller.getFactsFromConsumer(new Consumer());
+    assertNull(conduitFacts.getThreadsPerCore());
+  }
 }
--- a/clients/rhsm-client/src/main/java/org/candlepin/subscriptions/conduit/rhsm/client/StubRhsmApi.java
+++ b/clients/rhsm-client/src/main/java/org/candlepin/subscriptions/conduit/rhsm/client/StubRhsmApi.java
@@ -67,6 +67,8 @@ public class StubRhsmApi extends RhsmApi {
     consumer1.getFacts().put("Mac-addresses", "00:00:00:00:00:00, ff:ff:ff:ff:ff:ff");
     consumer1.getFacts().put("cpu.cpu_socket(s)", "2");
     consumer1.getFacts().put("cpu.core(s)_per_socket", "2");
+    consumer1.getFacts().put("cpu.cpu(s)", "3");
+    consumer1.getFacts().put("cpu.thread(s)_per_core", "5");
 
     consumer1.getFacts().put("memory.memtotal", "32757812");
     consumer1.getFacts().put("uname.machine", "x86_64");
--- a/swatch-system-conduit/schemas/inventory/hbi-system-profile.yaml
+++ b/swatch-system-conduit/schemas/inventory/hbi-system-profile.yaml
@@ -23,6 +23,10 @@ properties:
     type: float
   number_of_sockets:
     type: integer
+  number_of_cpus:
+    type: integer
+  threads_per_core:
+    type: integer
   owner_id:
     type: string
   is_marketplace:
--- a/swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
+++ b/swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
@@ -101,6 +101,8 @@ public class InventoryController {
   public static final String NET_INTERFACE_LO_IPV6_ADDRESS = "net.interface.lo.ipv6_address";
   public static final String CPU_SOCKETS = "cpu.cpu_socket(s)";
   public static final String CPU_CORES_PER_SOCKET = "cpu.core(s)_per_socket";
+  public static final String NUMBER_OF_CPUS = "cpu.cpu(s)";
+  public static final String THREADS_PER_CORE = "cpu.thread(s)_per_core";
   public static final String MEMORY_MEMTOTAL = "memory.memtotal";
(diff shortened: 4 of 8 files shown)


mstead (changes requested):
I think that the changes in this PR are part of [SWATCH-401](https://issues.redhat.com/browse/SWATCH-401) and does not accomplish what is required by SWATCH-379.

I believe the intent of SWATCH-379 is to get the new data from the Consumer feed into HBI via swatch-system-conduit. Please correct me if I'm wrong! :)

This PR will pull the new system profile facts from HBI (if they exist), but the data will not exist in HBI until we send it via conduit.

I was expecting to see a change in swatch-system-conduit that pulled 'cpu.cpu(s)' and 'cpu.thread(s)_per_core' from the Consumer feed data (if present) and then sent them to HBI as part of the system profile under the new names (number_of_cpus, threads_per_core). See `InventoryService.createSystemProfile(ConduitFacts facts)`

After they are sent to HBI, the current code changes you've made would make them available to the FactNormalizer during a nightly tally, allowing us to implement [SWATCH-401](https://issues.redhat.com/browse/SWATCH-401)

I'm fine with keeping the current code as part of this PR as it will need to be done as part of 401 anyway.

Sgitario:
> I think that the changes in this PR are part of [SWATCH-401](https://issues.redhat.com/browse/SWATCH-401) and does not accomplish what is required by [SWATCH-379](https://issues.redhat.com/browse/SWATCH-379).
> 
> I believe the intent of [SWATCH-379](https://issues.redhat.com/browse/SWATCH-379) is to get the new data from the Consumer feed into HBI via swatch-system-conduit. Please correct me if I'm wrong! :)
> 
> This PR will pull the new system profile facts from HBI (if they exist), but the data will not exist in HBI until we send it via conduit.
> 
> I was expecting to see a change in swatch-system-conduit that pulled 'cpu.cpu(s)' and 'cpu.thread(s)_per_core' from the Consumer feed data (if present) and then sent them to HBI as part of the system profile under the new names (number_of_cpus, threads_per_core). See `InventoryService.createSystemProfile(ConduitFacts facts)`
> 
> After they are sent to HBI, the current code changes you've made would make them available to the FactNormalizer during a nightly tally, allowing us to implement [SWATCH-401](https://issues.redhat.com/browse/SWATCH-401)
> 
> I'm fine with keeping the current code as part of this PR as it will need to be done as part of 401 anyway.

I think you're totally right. I've removed the existing changes that should be part of SWATCH-401 and updated the pull request with the correct changes.
Many thanks @mstead !

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/InventoryControllerTest.java
```diff
@@ -946,4 +946,26 @@ void testSystemMemoryBytesOverMaxValueSetsToNull() throws ApiException {
     verify(inventoryService).scheduleHostUpdate(cfacts);
     verify(inventoryService, times(1)).flushHostUpdates();
   }
+
+  @Test
+  void testNumberOfCpusIsMappedToConduitFacts() {
+    int expectedNumberOfCpus = 5;
```
mstead, line 952: Being a little picky here, but why not make expectedNumberOfCpusa String to begin with? This would allow you to use the value directly when setting the fact, instead of appending the "" when setting the value to make it a string  (same with expectedThreadsPerCore below).

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/InventoryControllerTest.java
```diff
@@ -946,4 +946,26 @@ void testSystemMemoryBytesOverMaxValueSetsToNull() throws ApiException {
...
+    Consumer consumer = new Consumer();
+    consumer.getFacts().put("cpu.cpu(s)", "" + expectedNumberOfCpus);
+
+    ConduitFacts conduitFacts = controller.getFactsFromConsumer(consumer);
+    assertEquals(expectedNumberOfCpus, conduitFacts.getNumberOfCpus());
+  }
+
+  @Test
+  void testThreadsPerCoreIsMappedToConduitFacts() {
+    int expectedThreadsPerCore = 3;
```
mstead, line 963: See note above.

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/InventoryControllerTest.java
```diff
@@ -946,4 +946,26 @@ void testSystemMemoryBytesOverMaxValueSetsToNull() throws ApiException {
...
+
+    Consumer consumer = new Consumer();
+    consumer.getFacts().put("cpu.thread(s)_per_core", "" + expectedThreadsPerCore);
+
+    ConduitFacts conduitFacts = controller.getFactsFromConsumer(consumer);
+    assertEquals(expectedThreadsPerCore, conduitFacts.getThreadsPerCore());
+  }
```
mstead, line 970: I'd like to see tests covering the case where the fact is not set if it is not present.

mstead (changes requested):
Overall this looks good. Couple of minor requests for the tests. 

Have you been able to test this out against an actual Consumer feed API? Would like to see some "How To Test" steps to go along with this PR. Perhaps using smqe-tools if possible.

swatch-system-conduit/src/test/java/org/candlepin/subscriptions/conduit/InventoryControllerTest.java
Sgitario: Added

mstead:
> I've updated the PR by adding the tests you were asking. Also, I've updated the description to verify the changes using stubs.
Excellent!

> I have no idea about using a real Consumer feed API or use sqme-tools for doing this (I can ask QE if you really think it's useful).
Personally, I think that this is something that QE might be interested in and would be useful for you to get familiar with). 

> About the other two comments, I've not changed it as the facts expect a string, but then the type expects an integer. So, we can't avoid the casting. I chose the string because it's easier to parse from string to integer, and also because the type we are covering is really an integer. But let me know if you feel strongly about using string instead.

Ah, missed the fact that it was an int coming out of the facts. :+1:

Sgitario:
> > I have no idea about using a real Consumer feed API or use sqme-tools for doing this (I can ask QE if you really think it's useful).
> 
> Personally, I think that this is something that QE might be interested in and would be useful for you to get familiar with).

I've been speaking with @ntkathole and he explained the best approach to verify these changes in an ephemeral environment. However, since it involves new system facts to be added to the `swatch-client` command line, we would need to do it directly in Python. So, the QE Unneeded label has been removed and they will verify these changes.

mstead (approved):

Aurobinda55 (approved):

Sgitario:
I'm merging this pull request, so we can fully test the e2e feature as part of SWATCH-401.

PR #2814, opened 2023-11-22
SWATCH-1940: Support multiple brokers in spring-boot based components

Sgitario:
Jira issue: [SWATCH-1940](https://issues.redhat.com/browse/SWATCH-1940)

## Description
Instead of specifying spring.kafka.bootstrap-servers as a single hostname:port, we need to transform the broker config we read into a comma-separated list of all provided host:port.

## Testing
No QE verification is needed.


swatch-core/src/main/resources/swatch-core/application.yaml
```diff
@@ -8,8 +8,6 @@
 # result in an exception at start-up when the clowder property cannot be successfully resolved.
 # The values can still be overridden using environment variables as they have a higher precedence
 # than the defaults defined here.
-KAFKA_BOOTSTRAP_HOST: ${clowder.kafka.brokers[0].hostname:localhost}
-KAFKA_BOOTSTRAP_PORT: ${clowder.kafka.brokers[0].port:9092}
 KAFKA_USERNAME: ${clowder.kafka.brokers[0].sasl.username:}
 KAFKA_PASSWORD: ${clowder.kafka.brokers[0].sasl.password:}
```
awood, line 12: Do we need to worry about these properties have multiple values too?  I don't have the clowder JSON in front of me, but just want to confirm we won't have a problem trying to connect to `broker[1]` with the username for `broker[0]`

awood (changes requested):
This looks great, but I just want to confirm we won't have problems with the SASL credentials.

swatch-core/src/main/resources/swatch-core/application.yaml
Sgitario: Very fair point indeed! In my first version, I also ported the SASL properties, but then I thought I went too far. 
But if you're ok, I'm updating the PR including these changes as well.

Sgitario:
@awood I already updated the pull request. Can you take another look please?

swatch-core/src/main/resources/swatch-core/application.yaml
```diff
@@ -8,24 +8,8 @@
...
-# This needs to be a PEM encoded certificate string.  If you want to point to a file you can use
-# spring.kafka.ssl.trust-store-location property but that property is mutually exclusive with the
-# spring.kafka.ssl.trust-store-certificates property.  Since Clowder gives us the actual
-# certificate PEM, we have to use the trust-store-certificates property here.
-#
-# The value of this property must match a regular expression defined in Kafka's
-# DefaultSslEngineFactory.PemParser class.  The regular expression is (with Java escapes)
-# -+BEGIN\\s*.*CERTIFICATE[^-]*-+\\s+(?:\\s*[^\\r\\n]*:[^\\r\\n]*[\\r\\n]+)*([a-zA-Z0-9/+=\\s]*)-+END\\s*.*CERTIFICATE[^-]*-+\\s+")
-# A notable point about this expression is that the "----BEGIN CERTIFICATE-----" piece must be
-# followed by a newline which makes it challenging to place the value in a properties or YAML file.
```
awood, line 29: I'd like to retain this comment down below in the `kafka.properties.ssl` YAML section since it's still relevant.

awood (changes requested):

swatch-core/src/test/resources/test-clowder-config.json
```diff
@@ -31,7 +31,19 @@
...
+                    "username": "john",
+                    "password": "doe",
+                    "securityProtocol":"SASL_SSL",
+                    "saslMechanism":"PLAIN"
+                }
+            },
+            {
+                "hostname": "env-rhsm-kafka-secondary.rhsm.svc",
+                "port": 29093
             }
```
awood, line 47: Is this what the multi-broker configuration looks like in practice?  And if so, my reading of the code is that the SASL bit from the first item in the list will be applied to the second item in the list.  That's fine, but I just want to have some comments to that effect in the PropertySource class.

swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySource.java
```diff
@@ -273,4 +376,46 @@ private String findPropertySourceInsertionPoint(MutablePropertySources sources)
...
+    }
+
+    return null;
+  }
+
+  private static Map<String, Object> getKafkaBrokerWithSasl(
+      List<Map<String, Object>> brokerConfig) {
+    return brokerConfig.stream()
+        .filter(m -> "sasl".equals(m.get("authtype")))
+        .findFirst()
```
awood, line 415: I'd like to be a little stricter here.  I.e. throw an exception if there is more than one SASL configuration given.  I want to fail explicitly and fast if we get JSON like

```json
           {
                "hostname": "env-rhsm-kafka.rhsm.svc",
                "port": 29092
                "port": 29092,
                "cacert": "Dummy value",
                "authtype": "sasl",
                "sasl": {
                    "username": "john",
                    "password": "doe",
                    "securityProtocol":"SASL_SSL",
                    "saslMechanism":"PLAIN"
                }
            },
           {
                "hostname": "env-rhsm-kafka-2.rhsm.svc",
                "port": 29092
                "port": 29092,
                "cacert": "A different cert",
                "authtype": "sasl",
                "sasl": {
                    "username": "jane",
                    "password": "smith",
                    "securityProtocol":"SASL_SSL",
                    "saslMechanism":"PLAIN"
                }
            }
```

There's a Guava collector [`toOptional`](https://javadoc.scijava.org/Guava/com/google/common/collect/MoreCollectors.html#toOptional()) that is useful for this purpose.

awood (changes requested):

swatch-core/src/main/resources/swatch-core/application.yaml
Sgitario: Re-added it to the relevant property in spring.kafka.properties.ssl.trust-store-certificates

swatch-core/src/test/resources/test-clowder-config.json
Sgitario: Added a comment in ClowderJsonPathPropertySource.KAFKA_BROKERS_PROPERTIES.

swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySource.java
Sgitario: I don't follow what is the value that the Guava collector toOptional brings here (we could simply don't use the `orElse`). Also, I think the code looks more readable without using optionals, and the methods are private and meant to be used internally only). 

About being stricter here, this is a change of behavior since previously, this is how it was working and also it's how it works the Clowder config source for Quarkus. We can always do this change in a later PR if there is a need.

Sgitario:
@awood I've added the comments you've requested. Can you take another look at the changes? Thanks!

awood (approved):
Nice work!

PR #2818, opened 2023-11-23
SWATCH-1948: Catch ProcessingException and ApiException exceptions

Sgitario:
Jira issue: SWATCH-1948

## Description
Catching the two exceptions is only meant to be caught for Quarkus rest clients because it will throw ProcessingException instead of ApiException. 

The problem is that the generated sources by the Quarkus REST-Client Reactive extension can only deal with runtime exceptions because it uses reactive flows, however the generated ApiException class by the OpenApi plugin is not a RuntimeException, but Exception.

Extended context in the JIRA ticket. 

## Testing
No QE is needed here.


mstead (approved):
Looks good.

PR #2827, opened 2023-11-28
SWATCH-780: springboot-based services: consume configuration for TLS

Sgitario:
Jira issue: [SWATCH-780](https://issues.redhat.com/browse/SWATCH-780)

## Description
Update our clowder config source to mimic the logic that the clowder config source from Quarkus: https://github.com/RedHatInsights/clowder-quarkus-config-source
After updating our clowder config source for Spring, the usage is exactly the same as in Quarkus. For example: `${clowder.endpoints.<dependency name>-service.url}` instead of `http://${clowder.endpoints.capacity-ingress.hostname}:${clowder.endpoints.capacity-ingress.port}`. 
And also to properly declare the dependencies.

Note that this strongly depends on the ClowdEnvironment object to be updated with:

```
{
  "spec": {
    "providers": {
      "web": {
        "tls": {
          "enabled": true,
          "port": 8800
        }
      }
    }
  }
}
```

With `tls.enabled=true`, the envoy sidecar will be created for all the clowd apps. 
However, if the `tls` flag is not enabled, everything will keep working as usual but using HTTP. The clowder config source will see if the `tls` flag is enabled or not, and configure the rest client accordingly. 

## Testing
Ephemeral namespace already has the "tls.enabled=true" flag, so running the QE test suite to confirm everything works as before should be enough.

I've added a trace in all the services where the endpoint will be printed. For example:

```
023-11-28 10:25:18,375 [thread=main] [INFO ] [org.candlepin.subscriptions.clowder.ClowderJsonPathPropertySource] - Endpoint 'swatch-subscription-sync-service' using 'https://swatch-subscription-sync-service.ephemeral-romvrc.svc:8800'
2
```

For manual deployment:

1. `bonfire namespace reserve --pool default` -> output should be the oc project name
2. `oc project <project name>`
3. `oc patch env env-<project name> -p '{"spec":{"providers":{"sidecars":{"tokenRefresher":{"enabled":true}}, "web":{"tls":{"enabled": true, "port": 8800}}}}}' --type=merge`
4. `bonfire deploy rhsm --source=appsre --namespace <project name> ...`


ntkathole:
I think PR needs to rebase with main to fix test failure

PR #2839, opened 2023-11-30
SWATCH-1978: remove deprecated "/hosts/products/{product_id}" endpoint

Sgitario:
Jira issue: [SWATCH-1978](https://issues.redhat.com/browse/SWATCH-1978)

## Description
remove deprecated hosts API endpoint "__/hosts/products/{product_id}"

## Testing
Regression tests only + tests using this endpoint should be updated.


src/main/java/org/candlepin/subscriptions/resource/InstancesResource.java
```diff
@@ -79,6 +80,17 @@ public class InstancesResource implements InstancesApi {
   private final HostRepository hostRepository;
   private final PageLinkCreator pageLinkCreator;
 
+  @SuppressWarnings("linelength")
+  public static final Map<HostReportSort, String> HOST_SORT_PARAM_MAPPING =
```
kahowell, line 84: Unneeded.

api/rhsm-subscriptions-api-spec.yaml
```diff
@@ -586,126 +586,6 @@ paths:
...
-          schema:
-            type: string
-            format: date-time
-          description: "Defines the end of the report period. Dates should be provided in ISO 8601
-            format but the only accepted offset is UTC. E.g. 2017-07-21T17:32:28Z.
-            Only applicable for OpenShift-dedicated-metrics and OpenShift-metrics products"
-        - name: sort
-          in: query
-          schema:
-            $ref: "#/components/schemas/HostReportSort"
```
kahowell, line 658: Please remove this unused schema (HostReportSort) too.

src/test/java/org/candlepin/subscriptions/db/HostRepositoryTest.java
```diff
@@ -509,7 +509,7 @@ void testCanSortByDisplayName() {
                 0,
                 10,
                 Sort.Direction.ASC,
-                HostsResource.HOST_SORT_PARAM_MAPPING.get(HostReportSort.DISPLAY_NAME)));
+                InstancesResource.HOST_SORT_PARAM_MAPPING.get(HostReportSort.DISPLAY_NAME)));
```
kahowell, line 512: These should be updated to INSTANCE_SORT_PARAM mappings (if there are any that don't map we should remove the resulting dead code).

kahowell (changes requested):

api/rhsm-subscriptions-api-spec.yaml
Sgitario: Thanks for spotting!

src/main/java/org/candlepin/subscriptions/resource/InstancesResource.java
Sgitario: This was needed by some tests in HostRepositoryTest.
However, these tests were verifying the sort feature from Spring, so I don't think we really need them. Just delete this HOST_SORT_PARAM_MAPPING field and these tests.

Sgitario:
@kahowell PR updated after your comments.

kahowell (approved):

kahowell:
> Looks good. The test refers to the old host api so it's failing and is fixed by the below PR.

@Aurobinda55 can we go ahead and get that merged?

Aurobinda55 (approved):

PR #2850, opened 2023-12-04
Add script to generate new changeset xml

kahowell:
## Description

For a recent task I found myself wanting to prototype db changes quickly.

Invoked as `bin/liquibase-new-changeset.sh change set description`.

It takes any args passed, and lower cases them, and replaces spaces with `-`, and then adds the new changeset to our changelog.xml file.


Sgitario (approved):
lgtm

bin/liquibase-new-changeset.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
Sgitario, line 1:
```suggestion
#!/bin/bash
# Usage: liquibase-new-changeset.sh <text to append to the generated changeset>
# For example: running `liquibase-new-changeset.sh remove-account` would create a change set `202312050855-remove-account.xml` where `202312050855` is auto-populated from the current date.
```

Sgitario (commented):
Added a suggestion to give an example in the script.
Moreover, I think it would be a good idea to mention the script in CONTRIBUTING.md (section "DB migrations").

PR #2851, opened 2023-12-04
SWATCH-1924: Support querying prometheus/rhelemeter when "product" label has engIds

lindseyburnett:
Jira issue: SWATCH-1924

**Context**
Getting creative with how we can detect these RHEL payg products with rhelemeter, since the role available on the hosts isn't unique enough to differentiate between different variants.

This change enables hosts to report a list of comma-separated engineering Ids in its "product" label. We'll detect comma-separated engineering Ids and set the swatch event.productIds field accordingly.

During hourly tally, we already match events to swatch subscription definitions by using event.role or event.productIds anyway.

# summary

## promql query changes
Introduced a new promql query template.  There's a couple of differences than the default template
* removes "metadataMetric" parameter
* adds "productLabelRegex" parameter.
* removes instanceKey parameter (note that instanceKey still needs to be defined on the product config files because it's being used programmatically elsewhere)

The rhel-for-x86-els-payg product configuration yaml has the regex defined as any comma-separated list that includes "204".

## Event message updates
Leverage changes made in #2844 and populate `productTag` field when emitting a metric Event on the service instance topic.  This is a little redundant, because the ingestion of metric Events (takes place in swatch-tally) looks up the productTag based on productIds & role and will re-set the value.  There's room for improvement here to potentially stop looking up productTag if an Event has a productTag set in the payload already. 

Populate `productIds` field when emitting a metric Event on the service instance topic.  The value comes from the prometheus response in the "product" payload.  As previously mentioned, during Event ingestion there's a productTag lookup.  Since there isn't a role that corresponds to rhel-for-x86-els-payg, that productTag can only be resolved by having productIds present.

## misc

While working on these changes, I encountered an issue with the timeframes that recording rules are evaluated.  There are non-trivial updates required for this, as detailed in SWATCH-2075.  As a workaround, I've set the `metric` field in the product config file to use the expression that backs the recording rule.  This workaround is to enable testing, and isn't a long-term solution.  We ultimately don't want to be attempting to maintain usage queries that are defined/crafted/owned by other product teams.

# expected behavior
* a POST request to `/api/swatch-metrics/v1/internal/metering/rhel-for-x86-els-payg` successfully retrieves data from rhelemeter
* swatch-metrics (or swatch-metrics-rhel if you're in EE) emits Event messages for the metrics returned from rhelemeter
* swatch-tally successfully consumes those message, resulting in entries in the rhsm-subscriptions.events database table
* a PUT request to `/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots` should result in snapshots

# test steps

these are written for local, but similar approach should work in ephemeral by port forwarding pods.  token refresher will already be attached to the swatch-metrics-rhel pod.

### 1. start up a token-refresher for stage rhelemeter

Secrets can be found in vault under _secrets/insights-stage/rhsm-stage/rhel-token-refresher_

```
podman run --name=token-refresher-$1 --rm -ti -p 8082:8080 \
  quay.io/observatorium/token-refresher:master-2023-09-20-f5e3403 \
  --scope="$SCOPE" \
  --oidc.client-secret="$CLIENT_SECRET" \
  --oidc.client-id="$CLIENT_ID" \
  --oidc.issuer-url="$ISSUER_URL" \
  --url="https://observatorium-mst.api.stage.openshift.com/api/metrics/v1/rhel"
  ```

### 2. start up swatch-metrics-rhel
```bash
EVENT_SOURCE=rhelemeter PROM_URL="http://localhost:8082/api/v1/" ./gradlew :swatch-metrics:quarkusDev
```
### 3. kick off metering for test org
```bash
curl -X POST --location "http://localhost:8000/api/swatch-metrics/v1/internal/metering/rhel-for-x86-els-payg?orgId=16787820" \
    -d 'accept: */*
x-rh-swatch-synchronous-request: false'
```

if you're running this at DEBUG log level, you should see some log statements containing details about the range query being done against prometheus - promql, step, and start/end ranges.

### 5. start up swatch-tally and verify it consumes the messages
`SERVER_PORT=8001 DEV_MODE=true ./gradlew :bootRun`

you should see messages getting consumed after it starts up.

after the events are consumed, check the `data` column in the `events` table that the information from the Event message persisted properly (had issues at one point with productTag getting blown away before persisting)

### 6. kick off hourly tally
will have to adjust based on timestamps you're doing this

```
curl -X 'PUT' \
'http://localhost:8001/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots?begin=2024-01-08T19%3A00%3A00Z&end=2024-01-08T23%3A00%3A00Z' \
-H 'accept: application/json' \
-H 'x-rh-swatch-psk: placeholder'
```

### DIFF
--- a/swatch-metrics/src/main/java/com/redhat/swatch/metrics/service/PrometheusMeteringController.java
+++ b/swatch-metrics/src/main/java/com/redhat/swatch/metrics/service/PrometheusMeteringController.java
@@ -158,7 +158,8 @@ public class PrometheusMeteringController {
               metricProperties.step(),
               metricProperties.queryTimeout(),
               item ->
-                  sendEventFromData(item, eventsSent, tag, orgId, meteringBatchId, metric, subDef));
+                  createEventFromDataAndSend(
+                      item, eventsSent, tag, orgId, meteringBatchId, metric, subDef));
 
       if (StatusType.ERROR.equals(metricData.getStatus())) {
         throw new MeteringException(
@@ -183,7 +184,7 @@ public class PrometheusMeteringController {
     }
   }
 
-  private void sendEventFromData(
+  private void createEventFromDataAndSend(
       QueryResultDataResultInner item,
       Set<EventKey> eventsSent,
       String productTag,
@@ -250,7 +251,7 @@ public class PrometheusMeteringController {
       // Send if and only if it has not been sent yet.
       // Related to https://github.com/RedHatInsights/rhsm-subscriptions/pull/374.
       if (eventsSent.add(EventKey.fromEvent(event))) {
-        send(event);
+        sendToServiceInstanceTopic(event);
       }
     }
   }
@@ -262,7 +263,7 @@ public class PrometheusMeteringController {
       OffsetDateTime start,
       OffsetDateTime end,
       UUID meteringBatchId) {
-    send(
+    sendToServiceInstanceTopic(
         createCleanUpEvent(
             orgId,
             MeteringEventFactory.getEventType(tagMetric.getId(), productTag),
@@ -311,10 +312,12 @@ public class PrometheusMeteringController {
     return event;
   }
 
-  private void send(BaseEvent event) {
+  private void sendToServiceInstanceTopic(BaseEvent event) {
     if (event instanceof Event eventToSend) {
       log.debug(
-          "Sending event {} for organization {}", eventToSend.getEventId(), eventToSend.getOrgId());
+          "Sending event with id {} for organization {}",
+          eventToSend.getEventId(),
+          eventToSend.getOrgId());
     } else if (event instanceof CleanUpEvent) {
       log.debug("Sending clean-up event for organization {}", event.getOrgId());
     }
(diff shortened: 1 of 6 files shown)


swatch-metrics/src/main/java/com/redhat/swatch/metrics/service/PrometheusMeteringController.java
```diff
@@ -198,11 +202,19 @@ private void sendEventFromData(
     String product = labels.get("product");
     String resourceName = labels.get("resource_name");
 
+    // TODO update this note
```
lindseyburnett, line 205: Looking for some assistance on how to wordsmith this note update.

swatch-product-configuration/src/main/resources/subscription_configs/RHEL/RHEL_for_x86_els_payg.yaml
```diff
@@ -29,9 +27,8 @@ metrics:
     awsDimension: vCPU_hours
     azureDimension: vcpu_hours
     prometheus:
-      queryKey: default
+      queryKey: rhelemeter
       queryParams:
+        productLabelRegex: .*(^|,)(204)($|,).*
```
Sgitario, line 32: Can you add a comment about what the intention with this value is and where the value 204 comes from?
This is used by `product=~"#{metric.prometheus.queryParams[productLabelRegex]}"`, so I guess you want to only fetch metrics for product that contains the product ID 204, right?

Sgitario (commented):
Changes look good to me. 
The test steps to verify the changes worked for me.
I only added a few minor comments.

ssidhaye (approved):
- a metering POST request successfully retrieves data from rhelemeter
- swatch-metrics (or swatch-metrics-rhel if you're in EE) emits Event messages for the metrics returned from rhelemeter
- swatch-tally comsumes messages from above steps and populates events table (product_tag and product_ids) 
- a PUT request to /api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots results in snapshots

PR #2641, opened 2023-10-12
SWATCH-1810 Do not send usage to AWS when too old

mstead:
Jira issue: [SWATCH-1810](https://issues.redhat.com/browse/SWATCH-1810)

## Description
The AWS Marketplace API will only accept usage in which
the timestamp must be on or after the START_OF_CURRENT_HOUR - 6h.

In order to avoid sending requests to the AWS API that we know will
fail, the AWS producer should check that the incoming usage is within
this window before sending the usage to AWS.

If the usage is outside of this window, the usage should not
be sent to AWS, a WARNING message should be logged, and an
'ignored' micrometer counter should be incremented.

An AWS_MARKETPLACE_USAGE_WINDOW configuration property was added to
allow us to adjust our pre-send window quickly if we ever need to.
The default value for this property is 6h.

## Example
Let's say that it is currently 2023-10-26T13:30:00Z. Any message that has a timestamp that is on or after 2023-10-26T07:00:00Z will be accepted. Therefore, a message sent with a timestamp of 2023-10-26T07:00:00Z  would be accepted up until the current time is 2023-10-26T14:00:00Z.

### Setup
1. Refresh the database
```
dropdb -U rhsm-subscriptions rhsm-subscriptions && createdb -U rhsm-subscriptions rhsm-subscriptions
```
2. Launch the subscriptions API, running on a different port than the AWS producer will be.
```
DEV_MODE=true SERVER_PORT=8101 ./gradlew :bootRun
```
3. Load a subscription into the DB.
```
psql -U rhsm-subscriptions rhsm-subscriptions -c "insert into subscription (sku, org_id, subscription_id, quantity, start_date, end_date, billing_provider_id, subscription_number, billing_provider, billing_account_id) values ('BASILISK', 'org123', '111111112', 1, '2022-10-10T04:00:00Z', '2024-10-10T04:59:59.999Z', 'basilisk;customer1;awsSeller1', '222222222', 'aws', '123456');"

```

1. Launch the AWS billing producer, configuring it to point at the internal subscription API above.
```
SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT=http://localhost:8101/api/rhsm-subscriptions/v1  ENABLE_AWS_DRY_RUN=true ./gradlew swatch-producer-aws:quarkusDev
```

### Steps
**In the following examples, it is currently 18:40 UTC. Be sure to update the date/time of the examples below.**

[...]

PR #2856, opened 2023-12-05
Add example payload to openapi spec for POST /internal/subscription

lindseyburnett:
swagger ui just showed "string" as the example, since it's not defined by a schema.  we programatically map the string to another openapi generated "Subscription" object.  We should change that eventually, but in the meantime I added an example payload.  I pulled the example out of a PR comment - https://github.com/RedHatInsights/rhsm-subscriptions/pull/2641#pullrequestreview-1705980261 

Also marking `reconcileCapacity` queryParam as required so swagger-ui makes you set it, otherwise you'll get a NPE in the logs

Verification
- on main branch, see that http://localhost:8000/api/swatch-subscription-sync/internal/swagger-ui/index.html#/internalSubscriptions/saveSubscriptions just had "string"
- on this branch, see that there's an example payload that should actually POST successfully once you set the the `x-rh-swatch-psk` header to `placeholder`

### DIFF
--- a/src/main/spec/internal-subscriptions-sync-api-spec.yaml
+++ b/src/main/spec/internal-subscriptions-sync-api-spec.yaml
@@ -42,6 +42,33 @@ paths:
           application/json:
             schema:
               type: string
+            examples:
+              subscription:
+                "value": [
+                  {
+                    "id": "000149",
+                    "subscriptionNumber": "8838431",
+                    "subscriptionProducts": [
+                      {
+                        "sku": "MCT4121HR"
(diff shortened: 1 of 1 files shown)


Sgitario (changes requested):
I think the `repro.http` file addition was not intentional (not part of this pull request).

lindseyburnett:
> I think the `repro.http` file addition was not intentional (not part of this pull request).

good catch :laughing:

Sgitario (approved):
lgtm

PR #2876, opened 2023-12-08
SWATCH-1579 Remove retally concept and batch process events

mstead:
Jira issue: SWATCH-1579

## Description
This patch completely removes the concept of an hourly retally
from the hourly tally process. When the hourly tally process runs,
Events are batch processed based on the Event.record_date,
instead of relying on a date range.

Events are fetched based on the last event that was tailled
and are processed in batches (configurable). A tally_state
table was added to track the last known Event.record_date
that was successfully processed, and is used to limit the
Event query when the tally operation occurs.

The API date range parameters have been removed. When the
hourly tally operation runs, it will process any events that
have not yet been processed.

As well, the overall hourly tally host and calculation process
has changed. Events with the same timestamp are now applied
accumulatively. For example, two events with the same
measurement for 10am will result in a 10am hourly snapshot
with a value equals to the sum of the measurements of both
events.

This makes it possible to perform ammendments on a Host's
monthly totals and affected account Snapshots directly,
by applying negative measurement values to a host.

In order to support retry-on-error via the RetryTemplate,
host updates via events needed to be idempotent. This was
accomplished by adding a lastAppledEventRecordDate to the
Host record. Since Events are processed in order by
record_date, this allows for determining if an Event had
already been processed successfully. Also, an
AccountUsageCalculationCache was created to track the
calculations that were computed as an Event batch is
processed. On retry, the event record date is again used
to determine if the Event was already applied to a
calculation.

NOTE:
* A default batch size of 16000 Events is used to keep memory
  usage at a reasonable level.
* When an account is reset, all TallyState records will be
  removed.

### Determine Tally Data Baseline From A Set Of Events

1.Check out the main branch.

2. Drop the DB and deploy.
```
dropdb -U rhsm-subscriptions rhsm-subscriptions && \
createdb -U rhsm-subscriptions rhsm-subscriptions
```
```
SPRING_PROFILES_ACTIVE=api,worker,kafka-queue DEV_MODE=true ./gradlew :bootRun
```

3. I pieced a raw python script together that can import Event records directly into the DB. Download the attached [import-events-into-db.py.txt](https://github.com/RedHatInsights/rhsm-subscriptions/files/13620564/import-events-into-db.py.txt) and import a base set of events. 

The following will import an event for 5 different OpenShift instances on '2023-12-05T17:00:00+00:00'.
```
EXCLUDE_RECORD_DATE_FROM_JSON=True ./import-events-db.py --org org123 --instances 5 --start '2023-12-05T11:00:00+00:00' --end '2023-12-05T11:00:00+00:00'
```

4. Run the hourly tally. **Because the current tally relies on the record_date when tallying, ensure that the end date is the current hour**
```
http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=org123&start=2023-12-05T00:00Z&end=2023-12-08T23:00Z" x-rh-swatch-psk:placeholder
```

5. Capture hosts and tallied data. Run the following in an isolated folder so that the resulting files can be diffed later. You can copy/paste them all and run them at the same time.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select inventory_id, insights_id, org_id, display_name, subscription_manager_id, is_guest, hypervisor_uuid, hardware_type, num_of_guests, last_seen, is_unmapped_guest, is_hypervisor, cloud_provider, instance_type, instance_id, billing_provider, billing_account_id from hosts order by instance_id" > main-initial-tally-hosts;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, m.metric_id, m.value from hosts h, instance_measurements m where h.id = m.host_id order by h.instance_id;" > main-initial-tally-instance-measurements;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, t.month, t.metric_id, t.value from hosts h, instance_monthly_totals t where h.id = t.host_id order by h.instance_id" > main-initial-tally-instance-monthly-totals;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, b.usage, b.product_id, b.sla, b.as_hypervisor, b.cores, b.sockets, b.measurement_type, b.billing_provider, b.billing_account_id, version from hosts h inner join host_tally_buckets b on h.id = b.host_id order by h.instance_id, b.host_id, b.usage, b.sla, b.billing_provider, b.billing_account_id" > main-initial-tally-buckets;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select s.snapshot_date, s.granularity, s.usage, s.sla, s.billing_provider, s.billing_account_id, m.measurement_type, m.metric_id, m.value from tally_snapshots s inner join tally_measurements m on s.id = m.snapshot_id order by s.snapshot_date, s.granularity, s.usage, s.sla, s.billing_provider, s.billing_account_id, m.measurement_type, m.metric_id;" > main-initial-tally-snapshots-with-measurements;
```

### Test And Verify The New Tally Logic
1. Check out the branch for this PR and deploy with a custom event `batch size of 2` to ensure that we trigger the batching functionality.
```
HOURLY_TALLY_EVENT_BATCH_SIZE=2 \
SPRING_PROFILES_ACTIVE=api,worker,kafka-queue DEV_MODE=true \
./gradlew :bootRun
```

**CHECK THE DB MIGRATIONS**
1. Verify that the events table PK is now simply the `event_id` and that there is no longer a unique constraint.
```
psql -U rhsm-subscriptions rhsm-subscriptions -c "\d events"
```
> "events_pkey" PRIMARY KEY, btree (org_id, event_type, event_source, instance_id, record_date, "timestamp")

2. Verify that any existing events have the `record_date` property copied into the JSON data. **The values in both columns should contain the same timestamps**
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions \
-c "select record_date,
data->>'record_date' as copied_record_date from events;"
```
```
          record_date          |        copied_record_date        
-------------------------------+----------------------------------
 2023-12-08 18:56:06.687489+00 | 2023-12-08T18:56:06.687489+00:00
 2023-12-08 18:56:06.687767+00 | 2023-12-08T18:56:06.687767+00:00
 2023-12-08 18:56:06.687934+00 | 2023-12-08T18:56:06.687934+00:00
 2023-12-08 18:56:06.688081+00 | 2023-12-08T18:56:06.688081+00:00
 2023-12-08 18:56:06.688223+00 | 2023-12-08T18:56:06.688223+00:00
(5 rows)

```

3. Verify that a new lookup index was added to the  `events` table.
```
psql -U rhsm-subscriptions rhsm-subscriptions -c "\d events" | grep orgid_servicetype_record_date_idx
```
>  "orgid_servicetype_record_date_idx" btree (org_id, (data ->> 'service_type'::text), record_date)

4. Verify that the tally_state table was created and was populated with an entry per existing service type for each account with Events in the table. The `latest_event_record_date` was defaulted to NOW().

```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions \
-c "select * from tally_state"
```
>   org_id |   service_type    |   latest_event_record_date    
--------+-------------------+-------------------------------
 org123 | OpenShift Cluster | 2023-12-08 19:01:39.673134+00
(1 row)

5. Verify the `last_applied_event_record_date` column was added to the hosts table.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions \
-c "\d hosts" | grep last_applied_event_record_date
```
> last_applied_event_record_date | timestamp with time zone

6. Verify that a new lookup index was added to the `hosts` table.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions \
-c "\d hosts" | grep host_org_id_instance_id_idx
```
> "host_org_id_instance_id_idx" btree (org_id, instance_id)

#### TEST THAT THERE'S NO TALLY OF EXISTING EVENTS AFTER INITIAL DEPLOY

Running the tally immediately after deployment should not trigger a new tally since there should not be any Events found that have a  `record_date` greater than the `lastProcessedEventRecordDate` on the TallyState record.

NOTE:
* The date range API parameters are no longer availiable on the nightly tally API.
* Logs should show that the tally for `OpenShift Cluster` was attempted, but should not log about updating snapshots.

```
http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=org123" x-rh-swatch-psk:placeholder
```
> ...
> Producing hourly snapshots for orgId org123 for service type OpenShift Cluster
> Finished producing OpenShift Cluster hourly snapshots for orgId org123
> ...

#### TEST HOURLY TALLY
1. Reset the DB so that only the events exist, re-run the tally and verify that the DB ended up in the same state. Note that we manually reset the date on the TallyState record so that the events will be picked up during the tally.
```
psql -U rhsm-subscriptions rhsm-subscriptions \
-c "truncate hosts cascade; truncate tally_snapshots cascade; truncate billable_usage_remittance cascade; update tally_state set latest_event_record_date = '2023-11-28 14:00:00+00'"
```

1. Run the hourly tally.
```
http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=org123" x-rh-swatch-psk:placeholder
```
> Producing hourly snapshots for orgId org123 for service type OpenShift Cluster
> ...
> Produced 32 TallySummary messages

2. Verify that tallying with the baseline events should result in the same data. The following can be copied and run all at once.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select inventory_id, insights_id, org_id, display_name, subscription_manager_id, is_guest, hypervisor_uuid, hardware_type, num_of_guests, last_seen, is_unmapped_guest, is_hypervisor, cloud_provider, instance_type, instance_id, billing_provider, billing_account_id from hosts order by instance_id" > branch-initial-tally-hosts;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, m.metric_id, m.value from hosts h, instance_measurements m where h.id = m.host_id order by h.instance_id;" > branch-initial-tally-instance-measurements;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, t.month, t.metric_id, t.value from hosts h, instance_monthly_totals t where h.id = t.host_id order by h.instance_id" > branch-initial-tally-instance-monthly-totals;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, b.usage, b.product_id, b.sla, b.as_hypervisor, b.cores, b.sockets, b.measurement_type, b.billing_provider, b.billing_account_id, version from hosts h inner join host_tally_buckets b on h.id = b.host_id order by h.instance_id, b.host_id, b.usage, b.sla, b.billing_provider, b.billing_account_id" > branch-initial-tally-buckets;

PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select s.snapshot_date, s.granularity, s.usage, s.sla, s.billing_provider, s.billing_account_id, m.measurement_type, m.metric_id, m.value from tally_snapshots s inner join tally_measurements m on s.id = m.snapshot_id order by s.snapshot_date, s.granularity, s.usage, s.sla, s.billing_provider, s.billing_account_id, m.measurement_type, m.metric_id;" > branch-initial-tally-snapshots-with-measurements;
```

3. Diff each of the resulting `branch-*` files with the corresponding `main-*` files. They should be the same.

**IMPORTANT** Bug https://issues.redhat.com/browse/SWATCH-2009 might affect the measurement values when diffing the tally snapshot measurements.

```
diff main-initial-tally-hosts branch-initial-tally-hosts

diff main-initial-tally-buckets branch-initial-tally-buckets

diff main-initial-tally-instance-measurements branch-initial-tally-instance-measurements

diff main-initial-tally-instance-monthly-totals branch-initial-tally-instance-monthly-totals

diff main-initial-tally-snapshots-with-measurements branch-initial-tally-snapshots-with-measurements
```

#### ADD AN EVENT THAT IS A MONTH OLD FOR AN EXISTING INSTANCE
In this scenario, the instance should have its monthly totals updated, but should NOT the instance_measurements or basic meta-data, since a more recent representation of the host has already been processed (Event.timestamp < Host.lastSeen). The value should be correctly applied to the snapshots because the record_date implys that the Event was not yet tallied in any way.

**This event WILL however, get applied to the overall tally snapshots.**

1. Find an existing instance_id from a single Host. Look for YOUR_INSTANCE_ID in the commands below and be sure to replace it with your instance_id.
```
psql -U rhsm-subscriptions rhsm-subscriptions -c "select instance_id from hosts where org_id='org123' limit 1"
```

2. Insert a single event that is in a previous month with the instance_id above. Specify a measurement value of 100 so that we can be sure that it gets applied.
```
./import-events-db.py --org org123 --instance-id YOUR_INSTANCE_ID --start '2023-11-05T12:00:00+00:00' --end '2023-11-05T12:00:00+00:00' --value 100.0
```

3. Run the hourly tally.
```
http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=org123" x-rh-swatch-psk:placeholder
```

4. Verify that the instance_measurements value remains as the previous value and does NOT get updated to 100. You can find the existing value in the inital data captured earlier by running:
```
cat branch-initial-tally-instance-measurements | grep 'YOUR_INSTANCE_ID'
```
> 25904f11-6073-4766-911a-197508f43c45 | CORES     |   3.4

The new value can be found as follows:
```
psql -U rhsm-subscriptions rhsm-subscriptions \
-c "select h.instance_id, m.metric_id, m.value from hosts h, instance_measurements m where h.id=m.host_id and h.org_id='org123' and h.instance_id='YOUR_INSTANCE_ID'"
```
> 25904f11-6073-4766-911a-197508f43c45 | CORES     |   3.4

5. Verify that the monthly totals include a NEW entry for `2023-11` matching your instance_id with a value of 100. You should also see a total for `2023-12` which matches the previous value (3.4 in my case). There should also be entries for the other instances for `2023-12` from the initial tally.
```
psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, t.month, t.metric_id, t.value from instance_monthly_totals t, hosts h where t.host_id=h.id;"
```

6. Ensure that the tally snapshots and measurements reflect the addition of 100 Cores. First determine the previous value.
```
cat branch-initial-tally-snapshots-with-measurements | grep "| DAILY       | _ANY       | _ANY    | _ANY             | _ANY               | TOTAL"
```
Result:
```
2023-12-04 20:00:00-04 | DAILY       | _ANY       | _ANY    | _ANY             | _ANY               | TOTAL            | CORES     |  14.8

```

7. Verify that there is a new DAILY snapshot for November with a value of 100 and corresponding MONTHLY. There should also be a set of December snapshots with the total from after the initial tally.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select s.snapshot_date, s.granularity, s.usage, s.sla, s.billing_provider, s.billing_account_id, m.measurement_type, m.metric_id, m.value from tally_snapshots s inner join tally_measurements m on s.id = m.snapshot_id where s.usage='_ANY' and s.sla='_ANY' and s.billing_provider='_ANY' and s.billing_account_id='_ANY' and m.measurement_type='TOTAL' order by s.snapshot_date, s.granularity"
```
 Results:
```
     snapshot_date      | granularity | usage | sla  | billing_provider | billing_account_id | measurement_type | metric_id | value 
------------------------+-------------+-------+------+------------------+--------------------+------------------+-----------+-------
 2023-11-05 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   100
 2023-11-05 12:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   100
 2023-12-05 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |  14.8
 2023-12-05 11:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |  14.8
```

#### ADD A RECENT EVENT FOR AN EXISTING INSTANCE
In this scenario, the instance should be updated in full and the usage should get applied to the existing snapshots.

1. Insert a single event for the instance_id above, in which the event timestamp is current. Specify a measurement value of 200 so that we can be sure that it gets applied. Make sure that the start/end dates are AFTER the inial event creation from earlier.
```
./import-events-db.py --org org123 --instance-id YOUR_INSTANCE_ID --start '2023-12-08T12:00:00+00:00' --end '2023-12-08T12:00:00+00:00' --value 200.0
```

2. Run the hourly tally.
```
http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=org123" x-rh-swatch-psk:placeholder
```

3. Verify that the instance_measurements value is now 200.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions \
-c "select h.instance_id, m.metric_id, m.value from hosts h, instance_measurements m where h.id=m.host_id and h.org_id='org123' and h.instance_id='YOUR_INSTANCE_ID'"
```

4. Verify that the monthly totals entry for `2023-12` is incremented by a value of 200 (203.4 in my case). The total for `2023-11` should have remained at 100.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, t.month, t.metric_id, t.value from instance_monthly_totals t, hosts h where t.host_id=h.id and h.instance_id='YOUR_INSTANCE_ID';"
```

5. Ensure that the tally snapshots contain a new HOURLY and DAILY snapshot for the Event.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select s.snapshot_date, s.granularity, s.usage, s.sla, s.billing_provider, s.billing_account_id, m.measurement_type, m.metric_id, m.value from tally_snapshots s inner join tally_measurements m on s.id = m.snapshot_id where s.usage='_ANY' and s.sla='_ANY' and s.billing_provider='_ANY' and s.billing_account_id='_ANY' and m.measurement_type='TOTAL' order by s.snapshot_date, s.granularity"
```
```
     snapshot_date      | granularity | usage | sla  | billing_provider | billing_account_id | measurement_type | metric_id | value 
------------------------+-------------+-------+------+------------------+--------------------+------------------+-----------+-------
 2023-11-05 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   100
 2023-11-05 12:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   100
 2023-12-05 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |  14.8
 2023-12-05 11:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |  14.8
 2023-12-08 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   200
 2023-12-08 12:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   200
(6 rows)
```

#### AMEND AN EXISTING TALLY SNAPSHOT
1. Ammend the usage we just created by adding an event that subtracts the existing value (200), and another that sets it to 500.
```
./import-events-db.py --org org123 --instance-id YOUR_INSTANCE_ID --start '2023-12-08T12:00:00+00:00' --end '2023-12-08T12:00:00+00:00' --value -200.0
```
```
./import-events-db.py --org org123 --instance-id YOUR_INSTANCE_ID --start '2023-12-08T12:00:00+00:00' --end '2023-12-08T12:00:00+00:00' --value 500.0
```

2. Run the hourly tally.
```
http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=org123" x-rh-swatch-psk:placeholder
```

3. Verify that the instance_measurements value is now 500.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions \
-c "select h.instance_id, m.metric_id, m.value from hosts h, instance_measurements m where h.id=m.host_id and h.org_id='org123' and h.instance_id='YOUR_INSTANCE_ID'"
```

4. Verify that the monthly totals for `2023-12` was decremented by 200 and incremented by a value of 500. The value should be `PREVIOUS_TOTAL - 200 + 500`. In my case it was `203.4 - 200 + 500 = 503.4`
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select h.instance_id, t.month, t.metric_id, t.value from instance_monthly_totals t, hosts h where t.host_id=h.id and h.instance_id='YOUR_INSTANCE_ID';"
```

5. Ensure that the tally snapshots reflect the subtraction of 200, and the addition of 500.
```
PGTZ=utc psql -U rhsm-subscriptions rhsm-subscriptions -c "select s.snapshot_date, s.granularity, s.usage, s.sla, s.billing_provider, s.billing_account_id, m.measurement_type, m.metric_id, m.value from tally_snapshots s inner join tally_measurements m on s.id = m.snapshot_id where s.usage='_ANY' and s.sla='_ANY' and s.billing_provider='_ANY' and s.billing_account_id='_ANY' and m.measurement_type='TOTAL' order by s.snapshot_date, s.granularity"
```

Note, that the `2023-12-06` snapshots were updated from 200 to 500.
```
     snapshot_date      | granularity | usage | sla  | billing_provider | billing_account_id | measurement_type | metric_id | value 
------------------------+-------------+-------+------+------------------+--------------------+------------------+-----------+-------
 2023-11-05 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   100
 2023-11-05 12:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   100
 2023-12-05 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |  14.8
 2023-12-05 11:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |  14.8
 2023-12-08 00:00:00+00 | DAILY       | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   500
 2023-12-08 12:00:00+00 | HOURLY      | _ANY  | _ANY | _ANY             | _ANY               | TOTAL            | CORES     |   500
(6 rows)
```

### TEST AN ACCOUNT RESET
Start the contracts service (for some reason, the CONTRACT_USE_STUB is not working) so that the cleanup will complete successfully.
```
SERVER_PORT=8001 ./gradlew  :swatch-contracts:quarkusDev
```

Reset the account and make sure that it cleans out the tally_state table for the org.
```
http DELETE ":8000/api/rhsm-subscriptions/v1/internal/rpc/tally/org123" x-rh-swatch-psk:placeholder
```
```
psql -U rhsm-subscriptions rhsm-subscriptions \
-c "select * from tally_state"
```


src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
```diff
@@ -109,46 +126,75 @@ public void produceHourlySnapshotsForOrg(String orgId, DateRange snapshotRange)
...
+        TallyState currentState =
+            tallyStateRepository
+                .findById(new TallyStateKey(orgId, serviceType))
+                .orElseGet(
+                    () ->
+                        tallyStateRepository.save(
+                            new TallyState(orgId, serviceType, clock.startOfToday().minusDays(1))));
+
+        AccountUsageCalculationCache calcCache = new AccountUsageCalculationCache();
+        while (true) {
```
Sgitario, line 141: I'm a bit scared every time I see `while (true)` statements :)
What about counting the number of events to process and using a `for` statement with this count?

src/main/java/org/candlepin/subscriptions/tally/AccountUsageCalculationCache.java
```diff
@@ -0,0 +1,77 @@
...
+import java.time.OffsetDateTime;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import lombok.extern.slf4j.Slf4j;
+import org.candlepin.subscriptions.json.Event;
+import org.candlepin.subscriptions.util.DateRange;
+
+@Slf4j
+public class AccountUsageCalculationCache {
```
Sgitario, line 32: Can you add a javadoc to describe that is the intention of this class?
Also, as far as I understand, this is acting like an accumulator of calculations instead of a cache, but I might be mistaken.

src/main/resources/liquibase/202309281140-add-record-date-to-events-pk.xml
```diff
@@ -0,0 +1,20 @@
...
+    <comment>Include the record_date column in the PK.</comment>
+    <dropPrimaryKey tableName="events" />
+    <addPrimaryKey constraintName="events_pkey"
+      tableName="events"
+      columnNames="org_id,event_type,event_source,instance_id,record_date,timestamp"/>
+  </changeSet>
+
+  <changeSet id="202309011316-2" author="mstead" dbms="postgresql">
+    <comment>Copy the record_date value into the JSON data.</comment>
+        <sql>update events set data = jsonb_set(data, '{record_date}', to_jsonb(record_date))</sql>
```
Sgitario, line 17:
```suggestion
    <sql>update events set data = jsonb_set(data, '{record_date}', to_jsonb(record_date))</sql>
```

src/main/resources/liquibase/202309281140-add-record-date-to-events-pk.xml
```diff
@@ -0,0 +1,20 @@
...
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+  <changeSet id="202309011316-1" author="mstead">
+    <comment>Include the record_date column in the PK.</comment>
+    <dropPrimaryKey tableName="events" />
+    <addPrimaryKey constraintName="events_pkey"
+      tableName="events"
+      columnNames="org_id,event_type,event_source,instance_id,record_date,timestamp"/>
+  </changeSet>
```
Sgitario, line 13: This change set is missing the rollback statement.

src/main/resources/liquibase/202309281140-add-record-date-to-events-pk.xml
```diff
@@ -0,0 +1,20 @@
...
+
+  <changeSet id="202309011316-2" author="mstead" dbms="postgresql">
+    <comment>Copy the record_date value into the JSON data.</comment>
+        <sql>update events set data = jsonb_set(data, '{record_date}', to_jsonb(record_date))</sql>
+  </changeSet>
```
Sgitario, line 18: This change set is missing the rollback statement.

src/main/resources/liquibase/202309271140-create-tally-state-table.xml
```diff
@@ -0,0 +1,33 @@
...
+      <column name="latest_event_record_date" type="TIMESTAMP WITH TIME ZONE"/>
+    </createTable>
+
+    <addPrimaryKey constraintName="tally_state_pkey"
+      tableName="tally_state"
+      columnNames="org_id, service_type"/>
+
+  </changeSet>
+
+  <changeSet id="202309271140-2" author="mstead" dbms="postgresql">
```
Sgitario, line 24: This change set is missing the rollback statement.

src/main/resources/liquibase/202309271140-create-tally-state-table.xml
```diff
@@ -0,0 +1,33 @@
...
+      <column name="org_id" type="VARCHAR(32)"/>
+      <column name="service_type" type="VARCHAR(32)"/>
+      <column name="latest_event_record_date" type="TIMESTAMP WITH TIME ZONE"/>
+    </createTable>
+
+    <addPrimaryKey constraintName="tally_state_pkey"
+      tableName="tally_state"
+      columnNames="org_id, service_type"/>
+
+  </changeSet>
```
Sgitario, line 22: This change set is missing the rollback statement.

src/main/resources/liquibase/202309271135-add-event-lookup-index.xml
```diff
@@ -0,0 +1,25 @@
...
+        - Liquibase does not support using JSONB syntax while creating an index.
+        - This index will only be applied to postgres databases since HSQLDB does not
+          support the JSONB syntax, and the index isn't really important for testing
+          purposes.
+    </comment>
+    <sql>
+      create index orgid_servicetype_record_date_idx
+          on events using btree (org_id, (data->>'service_type'), record_date);
+    </sql>
+  </changeSet>
```
Sgitario, line 23: This change set is missing the rollback statement.

Sgitario (commented):
Many thanks for the big refactor that brings getting rid of the retally process!
I made a few suggestions, plus the rollback comments which I'm not sure how important they are (at least, liquibase complains when doing a rollback).

By the way, I could not perform the steps in the description because of some errors when running the python script:

```
[jcarvaja@localhost rhsm-subscriptions]$ EXCLUDE_RECORD_DATE_FROM_JSON=True ./import-events-db.py --org org123 --instances 5 --start '2023-12-05T11:00:00+00:00' --end '2023-12-05T11:00:00+00:00'
Traceback (most recent call last):
  File "/home/jcarvaja/sources/RedHatInsights/rhsm-subscriptions/./import-events-db.py", line 6, in <module>
    import psycopg2
ModuleNotFoundError: No module named 'psycopg2'
```

Before doing this, I had to amend the permissions, so we can execute the file.

mstead:
> By the way, I could not perform the steps in the description because of some errors when running the python script:
> 
> ```
> [jcarvaja@localhost rhsm-subscriptions]$ EXCLUDE_RECORD_DATE_FROM_JSON=True ./import-events-db.py --org org123 --instances 5 --start '2023-12-05T11:00:00+00:00' --end '2023-12-05T11:00:00+00:00'
> Traceback (most recent call last):
>   File "/home/jcarvaja/sources/RedHatInsights/rhsm-subscriptions/./import-events-db.py", line 6, in <module>
>     import psycopg2
> ModuleNotFoundError: No module named 'psycopg2'
> ```
> 
> Before doing this, I had to amend the permissions, so we can execute the file.

The following should fix this issue.
```
python -m pip install psycopg
```

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
mstead: I agree. I'll come up with something a little better.

src/main/java/org/candlepin/subscriptions/tally/AccountUsageCalculationCache.java
mstead: Javadoc will be updated in next patch. It is used to cache calculations that are initially populated from snapshots in the DB when required. The intent is to track the state of the calculations over the entirety of the tally run so that we do not have to send Tally Summaries for every batch of Events that gets processed. This would result in a ton of summary messages per tally depending on the tally event batch size.

san7ket:
Can we have a rebase, so we can build on latest

Sgitario:
> Can we have a rebase, so we can build on latest

Rebased and conflicts fixed.

mstead:
> Can we have a rebase, so we can build on latest

Done.

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
mstead: Please see if the latest changes are a little better.

src/main/java/org/candlepin/subscriptions/tally/AccountUsageCalculationCache.java
mstead: DONE

src/main/resources/liquibase/202309281140-add-record-date-to-events-pk.xml
mstead: DONE

src/main/resources/liquibase/202309281140-add-record-date-to-events-pk.xml
mstead: DONE

src/main/resources/liquibase/202309281140-add-record-date-to-events-pk.xml
mstead: DONE

src/main/resources/liquibase/202309271140-create-tally-state-table.xml
mstead: DONE

src/main/resources/liquibase/202309271135-add-event-lookup-index.xml
mstead: DONE

src/main/java/org/candlepin/subscriptions/tally/TallySnapshotController.java
mstead: I tried a couple variations of using a `for` loop but it got  a little messy. Hopefully this is a little better.

Sgitario: My main concern is that for some reason the events from `fetchEventsInBatch` are never empty, so the loop is going infinite. 
I'm unsure if a safe-ward would make sense here (a limit within the loop). I will propose another solution later today.

Sgitario: @mstead this is what I had in mind: https://privatebin.corp.redhat.com/?078a45c59b426048#EizuYWrzdKfzYtnxEGGytCKT31kPMzp7kU8tJV2DoCbf

Basically, stream the events (using QueryHints) instead of querying the database every X times until we have no more data). I've done this via the stream support of Spring JPA.

In the patch from the pastebin link, I'm doing all the changes in a single transaction. This needs to be refactored in separate transactions as you stated in this comment https://github.com/RedHatInsights/rhsm-subscriptions/pull/2876/files#diff-8005db7bcd7bb68e8259f0d4a7f02e0a0797bf75bb380b8c14cb47e5bf22ce75R114.

mstead: Ya, I hear you for sure, and honestly, this is something that I was thinking about while doing this. Thinking about the processing 1 million Events benchmark, batching allows us to also lookup the hosts related to the Event in the batch all at once, which is why I leaned this way.

> A disadvantage of using streams is that we can't customize the count of the batches using the hourlyTallyEventBatchSize property, but it's hardcoded to the value set in the QueryHint annotation. I've checked that we're using QueryHint in other streaming operations, so I think we can live with that.
Agreed.

> List<Event> events = eventRecords.map(EventRecord::getEvent).toList();

Unless I'm mistaken, the above line would pull all of the EventRecords into memory all at once. We could process the events one by one, but I'm not sure what it would do to the performance of Host updates.

Thanks for the suggestions! Gives me some things to investigate. I'll look into this some more today.

Sgitario: > > List events = eventRecords.map(EventRecord::getEvent).toList();
> 
> Unless I'm mistaken, the above line would pull all of the EventRecords into memory all at once. We could process the events one by one, but I'm not sure what it would do to the performance of Host updates.

Not all the events, but only the ones that are part of the current batch of the stream. 
Semantically, this is the same as you are doing now, but avoiding having to query several times to check whether there are more events to process.

mstead: @Sgitario I pushed an update that I think addresses your concern as a separate commit to make it easier to follow. It continues to support batch processing the events, however, it now streams from the DB. I tested the effect of changing the batch size of the Stream and it had less of an impact of adjusting the processing batch size, so I think it is fine as is, and still provides the flexibility of adjusting the overall processing batch size. It also keeps the transaction setup the same.

Let me know what you think.

Sgitario (commented):
lgtm!
The only minor remaining thing is to delete some unused methods from the EventRecordRepository.

mstead:
> The only minor remaining thing is to delete some unused methods from the EventRecordRepository.

DONE

Sgitario (approved):
lgtm!

san7ket:
Can someone rebase again, I cannot rebuild with conflicts. Thanks

Sgitario:
> Can someone rebase again, I cannot rebuild with conflicts. Thanks

Rebased and conflicts resolved.

san7ket:
From the investigation on thread with @Sgitario 
The product with two uom has issues when we query the instance api for the products, where only one uom will give the host and other return empty.

```
In [7]: application.rhsm_subscriptions.get_instances_report("OpenShift-dedicated-metrics",uom="Instance-hours")
2023-12-19 10:26:36.940 [    INFO] [root] REST: http://swatch-api-service.ephemeral-eeuigu.svc:8000/api/rhsm-subscriptions/v1/instances/products/OpenShift-dedicated-metrics
2023-12-19 10:26:36.954 [    INFO] [iqe.base.rest_client] REST: METHOD=GET, request_id=None, params=[('limit', 100), ('offset', 0), ('uom', 'Instance-hours')]
Out[7]: 
[{'id': '237b2483-a703-4f18-9a83-45a4373590e6',
  'instance_id': 'cd59c2b9-7bb7-48f5-bb79-d4cb97deca74',
  'display_name': 'automation_osd_cluster_cd59c2b9-7bb7-48f5-bb79-d4cb97deca74',
  'billing_provider': 'red hat',
  'measurements': [1.0, 1.0],
  'last_seen': datetime.datetime(2023, 12, 19, 7, 0, tzinfo=tzlocal()),
  'number_of_guests': 0,
  '_labeled_measurements': {'Cores': 1.0, 'Instance-hours': 1.0}}]

In [8]: application.rhsm_subscriptions.get_instances_report("OpenShift-dedicated-metrics",uom="Cores")
2023-12-19 10:26:49.513 [    INFO] [root] REST: http://swatch-api-service.ephemeral-eeuigu.svc:8000/api/rhsm-subscriptions/v1/instances/products/OpenShift-dedicated-metrics
2023-12-19 10:26:49.527 [    INFO] [iqe.base.rest_client] REST: METHOD=GET, request_id=None, params=[('limit', 100), ('offset', 0), ('uom', 'Cores')]
Out[8]: []
```

san7ket:
Conflicts in branch cannot rebuild, could someone take a look

src/main/java/org/candlepin/subscriptions/tally/AccountUsageCalculationCache.java
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * An object that temporarily caches instances of {@link AccountUsageCalculation} objects and Maps
+ * them by {@link OffsetDateTime}. This object is used to cache any current usage calculations for
+ * an associated usage timestamp (Event.timestamp) while the hourly tally is running. This ensures
+ * that the {@link MetricUsageCollector} will only need to initialize the usage calculation from the
+ * DB if it is not already in the cache, avoiding the need to loading the same calculation for each
+ * batch of events.
+ */
+@Slf4j
+public class AccountUsageCalculationCache {
```
kahowell, line 40: What are the bounds on this cache? Would it make sense to have it be implemented w/ jcache in future work? or become tuneable?

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -24,261 +24,140 @@
...
+        hostRepository
+            .findAllByOrgIdAndInstanceIdIn(
+                orgId, events.stream().map(Event::getInstanceId).collect(Collectors.toSet()))
+            .collect(Collectors.toMap(Host::getInstanceId, Function.identity()));
+
+    for (Event event : events) {
+      Host host = hostsByInstanceId.getOrDefault(event.getInstanceId(), new Host());
+
+      // Determine if the Event has already been applied to this host and skip it if it is.
+      // This is important in the case that we attempt to apply the same Event multiple times.
```
kahowell, line 95: Some additional context that would be nice to have in this comment:
1. An example of when we might attempt to apply the same event.
2. What happens otherwise.

src/main/resources/application-worker.yaml
```diff
@@ -6,6 +6,7 @@ SERVICE_INSTANCE_INGRESS_DEAD_LETTER_TOPIC: ${clowder.kafka.topics[?(@.requested
 
 rhsm-subscriptions:
   hourlyTallyDurationLimitDays: 90d
+  hourlyTallyEventBatchSize: ${HOURLY_TALLY_EVENT_BATCH_SIZE:16000}
```
kahowell, line 9: See previous comment about the documenting why 16k. Also note because providing a default here takes precedence over the field default, you could omit the field default I think.

swatch-tally/deploy/clowdapp.yaml
```diff
@@ -167,6 +167,8 @@ parameters:
     value: '1024'
   - name: USE_CPU_SYSTEM_FACTS_TO_ALL_PRODUCTS
     value: 'true'
+  - name: HOURLY_TALLY_EVENT_BATCH_SIZE
+    value: '16000'
```
kahowell, line 171: See previous comment about 16k being a non-obvious choice.

kahowell (changes requested):
Looking for a few minor comment modifications. Otherwise, lgtm :smile:

mstead:
**UPDATE:** I'm currently working through some test issues after resolving the conflicts on this branch.

src/main/java/org/candlepin/subscriptions/ApplicationProperties.java
```diff
@@ -169,4 +169,7 @@ public class ApplicationProperties {
    * the new system profile facts to the OpenShift Container products.
    */
   private boolean useCpuSystemFactsToAllProducts = true;
+
+  /** Defines the number of Events to process in a batch during the hourly tally. */
+  private int hourlyTallyEventBatchSize = 16000;
```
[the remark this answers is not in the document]
mstead: DONE

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
mstead: DONE

src/main/resources/application-worker.yaml
mstead: DONE

swatch-tally/deploy/clowdapp.yaml
mstead: DONE

mstead:
> **UPDATE:** I'm currently working through some test issues after resolving the conflicts on this branch.

Conflicts are now resolved. See the commit message for details on the issues I had after the deps in main were applied.
https://github.com/RedHatInsights/rhsm-subscriptions/pull/2876/commits/f3444767418747c8cda7edf7e28d0cb007eab31d

src/main/java/org/candlepin/subscriptions/tally/AccountUsageCalculationCache.java
mstead: In most cases, there should only be less than a handful of account calculations cached. Outside of cost, only the hour being processed. Personally, I think that using JCache would be overkill for what this is doing.

src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -93,21 +95,38 @@ public Stream<Event> fetchEventsInTimeRange(
...
-    // here ensures they won't be processed twice hence we don't have to retally again.
-    return repo.findByOrgIdAndServiceTypeAndTimestampGreaterThanEqualAndTimestampLessThanOrderByTimestamp(
-            orgId, serviceType, begin, end)
-        .filter(
-            eventRecord ->
-                Objects.isNull(asOfDateTime) || eventRecord.getRecordDate().isBefore(asOfDateTime))
-        .map(EventRecord::getEvent);
+      int batchSize,
+      Consumer<List<Event>> eventConsumer) {
+    List<Event> toProcess = new LinkedList<>();
```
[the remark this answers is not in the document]
mstead: Making this change definately had a positive impact on performance and memory usage. Testing against 100K events/hosts, it caused a reduction of 3s and 200M memory usage for the overall tally.

Will push a change. Nice catch!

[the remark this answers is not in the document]
mstead: DONE

kahowell (approved):
Changes look good, all my feedback's been addressed

mstead:
> I am automating this currently, and I wanted to check if we can leverage the new last lastEventRecordDate to improve our testing, if we can somehow print it to response of a endpoint we can be sure that the triggered tally was done and complete. This would greatly improve the speed of our testing. I can file a separate ticket for this, if necessary, but would like to hear your thoughts ?

I'm assuming that you'd like to see this in the repsonse of the InstanceApi? If so, this is doable. We would need to add this field to the Instance DB view and then update the API definition.

If it will help with testing this PR in particular, I'm personally Ok with doing this as part of this PR. @kahowell @lindseyburnett thoughts?

lindseyburnett:
> > I am automating this currently, and I wanted to check if we can leverage the new last lastEventRecordDate to improve our testing, if we can somehow print it to response of a endpoint we can be sure that the triggered tally was done and complete. This would greatly improve the speed of our testing. I can file a separate ticket for this, if necessary, but would like to hear your thoughts ?
> 
> I'm assuming that you'd like to see this in the repsonse of the InstanceApi? If so, this is doable. We would need to add this field to the Instance DB view and then update the API definition.
> 
> If it will help with testing this PR in particular, I'm personally Ok with doing this as part of this PR. @kahowell @lindseyburnett thoughts?

Sounds like a good plan to me, I'm good with it.

san7ket (approved):
Failures are related to the updated spec file, tested locally with new spec file and QE automation PR in place, So would want to simultaneously merge this to not affect other PR checks, co-ordinating with @mstead

mstead:
Thanks @san7ket. I'm currently trying to get the perf team to re-run the perf tests on this PR. It would be nice to know where we stand performance wise before pushing this into stage.

mstead:
Waiting until perf team completes their testing before merging this PR. Expect it to be done by early next week. Work is being tracked here: https://issues.redhat.com/browse/SWATCH-2112

san7ket:
All 4 failures are valid and are due to spec file change. I will run tests with updated and see if there are any issues

san7ket:
@mstead Do I need to retest this, with all the rebases ?

mstead:
> @mstead Do I need to retest this, with all the rebases ?

Please do. There has been a lot from the main branch that has been merged into this PR.

san7ket:
@mstead Looks like the test failure is related to this PR specify, will debug this today, will report back

mstead:
Remaining in HOLD until I test the DB migrations against stage/prod data set.

mstead:
> Remaining in HOLD until I test the DB migrations against stage/prod data set.

DB migration against the larger of stage/prod takes approximately 40s, which seems Ok to me.

Will merge after next prod release to give it time to soak in stage.

PR #2878, opened 2023-12-11
SWATCH-1566: API - Remove RHOSAK from API Configuration

Sgitario:
Jira issue: [SWATCH-1566](https://issues.redhat.com/browse/SWATCH-1566)

## Description
As the control plane for RHOSAK has now been turned off. We should remove all remaining back-end configurations for RHOSAK from subscription-watch. 

## Testing
Regression tests only. If there are tests using RHOSAK, these need to be updated accordingly.


src/test/java/org/candlepin/subscriptions/resource/api/SwaggerTest.java
```diff
@@ -29,12 +29,14 @@
...
 import org.springframework.boot.test.web.server.LocalServerPort;
 import org.springframework.http.ResponseEntity;
+import org.springframework.test.context.ActiveProfiles;
 
 /**
  * This test covers the logic in ResteasyConfiguration and the static swagger resources (to ensure
  * that point out to the correct openapi endpoints).
  */
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("test")
```
Sgitario, line 39: Note that this change is unrelated to this pull request because this test was failing for me locally (because there was no database up and running).

lindseyburnett (approved):

PR #2880, opened 2023-12-12
SWATCH-470: Internal subscriptions sync org rest api should return json

Sgitario:
Jira issue: [SWATCH-470](https://issues.redhat.com/browse/SWATCH-470)

## Description
- Rename DefaultResponse to RpcResponse
- Remove status field
- Add result field w/ type "any" (don't specify the type param in spec).
- Document for "result" field, may be null (if unneeded). 

## Testing
Regression testing only. If there are tests that expect the old format ("Sync started. "), then these tests are to be updated to expect the JSON format (to check if the result was 200 OK).

### DIFF
--- a/buildSrc/src/main/groovy/swatch.quarkus-rest-client-conventions.gradle
+++ b/buildSrc/src/main/groovy/swatch.quarkus-rest-client-conventions.gradle
@@ -45,6 +45,7 @@ openApiGenerate {
 dependencies {
     implementation enforcedPlatform(libraries["quarkus-bom"])
     api 'io.quarkus:quarkus-rest-client-reactive-jackson'
+    api libraries["jackson-databind-nullable"]
     testImplementation 'io.rest-assured:rest-assured'
 }
 
--- a/src/main/spec/internal-subscriptions-sync-api-spec.yaml
+++ b/src/main/spec/internal-subscriptions-sync-api-spec.yaml
@@ -134,13 +134,12 @@ paths:
       tags:
         - internalSubscriptions
       responses:
-        '202':
-          description: "The request for syncing organization's subscription is processing."
+        '200':
+          description: The request for syncing a single organization's subscription is processing.
           content:
-            text/plain:
+            application/json:
               schema:
-                type: string
-                example: Sync started.
+                $ref: "#/components/schemas/RpcResponse"
         '400':
           $ref: "../../../spec/error-responses.yaml#/$defs/BadRequest"
         '403':
--- a/src/main/spec/internal-subscriptions-sync-api-spec.yaml
+++ b/src/main/spec/internal-subscriptions-sync-api-spec.yaml
@@ -110,7 +110,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/DefaultResponse"
+                $ref: "#/components/schemas/RpcResponse"
         '400':
           $ref: "../../../spec/error-responses.yaml#/$defs/BadRequest"
         '403':
@@ -162,7 +162,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/DefaultResponse"
+                $ref: "#/components/schemas/RpcResponse"
         '400':
           $ref: "../../../spec/error-responses.yaml#/$defs/BadRequest"
         '403':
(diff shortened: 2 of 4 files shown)


buildSrc/src/main/groovy/swatch.quarkus-rest-client-conventions.gradle
```diff
@@ -45,6 +45,7 @@ openApiGenerate {
 dependencies {
     implementation enforcedPlatform(libraries["quarkus-bom"])
     api 'io.quarkus:quarkus-rest-client-reactive-jackson'
+    api libraries["jackson-databind-nullable"]
```
Sgitario, line 48: I had to add this dependency to avoid having the following compilation failure:
```
/home/jcarvaja/sources/RedHatInsights/rhsm-subscriptions/clients/swatch-internal-subscription-client/build/generated/src/main/java/com/redhat/swatch/clients/swatch/internal/subscription/api/model/RpcResponse.java:22: error: package org.openapitools.jackson.nullable does not exist
import org.openapitools.jackson.nullable.JsonNullable;
```

This dependency is in the spring rest-client conventions as well.

Sgitario:
Adding the QE label because the linked JIRA ticket does it.

lindseyburnett:
The endpoint for syncing subs for a single org was still returning a string, so I pushed an update for that.  @Sgitario please double check my work for that.

Otherwise looks good to me.  Here were my test steps:
```
DEV_MODE=true SPRING_PROFILES_ACTIVE=capacity-ingress ./gradlew :bootRun
```

test sync single org
```
http PUT :8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/16774238 x-rh-swatch-psk:placeholder

HTTP/1.1 200 
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Connection: keep-alive
Content-Length: 2
Content-Type: application/json
Date: Fri, 15 Dec 2023 00:33:25 GMT
Expires: 0
Keep-Alive: timeout=60
Pragma: no-cache
Set-Cookie: JSESSIONID=45D922DE1D18FED165C29DACD98EBA15; Path=/; HttpOnly
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-XSS-Protection: 0

{}
```

test sync all orgs
```
http PUT :8000/api/rhsm-subscriptions/v1/internal/rpc/subscriptions/sync forceSync==false x-rh-swatch-psk:placeholder
HTTP/1.1 200 
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Connection: keep-alive
Content-Length: 2
Content-Type: application/json
Date: Fri, 15 Dec 2023 00:37:20 GMT
Expires: 0
Keep-Alive: timeout=60
Pragma: no-cache
Set-Cookie: JSESSIONID=F8395416F2AF6ACDEE6DB2F433509832; Path=/; HttpOnly
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-XSS-Protection: 0

{}
```

lindseyburnett (approved):

src/main/spec/internal-subscriptions-sync-api-spec.yaml
```diff
@@ -134,13 +134,12 @@ paths:
       tags:
         - internalSubscriptions
       responses:
-        '202':
-          description: "The request for syncing organization's subscription is processing."
+        '200':
```
lindseyburnett, line 137: This actually returns a 200 even though it was listed as a 202 here.

Sgitario: Thanks!

himmAllRight:
Looks good to me. I ran it in an EE and seems to run as expected:

```
In [11]: app.rhsm_subscriptions.sync_capacity()
2023-12-15 20:36:41.538 [    INFO] [root] PUT request response url: http://swatch-subscription-sync-service:8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/3340851, status_code: 200
2023-12-15 20:36:41.538 [    INFO] [root] Response: {}
Out[11]: {}

In [12]: smqe_tools.sync_capacity_for_org(org)
2023-12-15 20:36:46.327 [    INFO] [root] PUT request response url: http://swatch-subscription-sync-service:8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/3340851, status_code: 200
2023-12-15 20:36:46.327 [    INFO] [root] Response: {}
Out[12]: {}
```
If that’s it, I think it should be good. Our sync function doesn't appear to currently capture the response, so I don’t think it would break any regression. In addition to that, I did a quick search to see if we were doing any matching for the old `Sync started` and I didn't see anything.

We should be good to merge (and have the post-stage-deploy regression run 😉

himmAllRight (approved):

PR #2886, opened 2023-12-13
SWATCH-2009: Fix round issue when dealing with double numbers

Sgitario:
Jira issue: [SWATCH-2009](https://issues.redhat.com/browse/SWATCH-2009)

## Description
When processing multiple events, we were seeing weird values in the "tally_measurements" table like "9.799999999999999". This issue was caused by manipulating double numbers (see this link for further information: https://floating-point-gui.de/).

To fix the issue, we should use the BigDecimal class to accumulate the values in the UsageCalculation. 

## Testing
Only regression testing is needed. JUnit tests have been added.

### DIFF
--- a/src/test/java/org/candlepin/subscriptions/tally/UsageCalculationTest.java
+++ b/src/test/java/org/candlepin/subscriptions/tally/UsageCalculationTest.java
@@ -20,6 +20,7 @@
  */
 package org.candlepin.subscriptions.tally;
 
+import static com.redhat.swatch.configuration.util.MetricIdUtils.getCores;
 import static org.candlepin.subscriptions.tally.collector.Assertions.assertHardwareMeasurementTotals;
 import static org.candlepin.subscriptions.tally.collector.Assertions.assertNullExcept;
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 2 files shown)


src/test/java/org/candlepin/subscriptions/tally/UsageCalculationTest.java
```diff
@@ -52,6 +53,21 @@ void testAddToTotal() {
...
+  @Test
+  void testAddToTotalShouldRoundDoubleValues() {
+    UsageCalculation calculation = new UsageCalculation(createUsageKey("Product"));
+    calculation.addToTotal(getCores(), 2.9);
+    calculation.addToTotal(getCores(), 3.0);
+    calculation.addToTotal(getCores(), 2.3);
+    calculation.addToTotal(getCores(), 1.4);
+    calculation.addToTotal(getCores(), 0.2);
+
+    // Using the above metrics, without rounding the doubles, we got 9.799999999999999 instead of
```
lindseyburnett, line 65: I don't think the context of this comment is going to age well.  I like having an explanation though, so mabye maybe move to a comment on the method that says something like "Added this test case due to a bug we've seen with doubles being rounded" or a reference to the SWATCH-2009 card

lindseyburnett (approved):

src/test/java/org/candlepin/subscriptions/tally/UsageCalculationTest.java
Sgitario: Good idea. I've just changed it.

awood:
Are there any issues that you saw when persisting the data to the database?  We're using the `double precision` datatype for the measurements value column which is also an implementation of IEEE Standard 754 (like `double` in Java).  I'm wondering if we're just moving the problem from when our code does arithmetic to when our code persists the `BigDouble`.  If you've seen any issues like this (and this might be a good usage of test containers so that we can do a 100% faithful test), we might need a follow up card to convert the column to [`numeric` type](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL) which is what the docs recommend for things which high precision requirements.

Sgitario:
> Are there any issues that you saw when persisting the data to the database? We're using the `double precision` datatype for the measurements value column which is also an implementation of IEEE Standard 754 (like `double` in Java). I'm wondering if we're just moving the problem from when our code does arithmetic to when our code persists the `BigDouble`. If you've seen any issues like this (and this might be a good usage of test containers so that we can do a 100% faithful test), we might need a follow up card to convert the column to [`numeric` type](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL) which is what the docs recommend for things which high precision requirements.

The data that is persisted in the database is correct after this fix. I did verify that following the steps in the JIRA description. Unless for the tally_measurements table and the tally report which are the tables that mentioned in the ticket.

PR #2930, opened 2024-01-05
SWATCH-2074: save dimension from umb message

kflahert:
Jira issue: SWATCH-2074

## Description

Contracts is failing to save UMB messages because dimensions are coming from the message itself rather than the partner API. This change allows us to get dimension from both the message and partnerApi.


lindseyburnett (approved):

TrayvonMcKnight (approved):
Looks good to me.

PR #2939, opened 2024-01-10
SWATCH-2066: Expose prometheus data on :9000/metrics for swatch-metrics

awood:
Jira issue: SWATCH-2066

## Description
This is necessary so that app-interface can actually find our Prometheus metrics!

### Setup
1. Write a Clowder configuration file to `/tmp/test-clowder.json`
   ```json
   {
   "kafka": {
    "brokers": [
      {
        "hostname": "localhost",
        "port": 9092
      }
    ],
    "topics": []
   },
   "metricsPath": "/metrics",
   "metricsPort": 9000,
   "privatePort": 10000,
   "publicPort": 8000,
   "webPort": 8000
   }
   ```
1. Start swatch-metrics using the clowder and prometheus settings
   ```
   export RHSMSUBSCRIPTIONS_METERING_PROMETHEUS_CLIENT_URL=http://localhost:9090/api/v1
   ACG_CONFIG=/tmp/test-clowder.json QUARKUS_MANAGEMENT_ENABLED=true ../gradlew quarkusDev
   ```

### Verification
1. Pull the metrics
   ```
   http :9000/metrics
   ```

### DIFF
--- a/swatch-metrics/deploy/clowdapp.yaml
+++ b/swatch-metrics/deploy/clowdapp.yaml
@@ -142,7 +142,7 @@ objects:
     metadata:
       name: swatch-metrics
       labels:
-        prometheus: rhsm
+        prometheus: quarkus
     spec: # The name of the ClowdEnvironment providing the services
       envName: ${ENV_NAME}
 
@@ -340,7 +340,7 @@ objects:
     metadata:
       name: swatch-metrics-rhel
       labels:
-        prometheus: rhsm
+        prometheus: quarkus
     spec:
       envName: ${ENV_NAME}
 


Sgitario:
Updated description to add `QUARKUS_MANAGEMENT_ENABLED=true` because it was disabled when running in dev mode and the endpoint from `http :9000/metrics`  to `http :9000/q/metrics`

swatch-metrics/src/main/resources/application.yaml
```diff
@@ -23,6 +31,8 @@ SERVICE_INSTANCE_INGRESS_TOPIC: platform.rhsm-subscriptions.service-instance-ing
   SPLUNKMETA_namespace: local
   SPLUNK_HEC_INCLUDE_EX: true
   quarkus:
+    management:
+      enabled: false
```
Sgitario, line 35: Why disable the management server in dev mode? I prefer starting up this server when running in DEV mode.

Sgitario (changes requested):
Changes look good to me aside from the above comments. 
I added a commit to fix the swatch metrics clowdapp.yaml file to use the management health endpoints (when enabling the management server, the health endpoints change). 
Also, I created this ticket SWATCH-2091 to address this issue in the Quarkus Clowder config source.

However, apart from the changes in the pull request, I don't see why if the app-interface does not read the exposed metrics, the Kafka topic .*platform.rhsm-subscriptions.metering-tasks starts growing, can you clarify this?

swatch-metrics/src/main/resources/application.yaml
awood: It's  not disabled in the way you are thinking.  That option is very confusingly named.  Setting `quarkus.management.enabled` just moves a bunch of management endpoints from `:8000/q/whatever` to `:9000/q/whatever`.  So setting `quarkus.management.enabled: true` here actually preserves the behavior you like.

swatch-metrics/deploy/clowdapp.yaml
```diff
@@ -131,11 +131,11 @@ parameters:
   - name: OTEL_EXPORTER_OTLP_ENDPOINT
     value: 'http://localhost:4317'
   - name: METRICS_HEALTH_LIVENESS_PATH
-    value: /q/health/live
+    value: /health/live
```
awood, line 134: Thanks for catching this @Sgitario

awood:
> Note that I don't see why if the app-interface does not read the exposed metrics, the Kafka topic .*platform.rhsm-subscriptions.metering-tasks starts growing, so maybe the changes are indeed necessary, can you clarify this?

This is strictly an attempt to gather more data about the issue.  Without the metrics, we don't have good visibility on what is happening to the degraded pods over time.

Sgitario:
> > Note that I don't see why if the app-interface does not read the exposed metrics, the Kafka topic .*platform.rhsm-subscriptions.metering-tasks starts growing, so maybe the changes are indeed necessary, can you clarify this?
> 
> This is strictly an attempt to gather more data about the issue. Without the metrics, we don't have good visibility on what is happening to the degraded pods over time. Right now `/app/rhsm-metrics/metrics` off turnpike isn't serving anything at all.

Ok! Thanks for the clarification.

> 
> > I think the root cause of the issue is having this label in the swatch metrics applications:
> > ```
> > prometheus: rhsm
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If we change it to:
> > ```
> > prometheus: quarkus
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, the following service monitor [rhsm.quarkus-servicemonitor.yml](https://gitlab.cee.redhat.com/service/app-interface/-/raw/master/resources/insights-stage/rhsm-stage/rhsm.quarkus-servicemonitor.yml?ref_type=heads) will be used with the correct paths.
> 
> That is an interesting point. I wasn't aware of that `servicemonitor` file. And that change is certainly less invasive than what I have here. My only question is whether we still need to move the management endpoints to port 9000 via `quarkus.management.enabled: true`. I don't see anything in that `servicemonitor` file about the port number and looking at the other RHSM one, it seems to me that it's implied that everything will be off port 9000. But I could at least take out the bits that change the path. What do you think?

To be honest, I would prefer to always disable the management server as it was until now. 
The main purpose of the management configuration was about security. So, the application endpoints can go public and the management endpoints are private. But we don't have this problem in our project. 
Therefore, I would revert all but the label thing, so the service monitor takes place and all our quarkus applications work the same.

Sgitario (approved):
I tried to find metrics from a Quarkus application (the swatch producer aws) and I could find the metrics using for example [this query](https://grafana.app-sre.devshift.net/explore?left=%7B%22datasource%22:%22PDD8BE47D10408F45%22,%22queries%22:%5B%7B%22expr%22:%22max%28http_server_connections_seconds_max%7Bservice%3D%5C%22swatch-producer-aws-service%5C%22%7D%29%22,%22format%22:%22time_series%22,%22groupBy%22:%5B%7B%22params%22:%5B%22$__interval%22%5D,%22type%22:%22time%22%7D,%7B%22params%22:%5B%22null%22%5D,%22type%22:%22fill%22%7D%5D,%22interval%22:%22%22,%22intervalFactor%22:1,%22orderByTime%22:%22ASC%22,%22policy%22:%22default%22,%22refId%22:%22A%22,%22resultFormat%22:%22time_series%22,%22select%22:%5B%5B%7B%22params%22:%5B%22value%22%5D,%22type%22:%22field%22%7D,%7B%22params%22:%5B%5D,%22type%22:%22mean%22%7D%5D%5D,%22tags%22:%5B%5D,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22PDD8BE47D10408F45%22%7D,%22editorMode%22:%22builder%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-3h%22,%22to%22:%22now%22%7D%7D&orgId=1). 

Perhaps, this is expected to work only in stage/prod, not in the ephemeral environment (where as you spot, the service monitor points into a wrong path).

kahowell:
> Perhaps, this is expected to work only in stage/prod, not in the ephemeral environment (where as you spot, the service monitor points into a wrong path).

Likely, we should probably consult with the consoledot devprod team on how to get this to work in ephemeral.

awood:
> I tried to find metrics from a Quarkus application (the swatch producer aws) and I could find the metrics using for example [this query](https://grafana.app-sre.devshift.net/explore?left=%7B%22datasource%22:%22PDD8BE47D10408F45%22,%22queries%22:%5B%7B%22expr%22:%22max%28http_server_connections_seconds_max%7Bservice%3D%5C%22swatch-producer-aws-service%5C%22%7D%29%22,%22format%22:%22time_series%22,%22groupBy%22:%5B%7B%22params%22:%5B%22$__interval%22%5D,%22type%22:%22time%22%7D,%7B%22params%22:%5B%22null%22%5D,%22type%22:%22fill%22%7D%5D,%22interval%22:%22%22,%22intervalFactor%22:1,%22orderByTime%22:%22ASC%22,%22policy%22:%22default%22,%22refId%22:%22A%22,%22resultFormat%22:%22time_series%22,%22select%22:%5B%5B%7B%22params%22:%5B%22value%22%5D,%22type%22:%22field%22%7D,%7B%22params%22:%5B%5D,%22type%22:%22mean%22%7D%5D%5D,%22tags%22:%5B%5D,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22PDD8BE47D10408F45%22%7D,%22editorMode%22:%22builder%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-3h%22,%22to%22:%22now%22%7D%7D&orgId=1).
> 
> Perhaps, this is expected to work only in stage/prod, not in the ephemeral environment (where as you spot, the service monitor points into a wrong path).

:+1: Should have checked that myself!

PR #2944, opened 2024-01-11
SWATCH-1911: Remove deprecated GET /tally/products/{productId}

Sgitario:
Jira issue: [SWATCH-1911](https://issues.redhat.com/browse/SWATCH-1911)

## Description
The endpoint "/tally/products/{product_id}" was marked as deprecated a while ago.
Use /tally/products/{product_id}/{metric_id} instead.


awood (approved):

ssidhaye (approved):
Based on the PR check tally test results acking this change

Sgitario:
I've also confirmed that we're not using the removed endpoint in the UI.

PR #2968, opened 2024-01-17
SWATCH-2109 Eliminate double billing scenarios & modify the PromQL for rhel metering to handle system label changes

lindseyburnett:
Jira issue: SWATCH-2109

## Summary

Human explanation of the new query can be found here: https://github.com/RedHatInsights/rhsm-subscriptions/blob/lburnett/swatch-2109-rhelemeter-promql/swatch-metrics/README.md#querytemplatesrhelemeter

We've opted to not use the recording rule for two reasons.  First, it was only further complicating an already complicated promql query.  Second, we get more accurate data by not using it.  When we requested a recording rule to be made, we expected that its evaluation would happen the same frequency as the openshift telemeter instance.  Since that's not going to be the case, using a recording rule doesn't benefit us much.

## Verification of behavior reported in the card

>The problem: we have two series: one for a first half of an hour with a label support=premium, and one for the the second half of an hour with a label support=self-service.
I would expect a value of 0.5 for the first series and 0.5 for the second one

There is a slight change of "expected behavior".  Because of the timing of metrics being reported to prometheus, and how the average value over the hour is calculated, it's possible that an instance has datapoints with different metadata labels that overlap.  For this particular scenario described in the card, there was an overlap where there was instance metadata indicating Premium SLA and another datapoint for Standard SLA.  The query that we landed on will take the first metadata label seen.

What this means functionally is if a metadata field is updated, it can take up to an hour for any change in support level or billing account for a system before it's reflected in prometheus.  This can also result in us underbilling for the period of time between when the metadata field was changed and top of the next hour.  To address underbilling, we'd have to get a recording rule to record `max(system_cpu_logical_count) by (_id)` and then have it run at least every 10m, and then replace system_cpu_logical_count in the numerator with it.

See https://docs.google.com/document/d/1b8Wk-4Joy3nD5Pd0sACgHFN10Vleb6CWKRHCps4ftJE/edit?usp=sharing for more details, example queries, and screenshots.

## Testing

To inspect/verify the generated query, you can use the metering-promql script and then change the 204 to 69 in the output.

`./bin/metering-promql.py --product rhel-for-x86-els-payg --org 11789772`

Expectations:
- Eliminate double billing scenarios
1.  Before this card, we would get multiple data points for an instance per hour when gathering metrics from rhelemeter.  We should now only gather one metric value per hour per instance to be used when hourly tallying.
2. If an instance has its metadata labels changed, we shouldn't double bill for that instance for that hour.

To test the exact example from the card you'll have to get creative.  The instance is only reporting product code 69.  You'll have to update the regex in `swatch-product-configuration/src/main/resources/subscription_configs/RHEL/RHEL_for_x86_els_payg.yaml` to be 69 instead of 204.  Then you'll want to use orgId 11789772 instead of 16787820.  And you'll need to play with the timestamps on the APIs for getting metrics and starting an hourly tally to make sure the metrics that were showing the problem are considered.

### Steps

Run a token refresher for stage rhelemeter

```
TOKEN_REFRESHER_VERSION=master-2023-09-20-f5e3403
SCOPE="openid offline_access"
CLIENT_SECRET=
CLIENT_ID=
ISSUER_URL=https://sso.redhat.com/auth/realms/redhat-external
URL=https://observatorium-mst.api.stage.openshift.com/api/metrics/v1/rhel
```

```
podman run --name=token-refresher --rm -ti -p 8082:8080 \
  quay.io/observatorium/token-refresher:$TOKEN_REFRESHER_VERSION \
  --scope="$SCOPE" \
  --oidc.client-secret="$CLIENT_SECRET" \
  --oidc.client-id="$CLIENT_ID" \
  --oidc.issuer-url="$ISSUER_URL" \
  --url="$URL"
```

Start swatch-metrics configured for rhelemeter
```
EVENT_SOURCE=rhelemeter PROM_URL="http://localhost:8082/api/v1" ./gradlew :swatch-metrics:quarkusDev
```

Kick off gathering metrics for now
```
curl -X POST --location "http://localhost:8000/api/swatch-metrics/v1/internal/metering/rhel-for-x86-els-payg?orgId=16787820" -H 'x-rh-swatch-synchronous-request: false' -H 'x-rh-swatch-psk: placeholder'
```

Kick off gathering metrics with optional API params
```
curl -X 'POST' \
  'http://localhost:8000/api/swatch-metrics/v1/internal/metering/rhel-for-x86-els-payg?orgId=16787820&endDate=2024-01-10T20%3A00%3A00Z&rangeInMinutes=120' \
  -H 'accept: */*' \
  -H 'x-rh-swatch-synchronous-request: false' \
  -H 'x-rh-swatch-psk: placeholder' \
  -d ''
```

If there were metrics to be found, you should see log messages indicating they're being sent as events.  At this point you can inspect the `platform.rhsm-subscriptions.service-instance-ingress` topic to see the events.

If you want to continue on in the process, you can turn on swatch-tally (worker profile).  You should see the messages being consumed and they end up in the `rhsm-subscriptions.events` table.  Then you can initiate an hourly tally and look at those results to verify accurate usage values.  If you still have swatch-metrics running, you'll need to change server ports.  Also keep in mind that the start/end dates for the internal/tally/hourly are going to compare against the **event.record_date** database column and not the timestamp of the actual metric fetched from prometheus.  

```
SERVER_PORT=8002 DEV_MODE=true ./gradlew :bootRun
```

```
curl 'http://localhost:8002/api/rhsm-subscriptions/v1/internal/tally/hourly?org=16787820&start=2024-01-16T00%3A00Z&end=2024-01-17T00%3A00Z' \
  -X 'POST' \
  -H 'x-rh-swatch-psk: placeholder' \
  -H 'x-rh-swatch-synchronous-request: false'
  ```

### DIFF
--- a/swatch-metrics/README.md
+++ b/swatch-metrics/README.md
@@ -1,90 +1,112 @@
 # swatch-metrics
 
-This project uses Quarkus, the Supersonic Subatomic Java Framework.
+## Running in "default" mode
 
-If you want to learn more about Quarkus, please visit its website: https://quarkus.io/ .
+Without additional configuration values, swatch-metrics will use "telemeter" as the EVENT_SOURCE, which indicates metrics are sourced from the OpenShift Telemeter observatorium instance.
 
-## Running the application in dev mode
-
-You can run your application in dev mode that enables live coding using:
-
-```shell script
-./gradlew :swatch-metrics:quarkusDev
+```
+EVENT_SOURCE=telemeter PROM_URL="http://localhost:8082/api/v1" ./gradlew :swatch-metrics:quarkusDev
 ```
 
-> **_NOTE:_**  Quarkus now ships with a Dev UI, which is available in dev mode only
-> at http://localhost:8080/q/dev/.
+## Running in "rhelemeter" mode
 
-## Packaging and running the application
-
-The application can be packaged using:
-
-```shell script
-./gradlew build
+```
+EVENT_SOURCE=rhelemeter PROM_URL="http://localhost:8082/api/v1" ./gradlew :swatch-metrics:quarkusDev
 ```
 
-It produces the `quarkus-run.jar` file in the `build/quarkus-app/` directory.
-Be aware that it’s not an _über-jar_ as the dependencies are copied into
-the `build/quarkus-app/lib/` directory.
+## PromQL Templates
 
-The application is now runnable using `java -jar build/quarkus-app/quarkus-run.jar`.
+The [application.yaml](./src/main/resources/application.yaml) file defines templates for promql queries used when fetching metrics from `PROM_URL` under the `rhsm-subscriptions.metering.prometheus.metric` section.
 
-If you want to build an _über-jar_, execute the following command:
+### accountQueryTemplates vs queryTemplates
 
-```shell script
-./gradlew build -Dquarkus.package.type=uber-jar
+**accountQueryTemplates** are used to determine which org ids have metrics for the timeframe we care about for a product we care about.
+
+**queryTemplates** are used to fetch the actual metrics and values we care about, filtered by a bunch of criteria.
+
+Parameterized values come from [swatch-product-configuration](../swatch-product-configuration/src/main/resources/subscription_configs) yaml files.
+
+### PromQL breakdown
+
+#### queryTemplates.rhelemeter
+
+Example
+```promql
+(
+      max by (_id) (
+        sum_over_time(system_cpu_logical_count[1h:10m])
+      )
+    /
+      scalar(count_over_time(vector(1)[1h:10m]))
+  )
+* on (_id) group_right
+  topk by (_id) (
+    1,
+    group without (swatch_placeholder_label) (
+      min_over_time(
+        system_cpu_logical_count{
+    product=~".*(^|,)(69)($|,).*",
+    external_organization="11789772",
+    billing_model="marketplace",
+    support=~"Premium|Standard|Self-Support|None"
+  }[1h]
+      )
+    )
+  )
 ```
 
-The application, packaged as an _über-jar_, is now runnable using `java -jar build/*-runner.jar`.
+It's easiest to think about this query by breaking it in half at the join.  The left side provides us with the usage values we care about, and the right side provides us with the associated metadata labels we care about while tallying.  
 
-## Creating a native executable
+## "Left Side" of the join
 
-You can create a native executable using:
+It calculates the maximum sum of `system_cpu_logical_count` over 1-hour windows (shifted by 10 minutes) for each `_id`, and normalizes this value by the number of 10-minute intervals in an hour.  This is a way to find the maximum average `system_cpu_logical_count` per 10-minute interval for each `_id` in the last hour.
 
-```shell script
-./gradlew build -Dquarkus.package.type=native
-```
+### Numerator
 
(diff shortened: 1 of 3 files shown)


mstead (approved):
Looks good and functions as described.

@lindseyburnett Amazing descriptions and test steps! Thanks for that!

ntkathole:
@lindseyburnett The query is working fine but I see difference in values in old vs new query. You can try 

Old query - `max(max by(_id) (sum_over_time(system_cpu_logical_count[1h:10m])) / scalar(count_over_time(vector(1)[1h:10m]))) by (_id) * on(_id) group_right min_over_time({product=~".*(^|,)(69)($|,).*", support=~"Premium|Standard|Self-Support|None"}[1h])` 

New query - `max by (_id) (sum_over_time(system_cpu_logical_count[1h:10m])) / scalar(count_over_time(vector(1)[1h:10m])) * on (_id) group_right topk by (_id) (1, group without (swatch_placeholder_label) (min_over_time(system_cpu_logical_count{product=~".*(^|,)(69)($|,).*", support=~'Premium|Standard|Self-Support|None'}[1h])))`

Old query returns value **4** "rhel-metered" instance while new query returns **2** for the same instance. (Check Prod RHOBS with 2024-01-22 10:00:00 timestamp)

lindseyburnett:
> Old query returns value **4** "rhel-metered" instance while new query returns **2** for the same instance. (Check Prod RHOBS with 2024-01-22 10:00:00 timestamp)

The old value "4" was actually incorrect and part of the double billing problem that we're resolving with this new query.

old query left-side:
This tells us that the average cpu-hours for an instance id.
`max(max by(_id) (sum_over_time(system_cpu_logical_count[1h:10m])) / scalar(count_over_time(vector(1)[1h:10m]))) by (_id)`
returns a value of **2** for `{_id=79e424f9-5010-496b-ab81-2c9d2d4d3f67}`.


old query right-side:
`min_over_time(system_cpu_logical_count{product=~".*(^|,)(69)($|,).*", support=~"Premium|Standard|Self-Support|None"}[1h])`
returns the two metrics (one for support Self-Support and one for Premium) with value of **2**

The join (`* on(_id) group_right`) multiplies the left-side values by the right-side values (2 * 2), which was the culprit giving us the incorrect value of 4.

By looking at the system_cpu_logical_count metric, I was able to verify that the expected result is 2 cpu-hours. (screenshot below is a little deceiving here because using the stacked view just to show the overlap, and that's making the y-axis jump to 4.00, but there's not actually any data points up there)

image

ntkathole:
Okay, I thought we do cores * sockets for vCPUs calculation. But yeah, I think we are good with new query considering socket_count same as system_cpu_logical_count.

ntkathole (approved):
Looks good,

Modified smqe tools helper to return correct PromQL.

`# swatch-client telemeter swatch-query -i 123456 --product-id rhel-for-x86-els-payg --uom vCPUs`

`max by (_id) (sum_over_time(system_cpu_logical_count[1h:10m])) / scalar(count_over_time(vector(1)[1h:10m])) * on (_id) group_right topk by (_id) (1, group without (swatch_placeholder_label) (min_over_time(system_cpu_logical_count{product=~".*(^|,)(204)($|,).*",external_organization='123456',billing_model='marketplace', support=~'Premium|Standard|Self-Support|None'}[1h])))`

PR #2976, opened 2024-01-19
SWATCH-2105: Constrain tally floorist exports to 1 day's range

Sgitario:
Jira issue: [SWATCH-2105](https://issues.redhat.com/browse/SWATCH-2105)

## Description
Currently, the where clause includes this expression
`snapshot_date > now() - interval '2 day';`
However, depending on timing, this sometimes gathers 1 days data, sometimes 2 days worth.

Instead, change the clause to
`snapshot_date > date_trunc('day', now() - interval '2 day')`
which will force the query to pull only data from tallies for 2 days ago.

## Testing
1.- `podman-compose up` to start up the PostgreSQL database instance.
2.- `./gradlew liquibaseUpdate` to perform all the migrations
3.- Insert the test data:

```sql
-- Create snapshot 000000000001 from today

INSERT INTO tally_snapshots(id, product_id, granularity, org_id, snapshot_date, sla, "usage", billing_provider, billing_account_id)
VALUES('00000000-0000-0000-0000-000000000001', 'rosa', 'DAILY', '01', now(), '_ANY', '_ANY', '_ANY', '_ANY');

-- Create snapshot 000000000002 from yesterday

INSERT INTO tally_snapshots(id, product_id, granularity, org_id, snapshot_date, sla, "usage", billing_provider, billing_account_id)
VALUES('00000000-0000-0000-0000-000000000002', 'rosa', 'DAILY', '02', now() - interval '1 day', '_ANY', '_ANY', '_ANY', '_ANY');

-- Create snapshot 000000000003 from two days ago

INSERT INTO tally_snapshots(id, product_id, granularity, org_id, snapshot_date, sla, "usage", billing_provider, billing_account_id)
VALUES('00000000-0000-0000-0000-000000000003', 'rosa', 'DAILY', '03', now() - interval '2 day', '_ANY', '_ANY', '_ANY', '_ANY');

-- Create snapshot 000000000004 from three days ago

INSERT INTO tally_snapshots(id, product_id, granularity, org_id, snapshot_date, sla, "usage", billing_provider, billing_account_id)
VALUES('00000000-0000-0000-0000-000000000004', 'rosa', 'DAILY', '04', now() - interval '3 day', '_ANY', '_ANY', '_ANY', '_ANY');

-- Create measurements for all the previous snapshots
INSERT INTO tally_measurements(snapshot_id, measurement_type, metric_id, value)
	select s.id, 'PHYSICAL', 'CORES', 1
	from tally_snapshots s
	left join tally_measurements m on m.snapshot_id = s.id 
	where id::text like '00000000-0000-0000-0000-00000000000%' and m.snapshot_id is null;
```

4.- Run the query that has changed. 

You should see only three entries:

```
01	2024-01-19 13:27:43.335 +0100	rosa	PHYSICAL	CORES	CORES	1.0	2024-01-17 00:00:00.000 +0100
02	2024-01-18 13:27:43.335 +0100	rosa	PHYSICAL	CORES	CORES	1.0	2024-01-17 00:00:00.000 +0100
03	2024-01-17 13:27:43.335 +0100	rosa	PHYSICAL	CORES	CORES	1.0	2024-01-17 00:00:00.000 +0100
```

The entry with orgId 04 should be out because is three days old. 

Before these changes, the entry with orgId 03 was also out.

### DIFF
--- a/swatch-tally/deploy/clowdapp.yaml
+++ b/swatch-tally/deploy/clowdapp.yaml
@@ -699,7 +699,7 @@ objects:
         and billing_account_id = '_ANY'
         and granularity='DAILY'
         and measurement_type!='TOTAL'
-        AND snapshot_date > now() - interval '2 day'; 
+        AND snapshot_date at time zone 'utc' = date_trunc('day', now() at time zone 'utc' - interval '2 day');
     - prefix: swatch/subscriptions
       query: >-
         select


swatch-tally/deploy/clowdapp.yaml
```diff
@@ -699,7 +699,7 @@ objects:
         and billing_account_id = '_ANY'
         and granularity='DAILY'
         and measurement_type!='TOTAL'
-        AND snapshot_date > now() - interval '2 day'; 
+        AND snapshot_date > date_trunc('day', now() - interval '2 day');
```
Sgitario, line 702: @kahowell the JIRA description mentioned `snapshot_date=date_trunc('day', now() - interval '2 day')` but I think the correct statement is to use greater than `snapshot_date>date_trunc('day', now() - interval '2 day')`

kahowell: No actually, the point is to export a specific single day's worth of data consistently (the export runs daily). Our daily granularity records use a snapshot date which has a snapshot date that is truncated. Try some queries with gabi (use `count(*)` and/or `select date_trunc('day', now() - interval '2 day');` and you'll see get a good idea data would be returned).

Sgitario: What I first did when I started working in this ticket was to import all the data from stage to local and use the query you provided, but I got empty results. 
I did check again and I see now that the snapshot date is sometimes truncated and I guess this is the data we want to export here. 
My problem was that this date was in a different time zone that my now() date, that's why I could not find any records. I've confirmed this is not happening when using gabi.
I updated the pull request with taking into account the time zone which works locally and with gabi. Yet this is only a problem when running the query locally, so let me know if you prefer removing the time zone bits.

kahowell (approved):

PR #2994, opened 2024-01-24
SWATCH-2083: Update contract ingestion to require less data

kahowell:
Jira issue: [SWATCH-2083](https://issues.redhat.com/browse/SWATCH-2083)

With this change, a contract can be ingested before the subscriptionNumber has been assigned.

I made changes to our handling of the end dates, so that the contract records no longer use `null` end date for the most current contracts, but use the passed end date.

I added `billingProviderId` to `ContractEntity`, and changed the upsert to use billingProviderId to lookup the existing record, instead of depending on subscriptionNumber.

I also updated the `bin/liquibase-new-changeset.sh` script to work for swatch-contracts.

I also updated WireMockResource to enable use of the static stubbing methods in any test class using WireMock.

Testing
-------

See AzureContractLifecycleIntegrationTest, which simulates the various states an Azure contract data goes through.

### DIFF
--- a/swatch-contracts/src/main/resources/application.properties
+++ b/swatch-contracts/src/main/resources/application.properties
@@ -75,7 +75,7 @@ quarkus.datasource.password=${DATABASE_PASSWORD}
 quarkus.datasource.jdbc.url=jdbc:postgresql://${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_DATABASE}
 
 %test.quarkus.datasource.db-kind=h2
-%test.quarkus.datasource.jdbc.url=jdbc:h2:mem:db
+%test.quarkus.datasource.jdbc.url=jdbc:h2:mem:db;NON_KEYWORDS=VALUE
 %test.quarkus.hibernate-orm.dialect=org.hibernate.dialect.H2Dialect
 
 quarkus.liquibase.database-change-log-lock-table-name=DATABASECHANGELOGLOCK_SWATCH_CONTRACTS
--- a/swatch-contracts/src/main/resources/application.properties
+++ b/swatch-contracts/src/main/resources/application.properties
@@ -75,6 +75,7 @@ quarkus.datasource.password=${DATABASE_PASSWORD}
 quarkus.datasource.jdbc.url=jdbc:postgresql://${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_DATABASE}
 
 %test.quarkus.datasource.db-kind=h2
+# NOTE: because some of the entities use columns named "value", it is necessary to configure h2 to *not* treat it as a keyword.
 %test.quarkus.datasource.jdbc.url=jdbc:h2:mem:db;NON_KEYWORDS=VALUE
 %test.quarkus.hibernate-orm.dialect=org.hibernate.dialect.H2Dialect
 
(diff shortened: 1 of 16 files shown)


kflahert (approved):

swatch-contracts/src/main/resources/application.properties
```diff
@@ -75,7 +75,7 @@ quarkus.datasource.password=${DATABASE_PASSWORD}
 quarkus.datasource.jdbc.url=jdbc:postgresql://${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_DATABASE}
 
 %test.quarkus.datasource.db-kind=h2
-%test.quarkus.datasource.jdbc.url=jdbc:h2:mem:db
+%test.quarkus.datasource.jdbc.url=jdbc:h2:mem:db;NON_KEYWORDS=VALUE
```
Sgitario, line 78: what is the property `NON_KEYWORDS` for? Can you add a comment for this?

swatch-contracts/src/main/resources/db/changeLog.xml
```diff
@@ -9,5 +9,7 @@
   <include file="db/202302030707_create_contracts_table.xml"/>
   <include file="db/202304061624_add_vendorproductcode_to_contracts_table.xml"/>
   <include file="db/202310091227_fix_contract_metrics_for_rosa.xml"/>
-
-</databaseChangeLog>
\ No newline at end of file
+  <include file="db/202401111039_make_subscription_number_nullable.xml"/>
+  <include file="db/202401121651_add_billingproviderid_to_contracts.xml"/>
+  <include file="/db/202401171354_add_subscription_and_offering_h2_only.xml"/>
```
Sgitario, line 14:
```suggestion
  <include file="db/202401171354_add_subscription_and_offering_h2_only.xml"/>
```

Sgitario (commented):
Most of the changes look good. I only have one doubt about the new extractBillingProviderId method in the contract mapper which I think it won't be ever used.

kahowell:
@Sgitario feedback addressed, please take another look when you have a chance.

Sgitario (approved):
Many thanks @kahowell !

TrayvonMcKnight (approved):
This looks alright to me. We currently have automation in progress that will handle testing for this; expanded from our AWS tests.

ssidhaye (approved):
After azure contract operations, I do not see 
`Empty value in non-null fields for contract entity` 
in the pod logss

TrayvonMcKnight:
To test this in IQE, you just need to create an Azure contract. The contract will be missing a subscription id due to PartnerAPI not having it present. With this fix, the contract should still go through, and the id should show up later(when it's available)  

1. Create an azure contract
`app.rhsm_subscriptions.create_contract(
        product_id=product_id,
        billing_provider=billing_provider,
        contract_info=contract_payload,
    )`   
 
2. Check that contract is created
`app.rhsm_subscriptions.get_contracts()
`

ssidhaye:
Minor addition to the above test steps :
Before creating the contract we need the payload which can be created using the following helper:
`contract_payload = app.rhsm_subscriptions.generate_contract_payload(product_id='rosa', billing_provider='aws')`

PR #2999, opened 2024-01-25
Fix the MetricsTaskDescriptor definition

Sgitario:
## Description
Rename the "product_id" name from the required list to the right name "product_tag" ("product_id" does not exist).

No need QE here.


Sgitario:
Self-merging since this is a minor fix.

PR #3007, opened 2024-01-29
SWATCH-2113 Add internal API to update Billable Usage remittance value

kartikshahc:
Jira issue: SWATCH-2113

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.


mstead (changes requested):
This works as expected and will provide the functionality we need.

I would expect to see this code hang around until we get the DLQ/retry functionality in place for our billable usage processors. Even though this may be throw away, lets rename the API call to reflect that we are resetting the value of remittance, instead of simply a PUT to billable_usage.

We should get a couple tests around some of this logic since it will likely be around for a short while.

kartikshahc:
> This works as expected and will provide the functionality we need.
> 
> I would expect to see this code hang around until we get the DLQ/retry functionality in place for our billable usage processors. Even though this may be throw away, lets rename the API call to reflect that we are resetting the value of remittance, instead of simply a PUT to billable_usage.
> 
> We should get a couple tests around some of this logic since it will likely be around for a short while.

@mstead I have added a failing test and a non failing test in a method. Also I have renamed the API endpoint to `reset_billable_usage_remittance`. Please let me know it looks good to you. Thanks!

src/main/spec/internal-billing-api-spec.yaml
```diff
@@ -83,6 +83,52 @@ paths:
           $ref: "../../../spec/error-responses.yaml#/$defs/InternalServerError"
       tags:
         - internalBilling
+  /internal/rpc/remittance/reset_billable_usage_remittance:
+    put:
+      summary: Update billable usage remittance records.
```
mstead, line 88: Please add some details of what it means to 'reset' billable usage.

mstead (changes requested):
One more tiny request and it'll be good to go.

src/main/spec/internal-billing-api-spec.yaml
kartikshahc: Done

mstead (approved):

PR #3038, opened 2024-02-07
SWATCH-2192: Not using the path "/q" for swatch metrics ++ add gauge events counter

Sgitario:
Jira issue: [SWATCH-2192](https://issues.redhat.com/browse/SWATCH-2192)

## Description
Configure swatch-metrics to expose the metrics endpoint via the path "/metrics" instead of "/q/metrics" on port 9000 as in Spring services. Note that this implies that the health check endpoints changed as well. 

Moreover, we have added a new gauge metrics to count the number of events fetched from prometheus.

## Testing
This will be tested in stage.

### DIFF
--- a/swatch-metrics/deploy/clowdapp.yaml
+++ b/swatch-metrics/deploy/clowdapp.yaml
@@ -131,11 +131,11 @@ parameters:
   - name: OTEL_EXPORTER_OTLP_ENDPOINT
     value: 'http://localhost:4317'
   - name: METRICS_HEALTH_LIVENESS_PATH
-    value: /q/health/live
+    value: /health/live
   - name: METRICS_HEALTH_READY_PATH
-    value: /q/health/ready
+    value: /health/ready
   - name: METRICS_HEALTH_PORT
-    value: '8000'
+    value: '9000'
 objects:
   - apiVersion: cloud.redhat.com/v1alpha1
     kind: ClowdApp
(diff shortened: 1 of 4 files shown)


Sgitario:
> According to [kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/), `app.kubernetes.io/managed-by` is for "the tool being used to manage the operation of an application" (e.g. helm). Let's use a different label. We could use anything, so I'd suggest `stack` or `framework` to be more explicit. Additionally, if we're going to make these changes, I'd rather remove the existing `prometheus` labels. (This does require us to have servicemonitors for both for a short period). One other note for deployment, we'll likely have to work with devprod to manually remove the existing deployments and redeploy, as the labels are currently immutable.

Quarkus uses the 'app.kubernetes.io/managed-by' for this purpose, but I'm ok to change it to 'framework'. 

However, removing the existing label implies to manually do things and the idea was to include a quick fix for this, since we are missing a lot of metrics. If devprod can provide a quick fix for this, then we would simply rename the existing label to quarkus (which the original change that was reverted).

kahowell:
> If devprod can provide a quick fix for this, then we would simply rename the existing label to quarkus (which the original change that was reverted).

The workaround for adding or changing any labels is currently to have the devprod team remove the existing k8s deployments. Otherwise, we need to wait for [RHCLOUD-30608](https://issues.redhat.com/browse/RHCLOUD-30608) to be implemented.

Sgitario:
@kahowell I've updated the pull request with a different solution which is only for swatch-metrics and that does not rely on pod labels and service monitors.

kahowell (approved):

PR #3044, opened 2024-02-08
SWATCH-2203: Get product tags by product name when no role or eng ids

Sgitario:
Jira issue: [SWATCH-2203](https://issues.redhat.com/browse/SWATCH-2203)

## Description
In https://github.com/RedHatInsights/rhsm-subscriptions/pull/2905 we changed how to resolve the offering product tags. Before, we were using the product name to get the linked product tags. After these changes, we started using the offering role AND/OR the offering eng IDs. However, it seems that this change was not aligned with the upstream offering data where offerings were missing both role and eng IDs. 
Therefore, we need to keep using the product name when the offering is missing the role and the eng IDs to not break the contract ingestion. 

## Testing
1.- podman-compose up
2.- DEV_MODE=true SPRING_PROFILES_ACTIVE=worker ./gradlew :bootRun
3.- Create offering:

`INSERT INTO OFFERING(sku,product_name,product_family,sla,description,has_unlimited_usage, metered) VALUES ('MW02393','OpenShift Online','OpenShift Enterprise','Premium','Red Hat OpenShift Service on AWS Hosted Control Planes (Hourly)','false', 'true');`

4.- Check for offering tags:

```
curl -X GET -H 'Content-Type: application/json' -H 'Origin:console.redhat.com' -H 'x-rh-swatch-psk:placeholder' http://localhost:8000/api/rhsm-subscriptions/v1/internal/offerings/MW02393/product_tags
```

It should return:

```
{"data":["rosa"]}
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java
@@ -48,6 +48,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.candlepin.clock.ApplicationClock;
 import org.candlepin.subscriptions.capacity.CapacityReconciliationController;
 import org.candlepin.subscriptions.capacity.files.ProductDenylist;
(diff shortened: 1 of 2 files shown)


kahowell (approved):

kartikshahc (approved):
Since only Contracts (payg) use the findProductTags method, this PR should currently function properly. However for longer term with my current card SWATCH-2056, I am going to add the capability of non-payg identification and payg in the same method.

PR #3021, opened 2024-02-02
SWATCH-2089: Bump com.netflix.nebula:nebula-release-plugin from 18.0.8 to 19.0.4

Sgitario:
Jira issue: [SWATCH-2089](https://issues.redhat.com/browse/SWATCH-2089)

## Description
Bumps [com.netflix.nebula:nebula-release-plugin](https://github.com/nebula-plugins/nebula-release-plugin) from 18.0.8 to 19.0.4. 

This is related to this change: https://github.com/nebula-plugins/nebula-release-plugin/pull/253
Therefore, we need to configure the user we use in our actions before running the build. 

I also tried to do this as part of a gradle task, but they are running the verification of the git user as part of the release initial task, so we can't configure it. 

### More details
<details>
<summary>Release notes</summary>
<p><em>Sourced from <a href="https://github.com/nebula-plugins/nebula-release-plugin/releases">com.netflix.nebula:nebula-release-plugin's releases</a>.</em></p> <blockquote>
<h2>v19.0.4</h2>
<ul>
<li><a href="https://github.com/nebula-plugins/nebula-release-plugin/commit/4e96a060718c93a231a2f04612e47ddb9f5691f5">Improve error logging for tag push</a></li> </ul>
<p><strong>Full Changelog</strong>: <a href="https://github.com/nebula-plugins/nebula-release-plugin/compare/v19.0.3...v19.0.4">https://github.com/nebula-plugins/nebula-release-plugin/compare/v19.0.3...v19.0.4</a></p>
<h2>v19.0.3</h2>
<ul>
<li><a href="https://github.com/nebula-plugins/nebula-release-plugin/commit/714cd39440b91176b175a2271c64dea64dbfba4a">GitReadCommand: DescribeHeadWithTag handles tags known externally as warning</a></li>
</ul>
<p><strong>Full Changelog</strong>: <a href="https://github.com/nebula-plugins/nebula-release-plugin/compare/v19.0.2...v19.0.3">https://github.com/nebula-plugins/nebula-release-plugin/compare/v19.0.2...v19.0.3</a></p>
<h2>v19.0.2</h2>
<ul>
<li><a href="https://github.com/nebula-plugins/nebula-release-plugin/commit/5e778b63a1aa0f3a65c8a6ca06d2a16eb6a3dea0"><code>GetGitConfigValue</code> should use <code>--includes</code> to respect <code>include.*</code> directives in config files when looking up values</a></li> </ul>
[...]

PR #3045, opened 2024-02-08
Run an init script in the container builds to set git email

awood:
This is a result of the update of
com.netflix.nebula:nebula-release-plugin from 18.0.8 to 19.0.4.  For
some reason the new plugin demands that the user.name and user.email be
set in the git configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,17 @@ Ensure the checkout has the HBI submodule initialized:
 git submodule update --init --recursive
 ```
 
+Make sure you have `user.email` and `user.name` set in the local `.git/config`.
+The Nebula release plugin has an annoying property where it introspects the Git
+config for the name and email address.  When doing container builds, there is no
+global Git config so if you haven't set the name and email locally, the
+container build will fail.  Use `git config --local` to set these values.
+
+```
+git config --local user.name "John Doe"
+git config --local user.email johndoe@example.com
+```
+
 ### Dependent services
 
 NOTE: in order to deploy insights-inventory (not always useful), you'll need to login to quay.io first.


Sgitario (changes requested):
Why do we need to do this? Theoritically, the only needed changes was done in https://github.com/RedHatInsights/rhsm-subscriptions/pull/3021 and the build of the images was fixed.

Sgitario:
Update: QE just noticed that the deployment to stage was failing. I fixed that using this commit https://github.com/RedHatInsights/rhsm-subscriptions/commit/50a0a9868324a7736882a14dfeafee77933a7009. 
The solution is the same as in the linked pull request: https://github.com/RedHatInsights/rhsm-subscriptions/pull/3021.

Did you encounter this issue in somewhere else?

awood:
Yes, you can't build the images by hand from the Dockerfiles without this.  So the `bin/build-images.sh` script that can be used for building images to deploy into the ephemeral environments by developers was broken.

Sgitario:
> Yes, you can't build the images by hand from the Dockerfiles without this. So the `bin/build-images.sh` script that can be used for building images to deploy into the ephemeral environments by developers was broken.

Developers should have configured their user name and email locally. Actually, I didn't hit the issue when I tried to run this script. The problem with these changes is that we will always use the user of the last commit.

awood:
> Developers should have configured their user name and email locally. Actually, I didn't hit the issue when I tried to run this script. The problem with these changes is that we will always use the user of the last commit.

Yes, I agree.  I do actually have the name and email configured locally.  The problem is that I had that information configured in `~/.gitconfig`.  Consequently, those values were not in `.git/config` in the project.  I made the change to add the value to the local config, but this is a bit of a landmine.  I can see it blowing up in someone else's face in the future (because after all, why bother setting your name and email for each repo when you can just do it globally?).

Any thoughts on the best way to warn people about this?



when the build is performed *in the container*, those settings are not present.  You're right that the name is just going to be whatever is on HEAD.  What do you think about just statically setting the user name to "Subscription Watch" and the email to the team list.  I would like to get this merged as currently the `bin/build_images.sh` script is broken and I use that script when deploying to the EEs.

Would you mind testing and seeing if 

```
podman build . -t scratch
```

works for you?

Sgitario:
> > Developers should have configured their user name and email locally. Actually, I didn't hit the issue when I tried to run this script. The problem with these changes is that we will always use the user of the last commit.
> 
> Yes, I agree. I do actually have the name and email configured locally. The problem is that I had that information configured in `~/.gitconfig`. Consequently, those values were not in `.git/config` in the project. I made the change to add the value to the local config, but this is a bit of a landmine. I can see it blowing up in someone else's face in the future (because after all, why bother setting your name and email for each repo when you can just do it globally?).
> 
> Any thoughts on the best way to warn people about this?

Indeed, this is a very annoying change by the Nebula release plugin (why are we using this plugin? :S)

To be honest, since we're not really using the GIT user for anything, I would be ok to add the following:

```
# Initialize the GIT config which is required by the Gradle Nebula plugin to build the images
git config user.name "$(git --no-pager log --format=format:'%an' -n 1)"
git config user.email "$(git --no-pager log --format=format:'%ae' -n 1)"
```

as part of the build-images.sh script as done also in https://github.com/RedHatInsights/rhsm-subscriptions/commit/50a0a9868324a7736882a14dfeafee77933a7009.

This is mostly to avoid having to modify all the Dockerfile (and hence increasing the complexity of those) to deal with a suspicious change of the Nebula plugin.

awood:
> To be honest, since we're not really using the GIT user for anything, I would be ok to add the following:
> 
> ```
> # Initialize the GIT config which is required by the Gradle Nebula plugin to build the images
> git config user.name "$(git --no-pager log --format=format:'%an' -n 1)"
> git config user.email "$(git --no-pager log --format=format:'%ae' -n 1)"
> ```
> 
> as part of the build-images.sh script as done also in [50a0a98](https://github.com/RedHatInsights/rhsm-subscriptions/commit/50a0a9868324a7736882a14dfeafee77933a7009).

I thought about that, but wouldn't that end up affecting the person's local repo?  It'd end up setting the username and email in their local repo to whatever HEAD was.  Honestly, just getting rid of that plugin seems the best.  I don't think we actually use it for anything.

Sgitario (approved):

PR #3050, opened 2024-02-09
SWATCH-2206: Add retries/debug logs to Partner Gateway processing

kahowell:
Jira issue: [SWATCH-2206](https://issues.redhat.com/browse/SWATCH-2206)

Description
===========

In order to handle transient issues with the data coming back from the partner gateway API, as well as provide debug logging to assist the partner gateway engineers in debugging issues.

Note that I added a custom debug logger `DebugClientLogger` that can be configured to log the request and response bodies. See comments in `application.properties` for details. (It's configured for partnerEntitlement requests for `dev`, `test`, and `stage` profiles only).

I changed the contract entity validation to use hibernate validator, to get better messages about which fields are missing. (Each failed field is now logged at a `WARN` level).

I tweaked the API enrichment so that if the contract fails validation afterward, the enrichment will be retried up to 10 times, with a 500ms delay between each retry (adding ~5 seconds total for this unexpected case). Given we expect the initial message (not having an orgId for the contract) to fail until the contract is linked, this type of failure is _not_ retried.

Testing
=======

Run the AzureContractLifecycleIntegrationTest to see changes in logging and retries:

```shell
./gradlew :swatch-contracts:test -i --tests=AzureContractLifecycleIntegrationTest
```

Observe:
* Retries are verified by wiremock `verify` calls.
* Request and response bodies are logged (look for `DebugClientLogger` logs in the output).

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/config/DebugClientLogger.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/config/DebugClientLogger.java
@@ -25,7 +25,6 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpClientRequest;
 import io.vertx.core.http.HttpClientResponse;
 import jakarta.enterprise.context.ApplicationScoped;
-import java.util.Objects;
 import java.util.regex.Pattern;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
@@ -40,8 +39,6 @@ public class DebugClientLogger implements ClientLogger {
   @ConfigProperty(name = "rest-client-debug-logging.uri-filter")
   Pattern uriFilterPattern;
 
-  private ThreadLocal<Boolean> shouldLog = new ThreadLocal<Boolean>();
-
   @Override
   public void setBodySize(int bodySize) {
     // intentionally ignored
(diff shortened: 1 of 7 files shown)


swatch-contracts/src/main/java/com/redhat/swatch/contract/config/DebugClientLogger.java
```diff
@@ -0,0 +1,67 @@
...
+      response.bodyHandler(body -> log.debug("Response: \n{}", body.toString()));
+    }
+  }
+
+  @Override
+  public void logRequest(HttpClientRequest request, Buffer body, boolean omitBody) {
+    if (uriFilterPattern.matcher(request.getURI()).matches()) {
+      shouldLog.set(true);
+      log.debug(
+          "Request method={} URI={}: \n{}", request.getMethod(), request.getURI(), body.toString());
```
awood, line 62: Would it make sense to log the body as well if there is one?

awood (approved):

swatch-contracts/src/main/java/com/redhat/swatch/contract/config/DebugClientLogger.java
```diff
@@ -0,0 +1,67 @@
...
+  private ThreadLocal<Boolean> shouldLog = new ThreadLocal<Boolean>();
+
+  @Override
+  public void setBodySize(int bodySize) {
+    // intentionally ignored
+  }
+
+  @Override
+  public void logResponse(HttpClientResponse response, boolean redirect) {
+    if (Objects.equals(shouldLog.get(), Boolean.TRUE)) {
```
Sgitario, line 52: I don't think this would work in resteasy reactive (I know we're using resteasy legacy) since the thread that handles the body is not guaranteed to be the same that the request one. 
I would avoid dealing with this and check the request get URI again here.

Sgitario (changes requested):

PR #3083, opened 2024-02-16
SWATCH-2223: Update swatch-producer-aws to process billable-usage-hourly-aggregates

kflahert:
Jira issue: SWATCH-2223

## Description

AWS has a limitation of 1 usage record an hour for a given resource/subscription/org. If remittances are retried then some will fail since they will be billed for the same hour. To prevent this issue we should consume messages from the billable-usage-hourly-aggregate topic instead of from billable-usage topic.

**Deployment Issue**
When this is deployed there will be an overlap of the old swatch-producer-azure and the new billable-usage-hourly-aggregate consumer are reading the same messages. We need to come up with a strategy for preventing duplicate billing during this time.

### Setup

1. Insert test data:
```
INSERT INTO public.offering VALUES ('BASILISK', 'BASILISK', NULL, 0, 0, 0, 0, 'BASLISK', 'Premium', 'Production', NULL, NULL, NULL, NULL);

INSERT INTO subscription(
    sku, org_id, subscription_id, quantity, start_date, end_date, subscription_number, billing_provider, billing_provider_id, billing_account_id)
    VALUES ('BASILISK', '123', '1234436512', '3', '2021-11-30T05:00:00Z', '2031-01-01T01:10:00Z', 123511, 'aws', '12345;test;test', '1234567891234');
```
1. In 3 separate terminals run following commands to start services:

`DEV_MODE=true SERVER_PORT=8101 ./gradlew :bootRun`

```
SERVER_PORT=8001 SWATCH_INTERNAL_SUBSCRIPTION_ENDPOINT=http://localhost:8000 KSTREAM_BILLABLE_USAGE_AGGREGATION_WINDOW_DURATION=10s KSTREAM_BILLABLE_USAGE_AGGREGATION_GRACE_DURATION=2s ./gradlew :swatch-producer-azure:quarkusDev
```

`./gradlew  :swatch-producer-aws:quarkusDev`

### Steps

1. Produce the following message twice in a row onto platform.rhsm-subscriptions.billlable-usage topic:
`{"org_id":"123","id":null,"billing_provider":"aws","billing_account_id":"1234567891234","snapshot_date":"2024-01-01T01:10:28Z","product_id":"BASILISK","sla":"Premium","usage":null,"uom":"Storage-gibibyte-months","value":42.0}`
2. Wait at least 12 seconds.
3. Produce the message one more time to the topic (this triggers the previous hourly aggregate message to send)

### Verification

1. Check logs of swatch-producer-azure for message like this:
` Sending usage request to AWS: BatchMeterUsageRequest(UsageRecords=[UsageRecord(Timestamp=2024-02-16T21:00:00Z, CustomerIdentifier=test, Dimension=storage_gb, Quantity=84)], ProductCode=12345), organization=123, product_id=BASILISK`


Sgitario (changes requested):
This solution strongly depends on the Swatch Azure being up and running because this service is consuming the billable usage topic and producing the aggretations onto the `platform.rhsm-subscriptions.billable-usage-hourly-aggregate` topic. 

Each Swatch Azure and Swatch AWS services should be independent with each other, so both services should consume the billable usage topic and product to the hourly aggregate topic. 

Also, the `deploy/clowadpp.yaml` file for the Swatch AWS service needs to be updated to add the new topics that it needs. 

Another suspicious thing is that when I did the steps from the PR description, I expected a final total value of 126 because we're sending three messages with a value of `42`, but I got 84 (as said in the PR description). So, a couple of questions:
- why is this happening?
- why do we need to send a third message to trigger the billable usage aggregation?

kflahert:
@Sgitario The kstreams aggregation logic in swatch-producer-azure is going to be moved into its own deployment but we did this in the short term to hit deadlines. 

As for why the final value is 84 and not 126, this is because the aggregate window is for 10 seconds (plus a 2 second grace period to get out of order messages) and so only the first 2 messages land in the first window. Kstreams works off of event time and only knows that the window has passed once we get a new message. So when the third message comes in then the original 10 second windowed aggregate is published since now kstreams knows that the window  is closed. Then that 3rd message is stored as its own windowed aggregate that is ready to be published again once the window is closed and another message is put onto the queue so kstreams' internal clock can advance. This would have a total value of just 42 for the single message(Not in test steps just to elaborate further).

Also, added the missing topic to clowdapp.

Sgitario:
> @Sgitario The kstreams aggregation logic in swatch-producer-azure is going to be moved into its own deployment but we did this in the short term to hit deadlines.

Ok, if this is the case, I would add it as a dependency in the clowdapp.yaml file of our aws service. So, we clearly see that the aws service strongly depends on the azure service and a comment saying that this is to be refactored in any SWATCH-XXX ticket. 

> 
> As for why the final value is 84 and not 126, this is because the aggregate window is for 10 seconds (plus a 2 second grace period to get out of order messages) and so only the first 2 messages land in the first window. Kstreams works off of event time and only knows that the window has passed once we get a new message. So when the third message comes in then the original 10 second windowed aggregate is published since now kstreams knows that the window is closed. Then that 3rd message is stored as its own windowed aggregate that is ready to be published again once the window is closed and another message is put onto the queue so kstreams' internal clock can advance. This would have a total value of just 42 for the single message(Not in test steps just to elaborate further).

Many thanks for your explanation! Really helpful. Yet I see a risk of billable usages might be sent too late (out of the hour) because the window is closed when and only when a new message is received. Might this happen? If so, I'm not an expert on Kafka streams, but I think we can configure a scheduler to not depend on messages coming. I can help trying to investigate this if you want. 

> Also, added the missing topic to clowdapp.

Thanks!

kflahert:
@Sgitario added the dependency to clowdapp w/ comment as suggested.

As for the messages not getting sent on time- This shouldn't be an issue because we will be using the time of the window(bottom of current hour the first usage is consumed) to bill and the cloud providers allow us to bill within the past 24hrs. Since we are consistently billing every hour there would only ever be a delay of 1 hr for when the billable-usage-hourly-aggregate message gets sent.

Sgitario:
> @Sgitario added the dependency to clowdapp w/ comment as suggested.

Thanks!

> ... the cloud providers allow us to bill within the past 24hrs...

So, as long as a new message is coming within the next 24 hours, we're safe. I understand that the current solution will work and it has a low risk to fail. Though, I've created SWATCH-2227 to improve it if possible in a separate PR.

I'm approving these changes!

Sgitario (approved):

TrayvonMcKnight:
The fix the PR build test is currently in PR:

https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/596

TrayvonMcKnight (approved):
This PR has an impact on the failing IQE smoke_test test case "test_verify_rhel_payg_kafka_event". This PR has been verified in multiple EEs, multiple times. 

https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/596

PR #3085, opened 2024-02-19
SWATCH-2190: Change Splunk certificate verification in dev to false

awood:
When using Splunk in the dev profile, developers will be sending their events to a Splunk docker container that will not have a server certificate signed by a recognized certificate authority.  Instead the certificate will be self-signed.  Configuring the applications to disable certificate validation in the dev (and only in the dev) profile will be make life easier for developers while still maintaining security in non-development environments.

Jira issue: SWATCH-2190

## Description
Correct some of the documentation and configuration around running Splunk in a development capacity.

### Setup
1.  Start a Splunk docker container
    ```
    podman-compose -f config/splunk/docker-compose.yml up -d
    ```

### Steps
1.  Run `swatch-producer-azure` as follows:
    ```
    SERVER_PORT=8001 \
    ENABLE_SPLUNK_HEC=true \
    SPLUNK_HEC_URL=https://localhost:8088 \
    SPLUNK_HEC_TOKEN=29fe2838-cab6-4d17-a392-37b7b8f41f75 \
    ./gradlew :swatch-producer-azure:quarkusDev
    ```

### Verification
1.  Wait a bit (~30 seconds) since Splunk sends log messages in batches.  You should not see any TLS/SSL errors/exceptions.  The root exception to look out for is `Caused by: sun.security.provider.certpath.SunCertPathBuilderException`
1. Go to https://localhost:8000 and log in with `admin`/`admin123`.  Then search `namespace="local"` and you should see a bunch of log messages that got passed in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ as idle.
 
 ### RabbitMQ
 
-Some services like swatch-contracts need an AMQ service to handle the UMB messages. 
+Some services like swatch-contracts need an AMQ service to handle the UMB messages.
 For these services, we can start RabbitMQ as an AMQ service locally via:
 
 ```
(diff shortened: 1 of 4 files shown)


Sgitario (approved):
Many thanks for these changes. It worked for me.
I've updated the PR description because to verify the logs in Splunk, we need to use `http://localhost:8000` instead of `https://localhost:8000`, otherwise my browser rejects the connection because the SSL is invalid. But I was able to see the app logs +100.

PR #3096, opened 2024-02-26
SWATCH-2063: Remove json-path dependency

Sgitario:
Jira issue: SWATCH-2063

## Description
- Uses directly the JsonNode from jackson.
- Replace `${clowder.endpoints[?(@.app == 'rbac')].hostname` to `${clowder.endpoints.rbac-service.url}`
- Replace `${clowder.kafka.topics[?(@.requestedName == 'platform.rhsm-subscriptions.offering-sync')].name}` to `${clowder.kafka.topics."platform.rhsm-subscriptions.offering-sync".name}`

## Testing
Regression testing only.

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySource.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySource.java
@@ -45,8 +45,24 @@ import org.springframework.web.context.support.StandardServletEnvironment;
 
 /**
  * Property source that attempts to resolve properties against the Clowder JSON using JSON Path.
- * This class resolves any property starting with "clowder." Everything after the "clowder." prefix
- * is evaluated as a JSON Path expression against the Clowder JSON and the result is returned.
+ * This class resolves any property starting with "clowder.". Everything after the "clowder." prefix
+ * is evaluated by using the following patterns:
+ *
+ * <ul>
+ *   <li>clowder.kafka.brokers
+ *   <li>clowder.kafka.brokers.(sasl.securityProtocol|sasl.mechanism|sasl.jaas.config|cacert|cacert.type)
+ *   <li>clowder.kafka.topics.<topics[?].requestedName(*)>.name
+ *   <li>clowder.endpoints.<endpoints[?].app>-<endpoints[?].name>.(url|trust-store-path|trust-store-password|trust-store-type).
+ *       For example: "clowder.endpoints.index-service.url"
+ *   <li>clowder.privateEndpoints.<privateEndpoints[?].app(*)>-<privateEndpoints[?].name(*)>.(url|trust-store-path|trust-store-password|trust-store-type).
+ *       For example: "clowder.privateEndpoints.index-service.url"
+ *   <li>When there is no matching rule from above, we navigate from (*) and extract the string
+ *       value directly from the clowder config file. For example: "clowder.database.name".
+ * </ul>
+ *
+ * (*) Coming from the clowder config file. For example: <a
+ * href="https://raw.githubusercontent.com/RedHatInsights/rhsm-subscriptions/main/swatch-core/src/test/resources/test-clowder-config.json">see
+ * example of a clowder config file</a>
  */
 @EqualsAndHashCode(callSuper = true)
 @Slf4j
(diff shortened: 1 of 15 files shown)


swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySource.java
```diff
@@ -135,30 +126,23 @@ public ClowderJsonPathPropertySource() throws IOException {
...
-    // handling for special properties like kafka brokers
-    if (name.contains(KAFKA_BROKERS)) {
-      String brokerConfig = name.substring(PREFIX.length() + KAFKA_BROKERS.length());
-      // special logic to provide a comma-separated list of kafka brokers
-      // value is expected to be a list of key-value collection with the broker configuration.
-      KafkaBrokerConfigMapper configFunction = KAFKA_BROKERS_PROPERTIES.get(brokerConfig);
-      if (configFunction != null) {
-        return getKafkaBrokerConfig(name, configFunction);
+    for (var strategy : handledPropertyStrategies.entrySet()) {
+      if (name.contains(strategy.getKey())) {
```
awood, line 134: How would you feel about altering this to strip the `clowder.` off the name and then use `startsWith`.  I'm a little nervous that we might someday end up with something like `clowder.kafka-brokers.topics.endpoints` or something along those lines that would end up resolving to the `endpoints` path instead of the `kafka-brokers` path.  Plus, if we expand the map of strategies in the future, that increases the odds of this type of ambiguity.

swatch-core/src/test/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySourceTest.java
```diff
@@ -279,7 +122,9 @@ void testKafkaBrokersPropertyWhenNoBrokersConfiguration() throws Exception {
         "clowder.endpoints.index-service.trust-store-path|file:/tmp/truststore.*.trust",
         "clowder.endpoints.index-service.trust-store-password|.+",
         "clowder.endpoints.index-service.trust-store-type|PKCS12",
-        "clowder.privateEndpoints.export-service-service.url|http://export-service-service.svc:10000"
+        "clowder.privateEndpoints.export-service-service.url|http://export-service-service.svc:10000",
```
awood, line 125: Just for complete coverage, would you mind adding

```
        "clowder.endpoints.rbac-service.url|http://rbac-service.rhsm.svc:8000",
```

to the test case here?

awood (changes requested):
Would you also mind adding some comments somewhere about how the endpoint resolution works?

Something explaining that the class is going to take `clowder.endpoints.$TOKEN.url` and for each endpoint will join the `name` and `app` with a hypen and then compare that to `$TOKEN` and then that the `.url` prefix isn't an actual JSON coordinate but a directive to put the host and port together along with a choice of protocol based on the presence of the `tlsPort` property.

swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySource.java
Sgitario: Yes, this makes totally sense +1

swatch-core/src/test/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySourceTest.java
Sgitario: I don't see how adding this scenario will increase the coverage. I guess you're asking this because the rbac endpoint is not used, if so, I rather remove it.

Sgitario:
@awood PR updated

swatch-core/src/test/java/org/candlepin/subscriptions/clowder/ClowderJsonPathPropertySourceTest.java
awood: Just because `endpoints` and `privateEndpoints` have slightly different code paths.  `getPrivateEndpointConfig` vs `getEndpointConfig`

awood:
> @awood PR updated. I've added some documentation about how to use it in the clowder config property source class javadocs. Let me know if you prefer this documentation somewhere else.

Thank you!  The javadocs are a great place for it.

awood (approved):

PR #3118, opened 2024-03-04
SWATCH-1896: exclude RHEL rhel ELS/HA/RS from OpenShift and Satellite

Sgitario:
With new RHEL add-ons configs, We need to update includedSubscriptions from OpenShift Container Platform  and Satellite products to include new rhel ELS/HA/RS add-ons.

## Testing notes

I've added a test case that covers these changes. The test case and functionality must be reviewed.

Steps for manual testing (the automatic test case mimics the below steps):

1.- podman-compose up
2.- start service using stubs:

```
PRODUCT_USE_STUB=true SUBSCRIPTION_USE_STUB=true DEV_MODE=true ./gradlew :bootRun
```

Note that as part of these changes, I've added a stub that includes an orgId 790 with two products: 290 (openshift-container-platform) and 70 (rhel-for-x86-eus).

3.- Synchronize the org ID:

```
http PUT ":8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/790" \
Origin:console.redhat.com \
x-rh-swatch-psk:placeholder
```

4.- Verification in RHSM subscription database:

```
select product_id from subscription_product_ids where subscription_id='790'
```

It should return:

```
OpenShift Container Platform
```

Before these changes, it was returning:

```
OpenShift Container Platform
rhel-for-x86-eus
```

Jira issue: SWATCH-XXXX

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.


kahowell (approved):

ntkathole (approved):
Looks good 👍 
Automation coverage - test_tally_for_included_subscriptions and test_tally_for_non_included_subscriptions

PR #3126, opened 2024-03-06
SWATCH-2214 Switch from RateLimiter to BulkLimiter when calling RSHM API

Sgitario:
Jira issue: [SWATCH-2214](https://issues.redhat.com/browse/SWATCH-2214)

After digging into the RateLimiter, I see a couple of disadvantages to limit / contain the RHSM API calls:
1. We did set the limit of calls to 150, but at the same time, this limit is reset every 100 milliseconds, so the limit is too permisive and likely it will ever be reached out.
2. We can't increase the limitRefreshPeriod property to a higher value because we would be too strict and the RateLimiter does not allow to decrement the counter to permit more calls. 

Therefore, I agree with the reporter of SWATCH-2214 that the bulkhead pattern matches better for the purpose of not exhaust the RHSM services. 

This PR replaces the env vars from https://github.com/RedHatInsights/rhsm-subscriptions/pull/2010 with:

* `RHSM_API_MAX_CONCURRENT_CALLS` - how many concurrent calls we will allow for the RHSM API service.
* `RHSM_API_MAX_WAIT_DURATION` - the max time a pending RHSM API interaction should wait before timing out.

These are set by default to values that enforce an effective limit of 150 requests/second:

* max concurrent calls: 5
* max wait duration: 30m

Timeout chosen as 30m, as this corresponds to our kafka max poll interval (if it takes longer than that the kafka message is going to be rejected anyways)

## Testing

1. podman-compose up
2. Run the application with some overrides to test the bulkhead limiting:

```shell
DEV_MODE=true \
  RHSM_USE_STUB=true \
  RHSM_MAX_ATTEMPTS=1 \
  RHSM_API_MAX_CONCURRENT_CALLS=1 \
  RHSM_API_MAX_WAIT_DURATION=10s \
  ./gradlew :swatch-system-conduit:bootRun
```

Check the metrics for the limiter:

```shell
http :9000/metrics | grep resilience4j | grep rhsmApi
```

Output:
```
resilience4j_bulkhead_available_concurrent_calls{name="rhsmApi",} 1.0
resilience4j_bulkhead_max_allowed_concurrent_calls{name="rhsmApi",} 1.0
```

Next, invoke conduit operations:

```shell
http :8000/api/rhsm-subscriptions/v1/internal/rpc/syncOrg \
  org_id=slowOrg123 \
  x-rh-swatch-psk:placeholder
```

and within 5 seconds, open a new terminal and check again the metrics, where you should see:

```
resilience4j_bulkhead_available_concurrent_calls{name="rhsmApi",} 0.0
resilience4j_bulkhead_max_allowed_concurrent_calls{name="rhsmApi",} 1.0
```

The resilience4j_bulkhead_available_concurrent_calls counter is 0.0 because there is an ongoing operation. If we consult again the metrics (when the above conduit operation is done), we should see that the counter is back to 1.0:

```
resilience4j_bulkhead_available_concurrent_calls{name="rhsmApi",} 1.0
resilience4j_bulkhead_max_allowed_concurrent_calls{name="rhsmApi",} 1.0
```

### DIFF
--- a/swatch-system-conduit/deploy/clowdapp.yaml
+++ b/swatch-system-conduit/deploy/clowdapp.yaml
@@ -98,11 +98,9 @@ parameters:
     value: '3'
   - name: HOST_LAST_SYNC_THRESHOLD
     value: '30h'
-  - name: RHSM_API_LIMIT_FOR_PERIOD
-    value: '150'
-  - name: RHSM_API_LIMIT_REFRESH_PERIOD
-    value: 100ms
-  - name: RHSM_API_LIMIT_TIMEOUT_DURATION
+  - name: RHSM_API_MAX_CONCURRENT_CALLS
+    value: '5'
+  - name: RHSM_API_MAX_WAIT_DURATION
     value: 30m
   - name: OTEL_SERVICE_NAME
     value: swatch-system-conduit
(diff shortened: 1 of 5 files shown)


swatch-system-conduit/deploy/clowdapp.yaml
```diff
@@ -98,11 +98,9 @@ parameters:
     value: '3'
   - name: HOST_LAST_SYNC_THRESHOLD
     value: '30h'
-  - name: RHSM_API_LIMIT_FOR_PERIOD
-    value: '150'
-  - name: RHSM_API_LIMIT_REFRESH_PERIOD
-    value: 100ms
-  - name: RHSM_API_LIMIT_TIMEOUT_DURATION
+  - name: RHSM_API_MAX_CONCURRENT_CALLS
+    value: '5'
```
Sgitario, line 102: I think this configuration is too restrictive (only 5 concurrent calls with a timeout of 30m). Looking forward to what you think about it and suggestions.

kahowell: We run 3 replicas and each has a single kafka consumer, so I don't any issue with 5. I'm more interested to see whether this is sufficient to limit API calls enough to keep from overwhelming the RHSM API endpoint.

kahowell (approved):

PR #3136, opened 2024-03-11
SWATCH-2214: Synchronize orgs in conduit using tasks

Sgitario:
Jira issue: [SWATCH-2214](https://issues.redhat.com/browse/SWATCH-2214)

## Description
We will allow to enable a synchronous operation to sync the organization by skipping the Kafka queue by running the swatch conduit with ENABLE_SYNCHRONOUS_OPERATIONS=true, plus adding our standard "x-rh-swatch-synchronous-request" header when calling "/internal/rpc/syncOrg" endpoint.

## Testing
Added unit tests to cover this change.

### DIFF
--- a/swatch-system-conduit/deploy/clowdapp.yaml
+++ b/swatch-system-conduit/deploy/clowdapp.yaml
@@ -49,6 +49,8 @@ parameters:
     value: 1s
   - name: RHSM_BACK_OFF_MULTIPLIER
     value: '2'
+  - name: ENABLE_SYNCHRONOUS_OPERATIONS
+    value: 'false'
   - name: DATABASE_CONNECTION_TIMEOUT_MS
     value: '30000'
   # TODO This has been lowered from what it was in the previous environment (from 25 to 10)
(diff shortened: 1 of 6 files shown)


Sgitario:
> @Sgitario Need to enable async operations in EE and stage `Synchronous conduit operations are not enabled.`

Yes, I know this was the missing bit. I can do this after merging the pull request. However, you're not going to use the sync request, but the Kafka task instead, right?

ntkathole:
> > @Sgitario Need to enable async operations in EE and stage `Synchronous conduit operations are not enabled.`
> 
> Yes, I know this was the missing bit. I can do this after merging the pull request. However, you're not going to use the sync request, but the Kafka task instead, right?

Currently we have enabled automation to use `x-rh-swatch-synchronous-request` true ... Are you suggesting to use `x-rh-swatch-use-thread-pool-executor` instead ?

Sgitario:
> > > @Sgitario Need to enable async operations in EE and stage `Synchronous conduit operations are not enabled.`
> > 
> > 
> > Yes, I know this was the missing bit. I can do this after merging the pull request. However, you're not going to use the sync request, but the Kafka task instead, right?
> 
> Currently we have enabled automation to use `x-rh-swatch-synchronous-request` true ... Are you suggesting to use `x-rh-swatch-use-thread-pool-executor` instead ?

Using "x-rh-swatch-synchronous-request: true" is exactly how it was working before these changes and it will cause gateway timeout issues. 
You need to use x-rh-swatch-synchronous-request: false or not use any header to make it async.

kflahert (approved):

ntkathole (approved):

PR #3139, opened 2024-03-11
Replace hardcoded list of bonfire components in pr_check.sh

lindseyburnett:
Replace hardcoded list of components with a command to derive them based on clowdapp.yaml files found in the repo

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -58,7 +58,7 @@ OPTIONAL_DEPS_METHOD=none
 export COMPONENTS_W_RESOURCES="app:rhsm"
 
 # NOTE: this ensures that all of the other services end up deployed with the latest template
-export EXTRA_COMPONENTS="rhsm swatch-api swatch-contracts swatch-producer-aws swatch-producer-red-hat-marketplace swatch-metrics swatch-subscription-sync swatch-system-conduit swatch-tally swatch-producer-azure"
+export EXTRA_COMPONENTS="rhsm $(find -name clowdapp.yaml -exec dirname {} \; | cut -d'/' -f2 | xargs)"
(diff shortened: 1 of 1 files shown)


pr_check.sh
```diff
@@ -58,7 +58,7 @@ OPTIONAL_DEPS_METHOD=none
 export COMPONENTS_W_RESOURCES="app:rhsm"
 
 # NOTE: this ensures that all of the other services end up deployed with the latest template
-export EXTRA_COMPONENTS="rhsm swatch-api swatch-contracts swatch-producer-aws swatch-producer-red-hat-marketplace swatch-metrics swatch-subscription-sync swatch-system-conduit swatch-tally swatch-producer-azure"
+export EXTRA_COMPONENTS=$(find -name clowdapp.yaml -exec dirname {} \; | sed 's/^.//' | sed 's#^/##' | sed 's#/.*##' | tr '\n' ' ')
```
awood, line 61:
```
"$(find -name clowdapp.yaml -exec dirname {} \; | cut -d'/' -f2 | xargs)"
```

is a little more succinct.  Not sure how it compares on runtime performance (if that is an issue).  But it is basically reliant on the directory being only one level deep.  I think the chain of `sed`s makes a similar assumption though since if you create `foo/bar/deploy/clowdapp.yaml` the end result is `foo` for both pipelines.

The command you have (and the modified one I posted) are both missing `rhsm`.  So you might want to do something like

```
export EXTRA_COMPONENTS="rhsm $(find -name clowdapp.yaml -exec dirname {} \; | cut -d'/' -f2 | xargs)" which will append `rhsm` to the beginning.

awood (changes requested):

awood (approved):

PR #3145, opened 2024-03-12
SWATCH-2186: NOT MERGE! Investigate exports uploads

Sgitario:
Jira issue: [SWATCH-2186](https://issues.redhat.com/browse/SWATCH-2186)

## Description
I've opened up this pull request with the only purpose to investigate SWATCH-2186: Investigate Performance Optimization For Export Fetching. **NOT MERGE**

What I did here is first to insert up to 50000 subscriptions in database, and then manipulate the uploadJson method from ExportSubscriptionListener to provide Stream<Subscription> data. This data will be used to write the final object to be sent to the Export service (see example [here](https://github.com/RedHatInsights/export-service-go/blob/main/example_export_upload.json)).
With this data, we'll use Jackson API to construct a file on the fly (without keeping any objects on memory).

The results prove that we can handle a large number of subscription in a constant time:

| Number of Subscriptions  | Execution Time | Size of the request |
| ----------- | ----------- | ----------- |
| 1      | 21 sec       | 356 bytes |
| 100   | 16 sec        | 26492 bytes |
| 1000   | 16 sec        | 264092 bytes |
| 10000   | 16 sec        | 2.64 MB |
| 50000   | 14 sec        | 13.2 MB |

Find the following memory usage chart where we can see that the amount of memory remained almost constant when executing the upload for the different size of subscriptions:

Captura de pantalla de 2024-03-12 14-28-15

Important technical details:
- We need to directly use the Stream<Subscription> in the upload methods where Subscription is the hibernate entity instead of using the pojo Subscription. 
- I've changed the openapi definition of the export service, so we can directly provide the file. I've confirmed that there is no different in terms of the HTTP request (we're still sending an application/json request as expected).
- I've added a test case named "verifyExportUploadWithSingleSubscription" to verify that request we're sending to the export service. The other test case named "verifyExportUpload" is hiding the body because it's too large to be printed in the terminal.

### DIFF
--- a/clients-core/src/main/java/org/candlepin/subscriptions/http/HttpClient.java
+++ b/clients-core/src/main/java/org/candlepin/subscriptions/http/HttpClient.java
@@ -112,8 +112,8 @@ public class HttpClient {
     if (isDebugging) {
       clientConfig.register(org.jboss.logging.Logger.class);
     }
-    ClientBuilder clientBuilder = ClientBuilder.newBuilder().withConfig(clientConfig);
 
+    ClientBuilder clientBuilder = ClientBuilder.newBuilder().withConfig(clientConfig);
     return ((ResteasyClientBuilder) clientBuilder).httpEngine(engine).build();
   }
 
--- a/clients/export-client/export-api-spec.yaml
+++ b/clients/export-client/export-api-spec.yaml
@@ -41,7 +41,8 @@ paths:
         content:
           application/json:
             schema:
-              type: object
+              type: string
+              format: binary
           application/csv:
             schema:
               type: string
--- a/clients/export-client/src/main/java/com/redhat/swatch/clients/export/api/client/StubExportApi.java
+++ b/clients/export-client/src/main/java/com/redhat/swatch/clients/export/api/client/StubExportApi.java
@@ -22,6 +22,7 @@ package com.redhat.swatch.clients.export.api.client;
 
 import com.redhat.swatch.clients.export.api.model.DownloadExportErrorRequest;
 import com.redhat.swatch.clients.export.api.resources.ExportApi;
+import java.io.File;
 import java.util.UUID;
 
 public class StubExportApi extends ExportApi {
@@ -35,7 +36,7 @@ public class StubExportApi extends ExportApi {
   }
 
   @Override
-  public void downloadExportUpload(UUID id, String application, UUID resource, Object body) {
+  public void downloadExportUpload(UUID id, String application, UUID resource, File body) {
     // do nothing on purpose.
   }
 }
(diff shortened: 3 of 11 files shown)


Sgitario:
@lindseyburnett I've added you as a reviewer since I think you're already working on the uploadJson method implementation as part of another task.

barnabycourt:
Given what you have seen with the memory usage. Would it make sense to try to push it to breaking? e.g. Do a test with 5 or 50 million records?

Sgitario:
> Given what you have seen with the memory usage. Would it make sense to try to push it to breaking? e.g. Do a test with 5 or 50 million records?

At the moment, the limit is 200.000 subscriptions because it reaches the file limit of 50 MB to be sent via REST client (though, we can configure it). 

There was no performance difference when dealing with these subscriptions.

barnabycourt:
Is that 200k record limit something that would follow through for instances as well. if so, this would be a blocker and we would need to work with the export service for an alternate means to get the data across.

Sgitario:
> Is that 200k record limit something that would follow through for instances as well. 

I don't understand this statement. Can you clarify?

> If we are streaming the data out as we read it, what is the source of the limit?

This: 

> At the moment, the limit is 200.000 subscriptions because it reaches the file limit of 50 MB to be sent via REST client (though, we can configure it).

The HTTP connection file limit only allows to send a file up to 50 MB. Having said that, this is configurable, so we can increase it. We should clarify first the size the export service can handle and also if sending 200k subscriptions is a realistic scenario (atm, there are _only_ 50k in stage).

Sgitario:
> For subcriptions I'm less worried, my larger concern is about the pattern for instance data where our largest customers absolutely have hundreds of thousands of systems reporting in

I've asked for this in https://redhat-internal.slack.com/archives/C02317RKT6H/p1710406103536879. We should find out what the max size, the export service support.

Sgitario:
@barnabycourt the export service application seems not to have any size limit to process large data. We agreed to use 500MB instead of 50MB. The problem is that this is configurable via https://github.com/resteasy/resteasy/blob/770297a2dfada0ab76c367405fe61a55258ca1cb/docbook/reference/en/en-US/modules/Multipart.xml#L787 which can be only provided via MicroProfile Config (which in Spring, we don't use it). I will  see how we can workaround this.

Sgitario:
> @barnabycourt the export service application seems not to have any size limit to process large data. We agreed to use 500MB instead of 50MB. The problem is that this is configurable via https://github.com/resteasy/resteasy/blob/770297a2dfada0ab76c367405fe61a55258ca1cb/docbook/reference/en/en-US/modules/Multipart.xml#L787 which can be only provided via MicroProfile Config (which in Spring, we don't use it). I will see how we can workaround this.

It seems that the only way to configure this property is via system property (because we don't have a Microprofile Config context in Spring services). I asked for this in https://github.com/orgs/resteasy/discussions/4085. 

After increasing the file size limit to 500 MB, we were able to send more than 500K subscriptions in few seconds (the memory and execution time remains constant as expected). 

I've created SWATCH-2328 to track the work necessary to increase the file size limit.

Sgitario:
PR update by using "-1" to remove the RESTEASY client file size limitation which works as expected: https://github.com/RedHatInsights/rhsm-subscriptions/pull/3145/files#diff-b2c9e5db6408b868cac60974790ecb7f80c46ca334d28c835c8b5312be7d07c5R56

Morever, I refactored the performance of the solution to directly use the serializer config instead of the object mapper proxy: https://github.com/RedHatInsights/rhsm-subscriptions/pull/3145/files#diff-833f337360b4cd2bf2d15aa9b27f1c30f730be45919f54ed5927d44c308f942eR157. Surprisingly, now the code goes twice of fast than before.

Sgitario:
@lindseyburnett confirmed she is good with these changes. Since this PR was only about to investigate how to deal with huge data and the export service, I'm closing the PR and the ticket.

PR #3178, opened 2024-03-22
SWATCH-2277: Generate Subscription Sync API to swatch-contracts

Sgitario:
Jira issue: [SWATCH-2277](https://issues.redhat.com/browse/SWATCH-2277)

## Description
Generate the Subscription Sync API in swatch-contracts Leave the implementation with defaults (not implemented yet) Methods will always throw NotImplementedYet (special reactive exception to properly map the response status)

## Testing
1.- start contracts services:

```
./gradlew swatch-contracts:quarkusDev
```

2.- call to any of the new endpoints, for example:

```
curl -v http://localhost:8000/api/swatch-contracts/internal/tags/rosa/metrics
```

and you should see:

```
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /api/swatch-contracts/internal/tags/rosa/metrics HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/8.0.1
> Accept: */*
> 
< HTTP/1.1 501 Not Implemented
< content-length: 0
```

### Using QE tool

1.- Import call_internal_rest_api

```python
from smqe_tools import call_internal_rest_api
```

Execute the test:

```python
@pytest.mark.ephemeral
@pytest.mark.post_stage_deploy
@pytest.mark.parametrize("endpoint",
                         ["POST:/internal/subscriptions",
                          "PUT:/internal/rpc/subscriptions/sync",
                          "PUT:/internal/subscriptions/sync/org/123",
                          "DELETE:/internal/subscriptions/prune",
                          "GET:/internal/offerings/MW123/product_tags"])
def test_contracts_api_sync_api_returns_http_not_implemented(application, endpoint):
    """Verify SWatch SaaS Contracts service APIs returns HTTP/1.1 501 Not Implemented for the new Subscription Sync API endpoints.
    metadata:
        assignee: jcarvaja
        negative: false
        test_steps:
            1. Call to the SWatch Contracts API with the path.
        expected_results:
            1. All calls should return HTTP 501 status.
    """
    parts = endpoint.split(":")
    try:
        call_internal_rest_api(
            app="swatch-contracts",
            endpoint=parts[1],
            method=parts[0],
        )
        assert False, f"The endpoint '{endpoint}' is expected to fail with 501 HTTP status"
    except Exception as ex:
        assert "501 Server Error: Not Implemented" in str(ex)

```

Note that the test is not checking all the new endpoints, but a few.


swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/CapacityResource.java
```diff
@@ -52,6 +53,6 @@ public CapacityReportByMetricId getCapacityReportByMetricId(
       ServiceLevelType sla,
       UsageType usage)
       throws ApiException, ProcessingException {
-    throw new UnsupportedOperationException("Not yet implemented");
+    throw new NotImplementedYet();
```
Sgitario, line 56: @kahowell I think you added this endpoint a while ago and it's still not implemented. Do you know if we can fully remove it?

Sgitario:
MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/620

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/CapacityResource.java
kahowell: We can remove it for now. The intent was to move the customer facing APIs for capacity/subscriptions here, but it's not particularly helpful to have a stub just sitting :-).

Sgitario: This would imply to undo much more changes than I initially expected (see https://github.com/RedHatInsights/rhsm-subscriptions/commit/9d2cb2288349898c632ce017ffca22c2d9a23436). So, I think it's best to keep the stub as it is and if it no longer makes sense to have it, do it as part of another ticket. Do you want me to create this ticket or prever keeping the stub as it is?

liwalker-rh (approved):
Manually verified and everything looks good. Happy path automation has also been approved to merge: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/620

lindseyburnett (approved):

PR #3187, opened 2024-03-26
SWATCH-2366: Unify resteasy dependencies into one single property

Sgitario:
Jira issue: [SWATCH-2366](https://issues.redhat.com/browse/SWATCH-2366)

## Description
The following dependencies are declared in gradle individually:

org.jboss.resteasy:resteasy-multipart-provider
org.jboss.resteasy:resteasy-client
org.jboss.resteasy:resteasy-validator-provider
org.jboss.resteasy:resteasy-jackson2-provider
So, when a new version of resteasy is released, dependabot creates four pull requests for each one.

Unify the resteasy dependencies to use a single property `resteasyVersion` to declare the resteasy version.

## Testing
Only regression testing.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -14,6 +14,7 @@ buildscript {
 ext {
     quarkusVersion='3.6.6'
     springVersion='3.2.4'
+    resteasyVersion='6.2.8.Final'
     libraries = [:]
     plugins = []
 }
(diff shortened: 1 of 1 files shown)


Sgitario:
Auto-merging since this is related to some bump dependencies by dependabot.

PR #3197, opened 2024-03-28
SWATCH-2381: Clear billable usage suppress store topic during account

kflahert:
Jira issue: SWATCH-2381

## Description

The windowed billable usage aggregate kafka messages cause issues with IQE tests because they leave lingering messages between tests. This update allows us to delete all messages from that topic between tests.

### Setup

1. Start contract service: `SERVER_PORT=8001 MANAGEMENT_PORT=9002 ./gradlew :swatch-contracts:quarkusDev`
1. Start tally service: `DEV_MODE=true  ./gradlew :bootRun`

### Steps

1. Insert a record (any or no data is fine) into `platform.rhsm-subscriptions.swatch-billable-usage-aggregator-billable-usage-suppress-store-changelog`
1. Run curl command: 
`curl -X 'DELETE'     "localhost:8000/api/rhsm-subscriptions/v1/internal/rpc/tally/org345"     -H 'accept: application/json'     -H 'x-rh-swatch-psk: placeholder'`

### Verification

1. Using a kafka consumer starting at the earliest offset (or past hour) consume messages from `platform.rhsm-subscriptions.swatch-billable-usage-aggregator-billable-usage-suppress-store-changelog` and none should appear.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
+++ b/src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
@@ -34,6 +34,9 @@ import com.redhat.cloud.event.parser.ConsoleCloudEventParser;
 import io.micrometer.core.aop.TimedAspect;
 import io.micrometer.core.instrument.MeterRegistry;
 import jakarta.validation.Validator;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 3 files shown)


kflahert:
@san7ket this change would delete all messages regardless of orgId from the topic. Does this endpoint get called in stage? if so we likely need to edit this to only happen in EE

PR #3201, opened 2024-04-01
SWATCH-2381: Add heartbeat service to produce billable usages

kflahert:
Jira issue: SWATCH-2381

## Description

Billable usages being aggregated by kstreams cause issues for our IQE tests because they often leave messages behind between tests. In order to ensure no kafka messages are suppressed between tests this new scheduled producer will create billable-usage messages every half second.

### Setup

1. Start app with `DEV_MODE=true BILLABLE_USAGE_HEARTBEAT_ENABLED=true ./gradlew :bootRun`

### Verification

1. View messages coming in on `platform.rhsm-subscriptions.billable-usage`

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageHeartbeatProducer.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageHeartbeatProducer.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


Sgitario (changes requested):
I don't think this should be part of the production code for multiple reasons:
- the billing logic would lead into different behaviours when running in an ephemeral/stage environment (for testing) and the production environment. 
- we would need to maintain this logic when it strongly depends on how tests work (if the test changes, this can be quickly deprecated).

I do see this is for testing convenience only, but for the same reason, this logic should be part of the testing component/test/framework. So, the test that is asserting the billing logic should send this message before triggering the scenario.

PR #3154, opened 2024-03-14
swatch-1382 add metric_id to tally_summary kafka messages

kflahert:
Jira issue: SWATCH-1382

## Description

Add metric_id to tally summary kafka messages and deprecate UOM.

### Setup

1. Start a kafka consumer for platform.rhsm-subscriptions.tally

### Steps

1. Run a tally with any valid data. Can follow steps if you don't have the data already: https://github.com/RedHatInsights/rhsm-subscriptions/pull/3120

### Verification

1. Check kafka consumer for messages that should have metric_id value populated matching UOM.

PR #3207, opened 2024-04-03
Fix tally_snapshot.yaml to remove duplicated metric_id property

Sgitario:
## Description
Partially reverts https://github.com/RedHatInsights/rhsm-subscriptions/pull/3154 that added a duplicated "metric_id" property.

## Testing
Only regression testing.


Sgitario:
Self-merging this as it's only a clean up.

PR #3208, opened 2024-04-03
SWATCH-2285: Add swatch-billable-usage deployment

Sgitario:
Jira issue: [SWATCH-2285](https://issues.redhat.com/browse/SWATCH-2285)

## Description
- copy the swatch-producer-azure service into the new swatch-billable-usage service
- this is necessary to prepare the deployment of the new service in the ephemeral/stage/prod environments

## Testing
Only regression testing.


Sgitario:
Self-merging this PR to see if we can unblock the update of the template.

PR #3220, opened 2024-04-05
SWATCH-2336: Increased max header size to accomodate actual use

wottop:
Jira issue: SWATCH-2336

## Description
Some customers have been getting the following error on the call /api/rhsm-subscriptions/v1/tally/products/RHEL for x86/Sockets:
java.lang.IllegalArgumentException: Request header is too large

As this has been "random", it is assumed that the size in error is just above the set limit of 48K. This number has been doubled. This effectiveness of the solution will be reflected in the future frequency of the exception.

### DIFF
--- a/swatch-api/deploy/clowdapp.yaml
+++ b/swatch-api/deploy/clowdapp.yaml
@@ -20,7 +20,7 @@ parameters:
   - name: IMAGE_TAG
     value: latest
   - name: SERVER_MAX_HTTP_HEADER_SIZE
-    value: '48000'
+    value: '96000'
   - name: LOGGING_LEVEL_ROOT
     value: WARN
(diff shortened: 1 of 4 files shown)


Sgitario (commented):
Do we know the header that is causing the problem? My concern is that by increasing the header size, we're hiding the really root cause of the issue and increasing the header size also brings disadvantages (more to process by our server).

wottop:
Based on a convo with @barnabycourt we would rather accept a slightly larger message than throw an error.
If the increase is insufficient, we can come back and give it a closer look.

Sgitario (approved):

ssidhaye (approved):
I don't see stack trace in the pod log and no errors in browser console.

PR #3226, opened 2024-04-09
SWATCH-2412: Use saveAndFlush when consuming tally snapshots

Sgitario:
Jira issue: [SWATCH-2412](https://issues.redhat.com/browse/SWATCH-2412)

## Description
- Use saveAndFlush to validate the db state when consuming tally snapshots
- Added happy path integration test

Also, the kafka consumer already deals with the error handling, so instead of using a custom RetryTemplate, I've configured the existing mecanism to use the retry related configuration.

## Testing
Only regression testing.

To verify these changes is a bit tricky since you need to do manual changes:
1.- podman-compose up
2.- apply the changes from the patch: [swatch-2412.PATCH](https://github.com/RedHatInsights/rhsm-subscriptions/files/14916849/swatch-2412.PATCH)
3.- start the service: `DEV_MODE=true CONTRACT_USE_STUB=true ./gradlew :bootRun`
4.- create the following message into the topic `platform.rhsm-subscriptions.tally`:

```
{"org_id": "org123", "tally_snapshots": [{"billing_provider":"aws", "billing_account_id": "123", "snapshot_date": "2023-12-21T01:10:28Z", "product_id": "rosa", "sla": "Premium", "usage": "Production", "granularity": "Hourly", "tally_measurements": [{"hardware_measurement_type": "AWS", "metric_id": "Cores", "value": 12}]}]}
```

5.- the app should react to the above message and throw the following UUID constraint exception.
6.- since the message failed to be processed, no messages are expected in the topic `platform.rhsm-subscriptions.billable-usage`: http://localhost:3030/#/cluster/default/topic/n/platform.rhsm-subscriptions.billable-usage/

You can try again without the change from the entity and check that the topic `platform.rhsm-subscriptions.billable-usage` contains a message now.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageController.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/billing/BillableUsageController.java
@@ -232,7 +232,9 @@ public class BillableUsageController {
     newRemittance.setRemittedPendingValue(usageCalc.getRemittedValue());
     newRemittance.setRemittancePendingDate(usageCalc.getRemittanceDate());
     log.debug("Creating new remittance for update: {}", newRemittance);
-    billableUsageRemittanceRepository.save(newRemittance);
+    // using saveAndFlush to validate the entity against the database and raise constraints
+    // exception before moving forward.
+    billableUsageRemittanceRepository.saveAndFlush(newRemittance);
   }
 
   private Double getCurrentlyMeasuredTotal(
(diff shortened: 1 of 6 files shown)


kflahert (approved):

Sgitario:
@kflahert I've updated the whole solution to use `saveAndFlush` instead versus trying to configure the chained transaction. This is because for some reason, when running the monolith with the profile `worker`, the chained transactions of kafka and the database stops working (I could reproduce this issue in the integration test that is part of this pull request, but I could not spot anything from the logs to troubleshoot why this was happening). Therefore, I decided to use `saveAndFlush` which does the same thing (and it also rollbacks the data if the message could not be sent).

kflahert (approved):

PR #3246, opened 2024-04-12
SWATCH-2444: Fix bug with multiple azure API credentials 

kflahert:
Jira issue: SWATCH-2444

## Description

When we have multiple sets of api credentials to use against azure, the code was calling the api with both credentials every time. This code change makes it so we only try the next set of credentials if the first one fails instead of continuing to try more credentials.

To test this you need to use credentials from stage, updated to have another invalid entry.

### DIFF
--- a/swatch-producer-azure/src/main/java/com/redhat/swatch/azure/service/AzureMarketplaceClientFactory.java
+++ b/swatch-producer-azure/src/main/java/com/redhat/swatch/azure/service/AzureMarketplaceClientFactory.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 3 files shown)


swatch-producer-azure/src/test/java/com/redhat/swatch/azure/service/AzureMarketplaceServiceTest.java
```diff
@@ -0,0 +1,98 @@
...
+import com.redhat.swatch.clients.azure.marketplace.api.resources.AzureMarketplaceApi;
+import io.quarkus.test.InjectMock;
+import io.quarkus.test.junit.QuarkusTest;
+import jakarta.ws.rs.ProcessingException;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+@QuarkusTest
```
Sgitario, line 44: I don't think this test needs to use `@QuarkusTest`.

swatch-producer-azure/src/main/java/com/redhat/swatch/azure/service/AzureMarketplaceClientFactory.java
```diff
@@ -0,0 +1,75 @@
...
+    this.azureMarketplaceCredentials = azureMarketplaceCredentials;
+    this.oidcClients = oidcClients;
+  }
+
+  public List<AzureMarketplaceApi> createClientForEachTenant() {
+    return azureMarketplaceCredentials.getClients().stream()
+        .map(
+            client -> {
+              try {
+                return RestClientBuilder.newBuilder()
```
Sgitario, line 61: Use QuarkusRestClientBuilder instead (this is the recommended builder for Quarkus Reactive REST Client).

Sgitario (changes requested):
Very well spotted! 
Only a few minor things in my review.

kflahert:
> Very well spotted! Only a few minor things in my review.

Thanks

Sgitario (approved):

PR #3264, opened 2024-04-17
SWATCH-2453: Add the contracts-client module for Quarkus

Sgitario:
Jira issue: [SWATCH-2453](https://issues.redhat.com/browse/SWATCH-2453)

## Description
Create the contracts-client for Quarkus.
Copy the StubContractsApi and ContractsApiFactory classes into the new module Since we don't have the concept of FactoryBean in Quarkus and also the rest clients are injected via client proxies in Quarkus. I figured out to inject the DefaultApi instances via the qualifier ContractsClient annotation. This way, the ContractsApiFactory plays to produce either the stub contracts api or the generated default api.

Note that to fully test these changes, I had to do some minimal configuration in the swatch-billable-usage because it will eventually use it.

## Testing
Fully covered using the new tests.


kartikshahc (approved):
LGTM

PR #3278, opened 2024-04-22
SWATCH-2364: Append azureCustomerId to the end of billingProviderId

kflahert:
Jira issue: SWATCH-2364

## Description

For testing purposes we need to save the azureCustomerId which will be used to clean up test accounts.

## Testing
Covered in Integration test.


Sgitario (approved):
lgtm. 
I also checked that when reading the billing provider ID in https://github.com/RedHatInsights/rhsm-subscriptions/blob/644152909d84884dbbf7d9acdcfaf468d2d5584f/src/main/java/org/candlepin/subscriptions/capacity/admin/InternalSubscriptionResource.java#L251, it will keep working fine.

TrayvonMcKnight (approved):
Looks good. Shouldn't have any impact on current contract tests.

PR #3287, opened 2024-04-23
SWATCH-1385 Drop UOM from billable usage messages

kartikshahc:
Jira issue: SWATCH-1385

## Description

UOM field dropped from Billable usage message.


swatch-producer-aws/src/main/java/com/redhat/swatch/aws/processors/BillableUsageAggregateProcessor.java
```diff
@@ -255,7 +252,7 @@ public BatchMeterUsageResponse send(
 
   private UsageRecord transformToAwsUsage(
       AwsUsageContext context, BillableUsageAggregate billableUsageAggregate, Metric metric)
-      throws AwsDimensionNotConfiguredException, UsageTimestampOutOfBoundsException {
+      throws UsageTimestampOutOfBoundsException {
```
mstead, line 255: I'm curious as to why the `AwsDimensionNotConfiguredException` is no longer required since it is not referencing the `uom`?

src/test/java/org/candlepin/subscriptions/rhmarketplace/RhMarketplacePayloadMapperTest.java
```diff
@@ -81,7 +81,6 @@ void testUsesMetricIdRatherThanUom() {
 
     rhMarketplacePayloadMapper.produceUsageMeasurement(mockUsage);
     verify(mockUsage).getMetricId();
-    verify(mockUsage, never()).getUom();
```
mstead, line 84: Is this test still relevant now that `uom` is no longer in the billable_usage schema?

mstead (commented):
Looks good, couple of questions before I approve.

swatch-producer-aws/src/main/java/com/redhat/swatch/aws/processors/BillableUsageAggregateProcessor.java
kartikshahc: This code should have been removed when we moved the way we used to read tagprofile. 
```
.dimension(
            tagProfile.getAwsDimension(tallySnapshot.getProductId(), measurement.getUom().name())
```

kartikshahc:
> Looks good, couple of questions before I approve. Pr-checks are failing, but I believe QE are working on some flaky tests. All unit tests run fine locally for me.

Replied to both of your questions. Let me know if you still have questions.

mstead (changes requested):
Just need to remove the unneeded test and should be good to go.

src/test/java/org/candlepin/subscriptions/rhmarketplace/RhMarketplacePayloadMapperTest.java
mstead: Let's remove it then.

kartikshahc:
> Just need to remove the unneeded test and should be good to go.

Done

mstead (approved):

ntkathole:
> @ntkathole is only-regression sufficient in this case?

Currently moto billing tests not working in EE, so I think it’s good to have it verified  by QE manually this time. But we do have coverage, so no need of new automation.

himmAllRight:
@ntkathole , do you know if the manual verification has been done already, or is that something which still needs to be completed?

ntkathole:
> @ntkathole , do you know if the manual verification has been done already, or is that something which still needs to be completed?

@himmAllRight Since there is no qe contact on Jira as well as no qe assignee on PR, I see it is not yet picked up by anyone for verification.

liwalker-rh:
Confirmed through manual tests that billable usage message only has metric id and no longer has UOM, so looks good to me.

liwalker-rh (approved):

PR #3288, opened 2024-04-23
Drop productName from rhel els identification

kahowell:
Unneeded and misidentifies some things it shouldn't.


kahowell:
Self-merge since it's a trivial change

PR #3294, opened 2024-04-25
SWATCH-2333: Added attribute to subcription definition for isVdcType

wottop:
- This replaces the use of the id field as the indicator in the factory of which  UsageCollector is needed for processing

Jira issue: SWATCH-2333

## Description
The usage collection method's determination is based on the variant tag in the subscription definition. This method is not sufficient as this tag varies between "RHEL for ..." and "rhel-for-...". Instead of relying on that field which is not constrained, the field 'vdcType' is introduced in the definition. This way the determination is based on a boolean value with a specific purpose.

## Testing
IQE Test MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/672

### Verification
1. The test is structured to use the known variations of the variant tag. The test will show that the hypervisor usage is properly collected in both cases.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/collector/ProductUsageCollectorFactory.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/collector/ProductUsageCollectorFactory.java
@@ -20,6 +20,7 @@
  */
 package org.candlepin.subscriptions.tally.collector;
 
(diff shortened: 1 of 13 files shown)


Sgitario (approved):
lgtm

PR #3299, opened 2024-04-26
SWATCH-2483: Fix export subscriptions service to filter by payg

Sgitario:
Jira issue: [SWATCH-2483](https://issues.redhat.com/browse/SWATCH-2483)

## Description
The unlimited subscriptions query is used to populate the total capacity of the skus, and since we are not gathering this information in the export service, we don't need it.
To retrieve the payg subscriptions, the difference is how to filter by product (it's being done one way if the product is non payg, and in a different way if the product is payg). I have a fix for this to retrieve both.

## Testing

Steps to reproduce and verify using QE tooling:

1.- app.rhsm_subscriptions.create_mock_subscription(product_id="rosa",billing_provider="aws",quantity=10)
2.- app.rhsm_subscriptions.get_skus_report(product_id="rosa")
3.- filters = {"product_id": "rosa"}
4.- export = app.rhsm_subscriptions.create_swatch_export_request(filters=filters)
5.- data, _ = app.rhsm_subscriptions.download_export_to_json(export.id)

The subscriptions from step 2 and step 5 should match.

IQE Test MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/670

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/SubscriptionTableController.java
@@ -192,12 +192,7 @@ public class SubscriptionTableController {
       }
     }
 
-    boolean isOnDemand =
-        SubscriptionDefinition.lookupSubscriptionByTag(productId.toString())
-            .filter(SubscriptionDefinition::isPrometheusEnabled)
-            .stream()
-            .findFirst()
-            .isPresent();
+    boolean isOnDemand = SubscriptionDefinition.isPrometheusEnabled(productId.toString());
 
     SubscriptionType subscriptionType =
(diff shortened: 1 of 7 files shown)


san7ket:
Updated the automation, when the job passes we can merge this

san7ket:
In the recent failure, we are missing capacity in exported data, I will add some logging to be more sure. this tests is passing on stage.

Sgitario:
> In the recent failure, we are missing capacity in exported data, I will add some logging to be sure. this tests is passing on stage.

This is because the API logic differenciates between capacity and physical hypervisor to set the capacity. Since we're not providing these two type of capacities in the output, I'm reverting this change.

san7ket (approved):

kflahert (approved):

PR #3308, opened 2024-04-30
Configure logging SQL queries by deployment property

Sgitario:
## Description
Adding a common property `LOGGING_SHOW_SQL_QUERIES` in spring boot and quarkus services which is by default false.

In the Spring Boot services, this property will configure:
```
spring.jpa.show-sql: true
spring.jpa.properties.hibernate.format_sql: true
```

In the Quarkus services:
```
quarkus.hibernate-orm.log.sql=true
```

When enabled, this property will be helpful when troubleshooting issues in EE environments.

## Testing
No needed.

### DIFF
--- a/swatch-api/deploy/clowdapp.yaml
+++ b/swatch-api/deploy/clowdapp.yaml
@@ -25,6 +25,8 @@ parameters:
     value: WARN
   - name: LOGGING_LEVEL
     value: INFO
+  - name: LOGGING_SHOW_SQL_QUERIES
+    value: 'false'
   - name: REPLICAS
     value: '1'
   - name: MEMORY_REQUEST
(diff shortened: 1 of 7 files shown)


san7ket:
Would this work for swatch-metrics service aswell ??

Sgitario:
> Would this work for swatch-metrics service aswell ??

swatch-metrics does not run SQL queries. 
This is only effective in Spring Boot/Quarkus services that uses Hibernate.

san7ket:
ahh yea, i guess those are promql . ACK

lindseyburnett (approved):

lindseyburnett:
LGTM once PR checks are passing

PR #3313, opened 2024-05-01
SWATCH-2475: Update docker-compose

awood:
Jira issue: Part of SWATCH-2475

## Description
This patch adds several items:

* The unleash server now starts and runs properly.
* Health checks have been added to the db and kafka containers
* The command for the inventory containers has been altered to better reflect the commands used by the inventory development Dockerfile
* The postgresql container now listens for connections external to the compose network and internal to the compose network
* We no longer override the set_passwords.sh script in the postgresql container.  Instead we add an additional script into a directory where it will be run automatically by the container entrypoint.

### Steps
1.  If you have a compose already running, `podman-compose down`.  This will **destroy** your database, so if you have data you want to save, do a `pg_dump`
1. `podman-compose up -d`

### Verification
1. `podman-compose ps` should show that all the containers are up and running.  You can also go to `http://localhost:4242` to see the Unleash web UI.

### DIFF
--- a/bin/wait-for
+++ b/bin/wait-for
@@ -1,0 +1,192 @@
+#!/bin/sh
+
+# Courtesy of https://github.com/Unleash/unleash-docker
+# The MIT License (MIT)
+#
+# Copyright (c) 2017 Eficode Oy
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


Sgitario (approved):
lgtm

PR #3315, opened 2024-05-02
SWATCH-2482: Skip metered subscriptionUMB messages that do not have a billing provider

kflahert:
Jira issue: SWATCH-2482

## Description

When an azure subscription is created we get both a contract UMB message and a subscription UMB message. The contract message has the correct start_date and billing provider information while the subscription message has a truncated start_date with missing billing_provider. If the subscription UMB comes in before the contract UMB we end up with an extra subscription record with an incorrect start_date. 

In order to prevent this we should ignore subscriptionUMB messages where we do not have any billing_provider on either the message or an existing subscription (for metered offerings only). This way we will skip the initial subscription UMB and allow contract UMB to create the correct record.

## Testing
This requires testing in stage. We need to create new a new azure contract and test updating and inactivating the contract. We should do the same for an aws subscription.

The liquibase can be tested as follows:
1. Insert test data:
```
INSERT INTO public.offering VALUES ('RH00003', 'RHEL Server', 'Red Hat Enterprise Linux', NULL, 2, NULL, NULL, 'Red Hat Enterprise Linux Server', 'Premium', 'Production', 'Red Hat Enterprise Linux Server, Premium (Physical or Virtual Nodes)', false, NULL, false);
INSERT INTO public.offering VALUES ('RH02781HR', 'RHEL Server', 'RHEL', NULL, NULL, NULL, NULL, NULL, 'Premium', 'Production', 'Red Hat Enterprise Linux for Third Party Linux Migration with Extended Life Cycle Support (On-Demand, Hourly Production Support)', false, NULL, false);


INSERT INTO public.subscription VALUES ('RH02781HR', 'org123', '14746723', 1, '2024-04-11 10:23:22.799162+00', '2024-05-11 10:23:22.799162+00', 'testing', '14746880', 'azure', 'testing');
INSERT INTO public.subscription VALUES ('RH02781HR', 'org123', '14746723', 1, '2024-04-11 04:00:00+00', '2024-05-11 03:59:59+00', NULL, '14746880', NULL, NULL);
INSERT INTO public.subscription VALUES ('RH00003', 'org123', '14389751', 1, '2024-02-19 11:46:34.883958+00', '2024-03-19 11:46:34.883958+00', 'testing', '14389908', 'azure', 'testing');
INSERT INTO public.subscription VALUES ('RH00003', 'org123', '14389751', 1, '2024-02-19 05:00:00+00', '2024-05-16 03:59:59+00', NULL, '14389908', NULL, NULL);

INSERT INTO public.subscription VALUES ('RH00003', 'org123', '12512', 1, '2023-12-14 05:00:00+00', '2024-04-16 04:59:59.999+00', NULL, '578', NULL, NULL);
INSERT INTO public.subscription VALUES ('RH00003', 'org123', '123521', 1, '2023-12-14 05:00:00+00', '2024-04-16 04:59:59.999+00', NULL, '578', 'aws', 'testawsBillingAccount');
```
2. Start the app with `DEV_MODE=true ./gradlew :bootRun`
3. Verify that all subscription with billing_provider remain as well as both records where subscription_number='578' since subscriptions records should be left unaffected.


Sgitario (approved):

TrayvonMcKnight (approved):
https://drive.google.com/file/d/1afXyBKGP4lZPFfgsciSIaVUuutBJ1CMJ/view?usp=drive_link

looks good to me. we will follow up on the testing in stage. the thing to test will be canceling subscriptions using the cloud provider portals and checking to make sure that it gets reflected on our end

PR #3330, opened 2024-05-07
SWATCH-2145: Prevent tallies from creating duplicate system/host entries

Sgitario:
Jira issue: [SWATCH-2145](https://issues.redhat.com/browse/SWATCH-2145)

## Description
- Add a migration script to remove the duplicates based on (1) the presence of the billing_account_id, (2) the more recent last_seen.
- Add an unique constraint for org_id and inventory_id.
- Add an unique constraint for org_id and instance_id.
- When upserting a system during HBI tally, look for possible duplicates via org_id/instance_id. If there is an existing record, it should be treated as a matching system.

Note that this PR adds a performance killer when reconciling the hosts with HBI. 
Also note that I added the list of hosts to be deleted in the linked JIRA ticket. 

I ran the migration scripts using the stage data in local and it took less than 60 seconds. 

### Verify the migration scripts
1.- podman-compose up
2.- load the stage data into your local database
3.- check that you have duplicated hosts by:

```sql
select * from hosts h inner join (select count(*) as duplicated, org_id, instance_id from hosts group by org_id,instance_id) d on d.org_id=h.org_id and d.instance_id=h.instance_id where d.duplicated > 1 order by d.org_id, d.instance_id;
```

4.- run the migration script by `./gradlew liquibaseUpdate`
5.- run again the above query and now you should get no results.

### Verify that the reconciliation of hosts work with the new unique constraints

1.- podman-compose up
2.- insert two hosts with the same provider ID (which is our instance_id in swatch) into the insights db:

```sql
insert into hosts(id, org_id, created_on, modified_on, account, display_name, facts, system_profile_facts,
                  canonical_facts, stale_timestamp, reporter, "groups")
values ('c0000000-0000-0000-0000-000000000000', 'org123', now(), now(), 'account123', 'system1', '{
  "satellite": {
    "system_purpose_role": "Red Hat Enterprise Linux Server"
  }
}', '{
  "infrastructure_type": "virtual",
  "cores_per_socket": "2",
  "number_of_sockets": "1",
  "arch": "x86_64",
  "installed_products": [
    {
      "id": "69"
    },
    {
      "id": "408"
    }
  ]
}', '{
  "subscription_manager_id": "05000000-0000-0000-0000-000000000000",
  "insights_id": "80000000-0000-0000-0000-000000000000",
  "provider_id": "instance_id_xxx"
}', now() + interval '30 days', 'reporter', '{}');

insert into hosts(id, org_id, created_on, modified_on, account, display_name, facts, system_profile_facts,
                  canonical_facts, stale_timestamp, reporter, "groups")
values ('40000000-0000-0000-0000-000000000000', 'org123', '2023-06-07T14:40:37.079023Z', '2023-06-07T14:40:37.079023Z',
        'account123', 'system2', '{
    "satellite": {
      "virtual_host_uuid": "05d00000-0000-0000-0000-000000000000"
    }
  }', '{
    "infrastructure_type": "virtual",
    "cores_per_socket": "6",
    "number_of_sockets": "1",
    "arch": "x86_64",
    "installed_products": [
      {
        "id": "69"
      },
      {
        "id": "408"
      }
    ]
  }', '{
    "subscription_manager_id": "b0000000-0000-0000-0000-000000000000",
    "insights_id": "70000000-0000-0000-0000-000000000000",
	"provider_id": "instance_id_xxx"
  }', now() + interval '30 days', 'reporter', '{}');
```

3.- start the service `DEV_MODE=true ./gradlew :bootRun`
4.- run the reconciliation: `http PUT :8000/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots/org123 x-rh-swatch-psk:placeholder`

Verify that no unique constraints are thrown and only one host record in our swatch database exists.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
+++ b/src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
@@ -74,7 +74,7 @@ import lombok.Setter;
             @ColumnResult(name = "guest_id"),
             @ColumnResult(name = "subscription_manager_id"),
             @ColumnResult(name = "insights_id"),
-            @ColumnResult(name = "provider_id"),
+            @ColumnResult(name = "instance_id"),
             @ColumnResult(name = "cloud_provider"),
             @ColumnResult(name = "stale_timestamp", type = OffsetDateTime.class),
             @ColumnResult(name = "hardware_subman_id")
@@ -122,7 +122,13 @@ import lombok.Setter;
         h.system_profile_facts->>'is_marketplace' as is_marketplace,
         h.canonical_facts->>'subscription_manager_id' as subscription_manager_id,
         h.canonical_facts->>'insights_id' as insights_id,
-        h.canonical_facts->>'provider_id' as provider_id,
+        coalesce(
+            h.canonical_facts->>'provider_id',
+            -- fallback logic to set the instance ID if and only if the instanceId is not set yet:
+            -- We assume that the instance ID for any given HBI host record is the inventory ID; compare
+            -- to an OpenShift Cluster from Prometheus data, where we use the cluster ID.
+            cast(h.id as text)
+        ) as instance_id,
         rhsm_products.products,
         qpc_prods.qpc_products,
         system_profile.system_profile_product_ids,
@@ -150,7 +156,7 @@ import lombok.Setter;
            and (h.system_profile_facts->>'host_type' IS NULL OR h.system_profile_facts->>'host_type' <> 'edge')
            and NOW() < stale_timestamp + make_interval(days => :culledOffsetDays)
         -- NOTE: ordering is crucial for correct streaming reconciliation of HBI data
-        order by provider_id, hardware_subman_id, any_hypervisor_uuid, inventory_id
+        order by instance_id, hardware_subman_id, any_hypervisor_uuid, inventory_id
     """,
     resultSetMapping = "inventoryHostFactsMapping")
 @Getter
--- a/src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
+++ b/src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
@@ -150,7 +150,7 @@ import lombok.Setter;
            and (h.system_profile_facts->>'host_type' IS NULL OR h.system_profile_facts->>'host_type' <> 'edge')
            and NOW() < stale_timestamp + make_interval(days => :culledOffsetDays)
         -- NOTE: ordering is crucial for correct streaming reconciliation of HBI data
-        order by hardware_subman_id, any_hypervisor_uuid, inventory_id
+        order by provider_id, hardware_subman_id, any_hypervisor_uuid, inventory_id
     """,
     resultSetMapping = "inventoryHostFactsMapping")
 @Getter
--- a/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
@@ -247,28 +247,39 @@ public class InventoryAccountUsageCollector {
       Set<String> applicableProducts,
       List<Host> hosts) {
     log.debug(
-        "Reconciling HBI inventoryId={} & swatch inventoryId={}",
+        "Reconciling HBI inventoryId={} & swatch inventoryId={} & swatch instanceId={}",
         Optional.ofNullable(hbiSystem).map(InventoryHostFacts::getInventoryId),
-        Optional.ofNullable(swatchSystem).map(Host::getInventoryId));
+        Optional.ofNullable(swatchSystem).map(Host::getInventoryId),
+        Optional.ofNullable(swatchSystem).map(Host::getInstanceId));
     boolean isMetered = swatchSystem != null && swatchSystem.isMetered();
     if (hbiSystem == null && swatchSystem == null) {
       log.debug("Unexpected, both HBI & Swatch system records are empty");
-    } else if (hbiSystem == null) {
+    } else if (hbiSystem == null
+        && HBI_INSTANCE_TYPE.equalsIgnoreCase(swatchSystem.getInstanceType())) {
       if (!isMetered) {
-        log.info("Deleting system w/ inventoryId={}", swatchSystem.getInventoryId());
+        log.info(
+            "Deleting system w/ inventoryId={} and instanceId={}",
+            swatchSystem.getInventoryId(),
+            swatchSystem.getInstanceId());
         hostRepository.delete(swatchSystem);
       }
-    } else {
+    } else if (hbiSystem != null) {
       NormalizedFacts normalizedFacts =
           factNormalizer.normalize(hbiSystem, orgHostsData, isMetered);
       Set<Key> usageKeys = createHostUsageKeys(applicableProducts, normalizedFacts);
       if (swatchSystem != null) {
-        log.debug("Updating system w/ inventoryId={}", hbiSystem.getInventoryId());
+        log.debug(
+            "Updating system w/ inventoryId={} and instanceId={}",
+            hbiSystem.getInventoryId(),
+            hbiSystem.getProviderId());
         Host updatedSwatchSystem =
             updateSwatchSystem(hbiSystem, normalizedFacts, swatchSystem, usageKeys);
         hosts.add(updatedSwatchSystem);
       } else {
-        log.debug("Creating system w/ inventoryId={}", hbiSystem.getInventoryId());
+        log.debug(
+            "Creating system w/ inventoryId={} and instanceId={}",
+            hbiSystem.getInventoryId(),
+            hbiSystem.getProviderId());
         swatchSystem = createSwatchSystem(hbiSystem, normalizedFacts, usageKeys);
         hosts.add(swatchSystem);
       }
--- a/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
@@ -180,22 +180,10 @@ public class InventoryAccountUsageCollector {
   public static void populateHostFieldsFromHbi(
       Host host, InventoryHostFacts inventoryHostFacts, NormalizedFacts normalizedFacts) {
 
-    if (inventoryHostFacts.getProviderId() != null) {
-      // will use the provider ID from HBI
-      host.setInstanceId(inventoryHostFacts.getProviderId());
-    }
-
+    host.setInstanceId(inventoryHostFacts.getInstanceId());
     if (inventoryHostFacts.getInventoryId() != null) {
       host.setInventoryId(inventoryHostFacts.getInventoryId().toString());
-
-      // fallback logic to set the instance ID if and only if the instanceId is not set yet:
-      if (host.getInstanceId() == null) {
-        // We assume that the instance ID for any given HBI host record is the inventory ID; compare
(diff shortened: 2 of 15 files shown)


src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
```diff
@@ -157,6 +157,26 @@ will be processed (and the other will be saved for the next iteration).
...
+        // This won't be necessary after SWATCH-512 is done.
+        activeSwatchSystem =
+            hostRepository
+                .findByOrgIdAndInstanceIdOrInventoryId(
+                    orgId,
+                    activeHbiSystem.getProviderId(),
+                    Optional.ofNullable(activeHbiSystem.getInventoryId())
+                        .map(Object::toString)
+                        .orElse(null))
+                .orElse(null);
```
kahowell, line 179: Rather than doing this query for each host, we should rework the query that produces the stream of swatch systems to fetch these systems as well.

src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
```diff
@@ -150,7 +150,7 @@ from jsonb_array_elements(h.system_profile_facts->'installed_products') as items
            and (h.system_profile_facts->>'host_type' IS NULL OR h.system_profile_facts->>'host_type' <> 'edge')
            and NOW() < stale_timestamp + make_interval(days => :culledOffsetDays)
         -- NOTE: ordering is crucial for correct streaming reconciliation of HBI data
-        order by hardware_subman_id, any_hypervisor_uuid, inventory_id
+        order by hardware_subman_id, any_hypervisor_uuid, inventory_id, provider_id
```
kahowell, line 153: I think this can work, but we should put provider_id first for it to have an impact. (Because this query should not return any duplicate inventory_id values).

swatch-core/src/main/java/org/candlepin/subscriptions/db/HostRepository.java
```diff
@@ -71,11 +71,23 @@ public interface HostRepository
       left join fetch h.monthlyTotals
       where h.orgId=:orgId
         and h.instanceType='HBI_HOST'
-      order by coalesce(h.hypervisorUuid, h.subscriptionManagerId), h.hypervisorUuid, h.inventoryId, h.id
+      order by coalesce(h.hypervisorUuid, h.subscriptionManagerId), h.hypervisorUuid, h.inventoryId, h.instanceId, h.id
```
kahowell, line 74: Two thoughts here...
1. We should adjust instanceType to include "RHEL System", since we are looking to deduplicate other RHEL telemetry (this will ensure that systems brought in via metering are included). This will avoid a need to query for every new HBI system.
2. We should sort by instanceId first (see related comment around inventoryHost).

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -247,24 +247,34 @@ protected void reconcileHbiSystemWithSwatchSystem(
...
+        Optional.ofNullable(swatchSystem).map(Host::getInventoryId),
+        Optional.ofNullable(swatchSystem).map(Host::getInstanceId));
     if (hbiSystem == null && swatchSystem == null) {
       log.debug("Unexpected, both HBI & Swatch system records are empty");
     } else if (hbiSystem == null) {
-      log.info("Deleting system w/ inventoryId={}", swatchSystem.getInventoryId());
+      log.info(
+          "Deleting system w/ inventoryId={} and instanceId={}",
+          swatchSystem.getInventoryId(),
+          swatchSystem.getInstanceId());
```
kahowell, line 260: We'll need to add an additional check here: if we have seen a system in metering data (instance type "RHEL System"), but there's no HBI record, it should be kept/skipped.

kahowell (changes requested):
FYI, I still need to review the migrations, but wanted to get some feedback to you sooner.

src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
Sgitario: We can't do this using the stream of swatch systems because the duplicated host might come from the new hosts from the stream HBI query. 

Let me try to explain this better:

- hosts from HBI

hostA
hostB
hostB (host with the same instance ID or inventory ID)

- hosts from SWATCH

hostA

When processing the reconciliation:
- the hostA will be properly updated
- the first hostB will be properly added
- the second hostB will throw an unique constraint because it was added right in the previous iteration

As a result of the unique constraint, neither the hostA nor the hostB will be added/updated.

src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java
Sgitario: I don't follow why we should put the provider_id first to have an impact. Do you mean before `hardware_subman_id`? or `inventory_id`?

swatch-core/src/main/java/org/candlepin/subscriptions/db/HostRepository.java
Sgitario: About 1., this sounds to me like another unrelated task. 
About 2., let's discuss this in the previous comment.

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
Sgitario: This sounds to me like an unrelated task. Let's clarify this point internally.

Sgitario:
@kahowell I've addressed all your comments. 
Btw, finally I haven't used the OrgsData because this context is re-created just before of calling the consumer, so I could not compare and I encounter much simpler to simply keep a string for the previous provider Id, but let me know your thoughts about it.

src/main/resources/liquibase/202405061230-remove-duplicated-hosts.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
```
kahowell, line 2: It would be worth noting this link somewhere, so that others can quickly see the dupe deletion queries are applying a common pattern: https://wiki.postgresql.org/wiki/Deleting_duplicates

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
```diff
@@ -248,29 +258,31 @@ public static SortKey fromHbiSystem(InventoryHostFacts system) {
...
       if (hypervisorUuidResult != 0) {
         return hypervisorUuidResult;
       }
-      return Objects.compare(
-          inventoryId, other.getInventoryId(), Comparator.nullsLast(Comparator.naturalOrder()));
+
+      var instanceIdResult = compare(instanceId, other.getInstanceId());
+      if (instanceIdResult == 0) {
+        return 0;
+      }
```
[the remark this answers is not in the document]
Sgitario: I don't think this is super necessary, but I've changed it.

src/main/resources/liquibase/202405061230-remove-duplicated-hosts.xml
```diff
@@ -0,0 +1,54 @@
...
+      <dropUniqueConstraint tableName="hosts" constraintName="hosts_org_id_inventory_id_unique"/>
+    </rollback>
+  </changeSet>
+
+  <changeSet id="202405061230-2" author="jcarvaja" dbms="postgresql">
+    <comment>Remove the duplicated hosts by instance_id based on the billing_account_id and the more recent last_seen and add the unique constraint</comment>
+    <delete tableName="hosts">
+      <where>
+        id in (select d.id from (
+          select h.id, row_number() over (partition by h.org_id, h.instance_id order by h.billing_account_id desc, last_seen desc) as duplicated_row_number
```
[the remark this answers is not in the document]
Sgitario: I tested the order and it works fine using the `desc` sort order (note we'll keep only the row_number = 1). 
Using desc order:
Captura de pantalla de 2024-05-10 19-34-46

Using asc order:
Captura de pantalla de 2024-05-10 19-35-41

Do we want to keep the records that has the billing_account_id not null, or did I misunderstand this wrong?

src/main/resources/liquibase/202405061230-remove-duplicated-hosts.xml
Sgitario: I didn't know about this link, good to be added +1

src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
Sgitario: @kahowell PR updated by using "desc nulls last" instead.

kahowell: resolved offline

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
kahowell: given `isMetered` checks, no longer relevant

src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
```diff
@@ -248,29 +258,30 @@ public static SortKey fromHbiSystem(InventoryHostFacts system) {
...
     public int compareTo(@NotNull SortKey other) {
-      var hardwareSubmanIdResult =
-          Objects.compare(
-              hardwareSubmanId,
-              other.getHardwareSubmanId(),
-              Comparator.nullsLast(Comparator.naturalOrder()));
+      var instanceIdResult = compare(instanceId, other.getInstanceId());
+      if (instanceIdResult == 0) {
+        return 0;
+      }
```
kahowell, line 270:
```suggestion
      if (instanceIdResult != 0) {
        return instanceIdResult;
      }
```

src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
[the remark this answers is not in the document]
kahowell: Actually, this is critical because it affects whether the HBI system stream or swatch system stream is advanced when they are different systems. The collation breaks if the comparison logic is not the same as the database sorting.

kahowell (changes requested):

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
Sgitario: do you mean we don't need to check the instance type because we're already checking the isMetered flag?

kahowell (approved):

ntkathole:
@Sgitario` ERROR: duplicate key value violates unique constraint "hosts_org_id_inventory_id_unique"` in swatch-tally pod causing failures in tests

kahowell (approved):

Sgitario:
> @Sgitario` ERROR: duplicate key value violates unique constraint "hosts_org_id_inventory_id_unique"` in swatch-tally pod causing failures in tests

@ntkathole the latest commit fixes this issue. Now, all the tests are passing.

kahowell (approved):

ntkathole:
I will take a look at this after PR 3237.

ntkathole (approved):
ACK, This looks good with initial mocked data, however, I will take another look in stage with real data from different sources and add the automation accordingly.

ntkathole:
@Sgitario yes, I think we can merge. I am following up on flaky tests.

PR #3333, opened 2024-05-08
SWATCH-431: Update event data json to copy uom to metric_id

wottop:
Jira issue: SWATCH-431

## Description
In the transition from UOM to exclusively using metric_id, we need to update the event data. Where there is not currently a value for metric_id in the "measurements" block, but there is one for UOM, the data will be copied to the metric_id. 
The data will be left in the UOM field for now. The code has been modified to look for metricId first and then fall back to UOM.  

### Setup
1. Start the podman-compose for rhsm-subscriptions while using the main branch.
2. Transfer data from production using ```./import-from-gabi.py --org-id <org_id> --events``` from the swatch-support-scripts directory.
3. Examine the data value json to see that about half of them have uom but not metric_id:
    ```select count(*) from events;```
    ```select count(*) from events where data->'measurements'->0->'uom' = data->'measurements'->0->'metric_id';```

### Steps
1. Check out this branch
2. Run ```./gradle liquibaseUpdate```

### Verification

1. Examine the data values again to see that all data has uom and metric_id:
```select count(*) from events;```
```select count(*) from events where data->'measurements'->0->'uom' = data->'measurements'->0->'metric_id';```

## Note
I imported the events/latest/full_data.csv file from AWS S3 which contains 961,500 records to my local db.
```ChangeSet liquibase/202405080950-copy-uom-to-metric-id-events-table.xml::202405080950-01::wpoteat ran successfully in 15388ms```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
@@ -36,6 +36,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.apache.commons.lang3.StringUtils;
 import org.candlepin.clock.ApplicationClock;
 import org.candlepin.subscriptions.db.AccountServiceInventoryRepository;
 import org.candlepin.subscriptions.db.HostRepository;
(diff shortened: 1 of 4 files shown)


liwalker-rh:
Regression testing looks good, going to see if any other iqe test changes are needed before approving.

kflahert (approved):

liwalker-rh (approved):

PR #3336, opened 2024-05-09
SWATCH-2492: Remove use of Azure Tenent Id in the billing account id

wottop:
Jira issue: SWATCH-2492

## Description
The billing account id was a concatenation of the Azure Tenent id and the Azure Subscription id. This was used because the Tenent id was supposed to be immediate while the Subscription id can take as much as 72 hours. The existence of the Tenent id field has been deemed unreliable.

This update only uses the Subscription id for the billing account id field. If no value is available, then the operation is delayed until the value is in place.

## Testing

1. Run swatch-core:
`RHSM_SUBSCRIPTIONS_ENABLE_SYNCHRONOUS_OPERATIONS=true DEV_MODE=true PROM_URL=http://localhost:8082/api/metrics/v1/telemeter/api/v1 SWATCH_CONTRACTS_INTERNAL_SERVICE=http://localhost:8003 SERVER_PORT=8000 SPRING_PROFILES_ACTIVE=worker,kafka-queue,api,capacity-ingress,rh-marketplace,kafka-queue,rhsm-conduit ./gradlew clean :bootRun`
2. Run swatch-billable-usage:
`SERVER_PORT=8001 QUARKUS_MANAGEMENT_PORT=9001 KSTREAM_BILLABLE_USAGE_AGGREGATION_WINDOW_DURATION=10s KSTREAM_BILLABLE_USAGE_AGGREGATION_GRACE_DURATION=2s ./gradlew :swatch-billable-usage:quarkusDev`

### Steps
1. Produce Kafka message to service-instance-ingress
`{"sla": "Premium", "role": "Red Hat Enterprise Linux Server", "usage": "Production", "org_id": "13259775", "event_id": "527e29b6-321c-4021-a1f4-fcff4326af3b", "timestamp": "2024-05-15T19:00:00Z", "event_type": "snapshot", "expiration": "2024-05-15T20:00:00Z", "instance_id": "i-0d8d5297e354051db", "product_ids": ["69", "204"], "event_source": "cost-management", "measurements": [{"uom": "vCPUs", "value": 1.0}], "service_type": "RHEL System", "hardware_type": "Cloud", "cloud_provider": "Azure", "billing_provider": "azure", "azure_subscription_id": "746157280291"}`
3. Execute hourly tally
`http POST ":8000/api/rhsm-subscriptions/v1/internal/tally/hourly?org=13259775&start=2023-10-19T15:00Z&end=2023-10-19T16:00Z" x-rh-swatch-psk:placeholder`

### Verification
1. There will be a row in the events table for the above event.
2. The data field will have "billing_account_id" populated:
`  "cloud_provider": "Azure",
  "billing_provider": "azure",
  "billing_account_id": "746157280291",
  "azure_subscription_id": "746157280291"`
3. The billable_usage_remittance table will show a matching entry

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/admin/UsageContextSubscriptionProvider.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/admin/UsageContextSubscriptionProvider.java
@@ -25,11 +25,8 @@ import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.core.Response.Status;
 import java.time.OffsetDateTime;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.StringUtils;
 import org.candlepin.subscriptions.db.model.BillingProvider;
 import org.candlepin.subscriptions.db.model.ServiceLevel;
 import org.candlepin.subscriptions.db.model.Subscription;
@@ -67,19 +64,13 @@ public class UsageContextSubscriptionProvider {
       String billingAccountId,
       OffsetDateTime subscriptionDate) {
 
-    // If billingAccountId is multipart, find all matching first part (ex: azureTenantId)
-    var searchByBillingAccountId = billingAccountId;
-    if (StringUtils.isNotBlank(billingAccountId) && billingAccountId.contains(";")) {
-      searchByBillingAccountId = billingAccountId.split(";")[0];
-    }
-
     UsageCalculation.Key usageKey =
         new Key(
             productId,
             ServiceLevel.fromString(sla),
             Usage.fromString(usage),
             billingProvider,
-            searchByBillingAccountId);
+            billingAccountId);
 
     // Set start date one hour in past to pickup recently terminated subscriptions
     var start = subscriptionDate.minusHours(1);
@@ -87,11 +78,8 @@ public class UsageContextSubscriptionProvider {
         subscriptionSyncController.findSubscriptions(
             Optional.ofNullable(orgId), usageKey, start, subscriptionDate);
 
-    var bestMatchSubscriptions =
-        filterByClosestMatchingMultipartBillingAccountId(subscriptions, billingAccountId);
-
     var existsRecentlyTerminatedSubscription =
-        bestMatchSubscriptions.stream()
+        subscriptions.stream()
             .anyMatch(
                 subscription ->
                     subscription.getEndDate() != null
@@ -99,7 +87,7 @@ public class UsageContextSubscriptionProvider {
 
     // Filter out any terminated subscriptions
     var activeSubscriptions =
-        bestMatchSubscriptions.stream()
+        subscriptions.stream()
             .filter(
                 subscription ->
                     subscription.getEndDate() == null
@@ -107,7 +95,7 @@ public class UsageContextSubscriptionProvider {
                         || subscription.getEndDate().equals(subscriptionDate))
             .toList();
 
-    if (bestMatchSubscriptions.isEmpty()) {
+    if (subscriptions.isEmpty()) {
       missingSubscriptionCounter.increment();
       throw new NotFoundException();
     }
@@ -121,14 +109,6 @@ public class UsageContextSubscriptionProvider {
     }
 
     if (activeSubscriptions.size() > 1) {
-      // If this is a multipart ID, throw an error since correct result could not be determined.
-      if (billingAccountId.contains(";")) {
-        throw new SubscriptionsException(
-            ErrorCode.SUBSCRIPTION_CANNOT_BE_DETERMINED,
-            Status.NOT_FOUND,
-            "Subscription could not be determined. Likely multiple subscriptions exist missing partial billingAccountId information.",
-            "");
-      }
       ambiguousSubscriptionCounter.increment();
       log.warn(
           "Multiple subscriptions found for billing provider {} or for org {} with key {} and product tag {}."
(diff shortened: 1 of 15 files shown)


src/main/java/org/candlepin/subscriptions/capacity/admin/UsageContextSubscriptionProvider.java
```diff
@@ -142,32 +127,16 @@ public Optional<Subscription> getSubscription(
   }
 
   // Filter out subscriptions if we have an exact match on a multipart billingAccountId.
-  private List<Subscription> filterByClosestMatchingMultipartBillingAccountId(
+  private List<Subscription> filterByClosestMatchingBillingAccountId(
```
kflahert, line 130: We should actually be ok to just delete this method entirely. The bestMatch logic wont be necessary when its only subscription_id/not mulitpart.

kflahert:
@wottop Looks good! could you also add a liquibase script to update the existing azure subscriptions and contracts with the correct billing_account_id?

kflahert (changes requested):
Left a couple of inline comments to delete a method we won't need and add liquibase.

src/main/java/org/candlepin/subscriptions/capacity/admin/UsageContextSubscriptionProvider.java
wottop: I think the need to keep the whole list when there are no matches means we need to keep it to maintain the other behavior. I'll rename it to add clarity and add a comment.

kflahert: They should all be exact matches now. Azure was the only provider requiring us to use a multipart billing account ID like that so we shouldn't run into a situation where there are any to filter out here.

wottop: Will update

src/main/resources/liquibase/202405151547-update-events-azure-billable-account-id.xml
```diff
@@ -0,0 +1,18 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.20.xsd">
+  <changeSet id="202405151547-01" author="wpoteat">
+    <comment>
+      Copy data from azure_subscription_id field to billable_account_id field
+    </comment>
+    <sql dbms="postgresql">
+      update events
+      set data = jsonb_set(data::jsonb, '{"billable_account_id"}', data -> 'azure_subscription_id')
```
kflahert, line 13: @wottop I think we need to do the same for subscriptions,contracts,tally_snapshots, billable_usage_remittances and hosts/host_tally_buckets. I would also make sure to use billing_provider='azure' in the where clause  so we limit the query because those can be big tables.

diegomaranhao:
I'll test again after the code fix has been done.

src/main/java/org/candlepin/subscriptions/capacity/admin/UsageContextSubscriptionProvider.java
```diff
@@ -121,14 +118,6 @@ public Optional<Subscription> getSubscription(
     }
 
     if (activeSubscriptions.size() > 1) {
-      // If this is a multipart ID, throw an error since correct result could not be determined.
-      if (billingAccountId.contains(";")) {
```
kflahert, line 125: No change required in this PR but we use this SUBSCRIPTION_CANNOT_BE_DETERMINED error in swatch-azure-producer in order to determine if we can retry processing.  I created a follow up task for this because I think it's out of scope for this card.
https://issues.redhat.com/browse/SWATCH-2595

src/main/java/org/candlepin/subscriptions/capacity/admin/UsageContextSubscriptionProvider.java
```diff
@@ -67,47 +66,45 @@ public Optional<Subscription> getSubscription(
...
     var start = subscriptionDate.minusHours(1);
     List<Subscription> subscriptions =
         subscriptionSyncController.findSubscriptions(
             Optional.ofNullable(orgId), usageKey, start, subscriptionDate);
 
-    var bestMatchSubscriptions =
-        filterByClosestMatchingMultipartBillingAccountId(subscriptions, billingAccountId);
+    var matchedSubscription =
+        subscriptions.stream()
+            .filter(
```
kflahert, line 85: We don't need to do this filtering because the billing_account_id is part of the query so they should all be equal.

kflahert (changes requested):

kflahert:
@diegomaranhao can we also make sure to test a new contract created in stage for this? I think we're lacking the capability to fully test the UMB contract syncing logic in EE currently.

src/main/resources/liquibase/202405151547-update-events-azure-billable-account-id.xml
```diff
@@ -0,0 +1,99 @@
...
+  <changeSet id="202405151547-07" author="wpoteat">
+    <comment>
+      Update billing_account_id field for azure billing account change
+    </comment>
+    <sql dbms="postgresql">
+      update hosts
+      set billing_account_id = regexp_replace(billing_account_id,'.*;(.*)', '\1')
+      where billing_provider='azure'
+        and org_id in
+            (select org_id from subscription where billing_provider = 'azure' union
```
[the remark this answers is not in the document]
kflahert: AWS doesn't use the ";" in billing_account_id(its only the CustomerAwsAccountId). We do use that pattern in billing_provider_id for both azure/aws however.

kflahert (approved):

diegomaranhao:
@kflahert Yes, we will run an e2e testing after this hits the stage environment.

PR #3342, opened 2024-05-10
SWATCH-2540 Port the purge-remittances cron job to the billable-usage

Sgitario:
Jira issue: [SWATCH-2540](https://issues.redhat.com/browse/SWATCH-2540)

## Description
- Moved the remittances purge cronjob to the billable usage service
- Moved API to delete remittances from the tally service to the billable usage service

| Old | New  |
|---|---|
|http://swatch-tally-service:8000/api/rhsm-subscriptions/v1/internal/rpc/remittance/purge | http://swatch-billable-usage-service:8000/api/swatch-billable-usage/internal/rpc/remittance/purge |

## Testing
For all the verifications belong, you need to deploy the billable usage and the tally services into an EE environment:
```
bonfire deploy ...
```

### Verify remittances purge cronjob
1.- update the "swatch-billable-usage-purge-remittances" cronjob with:
```
spec:
  schedule: '* * * * *'
```

This will run the job every minute. 

2.- wait for the job to be triggered (up to 1 min). See jobs in `https://<OCP>/k8s/ns/ephemeral-xciphz/cronjobs/swatch-billable-usage-purge-remittances/jobs`

3.- the pod should see that the HTTP call finished successfully.


src/main/resources/application-worker.yaml
```diff
@@ -108,7 +105,7 @@ rhsm-subscriptions:
     seek-override-end: ${KAFKA_SEEK_OVERRIDE_END:false}
     seek-override-timestamp: ${KAFKA_SEEK_OVERRIDE_TIMESTAMP:}
   enabled-orgs:
-    target-topics-allowed: platform.rhsm-subscriptions.subscription-prune-task, platform.rhsm-subscriptions.subscription-sync-task
+    target-topics-allowed: platform.rhsm-subscriptions.subscription-prune-task, platform.rhsm-subscriptions.subscription-sync-task, platform.rhsm-subscriptions.remittances-purge-task
```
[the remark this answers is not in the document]
Sgitario: Yes, it was created for the parent JIRA task: https://github.com/RedHatInsights/platform-mq/pull/347

Sgitario:
@mstead this should be already fixed (see latest commit). I've also added some tests to prevent this to happen again.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/services/RemittancesPurgeTaskConsumer.java
```diff
@@ -0,0 +1,71 @@
...
+@AllArgsConstructor
+public class RemittancesPurgeTaskConsumer {
+  private final ApplicationClock applicationClock;
+  private final ApplicationConfiguration configuration;
+  private final BillableUsageRemittanceRepository remittanceRepository;
+
+  @Blocking
+  @Incoming(REMITTANCES_PURGE_TASK)
+  public void consume(EnabledOrgsResponse message) {
+    log.info("Received task for remittances purge with org ID: {}", message.getOrgId());
```
mstead, line 47: Change this to a debug statement to indicate that the message was received.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/services/RemittancesPurgeTaskConsumer.java
```diff
@@ -0,0 +1,71 @@
...
+    OffsetDateTime cutoffDate = getCutoffDate();
+    if (cutoffDate == null) {
+      return;
```
mstead, line 50: Add a warning log here to indicate the the cutoff date was not configured and that nothing will be purged.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/services/RemittancesPurgeTaskConsumer.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+    remittanceRepository.deleteAllByOrgIdAndRemittancePendingDateBefore(
```
mstead, line 52: Add an info log here to indicating that a purge will be done for the target org and cutoff date.

mstead (changes requested):
> @mstead this should be already fixed (see latest commit). I've also added some tests to prevent this to happen again.

The latest fix addresses the issue, however the tests that were added does not actually test the issue. Removing `@Blocking` from the consumer does not cause the test to fail. While I think it was a useful test to add, I'd feel more comfortable if we had a unit and/or IQE test that covers the issue that was encountered.

We are lacking logging around the purge logic in the consumer. Specifically around the no-op when the cutoff date was not configured (which puzzled me while testing when it didn't delete any records) and when the operation was actually performed (see comments below). I don't think this is directly related to your changes, but we may as well address this now.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/services/RemittancesPurgeTaskConsumer.java
Sgitario: done

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/services/RemittancesPurgeTaskConsumer.java
Sgitario: done

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/services/RemittancesPurgeTaskConsumer.java
Sgitario: done

Sgitario:
> > @mstead this should be already fixed (see latest commit). I've also added some tests to prevent this to happen again.
> 
> The latest fix addresses the issue, however the tests that were added does not actually test the issue. Removing `@Blocking` from the consumer does not cause the test to fail. While I think it was a useful test to add, I'd feel more comfortable if we had a unit and/or IQE test that covers the issue that was encountered.
> 
> We are lacking logging around the purge logic in the consumer. Specifically around the no-op when the cutoff date was not configured (which puzzled me while testing when it didn't delete any records) and when the operation was actually performed (see comments below). I don't think this is directly related to your changes, but we may as well address this now.

Good idea about adding more logs +1
Moreover, I'm also checking if the policy is configured, and if not, it won't emit any task (so we don't overload the systems with unnecessary tasks).
About the test, I could verify it was working via the logs, but you're right and better to update the test to fail if this is not working. To test it, I replaced the `@Blocking` annotation with the `@Transactional`, and removed the `@Transactional` from the repository.

mstead (approved):

mstead:
Re-tested the job in EE and all is well now. Once the PR checks all complete, this PR is good.

Sgitario:
> Re-tested the job in EE and all is well now. **Once the PR checks all complete, this PR is good.**

Many thanks!

PR #3352, opened 2024-05-16
SWATCH-2511: Sorting as per cores and sockets does not return all OCP instances

wottop:
Jira issue: SWATCH-XXXX

## Description
- Removed search criteria that excludes instances that do not include sort metric
- Updated tally_instance_view to store 0 for sockets and cores colums instead of null

## Testing
QE Test MR: Pending

### Setup
TBD

### Steps
TBD

### Verification
TBD


liwalker-rh:
There was one related failure running regression so this is the MR to update the test once everything is merged: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/714. Test passes with this change.

src/test/java/org/candlepin/subscriptions/tally/export/InstancesDataExporterServiceTest.java
```diff
@@ -164,27 +164,6 @@ void testFiltersFoundData(String filterName, String exists) {
...
-      value = {
-        "usage,disaster recovery",
-        "category,virtual",
-        "sla,standard",
-        "metric_id,Instance-hours",
-        "billing_provider,azure",
-        "billing_account_id,345",
-        "display_name_contains,another host"
-      })
-  void testFiltersDoesNotFoundDataAndReportIsEmpty(String filterName, String doesNotExist) {
```
Sgitario, line 178: why do we need to remove this test?

Sgitario (changes requested):
I don't see why specifying a metric ID to sort, it results into also filter by this metric ID. So, I would say that what is wrong here is: https://github.com/RedHatInsights/rhsm-subscriptions/blob/68e7cc58f142f3785e0490b746392fd88bbfeaa5/src/main/java/org/candlepin/subscriptions/resource/InstancesResource.java#L213

Moreover, the proposed solution implies that we're fetching **ALL** the data from database (not only the entries that have the metric ID) and then we'll filter by the metric ID one by one (discarding the entries that do not have the wanted metric ID), so there is a very important difference in terms of performance.

Sgitario (commented):
The solution looks correct to me. Though, do we still need to update the view for some reason? I recall raising this question with @kflahert and he told that null is a valid value and has a different meaning than zero.

kflahert:
> The solution looks correct to me. Though, do we still need to update the view for some reason? I recall raising this question with @kflahert and he told that null is a valid value and has a different meaning than zero

@Sgitario @wottop We shouldn't update the view to default to 0 for cores/sockets because only systems that report those metrics should have values there. Some systems only report on cores or sockets and not both so we should only return those values if the system is counting them. Here's the original card for this change:
https://issues.redhat.com/browse/SWATCH-1917

Sgitario (approved):

PR #3354, opened 2024-05-17
SWATCH-2518: Create subscription capacity view for the export service

Sgitario:
Jira issue: SWATCH-2518

## Description
The new view is used by the export service to process and aggregate all the metrics in a single entry. In the future, the same logic should be used by the Subscription API.

## Testing
IQE Test MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/706

### Setup
1.- deploy the app into an EE
2.- create a contract via API. Instructions using the iqe tooling:

```
import random
product_id = "rosa"

// create contract:
contract_data = {
    "instance_hours": random.randint(0, 10),
    "cores": random.randint(0, 7),
    "billing_provider": "aws",
    "billing_account_id": "AAA",
}
contract = app.rhsm_subscriptions.create_contract(
    product_id=product_id,
    **contract_data,
)
```

3.- confirm that the above API creates two subscriptions
4.- create measurements for both subscriptions (there is one that does not have any). For example:

```
insert into subscription_measurements(subscription_id, start_date, metric_id, measurement_type)
values ('488052','2024-05-16 10:01:08.000 +0200', 'Cores', '50');
```

5.- confirm the capacity that is returned from the subscription API.

Assuming that the cores values for the subscriptions are 10 and 50 and org ID is 3340851:

```
echo -n '{"identity":{"account_number":"","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"3340851"}}}' | base64 -w 0
-- eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IiIsInR5cGUiOiJVc2VyIiwidXNlciI6eyJpc19vcmdfYWRtaW4iOnRydWV9LCJpbnRlcm5hbCI6eyJvcmdfaWQiOiIzMzQwODUxIn19fQ==
```

```
curl -X 'PUT' \
  'http://localhost:8000/api/rhsm-subscriptions/v1/opt-in' \
  -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IiIsInR5cGUiOiJVc2VyIiwidXNlciI6eyJpc19vcmdfYWRtaW4iOnRydWV9LCJpbnRlcm5hbCI6eyJvcmdfaWQiOiIzMzQwODUxIn19fQ=='
```

```
curl -X GET \
  'http://localhost:8000/api/rhsm-subscriptions/v1/subscriptions/products/rosa' \
  -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IiIsInR5cGUiOiJVc2VyIiwidXNlciI6eyJpc19vcmdfYWRtaW4iOnRydWV9LCJpbnRlcm5hbCI6eyJvcmdfaWQiOiIzMzQwODUxIn19fQ==' | jq
```

Output:

```json
[{'billing_provider': 'aws',
  'capacity': 60,
  'has_infinite_quantity': False,
  'hypervisor_capacity': 0,
  'metric_id': 'Cores',
  'next_event_date': datetime.datetime(2024, 6, 16, 7, 16, 7, tzinfo=tzlocal()),
  'next_event_type': 'Subscription End',
  'product_name': 'Red Hat OpenShift Service on AWS Hosted Control Planes '
                  '(Hourly)',
  'quantity': 1,
  'service_level': 'Premium',
  'sku': 'MW02393',
  'subscriptions': [{'id': '320888', 'number': '3938546'}],
  'total_capacity': 60,
  'uom': 'cores',
  'usage': ''}]
```

6.- finally, verify that the export service returns the same capacity. Using iqe tooling:

```
filters = {
    "product_id": product_id
}
export = app.rhsm_subscriptions.create_swatch_export_request(filters=filters)
data, _ = app.rhsm_subscriptions.download_export_to_json(export.id)
data
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/subscription/export/SubscriptionDataExporterService.java
+++ b/src/main/java/org/candlepin/subscriptions/subscription/export/SubscriptionDataExporterService.java
@@ -20,62 +20,72 @@
  */
 package org.candlepin.subscriptions.subscription.export;
 
+import static org.candlepin.subscriptions.resource.ResourceUtils.ANY;
+
 import com.redhat.swatch.configuration.registry.MetricId;
 import com.redhat.swatch.configuration.registry.ProductId;
-import com.redhat.swatch.configuration.registry.SubscriptionDefinition;
 import jakarta.ws.rs.core.Response;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
-import java.util.function.BiConsumer;
+import java.util.Optional;
+import java.util.function.Function;
 import java.util.stream.Stream;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
-import org.candlepin.clock.ApplicationClock;
 import org.candlepin.subscriptions.db.HypervisorReportCategory;
-import org.candlepin.subscriptions.db.SubscriptionRepository;
+import org.candlepin.subscriptions.db.SubscriptionCapacityViewRepository;
 import org.candlepin.subscriptions.db.model.BillingProvider;
-import org.candlepin.subscriptions.db.model.DbReportCriteria;
 import org.candlepin.subscriptions.db.model.ServiceLevel;
-import org.candlepin.subscriptions.db.model.Subscription;
+import org.candlepin.subscriptions.db.model.SubscriptionCapacityView;
+import org.candlepin.subscriptions.db.model.SubscriptionCapacityViewMetric;
+import org.candlepin.subscriptions.db.model.SubscriptionCapacityView_;
 import org.candlepin.subscriptions.db.model.Usage;
 import org.candlepin.subscriptions.exception.ExportServiceException;
 import org.candlepin.subscriptions.export.DataExporterService;
 import org.candlepin.subscriptions.export.DataMapperService;
 import org.candlepin.subscriptions.export.ExportServiceRequest;
+import org.candlepin.subscriptions.json.SubscriptionsExportJsonMeasurement;
 import org.candlepin.subscriptions.utilization.api.model.ReportCategory;
 import org.springframework.context.annotation.Profile;
+import org.springframework.data.jpa.domain.Specification;
 import org.springframework.stereotype.Component;
 
 @Slf4j
 @Component
 @Profile("capacity-ingress")
 @AllArgsConstructor
-public class SubscriptionDataExporterService implements DataExporterService<Subscription> {
+public class SubscriptionDataExporterService
+    implements DataExporterService<SubscriptionCapacityView> {
   static final String SUBSCRIPTIONS_DATA = "subscriptions";
   static final String PRODUCT_ID = "product_id";
-  private static final Map<String, BiConsumer<DbReportCriteria.DbReportCriteriaBuilder, String>>
+  static final String METRIC_ID = "metric_id";
+  static final String CATEGORY = "category";
+  static final String PHYSICAL = "PHYSICAL";
+  static final String HYPERVISOR = "HYPERVISOR";
+  private static final Map<String, Function<String, Specification<SubscriptionCapacityView>>>
       FILTERS =
           Map.of(
               PRODUCT_ID,
               SubscriptionDataExporterService::handleProductIdFilter,
               "usage",
               SubscriptionDataExporterService::handleUsageFilter,
-              "category",
+              CATEGORY,
               SubscriptionDataExporterService::handleCategoryFilter,
               "sla",
               SubscriptionDataExporterService::handleSlaFilter,
-              "metric_id",
+              METRIC_ID,
               SubscriptionDataExporterService::handleMetricIdFilter,
               "billing_provider",
               SubscriptionDataExporterService::handleBillingProviderFilter,
               "billing_account_id",
               SubscriptionDataExporterService::handleBillingAccountIdFilter);
 
-  private final SubscriptionRepository subscriptionRepository;
+  private final SubscriptionCapacityViewRepository repository;
   private final SubscriptionJsonDataMapperService jsonDataMapperService;
   private final SubscriptionCsvDataMapperService csvDataMapperService;
-  private final ApplicationClock clock;
 
   @Override
   public boolean handles(ExportServiceRequest request) {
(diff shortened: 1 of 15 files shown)


Sgitario:
@san7ket I think the failures are wrongly failing. I'm checking the data and there is actually 34 subscriptions with sku 'RH00006F5', sla 'Premium' and usage 'Production', not 17. 
The problem is that half of subscriptions are for the product 'RHEL for x86' and the rest for 'rhel-for-x86-eus'.
Indeed, if when you change the export request to include product_id='RHEL for x86', then you will get the expected 17 capacity. 
To be honest, I'm not sure how this could be working before (maybe it wasn't working fine). 
Can you double check? If so, I can modify these tests in the linked MR.

src/main/java/org/candlepin/subscriptions/subscription/export/SubscriptionDataExporterService.java
```diff
@@ -122,63 +131,165 @@ private DbReportCriteria extractExportFilter(ExportServiceRequest request) {
...
-      DbReportCriteria.DbReportCriteriaBuilder builder, String value) {
-    builder.hypervisorReportCategory(
-        HypervisorReportCategory.mapCategory(ReportCategory.fromString(value)));
+  private static Specification<SubscriptionCapacityView> handleCategoryFilter(String value) {
+    String measurementType = getMeasurementTypeFromCategory(value);
+    if (measurementType != null) {
+      return (root, query, builder) ->
+          builder.like(
+              builder.upper(builder.function("jsonb_pretty", String.class, root.get("metrics"))),
+              "%" + measurementType.toUpperCase(Locale.ROOT) + "%");
```
awood, line 163: Could we do these `like` expressions using Postgresql's native JSON operations?  Pretty-printing the JSON and then doing a `LIKE` on the results seems potentially error-prone and likely inefficient.

```sql
select * from subscription_capacity_view where metrics @? '$[*] ? (@.measurement_type == "PHYSICAL")';
```

uses the Postgresql [JSON path](https://www.postgresqltutorial.com/postgresql-json-functions/postgresql-json-path/) syntax.

I realize that hammering that into the criteria query API isn't easy, but maybe you can call out to the `jsonb_path_exists` function?

```sql
select * from subscription_capacity_view where jsonb_path_exists(metrics, '$[*] ? (@.measurement_type == "PHYSICAL")')
```

I also recognize that this might complicate testing on HSQLDB but I don't think HSQLDB supports `jsonb_pretty` either so we shouldn't be in a worse position using the Postgresql JSON path functions.

awood (changes requested):

src/main/java/org/candlepin/subscriptions/subscription/export/SubscriptionDataExporterService.java
Sgitario: I'm not familiar with `jsonb_path_exists` but it seems to match much better. Will try to use it and update the PR accordingly afterwards. Thanks!

Sgitario: @awood PR updated including this change.

src/main/java/org/candlepin/subscriptions/subscription/export/SubscriptionDataExporterService.java
```diff
@@ -122,63 +131,165 @@ private DbReportCriteria extractExportFilter(ExportServiceRequest request) {
...
-      DbReportCriteria.DbReportCriteriaBuilder builder, String value) {
-    builder.billingAccountId(value);
+  private static SubscriptionsExportJsonMeasurement getOrCreateMeasurement(
+      List<SubscriptionsExportJsonMeasurement> metrics, SubscriptionCapacityViewMetric metric) {
+    return metrics.stream()
+        .filter(
+            m ->
+                Objects.equals(m.getMetricId(), metric.getMetricId())
+                    && Objects.equals(m.getMeasurementType(), metric.getMeasurementType()))
+        .findFirst()
```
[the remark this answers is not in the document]
awood: Correct me if I'm wrong, but we've already pulled in Guava for other stuff in this library, so it's not really adding another dependency.

But the bug I'm referencing above was relying on the guarantee where there will be only one or none...until that guarantee no longer held.

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/converters/SubscriptionCapacityViewMetricConverter.java
```diff
@@ -0,0 +1,96 @@
...
+                                .orElse(null))
+                        .build());
+              });
+      return metrics;
+    } catch (JsonProcessingException e) {
+      throw new IllegalArgumentException("Error parsing metrics", e);
+    }
+  }
+
+  private double toDouble(Object o) {
```
awood, line 85: Last thing: can you cover `float` here too?

awood (changes requested):

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/converters/SubscriptionCapacityViewMetricConverter.java
Sgitario: added

src/main/java/org/candlepin/subscriptions/subscription/export/SubscriptionDataExporterService.java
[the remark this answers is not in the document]
Sgitario: the source of the collection in the linked issue is the data, so it was not guaranteed that it was unique. 
Anyway, for sake of clarity, I've replaced the list of metrics by map, so I hope it's now clearer.

Sgitario:
@awood PR updated with the additional commit.

src/main/java/org/candlepin/subscriptions/subscription/export/SubscriptionDataExporterService.java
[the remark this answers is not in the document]
awood: I really like this map approach.  Nice!

san7ket:
Failure unrelated and is being tracked separately.
ACK , only one subscriptions is now shown

PR #3361, opened 2024-05-21
SWATCH-2492: Remove use of Azure Tenent Id in the billing account id

lindseyburnett:


src/main/java/org/candlepin/subscriptions/event/EventController.java
```diff
@@ -256,7 +256,13 @@ private Optional<Event> processEvent(Event eventToProcess) {
     }
 
     if (BillingProvider.AZURE.equals(eventToProcess.getBillingProvider())) {
-      eventToProcess.setBillingAccountId(eventToProcess.getAzureSubscriptionId());
+
+      Optional<String> azureSubscriptionId = eventToProcess.getAzureSubscriptionId();
+      if (azureSubscriptionId != null && azureSubscriptionId.isPresent()) {
```
lindseyburnett, line 261: Had to add a null check here even though getAzureSubscriptionId is supposed to be an Optional.  For some reason the deserialization when we're making batches isn't honoring the jdk8 data type module, even though it's registered on the objectMapper bean.  I started digging a little bit to see if it would be a quick fix, but doesn't seem like it will be so I'll just log a tech debt card for it.  Just pointing it out because my IDE is flagging this as "that's an optional, it should never be null!"....so I wouldn't be surprised if Sonar has the same complaint

PR #3384, opened 2024-05-30
SWATCH-2590: Split tally_instance_view into two versions for payg

Sgitario:
Jira issue: SWATCH-2590

## Description
Since the way to resolve the metrics is different depending on whether the product is payg or not, and also we need this logic to be done in the export service, we need to move this logic onto the query itself. 

Changes:
- Split tally_instance_view into tally_instance_payg_view and tally_instance_non_payg_view (because the way to resolve the metrics is different). 
- Ensure the export service also works

## Testing
Added tests to cover the changes.
For using the IQE tooling, see the JIRA ticket for the instructions.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resource/InstancesResource.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/InstancesResource.java
@@ -187,7 +187,7 @@ public class InstancesResource implements InstancesApi {
         getHardwareMeasurementTypesFromCategory(reportCategory);
 
     List<InstanceData> payload;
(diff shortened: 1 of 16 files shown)


san7ket:
https://ci.ext.devshift.net/job/RedHatInsights-rhsm-subscriptions-pr-check/7120/testReport/tests.swatch-tally.system_table/test_system_table_payg/
ACK

san7ket (approved):

mstead (approved):
LGTM

PR #3385, opened 2024-05-30
SWATCH-2594: Ensure ordering of streams when collating hosts

Sgitario:
Jira issue: SWATCH-2594

## Description
The problem is that when querying the hosts in stage, we're retrieving the following hosts at this order:

First iteration:
- Inventory Host with InventoryId=YYY
- Swatch Host with InventoryId=AAA

Second iteration: 
- Inventory Host with InventoryId=ANY
- Swatch Host with InventoryId=YYY (it should be found in the first iteration!)

I don't see how we could possible resolve the issue by tweaking the order by statement in the HostRepository.streamHbiHostsByOrgId since the InventoryId is ordered alphabetically, so the above is a realistic scenario).

The only solution that I can think of is to sorting the streams using the SortKey.

## Testing
I reproduced the scenario using unit tests.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
@@ -91,10 +91,15 @@ public class InventorySwatchDataCollator {
    */
   public int collateData(String orgId, int culledOffsetDays, Processor processor) {
     Stream<InventoryHostFacts> inventorySystemStream =
-        inventoryRepository.streamFacts(orgId, culledOffsetDays);
+        inventoryRepository
+            .streamFacts(orgId, culledOffsetDays)
+            .sorted(Comparator.comparing(SortKey::fromHbiSystem));
     Stream<String> activeSubmanIdStream =
         inventoryRepository.streamActiveSubscriptionManagerIds(orgId, culledOffsetDays);
-    Stream<Host> swatchSystemStream = hostRepository.streamHbiHostsByOrgId(orgId);
+    Stream<Host> swatchSystemStream =
+        hostRepository
+            .streamHbiHostsByOrgId(orgId)
+            .sorted(Comparator.comparing(SortKey::fromSwatchSystem));
 
     /*
     Setup peeking iterators for each of HBI systems, HBI subman IDs, and swatch systems.
--- a/src/test/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollatorTest.java
+++ b/src/test/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollatorTest.java
@@ -96,21 +96,26 @@ class InventorySwatchDataCollatorTest {
   }
 
   @Test
-  void testCollatorProcessesSameInventoryIdTogetherInOneIteration() {
+  void testCollatorProcessesSameInventoryIdTogetherInTwoIterations() {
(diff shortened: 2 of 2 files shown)


src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
```diff
@@ -91,10 +91,15 @@ public InventorySwatchDataCollator(
    */
   public int collateData(String orgId, int culledOffsetDays, Processor processor) {
     Stream<InventoryHostFacts> inventorySystemStream =
-        inventoryRepository.streamFacts(orgId, culledOffsetDays);
+        inventoryRepository
+            .streamFacts(orgId, culledOffsetDays)
+            .sorted(Comparator.comparing(SortKey::fromHbiSystem));
```
kahowell, line 96: This should be a no-op given the order by clause:
https://github.com/RedHatInsights/rhsm-subscriptions/blob/9050d6340388cc41ee15c3f771cd96ab15519555/src/main/java/org/candlepin/subscriptions/inventory/db/model/InventoryHost.java#L159

should be equivalent to the `SortKey`'s `compareTo` implementation:

https://github.com/RedHatInsights/rhsm-subscriptions/blob/9050d6340388cc41ee15c3f771cd96ab15519555/src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java#L268-L281

src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
```diff
@@ -91,10 +91,15 @@ public InventorySwatchDataCollator(
...
     Stream<String> activeSubmanIdStream =
         inventoryRepository.streamActiveSubscriptionManagerIds(orgId, culledOffsetDays);
-    Stream<Host> swatchSystemStream = hostRepository.streamHbiHostsByOrgId(orgId);
+    Stream<Host> swatchSystemStream =
+        hostRepository
+            .streamHbiHostsByOrgId(orgId)
+            .sorted(Comparator.comparing(SortKey::fromSwatchSystem));
```
kahowell, line 102: This should be a no-op given the order-by clause:
https://github.com/RedHatInsights/rhsm-subscriptions/blob/3fb6479130cd1118346f1065f396e6b235df3039/swatch-core/src/main/java/org/candlepin/subscriptions/db/HostRepository.java#L73

should be equivalent to the `SortKey`'s `compareTo` implementation:

https://github.com/RedHatInsights/rhsm-subscriptions/blob/9050d6340388cc41ee15c3f771cd96ab15519555/src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java#L268-L281

kahowell (changes requested):
Both of those sorts should have no effect other than increased memory usage. Are my assumptions incorrect?

kahowell:
Put another way, rather than invoking a sort using the `SortKey` as a comparator, the effect will be the same if we fix whatever is causing the `SortKey` `compareTo` to not be equivalent to the `order by` clauses (probably only need changes in the `compareTo` implementation), and then it won't require increased memory usage. I haven't had a chance to dig deeper to see exactly where they differ, but I think it should be feasible using a debugger on the unit test case you added.

Sgitario:
Closing to look for another solution.

PR #3391, opened 2024-06-03
Revert "SWATCH-2145: Prevent tallies from creating duplicate system/host entries (#3330)"

Sgitario:
Jira issue: SWATCH-2594

## Description
The existing reconciliation logic limits us to add unique constraints because the logic needs to first (1) insert a host and then (2) delete it, and the unique constraint will be issued at (1).
Since there seems not to be an easy way to fix it, we would need to first adopt the event-driven design logic that will be done in [SWATCH-512](https://issues.redhat.com/browse/SWATCH-512) to prevent inserting duplicated hosts in our systems. 

This pull request adds a migration script to drop the unique constraints.

## Testing
No need.

### DIFF
--- a/src/main/resources/liquibase/202406030700-drop-unique-constraints-in-hosts.xml
+++ b/src/main/resources/liquibase/202406030700-drop-unique-constraints-in-hosts.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


kahowell (changes requested):
It's unnecessary to revert the whole commit. Instead, just drop the unique constraint.

Sgitario:
> It's unnecessary to revert the whole commit. Instead, just drop the unique constraint.

I don't feel comfortable adding a partial solution of something...

kahowell (approved):

kahowell:
@Sgitario thanks for being flexible. If we find other issues with SWATCH-2145 changes (but I'm confident we won't because I have confidence in the code changes for SWATCH-2145 :smile:), we can always revisit and revert more.

I just want to make sure we don't lose the value of the underlying code changes.

PR #3394, opened 2024-06-03
SWATCH-2579 Prevent subscription duplication on quantity change

awood:
Jira issue: SWATCH-2579

## Description
A single quantity changes on a subscription was resulting in the subscription being duplicated on every sync.  When the quantity on a subscription changes, we segment the subscription.  I.e. we terminate the subscription, giving it an end date of `now()`, create a new subscription with the new quantity, same subscription number, and same end date as the original subscription.

Prior to this patch, if the quantity in the subscription service ever changed from the quantity we had on a `subscription` table row, we would segment the subscription on every sync.

## Testing
IQE Test MR: 

### Setup
1. Place the [swatch-2579.json](https://github.com/user-attachments/files/15537832/swatch-2579.json) wiremock scenario in `stub/mappings` as `swatch-2579.json`
1. Ensure you have `RH00001` in `offerings`
    ```sql
    INSERT INTO offering (sku, product_name, product_family, cores, sockets, hypervisor_cores, hypervisor_sockets, role, sla, usage, description, has_unlimited_usage, derived_sku, metered, special_pricing_flag) VALUES ('RH00001', 'RHEL Server', 'Red Hat Enterprise Linux', null, null, null, 2, 'Red Hat Enterprise Linux Server', 'Premium', 'Production', 'Red Hat Enterprise Linux for Virtual Datacenters, Premium', false, 'RH00049', false, null);
    ```
1. Create a wiremock container
    ```shell
    podman run -it --rm -p 8101:8080 --name wiremock -v $(git rev-parse --show-toplevel)/stub:/home/wiremock:z wiremock/wiremock:2.32.0 --verbose
    ```
   The scenario will automatically be loaded.

### Steps
1. Make sure you are on `main` and start the application like so
    ```shell
    DEV_MODE=true RHSM_SUBSCRIPTIONS_SUBSCRIPTION_URL=http://localhost:8101/svcrest/subscription/v5 ./gradlew :bootRun
    ```
1. Run
    ```shell
    for x in $(seq 1 6); do http PUT \
    :8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/123456 \
    x-rh-swatch-psk:placeholder; done
    ```
1. Examine the `subscription` table to see how many records there are for subscription 1111.  We are expecting 3: one for quantity 4, one for quantity 10, and one for quantity 20.
    ```shell
    psql -h localhost -U rhsm-subscriptions rhsm-subscriptions -c "select s.subscription_id, s.start_date, s.end_date, s.quantity from subscription s where s.subscription_id = '1111'"
    ```
    Instead there are 6 rows and 4 of those rows have quantity 20.
1. Reset the database
    ```shell
    psql -h localhost -U rhsm-subscriptions rhsm-subscriptions -c \
    "
    DELETE FROM subscription_product_ids WHERE subscription_id = '1111';
    DELETE FROM subscription_measurements WHERE subscription_id = '1111';
    DELETE FROM subscription WHERE subscription_id = '1111'
    "
    ```
    and reset WireMock
    ```shell
    http POST :8101/__admin/scenarios/reset
    ```
1. Switch to `awood/SWATCH-2579-subscription-duplication`
1. Restart the application and run again the same as in step 1.

### Verification
1. Re-run the `for` loop from step 2 in the Steps section.
1. Re-run the `select` query from step 3.  There should only be 3 rows.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/liquibase/LiquibaseCustomTask.java
+++ b/src/main/java/org/candlepin/subscriptions/liquibase/LiquibaseCustomTask.java
@@ -21,11 +21,11 @@
 package org.candlepin.subscriptions.liquibase;
 
 import java.io.Closeable;
-import java.io.IOException;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
@@ -57,8 +57,9 @@ public abstract class LiquibaseCustomTask
     this.preparedStatements = new HashMap<>();
   }
 
+  @SuppressWarnings("java:S112")
   @Override
-  public void close() throws IOException {
+  public void close() {
     try {
       for (PreparedStatement statement : this.preparedStatements.values()) {
         statement.close();
@@ -66,7 +67,7 @@ public abstract class LiquibaseCustomTask
 
       this.preparedStatements.clear();
     } catch (SQLException e) {
-      throw new IOException(e);
+      throw new RuntimeException(e);
     }
   }
 
@@ -156,7 +157,7 @@ public abstract class LiquibaseCustomTask
   }
 
   /**
-   * Executes the given SQL update/insert.
+   * Executes the given SQL update/insert/delete
    *
    * @param sql The SQL to execute. The given SQL may be parameterized.
    * @param argv The arguments to use when executing the given update.
@@ -188,9 +189,11 @@ public abstract class LiquibaseCustomTask
    * Executes this liquibase upgrade task.
    *
    * @throws DatabaseException if an error occurs while performing a database operation
+   * @throws CustomChangeException if an error occurs with changeset expectations
    * @throws SQLException if an error occurs while executing an SQL statement
    */
-  public abstract void executeTask(Database database) throws DatabaseException, SQLException;
+  public abstract void executeTask(Database database)
+      throws DatabaseException, SQLException, CustomChangeException;
 
   @Override
   public void execute(Database database) throws CustomChangeException {
@@ -222,7 +225,13 @@ public abstract class LiquibaseCustomTask
       } catch (DatabaseException e) {
         logger.severe("Could not reset autocommit");
       }
+      close();
     }
+  }
+
+  // Courtesy https://stackoverflow.com/a/2861510
+  public String preparePlaceHolders(int length) {
+    return String.join(",", Collections.nCopies(length, "?"));
   }
 
   @Override
--- a/src/main/java/org/candlepin/subscriptions/liquibase/PopulateSubscriptionMeasurements.java
+++ b/src/main/java/org/candlepin/subscriptions/liquibase/PopulateSubscriptionMeasurements.java
@@ -47,7 +47,6 @@ public class PopulateSubscriptionMeasurements extends LiquibaseCustomTask {
 
   @Override
   public void executeTask(Database database) throws DatabaseException, SQLException {
-
     int total = 0;
 
     ResultSet socketsSet =
--- a/src/main/java/org/candlepin/subscriptions/liquibase/RemoveIncorrectlySegmentedSubscriptions.java
+++ b/src/main/java/org/candlepin/subscriptions/liquibase/RemoveIncorrectlySegmentedSubscriptions.java
@@ -46,78 +46,44 @@ public class RemoveIncorrectlySegmentedSubscriptions extends LiquibaseCustomTask
       ORDER BY s2.subscription_id, s2.start_date ASC;
       """;
 
-  // language=PostgreSQL
-  public static final String DELETE_DEPENDENT_MEASUREMENTS =
-      """
-      DELETE
-      FROM subscription_measurements sm
-      WHERE (sm.subscription_id, sm.start_date) NOT IN
-          (
-              SELECT subscription_id, start_date
-              FROM subscription
-              WHERE subscription_id = ?
-              ORDER BY start_date ASC
-              LIMIT 2
-          )
-        AND sm.subscription_id = ?;
-      """;
-
-  // Possibly these statements could be improved using anti-joins (LEFT JOIN ... WHERE
-  // l.subscription_id IS NULL)
-  // but using the IN clause has pretty good performance
-
-  // language=PostgreSQL
-  public static final String DELETE_DEPENDENT_PRODUCT_IDS =
-      """
-      DELETE
-      FROM subscription_product_ids spi
-      WHERE (spi.subscription_id, spi.start_date) NOT IN
-          (
-              SELECT subscription_id, start_date
-              FROM subscription
-              WHERE subscription_id = ?
-              ORDER BY start_date ASC
-              LIMIT 2
-          )
-        AND spi.subscription_id = ?;
-      """;
-
   // Delete all segments except for the initial subscription and first
-  // segment swatch created
+  // segment after a quantity change
   // language=PostgreSQL
-  public static final String DELETE_ALL_BUT_TWO_SEGMENTS =
+  public static final String DELETE_INCORRECT_SEGMENTS =
       """
-      DELETE
-      FROM subscription
-      WHERE (subscription_id, start_date) NOT IN
-          (
-              SELECT subscription_id, start_date
+      DELETE FROM %s s
+      USING (
+          SELECT subscription_id, start_date
+          FROM (
+              SELECT subscription_id,
+                  start_date,
+                  quantity,
+                  ROW_NUMBER() OVER (
+                      PARTITION BY subscription_id, quantity ORDER BY subscription_id, start_date DESC
+                  ) AS duplicated_row_number
               FROM subscription
               WHERE subscription_id = ?
-              ORDER BY start_date ASC
-              LIMIT 2
-          )
-      AND subscription_id = ?
+          ) windowed
+          WHERE duplicated_row_number != 1
+      ) d
+      WHERE s.subscription_id = d.subscription_id
+      AND s.start_date = d.start_date;
       """;
 
-  // Set the end date of segment 1 to the start date of segment 2
+  // Set the end date of segment n to the start date of segment n+1
   // language=PostgreSQL
   public static final String UPDATE_SEGMENT =
       """
       UPDATE subscription
-      SET end_date=subquery.start_date
+      SET end_date = next_end
       FROM (
-          SELECT s1.start_date FROM subscription s1 WHERE s1.subscription_id = ? ORDER BY s1.start_date DESC LIMIT 1
-      ) subquery
-      WHERE (subscription.subscription_id, subscription.start_date) IN
-          (
-              SELECT s2.subscription_id, s2.start_date
-              FROM subscription s2
-              WHERE s2.subscription_id = ?
-              ORDER BY s2.start_date ASC
-              LIMIT 1
-          )
-        AND subscription_id = ?
+          SELECT subscription_id, quantity, start_date, LAG(start_date) OVER (ORDER BY start_date DESC) next_end
+          FROM subscription
+          WHERE subscription_id = ?
+      ) d
+      WHERE subscription.start_date = d.start_date
+          AND d.next_end IS NOT NULL
+          AND d.subscription_id = ?;
       """;
 
   public static final String LOG_PREFIX = "SWATCH-2579: ";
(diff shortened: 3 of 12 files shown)


Sgitario (changes requested):
Checking the solution, it seems to me overkilling having to query all the subscriptions when a quantity changed. Also, this query is not taking into account the start and end dates of the subscription that we got from the subscription search, so we might update an expected subscription. 

Another missing part is how to address the existing duplicated subscriptions (I would expect a migration script here).

And more thoughts:

As far as I understood, the issue is that having the following subscription for the same orgId:
- original: subscription with ID 1234, quantity 5, with start date 2023-10-31 and end date 2024-10-30

When we sync the subscriptions, we'll be calling the subscription search by subscription ID and start date, and we might receive a quantity change like:

```json
{
    "id": 13649362,
    "subscriptionNumber": "13649518",
    "quantity": 4, -- quantity changed from 5 to 4
    "effectiveStartDate": 1698724800000, --  Tuesday, 31 October 2023 4:00:00
    "effectiveEndDate": 1730260800000, -- Wednesday, 30 October 2024 4:00:00
    ...
  },
```

This would end up by us creating a duplicated subscription, and segment the previous subscription into two parts:

- original: subscription with ID 1234, quantity 5, with start date 2023-10-31 and end date now()
- segment: subscription with ID 1234, quantity 4, with start date now() and end date 2024-10-30

The problem is that when calling again the subscription search by subscription ID and start date, we'll keep matching the first original subscription (because of the start_date) and creating more duplications:

- original: subscription with ID 1234, quantity 5, with start date 2023-10-31 and end date now()
- segment: subscription with ID 1234, quantity 4, with start date yesterday() and end date 2024-10-30
- segment: subscription with ID 1234, quantity 4, with start date now() and end date 2024-10-30

And the "segment" subscriptions are kind of stale subscriptions (spite of these are the actual active ones) that we cannot sync because there is no matching start date with the subscriptions we receive from the subscription search API. 

About the solution, first of all, I don't understand why we need to segment the original subscription if the start date didn't change, so in this case, I would proceed with a simple update. 
Secondly, since we're modifying the start date, we can't use it to match the subscriptions from the subscription search API. So, in the case when the subscription changes the quantity and also the start date, the existing logic works just fine, but we should sort the subscription we have in our database by start date DESC (so, we get the most recent - this is done in the reconcileSubscriptionsWithSubscriptionService method) and change the map to lookup the subscriptions we received to remove the start_date (this is assuming the a single subscription ID can only have one subscription with a start date). 

Another solution would be to keep the original start_date in the segmented subscriptions, so we can still use it, but this would imply a migration script.

awood:
> Another missing part is how to address the existing duplicated subscriptions (I would expect a migration script here).

Yes, that is going to be a separate item.

> And more thoughts:
> 
> As far as I understood, the issue is that having the following subscription for the same orgId:
> 
>     * original: subscription with ID 1234, quantity 5, with start date 2023-10-31 and end date 2024-10-30
> 
> When we sync the subscriptions, we'll be calling the subscription search by subscription ID and start date, and we might receive a quantity change like:
> 
> ```json
> {
>     "id": 13649362,
>     "subscriptionNumber": "13649518",
>     "quantity": 4, -- quantity changed from 5 to 4
>     "effectiveStartDate": 1698724800000, --  Tuesday, 31 October 2023 4:00:00
>     "effectiveEndDate": 1730260800000, -- Wednesday, 30 October 2024 4:00:00
>     ...
>   },
> ```

Yes, but calls to the subscription service do not include the start date.  In this case, we're pulling the entire subscription history for the org, but if you look at the `subscription-api-spec.yaml` file you'll see in the other method calls that start date is never used as an argument.  Only the subscription number is pertinent.

```
https://subscription.api.redhat.com/svcrest/subscription/v5/search/citeria;web_customer_id=123456/options;products=ALL;showExternalReferences=true;firstResultIndex=0;maxResults=1000
```

> This would end up by us creating a duplicated subscription, and segment the previous subscription into two parts:
>
>  original: subscription with ID 1234, quantity 5, with start date 2023-10-31 and end date now()
>  segment: subscription with ID 1234, quantity 4, with start date now() and end date 2024-10-30
>
> The problem is that when calling again the subscription search by subscription ID and start date, we'll keep matching
> the first original subscription (because of the start_date) and creating more duplications:
>
>   original: subscription with ID 1234, quantity 5, with start date 2023-10-31 and end date now()
>   segment: subscription with ID 1234, quantity 4, with start date yesterday() and end date 2024-10-30
>   segment: subscription with ID 1234, quantity 4, with start date now() and end date 2024-10-30
>
> And the "segment" subscriptions are kind of stale subscriptions (spite of these are the actual active ones) that we cannot > sync because there is no matching start date with the subscriptions we receive from the subscription search API.

Start date is only something we track internally.  The subscription service itself should not ever have 2 subscriptions with the same subscription ID but different start dates (correct me if I am wrong @kahowell).

> About the solution, first of all, I don't understand why we need to segment the original subscription if the start date
> didn't change, so in this case, I would proceed with a simple update.

Because the quantity changed on the subscription.  And the way we represent that internally is by segmenting the subscription into 2 parts like you described earlier.  Currently, our code creates the situation you describe in your second example where it keeps creating a segment on every sync.

```
original: subscription with ID 1234, quantity 5, with start date 2023-10-31 and end date now()
segment: subscription with ID 1234, quantity 4, with start date now() and end date 2024-10-30
```

> Secondly, since we're modifying the start date, we can't use it to match the subscriptions from the subscription search
> API. So, in the case when the subscription changes the quantity and also the start date, the existing logic works just fine, > but we should sort the subscription we have in our database by start date DESC (so, we get the most recent - this is done > in the reconcileSubscriptionsWithSubscriptionService method) and change the map to lookup the subscriptions we 
> received to remove the start_date (this is assuming the a single subscription ID can only have one subscription with a
> start date).

The start date is never given to `SearchApi`, so the issue of not being able to match on start date is not something we need to worry about.  The subscription service may update the quantity of a subscription but it is my understanding that the start and end dates should not change (again, correct me if I am wrong @kahowell).

I didn't alter the map in `reconcileSubscriptionsWithSubscriptionService` so we still use the original subscription record in our table to match the results that come back from the subscription service.  E.g.

```
We have a subscription with a quantity change:
Sub 123, starts 1-Jul-2023, ends 30-Jun-2024, quantity 4
Sub 123, starts 1-Aug-2023, ends 30-Jun-2024, quantity 10

The subscription service is only going to return
Sub 123, starts 1-Jul-2023, ends 30-Jun-2024, quantity 10
```

You're suggesting that we sort our subscriptions by DESC, remove the start date from consideration, and match solely on subscription ID?  I like that idea.  I will look into that.

Sgitario:
> > Another missing part is how to address the existing duplicated subscriptions (I would expect a migration script here).
> 
> Yes, that is going to be a separate item.
> 

Why do we need a separate item for this?

> Yes, but calls to the subscription service do not include the start date. In this case, we're pulling the entire subscription history for the org, but if you look at the `subscription-api-spec.yaml` file you'll see in the other method calls that start date is never used as an argument. Only the subscription number is pertinent.
> 
> ```

I tried the search API myself and the effectiveStartDate is provided. Actually, it constructs a map with a compound key using the subscription ID and the start date.

> 
> You're suggesting that we sort our subscriptions by DESC, remove the start date from consideration, and match solely on subscription ID? I like that idea. I will look into that.

Yes, since we are going to receive only one subscription by the search API and we can't use the start date because we're going to overwrite it when we segment the subscriptions in our database, so I think it could work.

awood:
> > > Another missing part is how to address the existing duplicated subscriptions (I would expect a migration script here).
> > 
> > 
> > Yes, that is going to be a separate item.
> 
> Why do we need a separate item for this?

Because that's the way I want to do it.

> Yes, but calls to the subscription service do not include the start date. In this case, we're pulling the entire subscription
> history for the org, but if you look at the subscription-api-spec.yaml file you'll see in the other method calls that start
> date is never used as an argument. Only the subscription number is pertinent.
>
> I tried the search API myself and the effectiveStartDate is provided. Actually, it constructs a map with a compound key
> using the subscription ID and the start date.

Although it may be there, we don't use it.  Not only that, we can't use it right now.  The client we wrote does not provide a way to specify the start date.  I confirmed with @kahowell that start dates on a subscription do not change in the subscription service.

Sgitario:
> > 
> > 
> > Why do we need a separate item for this?
> 
> Because that's the way I want to do it.
> 

The SWATCH-2579 would only be partly fixed in this case until the another part is merged, but up to you. 

> > Yes, but calls to the subscription service do not include the start date. In this case, we're pulling the entire subscription
> > history for the org, but if you look at the subscription-api-spec.yaml file you'll see in the other method calls that start
> > date is never used as an argument. Only the subscription number is pertinent.
> > I tried the search API myself and the effectiveStartDate is provided. Actually, it constructs a map with a compound key
> > using the subscription ID and the start date.
> 
> Although it may be there, we don't use it. Not only that, we can't use it right now. The client we wrote does not provide a way to specify the start date. I confirmed with @kahowell that start dates on a subscription do not change in the subscription service.

Maybe we're speaking about different things? The effectiveStartDate is indeed used: https://github.com/RedHatInsights/rhsm-subscriptions/blob/cf0ba27a900657d12d1753e80fa1afa1fd09296f/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java#L345

awood:
> > 
> 
> Maybe we're speaking about different things? The effectiveStartDate is indeed used:
> 
> https://github.com/RedHatInsights/rhsm-subscriptions/blob/cf0ba27a900657d12d1753e80fa1afa1fd09296f/src/main/java/org/candlepin/subscriptions/subscription/SubscriptionSyncController.java#L345

The start date is only used internally in SubscriptionSyncController.  It's never sent to the `SearchApi` (AKA the external IT subscription service) via an HTTP request.  We don't even have start date as a query parameter in our client to the subscription service.  See [here](https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/clients/subscription-client/subscription-api-spec.yaml#L31-L53).

Our usage of the start date is completely internal to Swatch.  We're using the start date returned from the subscription service along with the subscription ID so that we match on the very first segment of a subscription (if the subscription has actually been segmented).  And operating on the first segment isn't even the right thing to do since that's what's causing the "quantity has changed" condition to execute every time a sync runs.  So, like you suggested, it's easier and more correct to just return the records for a subscription ID sorted by start date DESC and then we can perform the "quantity has changed" check on the correct segment.

awood:
@Sgitario I've updated this PR with the migration steps.  Originally,  I intended to get the fix pushed first and get to the migration second to alleviate the customer issue as quickly as possible.  But this PR didn't get merged as quickly as I had hoped, so I've gone ahead and attached the migration script to it.

I've made the alterations you suggested where we sort our subscriptions by DESC, remove the start date from consideration and match solely on subscription ID.

Sgitario (changes requested):
I found a few problems following the steps from the description:
- the wiremock port was 8101, not 8000 (as stated in the steps 1 and 4).
- the stub is using the sku RH00001 that does not exist (even using the product stubs).
- the curl in the step 2 is missing the origin header

After addressing the above bullets as:

### Setup

0. Modify the swatch-2579.json to use the sku "RH0180192" instead of "RH00001".

### Steps
1. Make sure you are on `main` and start the application like so
```shell
DEV_MODE=true PRODUCT_USE_STUB=true RHSM_SUBSCRIPTIONS_SUBSCRIPTION_URL=http://localhost:8101/svcrest/subscription/v5 ./gradlew :bootRun
```

2. Run
```shell
for x in $(seq 1 6); do http PUT :8000/api/rhsm-subscriptions/v1/internal/subscriptions/sync/org/123456 x-rh-swatch-psk:placeholder Origin:console.redhat.com; done
```

....
and reset WireMock
```shell
http POST :8101/__admin/scenarios/reset
```

Using the updated steps, I could reproduce the issue and saw up to 6 entries in the subscription tables as expected. 

When I switched to this branch and after doing the same steps, the issue is gone and now I see 3 entries where the one with the quantity 20 is the most recent. So, everything looks good.

The problem I found is about the migration script (I started over the steps using main and then switched to your branch to trigger the script). After the script was executed, there were only 2 entries where I expected the same 3 entries (the one with the quantity 20 was actually deleted).

I didn't try to find the issue with the migration script (actually, I would not use a Java source for the migration script because this is a Java source that we will need to maintain forever even though the migration was already executed -> but this is a biased comment!). 

Using the following three SQL statements fixed the issue:

```sql
delete from "subscription_measurements" s
using (select subscription_id, start_date
	from (
		select subscription_id, start_date, quantity, row_number() over (partition by subscription_id, quantity order by subscription_id, start_date desc) as duplicated_row_number
		from "subscription" 
	) a
	where duplicated_row_number != 1) d
where s.subscription_id = d.subscription_id and s.start_date = d.start_date;

delete from "subscription_product_ids" s
using (select subscription_id, start_date
	from (
		select subscription_id, start_date, quantity, row_number() over (partition by subscription_id, quantity order by subscription_id, start_date desc) as duplicated_row_number
		from "subscription" 
	) a
	where duplicated_row_number != 1) d
where s.subscription_id = d.subscription_id and s.start_date = d.start_date;

delete from "subscription" s
using (select subscription_id, start_date
	from (
		select subscription_id, start_date, quantity, row_number() over (partition by subscription_id, quantity order by subscription_id, start_date desc) as duplicated_row_number
		from "subscription" 
	) a
	where duplicated_row_number != 1) d
where s.subscription_id = d.subscription_id and s.start_date = d.start_date;
```

Now, the 3 subscriptions are kept with the one with the quantity 20 as the most recent. 

Feel free to use these SQL scripts or instead fix the java migration script accordingly.

awood:
> Out of curiosity: how do you know that the queries are deleting more records than the ones affected by this bug? The queries are deleting the subscriptions that are duplicated and the quantity didn't change.
> 
> Anyway, as long as the migration script's output is correct, feel free to address it as you feel more comfortable with.

I imported the entire `subscription` table (and descendant tables) from production and ran it against the subscription ID for the duplicate records we have that don't fall into this particular error case.

awood:
@Sgitario Were there any other open items in your opinion?  The discussion is pretty lengthy but I think everything has been covered.

Sgitario:
> @Sgitario Were there any other open items in your opinion? The discussion is pretty lengthy but I think everything has been covered.

The only remaining item was the issue about the migration script that I've just checked that has been already fixed, so approving.

Sgitario (approved):
lgtm!

awood:
> lgtm!

Thanks Jose!

awood:
@ntkathole D'oh.  I hit merge on this before QE signed off!  Sorry!  Let me know if you have thoughts/remarks on this and I'll do a follow-up.

PR #3397, opened 2024-06-04
Prevent unique constraints in the 202405151547-update-events-azure-billable-account-id.xml

Sgitario:
Jira issue: SWATCH-2492

## Description
Running the existing migration scripts fails on stage with:
```
2024-06-04 10:20:40,457 [thread=restartedMain] [ERROR] [liquibase.changelog] - ChangeSet liquibase/202405151547-update-events-azure-billable-account-id.xml::202405151547-06::wpoteat encountered an exception.
liquibase.exception.DatabaseException: ERROR: duplicate key value violates unique constraint "host_tally_bucket_pkey"
```

Tested using the stage data.


Sgitario:
Self-merging to fix stage.

PR #3445, opened 2024-06-18
SWATCH-2569: Add a metric attribute for type (counter or gauge)

Sgitario:
Jira issue: SWATCH-2569

## Description
- Fixed the validation by adding the `@Valid` annotation to the metrics and variants fields 
- Added a test to verify the validation works (I had to add some test dependencies to do this).

## Testing
Since we're not adding any functional change here, we can't manually verify these changes. Though, I've added a JUnit test that validates that the new type is mandatory to be set.

For QE, I think they are directly reading the yaml configuration from main (it should be by branch).


san7ket:
@Sgitario So, a regression test run for this change should be enough ?

Sgitario:
> @Sgitario So, a regression test run for this change should be enough ?

Yep

kartikshahc (approved):

Sgitario:
The failures are unrelated (same number as other PRs). Merging.

PR #3452, opened 2024-06-19
SWATCH-2665: Expose public API by version in different openapi defs

Sgitario:
Jira issue: SWATCH-2665

## Description
- Rename existing rhsm-subscriptions-api-spec.yaml to rhsm-subscriptions-api-v1-spec.yaml.
- Generate the v1 resources into the package "org.candlepin.subscriptions.utilization.api.v1"

To ease the review, I splitted the PR into two commits: the actual fix and the move/name of the packages to append the v1. 

## Testing
Verify that the openapi resources are now exposed:
- podman-compose up
- deploy the monolith `DEV_MODE=true ./gradlew :bootRun`
- call the openapi resource: `curl -X GET \
  'http://localhost:8000/api/rhsm-subscriptions/v1/openapi.yaml' \
  -H 'x-rh-identity: eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IiIsInR5cGUiOiJVc2VyIiwidXNlciI6eyJpc19vcmdfYWRtaW4iOnRydWV9LCJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxMjM0NTYifX19'`

We should now see the expected openapi definition.


mstead (approved):
This seems to address the URL issues reported in SWATCH-2665, and the changes look good to me. I'm assigning this to @san7ket to verify the fix in the QE env and labeling it with "QE" as such.

@Sgitario We are currently unable to get the swagger docs to load when we hit [http://localhost:8000/api/swatch-tally/internal/swagger-ui/index.html](http://localhost:8000/api/swatch-tally/internal/swagger-ui/index.html#/) locally. I realize this is out of scope for this PR, but do you know of another card that might address this? Or should a new one be created?

san7ket:
` curl -X GET 'http://swatch-api-service:8000/api/rhsm-subscriptions/v1/openapi.yaml'`

PR #3458, opened 2024-06-21
SWATCH-2323: Add BillableUsage.tally_id to replace BillableUsage.id

Sgitario:
Jira issue: SWATCH-2323

## Description
- Add BillableUsage.tally_id to replace BillableUsage.id
- Deprecate BillableUsage.id
- Found that the BillableUsage class was also generated in the swatch-producer-aws and swatch-producer-azure even when the used by the API is the expected one from the swatch-model-billable-usage. This seems to be a known behaviour of the openapi generator but it causes confusion, so to avoid it, I made use of https://github.com/OpenAPITools/openapi-generator/blob/master/docs/customization.md#ignore-file-format.

## Testing
Only regression testing since the billable_usage.id and billable_usage.tally_id are both supported. This is going to be fully removed in SWATCH-2674.

Adding the QE label to update the iqe steps to make use of the tally_id instead.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/rhmarketplace/RhMarketplacePayloadMapper.java
+++ b/src/main/java/org/candlepin/subscriptions/rhmarketplace/RhMarketplacePayloadMapper.java
@@ -31,6 +31,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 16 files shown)


TrayvonMcKnight (approved):
Update in smqe_tools side. This function isn't used in any tests currently. 

MERGED: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/769

lindseyburnett (approved):

PR #3487, opened 2024-07-01
SWATCH-2424: Limit the snapshots to purge in batches

Sgitario:
Jira issue: SWATCH-2424

## Description
I checked the production database and it seems that there are an huge number (we're speaking about more of millions for every month up to 2022) of snapshots that is being tried to be deleted in a single delete query. 

So, we can't split up the delete query to be done by org_id, because we would have the same problem. The only trustable approach is to limit the query by the number of snapshots. This can be defined using a property and by default is 1000000. 

I could test these changes with such amount of data locally, but yes using the stage data. 

Note that to use this limit, I had to turn the delete query into a native query.

### Using small dataset

0. podman-compose up
2. insert the small dataset:

```

```

### Using large data

0. download the data and script from https://drive.google.com/file/d/1vlHzCmPIoy8jov9g-qXGhYKC0VHRBX2g/view?usp=sharing.

6. start the postgresql instance:

```
podman-compose up
```

7. initialize the schema:

```
./gradlew liquibaseUpdate
```

8. load the data (script in the above link):

```
./load_data.sh
```

This file contains:
- 53255 orgs
- 2000000 snapshots
- 4000000 measurements

9. start the service:

```
```

10. run the purge:

```
curl --fail -H "Origin: https://swatch-tally-service.redhat.com" -H "x-rh-swatch-psk: placeholder" -X POST "http://localhost:8000/api/rhsm-subscriptions/v1/internal/rpc/tally/purge"
```

11. wait until the tally snapshot purge task is finished.
When the task is done, you should see the following in the app logs:

```
2024-03-19 16:05:39,969 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.security.LogPrincipalFilter] self- Internal API: /api/rhsm-subscriptions/v1/internal/rpc/tally/purge requested by user: self
2024-03-19 16:05:40,081 [thread=http-nio-8000-exec-1] [INFO ] [org.candlepin.subscriptions.tally.admin.InternalTallyResource] self- Initiating tally snapshot purge.
2024-03-19 16:05:40,086 [thread=purge-tally-snapshots-1] [INFO ] [org.candlepin.subscriptions.retention.TallyRetentionController] - Starting tally snapshot purge.
2024-03-19 16:06:05,634 [thread=purge-tally-snapshots-1] [INFO ] [org.candlepin.subscriptions.retention.TallyRetentionController] - Tally snapshot purge completed successfully.
```

Note that it took 30 seconds to finish.

12. Verify that the data in the tally_snapshots and in the tally_measurements tables.

```
select count(*) from tally_snapshots; -- expected: 2308

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/retention/TallyRetentionController.java
+++ b/src/main/java/org/candlepin/subscriptions/retention/TallyRetentionController.java
@@ -22,27 +22,21 @@ package org.candlepin.subscriptions.retention;
 
 import io.micrometer.core.annotation.Timed;
 import java.time.OffsetDateTime;
+import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.candlepin.subscriptions.db.TallySnapshotRepository;
 import org.candlepin.subscriptions.db.model.Granularity;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Component;
 
 /** Cleans up stale tally snapshots for an account. */
 @Slf4j
 @Component
+@AllArgsConstructor
 public class TallyRetentionController {
 
   private final TallySnapshotRepository tallySnapshotRepository;
   private final TallyRetentionPolicy policy;
-
-  @Autowired
-  public TallyRetentionController(
-      TallySnapshotRepository tallySnapshotRepository, TallyRetentionPolicy policy) {
-    this.tallySnapshotRepository = tallySnapshotRepository;
-    this.policy = policy;
-  }
 
   @Timed("rhsm-subscriptions.snapshots.purge")
   @Async("purgeTallySnapshotsJobExecutor")
@@ -54,8 +48,15 @@ public class TallyRetentionController {
         if (cutoffDate == null) {
           continue;
         }
-        tallySnapshotRepository.deleteAllByGranularityAndSnapshotDateBefore(
-            granularity, cutoffDate);
+
+        long count =
+            tallySnapshotRepository.countAllByGranularityAndSnapshotDateBefore(
+                granularity, cutoffDate);
+        while (count > 0) {
+          tallySnapshotRepository.deleteAllByGranularityAndSnapshotDateBefore(
+              granularity.name(), cutoffDate, policy.getSnapshotsToDeleteInBatches());
+          count -= policy.getSnapshotsToDeleteInBatches();
+        }
       }
       log.info("Tally snapshot purge completed successfully.");
     } catch (Exception e) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/main/java/org/candlepin/subscriptions/retention/TallyRetentionPolicy.java [hunk 1]
```diff
@@ -103,4 +103,8 @@ public OffsetDateTime getCutoffDate(Granularity granularity) {
             String.format("Unsupported granularity: %s", granularity));
     }
   }
+
+  public long getSnapshotsToDeleteInBatches() {
```
awood, line 107: Can we apply some validation to this setting?  I know we do validation in stuff like `ConduitFacts`, but I can't remember if we do it (yet) on any of the configuration settings.  Regardless, we definitely want to make sure this value is greater than zero somehow.  I prefer bean validation annotations, but if that's a pain then just a check in the setter and an exception thrown at startup is fine.  A typo like '000000' would result in an infinite loop.

awood (changes requested):

src/main/java/org/candlepin/subscriptions/retention/TallyRetentionController.java
```diff
@@ -54,8 +54,19 @@ public void purgeSnapshotsAsync() {
...
         }
-        tallySnapshotRepository.deleteAllByGranularityAndSnapshotDateBefore(
-            granularity, cutoffDate);
+
+        long count =
+            tallySnapshotRepository.countAllByGranularityAndSnapshotDateBefore(
+                granularity, cutoffDate);
+        if (count == 0) {
+          continue;
+        }
```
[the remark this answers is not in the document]
Sgitario: yeah, this makes totally sense. Removed.

src/main/java/org/candlepin/subscriptions/retention/TallyRetentionPolicy.java
Sgitario: Good point! Added.

Sgitario:
@awood PR updated, would you mind taking another look?

awood (approved):

san7ket:
```
2024-07-11 08:14:29,679 [thread=http-nio-8000-exec-3] [INFO ] [org.candlepin.subscriptions.security.LogPrincipalFilter] self- Internal API: /api/rhsm-subscriptions/v1/internal/rpc/tally/purge requested by user: self
2024-07-11 08:14:29,681 [thread=http-nio-8000-exec-3] [INFO ] [org.candlepin.subscriptions.tally.admin.InternalTallyResource] self- Initiating tally snapshot purge.
2024-07-11 08:14:29,685 [thread=purge-tally-snapshots-1] [INFO ] [org.candlepin.subscriptions.retention.TallyRetentionController] - Starting tally snapshot purge.
2024-07-11 08:15:03,038 [thread=purge-tally-snapshots-1] [INFO ] [org.candlepin.subscriptions.retention.TallyRetentionController] - Tally snapshot purge completed successfully.
```
```
swatch-tally-db=# select count(*) from tally_snapshots;
  count  
---------
 2000000
(1 row)

swatch-tally-db=# select count(*) from tally_snapshots;
 count 
-------
  2308
(1 row)
```



Looks good

PR #3496, opened 2024-07-02
SWATCH-2704 Cleanup Product Tag Determination

lindseyburnett:
Jira issue: SWATCH-2704

### Separate concerns around when product tags match isConverted/isMetered.
SubscriptionDefinition shouldn't own any extra logic required to ensure snapshots with proper tags get created for nightly vs hourly tally.

### Introduce ProductTagLookupParams to encapsulate the growing parameters
We keep finding ourselves adding parameters to the SubscriptionDefinition.getAllProductTags and it's tedious to maintain.

Introduced ProductTagLookupParams to encapsulate these.  When we need to add another parameter, we can do it here without modifying the signature of the method.

Then with the new predicate-building approach in that method, it will be easy enough to expand on with something like

```java
Predicate<Variant> params.newParameter == null ? variant -> true : variant -> whatever == params.getNewParameter()
```

## Testing
IQE tests and regression

rudimentary functional tests found in gitlab swatch-support-scripts on the `lburnett/functional-tests` branch

```
pytest -k test_offering_sync_product_tags
```


kartikshahc (approved):
LGTM.

kartikshahc (changes requested):
Regression is failing so have to check what is causing it to fail

PR #3530, opened 2024-07-16
SWATCH-2738: Drop subscription_product_ids table that is no longer in use

kflahert:
Jira issue: SWATCH-2738

## Description

Some subscriptions are failing to sync because the child subscription_product_ids records are not being deleted. Since we no longer utilize this table this PR drops it.

## Testing

IQE Test MR: 

Regresssion.

Tested floorist change against gabi.


lindseyburnett (approved):

ntkathole (approved):
Looks good! Org contract sync works fine, will re-check the issue in stage as well.

PR #3537, opened 2024-07-17
SWATCH-2744: Fix capacity reconciliation when another task is queued

kahowell:
https://issues.redhat.com/browse/SWATCH-2744

See inline comments for more details.

Testing
-------

See `CapacityReconciliationIntegrationTest`. To reproduce the issue, revert the fix and run the test.

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/CapacityReconciliationService.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/CapacityReconciliationService.java
@@ -29,6 +29,7 @@ import com.redhat.swatch.contract.repository.SubscriptionMeasurementEntity;
 import com.redhat.swatch.contract.repository.SubscriptionRepository;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.smallrye.reactive.messaging.MutinyEmitter;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
@@ -38,7 +39,6 @@ import java.util.Objects;
 import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.microprofile.reactive.messaging.Channel;
-import org.eclipse.microprofile.reactive.messaging.Emitter;
 
 @ApplicationScoped
 @Slf4j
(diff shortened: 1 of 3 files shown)


kflahert (approved):

kflahert:
LGTM, looking into that weird testcontainers error.

kahowell:
5472eab to e0be4b4: address sonarcloud issues

ntkathole (approved):
Looks good!

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/CapacityReconciliationService.java
```diff
@@ -78,8 +78,18 @@ public void reconcileCapacityForOffering(String sku, int offset, int limit) {
     subscriptions.forEach(this::reconcileCapacityForSubscription);
     if (subscriptions.size() >= limit) {
       offset = offset + limit;
-      reconcileCapacityByOfferingEmitter.send(
-          ReconcileCapacityByOfferingTask.builder().sku(sku).offset(offset).limit(limit).build());
+      reconcileCapacityByOfferingEmitter
+          .send(
```
Sgitario, line 82: There is an easier way to do this by changing the `Emitter` to use `MutinyEmitter` which has the method `sendAndAwait` that does the wait already. 
This is just a minor suggestion, tho we also use the `MutinyEmitter` in other producers in our project.

swatch-contracts/src/test/java/com/redhat/swatch/contract/CapacityReconciliationIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.time.Duration;
+import java.time.OffsetDateTime;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+@Disabled("Running kafka w/ kubedock broken, see https://github.com/joyrex2001/kubedock/issues/53")
+@QuarkusTest
+@QuarkusTestResource(value = KafkaCompanionResource.class, restrictToAnnotatedClass = true)
+class CapacityReconciliationIntegrationTest {
+  @InjectKafkaCompanion KafkaCompanion companion;
```
Sgitario, line 49: I know this issue cannot be reproduced using the smallrye in-memory component, so I like the idea of keeping this test here, but disabled, just to troubleshooting purposes.

Sgitario (commented):
@kahowell just a minor suggestion

kahowell:
Will merge after CI passes since the suggested change I pushed was semantically equivalent.

PR #3550, opened 2024-07-22
SWATCH-2726: Prevent send usages to other azure tenants if not forbidden

Sgitario:
Jira issue: SWATCH-2726

## Description
According to the Azure Marketplace API documentation: https://learn.microsoft.com/en-us/partner-center/marketplace-offers/marketplace-metering-service-apis, it can return the following HTTP error codes:
- 400: invalid request 
- 403: forbidden 
- 409: conflict, the usage event has already been successfully reported.

The problem is that we're always throwing the "AzureMarketplaceRequestFailedException: AZURE_MARKETPLACE_API_REQUEST_FAILED: Hint: Check Azure credentials" exception regardless the HTTP error code and also trying other client tenant. 

This exception only makes sense after receiving a 403 forbidden. For the rest, we should log a trace for troubleshooting and stop trying to send the usage (since azure will return the same response).

## Changes
- Do not try to send usages or throw the exception if azure returns with an error code other than 403.
- Add test using wiremock to cover integration with azure
- Disable auto discovery of providers for Quarkus rest-client in aws and azure producers (the existing clients already configure the providers using either the property and/or the annotation)

## Testing
Added junit test to reproduce the issue. 
Since we need a Azure service, I think this can't be reproduced locally or in an ephemeral environment.

### DIFF
--- a/swatch-producer-azure/src/main/java/com/redhat/swatch/azure/service/AzureMarketplaceService.java
+++ b/swatch-producer-azure/src/main/java/com/redhat/swatch/azure/service/AzureMarketplaceService.java
@@ -24,22 +24,24 @@ import com.redhat.swatch.azure.exception.AzureMarketplaceRequestFailedException;
 import com.redhat.swatch.azure.file.AzureMarketplaceProperties;
 import com.redhat.swatch.clients.azure.marketplace.api.model.UsageEvent;
 import com.redhat.swatch.clients.azure.marketplace.api.model.UsageEventOkResponse;
+import com.redhat.swatch.clients.azure.marketplace.api.model.UsageEventStatusEnum;
 import com.redhat.swatch.clients.azure.marketplace.api.resources.ApiException;
 import com.redhat.swatch.clients.azure.marketplace.api.resources.AzureMarketplaceApi;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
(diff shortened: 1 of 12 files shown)


swatch-producer-azure/src/main/java/com/redhat/swatch/azure/service/AzureMarketplaceService.java
```diff
@@ -60,10 +63,23 @@ public UsageEventOkResponse sendUsageEventToAzureMarketplace(UsageEvent usageEve
...
+              "Exception occurred during azure marketplace api request, likely expected since credentials are tried at random",
+              ex);
+        } else {
+          response =
+              new UsageEventOkResponse()
+                  .status(
+                      HTTP_STATUS_CONFLICT == status
+                          ? UsageEventStatusEnum.DUPLICATE
+                          : UsageEventStatusEnum.ERROR);
+          break;
```
Sgitario, line 79: @kflahert I think it makes sense to return this response when Azure returns these two errors, wdyt?

Sgitario:
Changing the PR to draft to:
- add an integration test that uses wiremock to stub azure
- fix the exception mapper so the ApiExceptionMapper implementation is used when submitting the usages to azure instead of DefaultApiExceptionMapper.

kflahert (approved):

liwalker-rh (approved):
Testing and latest PR checks look good to me.

PR #3557, opened 2024-07-23
SWATCH-2744: Refactor capacity reconciliation

kahowell:
Jira issue: SWATCH-2744

1. Enqueue all tasks up front.
2. Only reconcile from offering sync if capacity attributes change.

Jira issue: SWATCH-XXXX

## Testing

IQE Test MR: 

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.


lindseyburnett (approved):

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
```diff
@@ -171,12 +178,35 @@ private SyncResult syncOffering(Offering newState, Optional<Offering> persistedO
...
 
+  private boolean offeringChangeImpactsCapacityRecords(
+      Offering newState, Optional<Offering> persistedOffering) {
+    if (persistedOffering.isEmpty()) {
+      return true;
+    }
+
+    var existingState = persistedOffering.get();
+
+    return !Objects.equals(newState.getCores(), existingState.getCores())
```
kartikshahc, line 202: Seems good. Just verifying SLA and Usage are not part of capacity correct?

swatch-contracts/src/main/java/com/redhat/swatch/contract/repository/SubscriptionRepository.java
```diff
@@ -45,8 +45,17 @@ public List<SubscriptionEntity> findBySubscriptionNumber(String subscriptionNumb
...
-            Sort.by("startDate", Sort.Direction.Descending),
+            Sort.by(SubscriptionEntity_.START_DATE, Sort.Direction.Descending),
             subscriptionNumber);
     return query.list();
   }
+
+  public long countByOfferingSku(String sku) {
+    PanacheQuery<SubscriptionEntity> query =
+        find(
+            "offering.sku = ?1",
```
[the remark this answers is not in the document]
kahowell: Even better I realized we don't need sorts and can rely on the panache simplified syntax for this, will push up a change.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
kahowell: Correct.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
```diff
@@ -171,12 +178,35 @@ private SyncResult syncOffering(Offering newState, Optional<Offering> persistedO
       }
     }
 
-    // Existing capacities might need updated if certain parts of the offering was changed.
-    capacityReconciliationController.enqueueReconcileCapacityForOffering(newState.getSku());
+    // Existing capacities need to be updated if certain parts of the offering changed.
+    if (isCapacityImpacted) {
```
kartikshahc, line 182: Should this enqueue functionality occur before saveAndFlush in offeringRepo? This doesn't seem incorrec, as the save offering part is async and will also rollback if the enqueue transaction fails. If this implementation feels appropriate, I am fine.

kartikshahc:
Manual tested if we have a change in level 1, level 2, role then it doesn't affect capacity due to the check https://github.com/RedHatInsights/rhsm-subscriptions/blob/25623164e491243a89c8c2fd5586aba0b862ab7c/src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java#L202

Also checked if there is change in Sockets then it enqueues the task to update capacity

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
kahowell: Per [spring-kafka docs](https://docs.spring.io/spring-kafka/reference/kafka/transactions.html#container-transaction-manager):

> When the method exits, the database transaction will commit followed by the Kafka transaction.

In other words, the messages won't be sent if the DB transaction fails. If this was triggered as part of processing a task message, the message will be reprocessed. If it was triggered by an API, the API should show the error. In either case, we'd either manually (via API) or automatically (via kafka message) retry.

Also, it's been this way for a while, I just made the message send conditional.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
kahowell: Correct

swatch-contracts/src/main/java/com/redhat/swatch/contract/repository/SubscriptionRepository.java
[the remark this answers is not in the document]
kahowell: Even better, I realized the simplified panache syntax can be used. FWIW, I manually tested it with

```
quarkus.log.category."org.hibernate.SQL".level=DEBUG
```

in swatch-contracts application.properties; using

```shell
http PUT :8000/api/swatch-contracts/internal/rpc/offerings/reconcile/RH00604F5 x-rh-swatch-psk:placeholder
```

Both output equally:

```
select count(*) from subscription se1_0 where se1_0.sku=?
```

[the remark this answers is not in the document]
kartikshahc: yeah I was wondering about the same during the review. Thanks for the simplification.

src/main/java/org/candlepin/subscriptions/product/OfferingSyncController.java
kartikshahc: yup  :100: . I don't see any issue as I mentioned but, just wanted your thoughts on it.

kartikshahc (approved):
LGTM

PR #3560, opened 2024-07-24
SWATCH-2676: Return status code 404 for non-existing sku when syncing

wottop:
Jira issue: SWATCH-2676

## Description

The /internal/rpc/offerings/sync/{sku} always returns code 200 regardless of the outcome.
In scenarios where that is not the correct response the code will be what is expected

## Testing

IQE Test MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/809

### Steps

1. The unit tests have been updated for the new response. They should run correctly.
2. An IQE test is added as noted above.

### Verification
1. Tests pass


Sgitario (approved):
lgtm

PR #3572, opened 2024-07-26
SWATCH-2410: sort the metrics alphabetically in Subscription API v2

Sgitario:
Jira issue: SWATCH-2410

## Description
When filtering by different products, the order the metrics might be diferent: for example, for "rosa": ['Instance-hours', 'Cores'] and for "OpenShift-dedicated-metrics": ['Cores', 'Instance-hours'].

This is because the metrics are defined at this order. 

In the Instances API, we're sorting the metrics by alphabetical order, so we should do the same in the new Subscriptions API v2. This will simplify a lot the test automation.

## Testing
Only regression testing.


Sgitario:
fyi @mmusil

Sgitario:
Since @mmusil is being blocked by these changes and the change is very simple, I will self-merge it once CI is green.

PR #3579, opened 2024-07-30
SWATCH-2676: Update to error handling on SKU sync

wottop:
Jira issue: SWATCH-2676

## Description

The /internal/rpc/offerings/sync/{sku} always returns code 200 regardless of the outcome.
In scenarios where that is not the correct response the code will be what is expected

## Testing

IQE Test MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/809

### Steps

1. The unit tests have been updated for the new response. They should run correctly.
2. An IQE test is added as noted above.

### Verification
1. Tests pass

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsResource.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsResource.java
@@ -21,6 +21,7 @@
 package com.redhat.swatch.contract.resource;
 
 import com.fasterxml.jackson.dataformat.xml.XmlMapper;
+import com.redhat.swatch.clients.product.api.resources.ApiException;
 import com.redhat.swatch.contract.config.ApplicationConfiguration;
 import com.redhat.swatch.contract.model.SyncResult;
 import com.redhat.swatch.contract.openapi.model.AwsUsageContext;
@@ -37,7 +38,6 @@ import com.redhat.swatch.contract.openapi.model.RpcResponse;
 import com.redhat.swatch.contract.openapi.model.StatusResponse;
 import com.redhat.swatch.contract.openapi.model.SubscriptionResponse;
 import com.redhat.swatch.contract.openapi.model.TerminationRequest;
-import com.redhat.swatch.contract.openapi.resource.ApiException;
 import com.redhat.swatch.contract.openapi.resource.DefaultApi;
 import com.redhat.swatch.contract.product.umb.CanonicalMessage;
 import com.redhat.swatch.contract.product.umb.UmbSubscription;
@@ -67,7 +67,7 @@ import org.jboss.resteasy.reactive.common.NotImplementedYet;
 @Slf4j
 @ApplicationScoped
 @AllArgsConstructor
-public class ContractsTestingResource implements DefaultApi {
+public class ContractsResource implements DefaultApi {
 
   public static final String FEATURE_NOT_ENABLED_MESSAGE = "This feature is not currently enabled.";
   private static final String SUCCESS_STATUS = "Success";
@@ -321,14 +321,13 @@ public class ContractsTestingResource implements DefaultApi {
   @RolesAllowed({"test", "support", "service"})
   public OfferingResponse syncOffering(String sku) throws ProcessingException {
     var response = new OfferingResponse();
+    SyncResult result = null;
     try {
       log.info("Sync for offering {}", sku);
-      SyncResult result = offeringSyncService.syncOffering(sku);
+      result = offeringSyncService.syncOffering(sku);
       response.detail(String.format("%s for offeringSku=\"%s\".", result, sku));
-    } catch (Exception e) {
-      log.error("Error syncing offering", e.getMessage());
-      if (e.getCause() instanceof ApiException) {
-        ApiException apiException = (ApiException) e.getCause();
+    } catch (RuntimeException e) {
+      if (e.getCause() instanceof ApiException apiException) {
         switch (apiException.getResponse().getStatus()) {
           case 400:
             throw new BadRequestException(apiException.getMessage());
@@ -340,7 +339,12 @@ public class ContractsTestingResource implements DefaultApi {
             throw new InternalServerErrorException(apiException.getMessage());
         }
       }
-      response.setDetail("Error syncing offering");
+      throw new InternalServerErrorException(e.getMessage());
+    }
+    if (SyncResult.SKIPPED_NOT_FOUND.equals(result)) {
+      throw new NotFoundException(result.description());
+    } else if (SyncResult.SKIPPED_DENYLISTED.equals(result)) {
+      throw new ForbiddenException(result.description());
     }
     return response;
   }
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/DefaultExceptionMapper.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/DefaultExceptionMapper.java
@@ -23,6 +23,7 @@ package com.redhat.swatch.contract.resource;
 import com.redhat.swatch.contract.exception.ErrorCode;
 import com.redhat.swatch.contract.exception.ServiceException;
 import com.redhat.swatch.contract.openapi.model.Error;
+import jakarta.ws.rs.ClientErrorException;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.Response.Status;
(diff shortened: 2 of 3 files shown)


lindseyburnett (approved):

ntkathole:
@wottop does this fix https://issues.redhat.com/browse/SWATCH-2783 as well ?

Sgitario:
@wottop fyi, I had to revert the changes in https://github.com/RedHatInsights/rhsm-subscriptions/pull/3560 because it was preventing the service to be started.

So, as part of this PR, it would be needed to:
- integrate back the changes in https://github.com/RedHatInsights/rhsm-subscriptions/pull/3560, but fixing the mentioned issue (happy to be reviewer to help with this)
- updated this PR to do these changes in swatch contracts (it was moved from the monolith)

wottop:
Logging question still pending

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsTestingResource.java
```diff
@@ -321,26 +321,39 @@ public RpcResponse syncAllSubscriptions(Boolean forceSync) throws ProcessingExce
...
+            log.error("Error syncing offering", e.getMessage());
             throw new BadRequestException(apiException.getMessage());
           case 403:
+            log.warn("Warning syncing offering", e.getMessage());
             throw new ForbiddenException(apiException.getMessage());
           case 404:
+            log.warn("Warning syncing offering", e.getMessage());
             throw new NotFoundException(apiException.getMessage());
           default:
+            log.error("Error syncing offering", e.getMessage());
```
Sgitario, line 342: I don't think the new log traces added here add any value since we're throwing the exceptions with the actual information we need.

swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/ContractsTestingResourceTest.java
```diff
@@ -115,4 +127,46 @@ void forceSyncForOrgShouldReturnSuccess() {
...
+
+    when(offeringSyncService.syncOffering(any(String.class)))
+        .thenReturn(SyncResult.SKIPPED_NOT_FOUND);
+    assertThrows(NotFoundException.class, () -> resource.syncOffering("TEST_SKU"));
+
+    when(offeringSyncService.syncOffering(any(String.class)))
+        .thenReturn(SyncResult.SKIPPED_DENYLISTED);
+    assertThrows(ForbiddenException.class, () -> resource.syncOffering("TEST_SKU"));
+
+    ExternalServiceException ese = mock(ExternalServiceException.class);
```
Sgitario, line 150: The ExternalServiceException is never thrown by the OfferingSyncService, only by the SubscriptionSyncService. 
As far I see, the intention here was to mock the ApiException which is thrown by the ProductApi client. This client should directly return an instance of ApiException.

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsTestingResource.java
```diff
@@ -321,26 +321,39 @@ public RpcResponse syncAllSubscriptions(Boolean forceSync) throws ProcessingExce
...
     try {
       log.info("Sync for offering {}", sku);
-      SyncResult result = offeringSyncService.syncOffering(sku);
+      result = offeringSyncService.syncOffering(sku);
       response.detail(String.format("%s for offeringSku=\"%s\".", result, sku));
     } catch (Exception e) {
-      log.error("Error syncing offering", e.getMessage());
-      if (e.getCause() instanceof ApiException) {
-        ApiException apiException = (ApiException) e.getCause();
+      if (e.getCause() instanceof ApiException apiException) {
```
Sgitario, line 330: I think this is dead code which is never invoked. What we should do instead is:
- catch the ApiException (thrown by the ProductApi) and in this scenario, return an InternalServerErrorException exception (because the service is handing the connection to the ProductApi
- and don't catch any other exception

Sgitario (changes requested):
Apart from my comments, these changes will cause that the service became unavailable when the Kafka/UMB consumers handle messages that are not found or are denied (the OfferingSyncTaskConsumer methods should catch these exceptions to prevent this and log a warning trace when this is happening).

swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/ContractsTestingResourceTest.java
wottop: When I ran it with wiremock, this is exactly what showed up when the SKU was not found. If that is not a real live scenario, then what is? 
ExternalServiceException is a RuntimeException. I caught it and handled it for the reason above.

Sgitario: Maybe we're not speaking about the same use case. I tried to sync an offering without using stubs, so an exception is expected because the ProductApi is down and I see the ProcessingException instead of the ExternalServiceException:
Captura de pantalla de 2024-08-07 15-04-07

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsTestingResource.java
```diff
@@ -340,7 +339,17 @@ public OfferingResponse syncOffering(String sku) throws ProcessingException {
             throw new InternalServerErrorException(apiException.getMessage());
         }
       }
-      response.setDetail("Error syncing offering");
+      if (e instanceof NotFoundException
```
Sgitario, line 342: this `if` statement is no longer necessary, right?

wottop: It's in case we get another type of runtime exception of an unexpected type and converting to a 500. If you are good with just throwing it as is, then I can take it out.

Sgitario: 500 makes sense to me in this scenario. +1 to drop it

diegomaranhao:
```
In [2]: application = app

In [3]: sku = "NOTASKU"

In [4]: application.rhsm_subscriptions.sync_offering(sku=sku)
2024-08-13 21:26:09.603 [    INFO] [root] PUT request response url: http://swatch-contracts-service:8000/api/swatch-contracts/internal/rpc/offerings/sync/NOTASKU, status_code: 404
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 application.rhsm_subscriptions.sync_offering(sku=sku)

File ~/lib/python3.11/site-packages/iqe_rhsm_subscriptions/__init__.py:849, in ApplicationRhsmSubscriptions.sync_offering(self, sku)
    843 def sync_offering(self, sku):
    844     """
    845     Sync offering
    846 
    847     Returns: Updated offering definition in environment.
    848     """
--> 849     self.application.rhsm_subscriptions.rhsm_internal_api.contracts.sync_sku_for_capacity(sku)

File ~/lib/python3.11/site-packages/iqe_rhsm_subscriptions/internal_apis/swatch_contracts.py:355, in Contracts.sync_sku_for_capacity(self, sku)
    350 def sync_sku_for_capacity(self, sku):
    351     """Sync subscriptions for SKU ID
    352 
    353     :param str sku: Sku to sync capacity
    354     """
--> 355     self.call(
    356         app=swatch_contracts_app,
    357         endpoint=f"/internal/rpc/offerings/sync/{sku}",
    358         method="PUT",
    359     )
    360     logging.info(f"Offering {sku} synced.")

File ~/lib/python3.11/site-packages/iqe_rhsm_subscriptions/internal_apis/__init__.py:107, in InternalRestAPIClient.call(self, app, endpoint, request_params, payload_data, method, synchronous_request, thread_pool_executor)
    102 if payload_data:
    103     logging.debug(
    104         f"Payload: {json.dumps(payload_data, indent=4, sort_keys=True, ensure_ascii=False)}"
    105     )
--> 107 response.raise_for_status()
    108 if method == "DELETE":
    109     return None

File ~/lib/python3.11/site-packages/requests/models.py:1024, in Response.raise_for_status(self)
   1019     http_error_msg = (
   1020         f"{self.status_code} Server Error: {reason} for url: {self.url}"
   1021     )
   1023 if http_error_msg:
-> 1024     raise HTTPError(http_error_msg, response=self)

HTTPError: 404 Client Error: Not Found for url: http://swatch-contracts-service:8000/api/swatch-contracts/internal/rpc/offerings/sync/NOTASKU

In [5]: application.rhsm_subscriptions.sync_offering(sku="MCT4701")
2024-08-13 21:28:48.359 [    INFO] [root] PUT request response url: http://swatch-contracts-service:8000/api/swatch-contracts/internal/rpc/offerings/sync/MCT4701, status_code: 200
2024-08-13 21:28:48.359 [    INFO] [root] Response: {
    "detail": "syncResult=FETCHED_AND_SYNCED (Successfully fetched and synced updated value from upstream) for offeringSku=\"MCT4701\"."
}
2024-08-13 21:28:48.359 [    INFO] [root] Offering MCT4701 synced.
```

Sgitario (changes requested):
Thanks for renaming the ContractsTestingResource to ContractsResource +100
Yet the ContractsTestingResourceTest would also need to be renamed to ContractsResourceTest.

Sgitario (approved):

PR #3592, opened 2024-08-05
SWATCH-2790: Use saveOrUpdate when syncing subscriptions

Sgitario:
Jira issue: SWATCH-2790

## Description
The method `persist` will always perform an INSERT in database and this is causing issues when there are subscriptions with no subscription number, but the subscription ID correctly sets (it sounds weird, but it's what is happening in stage).

Using the method `saveOrUpdate` will correctly update the existing subscription if exists (so the subscription without the subscription number will be correctly fixed).

Note that because the `persist` method is a bit problematic (it happened the same when synchorinizing offerings), I replaced all the methods to `saveOrUpdate`. 

Apart from these changes:
- I created the stub search API which is by default false, because it was very convenient to reproduce the issue locally
- And disable the channels that consumed messages from umb if the property UMB_ENABLED is false: this is to avoid having a lot of exceptions locally

## Testing
1.- podman_compose up
2.- ./gradlew liquibaseUpdate
3.- load the stage data into your local database
4.- check that the following subscription is missing the subscription number:

```
select * from subscriptions where subscription_id ='6438190'
```

The subscription_number is empty.

5.- start the contracts service using the stubs: `SUBSCRIPTION_CLIENT_USE_STUB=true ./gradlew :swatch-contracts:quarkusDev`
6.- try to sync the problematic subscription:

```
<CanonicalMessage xmlns="http://esb.redhat.com/Canonical/6">
    <Payload>
        <Sync>
            <Subscription>
                <Identifiers>
                    <MasterSystem>SUBSCRIPTION</MasterSystem>
                    <Identifier entity-name="Install Base"
                        qualifier="number" system="EBS">39832705</Identifier>
                    <Identifier entity-name="Registration" qualifier="number">e401bce70edd43bd</Identifier>
                    <Identifier entity-name="Subscription"
                        qualifier="number" system="SUBSCRIPTION">6438195</Identifier>
                    <Reference entity-name="Installation" qualifier="number">c967ef3956ff3def</Reference>
                    <Reference entity-name="Customer" qualifier="number" system="EBS">customer123</Reference>
                    <Reference entity-name="Account" qualifier="number" system="EBS">account123</Reference>
                    <Reference entity-name="Customer" qualifier="id" system="WEB">7746627</Reference>
                    <Reference entity-name="Account" qualifier="id" system="WEB">16764251</Reference>
                </Identifiers>
                <Status>
                    <State>Active</State>
                    <StartDate>2023-04-24T00:00:00.000</StartDate>
                </Status>
[...]

PR #3597, opened 2024-08-06
SWATCH-2790: Use persist instead of saveOrUpdate in existing subs

Sgitario:
Jira issue: SWATCH-2790

## Description
These changes address some concerns from Alex in https://github.com/RedHatInsights/rhsm-subscriptions/pull/3592.  Also, the saveOrUpdate method will return the entity that is attached to the transaction context.

## Testing
Only regression testing.

### DIFF
--- a/swatch-common-panache/src/main/java/com/redhat/swatch/panache/PanacheSpecificationSupport.java
+++ b/swatch-common-panache/src/main/java/com/redhat/swatch/panache/PanacheSpecificationSupport.java
@@ -69,7 +69,7 @@ public interface PanacheSpecificationSupport<Entity, Id> extends PanacheReposito
     return query(clazz, specification, Page.of(0, 1)).getResultStream().findFirst();
   }
 
-  default void saveOrUpdate(Entity entity) {
-    JpaOperations.INSTANCE.getEntityManager().merge(entity);
+  default Entity merge(Entity entity) {
+    return JpaOperations.INSTANCE.getEntityManager().merge(entity);
   }
 }
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
@@ -413,7 +413,7 @@ public class ContractService {
     boolean areRecordsUpdated = !subscriptionsToPersist.isEmpty();
     if (areRecordsUpdated) {
       log.info("Persisting subscriptions: {}", subscriptionsToPersist);
-      subscriptionsToPersist.forEach(subscriptionRepository::saveOrUpdate);
+      subscriptionsToPersist.forEach(subscriptionRepository::persist);
     }
     return areRecordsUpdated;
   }
@@ -523,7 +523,7 @@ public class ContractService {
         .findContracts(ContractEntity.orgIdEquals(orgId))
         // we only want to update existing subscriptions, so we don't need to provide the
         // subscriptionId here.
-        .forEach(contract -> syncSubscriptionForContract(contract, null));
+        .forEach(this::syncSubscriptionForContract);
 
     statusResponse.setStatus(SUCCESS_MESSAGE);
     return statusResponse;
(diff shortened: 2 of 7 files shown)


awood (commented):
Specifically, I was talking about [here](https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/SubscriptionSyncService.java#L195).  I don't believe `saveOrUpdate` is needed at all here.  The changes will be persisted automatically after the transaction ends.

Generally, I'm also concerned about the identity properties of `Subscription` because of this unit test:

```java
  @Transactional
  @Test
  void testMergeExisting() {
    SubscriptionEntity s1 = createSubscription("888");
    s1.setSubscriptionNumber("123");
    // Uncomment and test will pass
    // s1.setStartDate(OffsetDateTime.of(2024, 1, 1, 10, 0, 0, 0, ZoneOffset.UTC));
    subscriptionRepo.persistAndFlush(s1);
    log.info("s1 {}", s1);

    subscriptionRepo.getEntityManager().detach(s1);
    s1.setQuantity(1000);
    var attached = subscriptionRepo.getEntityManager().merge(s1);
    subscriptionRepo.getEntityManager().flush();
    assertEquals(s1.getSubscriptionId(), attached.getSubscriptionId());
    assertEquals(s1.getStartDate(), attached.getStartDate());
    assertEquals(s1, attached);
    log.info("s1 {}", s1);
    log.info("attached {}", attached);

    assertEquals(1, subscriptionRepo.findBySubscriptionNumber("123").size());
  }
```

That test will actually create 2 records instead of just doing an `UPDATE`.  So if we start using `merge` in `saveOrUpdate` with `Subscription`, I'm worried we will have problems with duplicated records.

And even more generally, I don't think we should encourage a `saveOrUpdate` model of thought.  It just doesn't fit accurately into the JPA lifecyle. Hibernate has deprecated a `saveOrUpdate` method on `Session` since Hibernate 6 for exactly that [reason](https://github.com/hibernate/hibernate-orm/discussions/4520).  The Spring Data `save` method is pretty lenient 

```java
	@Override
	@Transactional
	public <S extends T> S save(S entity) {

		Assert.notNull(entity, "Entity must not be null");

		if (entityInformation.isNew(entity)) {
			entityManager.persist(entity);
			return entity;
		} else {
			return entityManager.merge(entity);
		}
	}
```

If the `entity` is already attached to the persistence context, Hibernate [just ignores](https://thorben-janssen.com/spring-data-jpa-save-saveandflush-and-saveall/#merging-an-existing-entity-object) the call to `merge`.  Thorben describes my concern pretty succinctly: "Unfortunately, during my audit and coaching sessions, I often see calls of the save method after an entity object was changed. This method call shows that the developer wasn’t familiar with [JPA’s lifecycle model](https://thorben-janssen.com/entity-lifecycle-model/)."

So, I'm all for having a 

```java
  default Entity merge(Entity entity) {
    return JpaOperations.INSTANCE.getEntityManager().merge(entity);
  }
```
method on `PanacheSpecificationSupport` and then using that `merge` when dealing with detached entities.  But I think having it named `saveOrUpdate` and then just replacing `persist` with `saveOrUpdate` isn't a good idea.  At best, we'll end up doing extra DB calls and at worst people are going to ignore the returned entity and just keep working with the passed argument and end up with confusing bugs when they try working with that entity once the transaction closes and they try to `persist` it in another transaction.

Sgitario:
> Specifically, I was talking about [here](https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/SubscriptionSyncService.java#L195). I don't believe `saveOrUpdate` is needed at all here. The changes will be persisted automatically after the transaction ends.
> 

Correct, I already fixed this.

> Generally, I'm also concerned about the identity properties of `Subscription` because of this unit test:
> 
> ```java
>   @Transactional
>   @Test
>   void testMergeExisting() {
>     SubscriptionEntity s1 = createSubscription("888");
>     s1.setSubscriptionNumber("123");
>     // Uncomment and test will pass
>     // s1.setStartDate(OffsetDateTime.of(2024, 1, 1, 10, 0, 0, 0, ZoneOffset.UTC));
>     subscriptionRepo.persistAndFlush(s1);
>     log.info("s1 {}", s1);
> 
>     subscriptionRepo.getEntityManager().detach(s1);
>     s1.setQuantity(1000);
>     var attached = subscriptionRepo.getEntityManager().merge(s1);
>     subscriptionRepo.getEntityManager().flush();
>     assertEquals(s1.getSubscriptionId(), attached.getSubscriptionId());
>     assertEquals(s1.getStartDate(), attached.getStartDate());
>     assertEquals(s1, attached);
>     log.info("s1 {}", s1);
>     log.info("attached {}", attached);
> 
>     assertEquals(1, subscriptionRepo.findBySubscriptionNumber("123").size());
>   }
> ```
> 
> That test will actually create 2 records instead of just doing an `UPDATE`. So if we start using `merge` in `saveOrUpdate` with `Subscription`, I'm worried we will have problems with duplicated records.

It's not possible to have 2 duplicated records because we would be hitting a duplicated primary key constraint exception. However, if we would be using auto generated IDs, this is a real issue, because even when you set the ID to update an existing entity, then, when using `merge`, it will ignore this ID, auto-generate a new ID and it will create another record. When using `persist`, this works fine and the entity is updated. 

> 
> And even more generally, I don't think we should encourage a `saveOrUpdate` model of thought. It just doesn't fit accurately into the JPA lifecyle. Hibernate has deprecated a `saveOrUpdate` method on `Session` since Hibernate 6 for exactly that [reason](https://github.com/hibernate/hibernate-orm/discussions/4520). The Spring Data `save` method is pretty lenient
> ...
> method on `PanacheSpecificationSupport` and then using that `merge` when dealing with detached entities. But I think having it named `saveOrUpdate` and then just replacing `persist` with `saveOrUpdate` isn't a good idea. At best, we'll end up doing extra DB calls and at worst people are going to ignore the returned entity and just keep working with the passed argument and end up with confusing bugs when they try working with that entity once the transaction closes and they try to `persist` it in another transaction.

Yes, I think your arguments here make a lot of sense +100
Initially, I needed the merge operation (the `saveOrUpdate` method) because we were querying a subscription by subscription number, but then using the subscription ID as primary key, so when the provided <subscription number> x <subscription ID> didn't match in our database, then we got the issue because the existing subscription was not found. I will revert the usage of the `saveOrUpdate` to workaround this and fix the logic to find subscriptions.

Sgitario:
@awood PR updated with:
- I renamed the `saveOrUpdate` to `merge` to avoid misunderstandings
- I only kept the `merge` operation only in the place that was needed and created SWATCH-2801 to address it

PR #3603, opened 2024-08-07
SWATCH-2742: key constraint exception during account reset

awood:
Jira issue: SWATCH-2742

## Description
Calls to ` DELETE .../api/rhsm-subscriptions/v1/internal/rpc/tally/{org_id}` were failing because the code attempted to delete the `account_services` rows while they were still referenced from `hosts`.

I also included a small fix to `insert-mock-hosts` since I discovered that script wasn't working when I was trying to replicate the issue.

## Testing
Unit test only


awood:
QE can probably skip this one since the issue is already causing a failing test.

lindseyburnett (approved):

PR #3606, opened 2024-08-07
SWATCH-2766: Error reading denylist

wottop:
Jira issue: SWATCH-2766

## Description

When the product id is null we need to return false instead of checking against the denylist

## Testing
Unit tests to cover places where this is found

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/capacity/files/ProductDenylist.java
+++ b/src/main/java/org/candlepin/subscriptions/capacity/files/ProductDenylist.java
@@ -65,7 +65,7 @@ public class ProductDenylist implements ResourceLoaderAware {
   }
 
   public boolean productIdMatches(String productId) {
-    if (source == null) {
+    if (source == null || productId == null) {
       return false;
     }
     try {
--- a/src/test/java/org/candlepin/subscriptions/files/ProductDenylistTest.java
+++ b/src/test/java/org/candlepin/subscriptions/files/ProductDenylistTest.java
@@ -52,6 +52,7 @@ class ProductDenylistTest {
     assertTrue(denylist.productIdMatches("I3"));
     assertFalse(denylist.productIdMatches("I111"));
     assertFalse(denylist.productIdMatches("I112"));
+    assertFalse(denylist.productIdMatches(null));
   }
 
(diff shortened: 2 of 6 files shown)


Sgitario (changes requested):
The root cause of this issue is that there are some tests trying to save subscriptions with a sku = null:

```json
[{"id": 951221, "subscriptionNumber": 8401831, "subscriptionProducts": [{"sku": null}], "webCustomerId": "17702870", "quantity": 1, "oracleAccountNumber": "", "effectiveStartDate": 1719792188216.0, "effectiveEndDate": 1753920188216.0, "externalReferences": {"azureMarketplace": {"accountID": null, "subscriptionID": "testBillingProvider", "azureResourceId": "a69ff71c-aa8b-43d9-dea8-822fab4bbb86", "partner": "azure_marketplace", "offerId": "azureProductCode", "planId": "rh-rhel-sub-1yr"}}}]
```

Splunk query: `index=rh_rhsm namespace="rhsm-stage" properties.trace_id="f81a1a50ad5b0701c156f08ec5541937"` (from 31 July). 

@ntkathole this is something that you might need to address (or maybe you already did because I don't see more errors after 31 July)

These changes make the `productIdMatches` method more robust which is good, but I bet that when trying to save a subscription with sku = null, it will fail in another part of the logic. 

Therefore, apart of making the method `productIdMatches` more robust as you already did, it's also needed to update the `extractSku` method from [here](https://github.com/RedHatInsights/rhsm-subscriptions/blob/5bf0182af12c32b1ac7a8f992c26132849f3ba20/swatch-contracts/src/main/java/com/redhat/swatch/contract/utils/SubscriptionDtoUtil.java#L48) to filter out sku that are empty or null.

After doing this change, it will fail with the expected exception: https://github.com/RedHatInsights/rhsm-subscriptions/blob/5bf0182af12c32b1ac7a8f992c26132849f3ba20/swatch-contracts/src/main/java/com/redhat/swatch/contract/utils/SubscriptionDtoUtil.java#L61

Sgitario (approved):
lgtm

ntkathole (approved):
Looks good!

PR #3610, opened 2024-08-08
SWATCH-2805: Bind the subscription client property in swatch-contracts

Sgitario:
Jira issue: SWATCH-2805

## Description
The offering sync fails in stage because the call to subscription client fails with 504 Bad Gateway. I confirmed that the subscription API is working fine for stage, so this is a configuration issue. 

By binding these subscription properties, the swatch-contracts should be able to overwrite the URL to use.

## Testing
Only reproducible in stage.

### DIFF
--- a/swatch-contracts/deploy/clowdapp.yaml
+++ b/swatch-contracts/deploy/clowdapp.yaml
@@ -56,6 +56,12 @@ parameters:
     value: prod
   - name: UMB_KEYSTORE_PATH
     value: /pinhead/keystore.jks
+  - name: UMB_ENABLED
+    value: 'false'
+  - name: UMB_PROCESSING_ENABLED
+    value: 'true'
+  - name: UMB_SERVICE_ACCOUNT_NAME
+    value: placeholder
   - name: PRODUCT_URL
     value: https://product.stage.api.redhat.com/svcrest/product/v3
   - name: SUBSCRIPTION_SYNC_ENABLED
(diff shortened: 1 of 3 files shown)


kahowell (changes requested):
There are some unrelated changes in the PR (related to UMB), but generally, this isn't necessary. Try:

```shell
RBAC_ENDPOINT=http://localhost:8080 RBAC_ENABLED=false SWATCH_SELF_PSK=placeholder QUARKUS_PROFILE=stage quarkus dev
```
and then visit http://localhost:8000/q/dev-ui/configuration-form-editor and you'll be able to confirm that the effective value for `SUBSCRIPTION_URL` is already `https://subscription.stage.api.redhat.com/svcrest/subscription/v5` when `QUARKUS_PROFILE=stage`

Sgitario:
> There are some unrelated changes in the PR (related to UMB), but generally, this isn't necessary. Try:
> 
> ```shell
> RBAC_ENDPOINT=http://localhost:8080 RBAC_ENABLED=false SWATCH_SELF_PSK=placeholder QUARKUS_PROFILE=stage quarkus dev
> ```
> 
> from swatch-contracts
> 
> and then visit http://localhost:8000/q/dev-ui/configuration-form-editor and you'll be able to confirm that the effective value for `SUBSCRIPTION_URL` is already `https://subscription.stage.api.redhat.com/svcrest/subscription/v5` when `QUARKUS_PROFILE=stage`

I know that these changes should not be necessary, but I was trying to keep the service as the subscription sync service was configured. 
Anyway, I will try a different strategy and keep you posted.

PR #3615, opened 2024-08-09
Remove unused SubscriptionDtoUtil from the monolith

Sgitario:
## Description
This class should have been ported/removed in https://github.com/RedHatInsights/rhsm-subscriptions/pull/3510

## Testing
Only regression testing.


Sgitario:
Self-merging since this was a leftover from another JIRA ticket.

PR #3616, opened 2024-08-09
SWATCH-2805: disable autodiscovery of providers in swatch-contracts

Sgitario:
Jira issue: SWATCH-2805

## Description
- Disable autodiscovery of providers in swatch-contracts
This is causing that the exceptions to be thrown by the clients are not the ones from its exception mappers.

- Enable logging of requests-response for the SearchAPI
This change might help to understand what is going on in SWATCH-2805

## Testing
Only regression testing.


Sgitario:
Self-merging this to see if stage gets fixed.

PR #3617, opened 2024-08-09
Add Spring Boot to Quarkus migration tips section

Sgitario:
## Description
This section aims to add some tips of known differences between Spring Boot and Quarkus, and how to migrate it.

## Testing
Not needed.


Sgitario:
Self-merging this since this is a documentation only thing.

PR #3620, opened 2024-08-09
SWATCH-2805: Add PRODUCT_KEYSTORE and SUBSCRIPTION_KEYSTORE in swatch-contracts

Sgitario:
Jira issue: SWATCH-2805

## Description
And replace TRUSTSTORE_RESOURCE by TRUSTSTORE_PATH which is the one used in the environment

## Testing
Only regression testing.


Sgitario:
Self-merging since this is just a minor enhancement to keep using the previous variables before porting this functionality to swatch-contracts.

PR #3626, opened 2024-08-12
SWATCH-2424: Delete inner query to purge the tally snapshots

Sgitario:
Jira issue: SWATCH-2424

## Description
Delete the inner query "(select distinct c.org_id from org_config c)" by queuing the tasks to purge the tally snapshots by org_id.

I checked the explain using the inner query and without using it, and the delete statement cost decreased significantly. Note that I could reproduce the issue locally (by loading a large dataset), but hopefully these changes and making the deletion by organization ID in a single transaction will improve this situation.

Related to https://github.com/RedHatInsights/rhsm-subscriptions/pull/3487

## Testing
0. podman-compose up
2. insert the small dataset:

```

```

### DIFF
--- a/src/main/resources/application-worker.yaml
+++ b/src/main/resources/application-worker.yaml
@@ -41,6 +41,8 @@ rhsm-subscriptions:
     # Five years' worth
     quarterly: ${TALLY_RETENTION_QUARTERLY:20}
     yearly: ${TALLY_RETENTION_YEARLY:5}
+    # Batch size to fetch organizations using pagination
+    organizations-batch-limit: 1000
     snapshots-to-delete-in-batches: 1000000
   tally-summary-producer:
     back-off-initial-interval: ${TALLY_SUMMARY_PRODUCER_BACK_OFF_INITIAL_INTERVAL:1s}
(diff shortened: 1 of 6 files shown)


kflahert (approved):

Sgitario:
The problem of these changes is that we will keep a long running transaction to iterate over all the orgIDs which is not a good practice ++ I'm sure it will fail in production because timeouts. 
I will be updating the PR to fetch the orgIds in batches instead.

Sgitario:
@kflahert I've changed the implementation to use pagination (load around 1000 organizations in memory) instead of keeping a single transaction running to stream the organizations and purge all snapshots since this is a bad practice and I'm sure it will time out in production.

src/main/resources/application-worker.yaml
```diff
@@ -41,6 +41,8 @@ rhsm-subscriptions:
     # Five years' worth
     quarterly: ${TALLY_RETENTION_QUARTERLY:20}
     yearly: ${TALLY_RETENTION_YEARLY:5}
+    # Load organizations in this batch limit. It needs to be
```
kflahert, line 44: Looks like an unfinished comment here :)

Sgitario: Ops! Fixed.

PR #3655, opened 2024-08-16
SWATCH-2826: Fix PAYG fingerprinting bug with role value

kahowell:
Jira issue: SWATCH-2826

Description
===========

Fix a bug where cost-managed populated role values were unintentionally causing payg values to not be properly identified.

This happened because we refactored the fingerprinting logic in #3496 so that the role/eng id/level/product name identifiers are consulted first, and the first match short-circuits the rest. Then further filtering is done later on to find a more specific match.

Given current product definitions, this meant that a PAYG rhel fingerprint could be inappropriately initially tagged with "RHEL for x86" and then later have that tag filtered out, leading to no matching tags.

This change moves the payg/non-payg filter up-front, so that this doesn't happen.

Testing
=======

IQE Test MR: TODO

See the added test:

```shell
./gradlew swatch-product-configuration:test --tests SubscriptionDefinitionTest.testCostManagementRhelElsAddon
```

### DIFF
--- a/swatch-product-configuration/src/main/java/com/redhat/swatch/configuration/registry/SubscriptionDefinition.java
+++ b/swatch-product-configuration/src/main/java/com/redhat/swatch/configuration/registry/SubscriptionDefinition.java
@@ -188,13 +188,12 @@ public class SubscriptionDefinition {
 
   public static Set<String> getAllProductTags(ProductTagLookupParams params) {
 
-    // First identify variants by one of the required product identifiers
-    Set<Variant> filteredVariants = getVariantsMatchingRequiredIdentifier(params);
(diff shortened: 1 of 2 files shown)


kflahert (approved):

lindseyburnett (approved):

liwalker-rh (approved):
Looks good in multiple PR checks (any failures are unrelated) and running tests locally. We do have some IQE coverage already to look up product tags but can look into adding more coverage if needed.

kahowell:
Assuming that `tests.swatch-metrics.test_events.test_event_conflict_resolution[rhel-for-x86-els-payg]` failure is unrelated, merging.

PR #3659, opened 2024-08-19
Konflux: Fix image specification for deprecated-base-image-check task

barnabycourt:
An update to the task changed the parameters that need to be passed to it. This is causing a failure to initialize the build on push.


barnabycourt:
Merging after discussion on morning call

PR #3660, opened 2024-08-19
SWATCH-2809: Add traceresponse header to quarkus HTTP responses

wottop:
Jira issue: SWATCH-2809

## Description

By adding the header, we can better track back to the logs to determine the source of errors

### Setup

1. Start swatch-contracts locally [with all that goes with it]

### Steps

1. use http to hit an endpoint in one of the quarkus based modules
```http  :8882/api/swatch-contracts/internal/rpc/offerings/sync/NOTASKU``` 

### Verification

1. The response will include the "traceresponse" header
```
TTP/1.1 405 Method Not Allowed
content-length: 0
traceresponse: 00-2a994a3866c77d96d5b7d8680162a387-5c09d8b8dd197a1e-01

```

### CODE REVIEW

**Review by awood — CHANGES_REQUESTED** [review]
I'm not actually seeing this get hit.  I'm running`swatch-billable-usage` like so:

```
./gradlew :swatch-billable-usage:quarkusDev
```

and then hitting it via 

```
http -v :8000/api/swatch-billable-usage/internal/remittance/accountRemittances orgId==org123 x-rh-swatch-psk:placeholder
```

And I'm not actually seeing the trace.  Nor the debug message.  Even with the debugger attached, I don't see it hitting `TraceResponseFilter`.

PR #3680, opened 2024-08-26
SWATCH-2282: Port remaining subscription logic to swatch-contracts

kahowell:
Jira issue: SWATCH-2282

Description
===========

This ports logic from the monolith into swatch-contracts. Namely, this makes the following APIs functional:

- `GET /api/swatch-contracts/internal/subscriptions/awsUsageContext`
- `GET /api/swatch-contracts/internal/subscriptions/azureUsageContext`
- `GET /api/swatch-contracts/internal/subscriptions/rhmUsageContext`
- `GET /api/swatch-contracts/internal/tags/{tag}/metrics`
- `POST /api/swatch-contracts/internal/subscriptions/terminate/{subscription_id}`

It was necessary to enhance `PanacheSpecificationSupport` to support sorting. I added some unit tests to validate that the generated queries are valid, but those tests don't validate the generated sql contains the `order by` clauses expected, because there wasn't an obvious way to verify this. Instead, I used logging to manually verify the generated queries.

It was necessary to update the ORM mapping for SubscriptionEntity to match that used in the monolith (namely the `subscriptionMeasurements` attribute). Unfortunately, this caused changes in a number of dependent classes.

I also performed some refactoring as follows:

1. Refactored the product tag lookups to use the internal swatch-product-configuration module.
2. Refactored the PAYG usage context lookups to use the repository to lookup existing state directly, rather than depending on SubscriptionSyncService to do this.
3. Refactored the metrics for missing and ambigious payg subscriptions to use a `provider` label, rather than creating unique metrics for each billing provider. (No existing dashboards or alerts use these).

Testing
=======

IQE Test MR: [!841](https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/841)

Note that unit tests and the IQE changes are expected to sufficiently test these changes, since there is no net new functionality.

Note that `com.redhat.swatch.panache.SpecificationTest` logs queries generated with `order by` clauses.


kahowell:
055f933 to c6bc65f: spotlessApply fixup

Sgitario (changes requested):
Changes look good to me, though I was expecting to the ported logic in InternalSubscriptionResource was also deleted and the dependant classes were moved instead of created (to preserve the GIT history and allow to perform rebases if needed). Some of these classes are:
- UsageContextSubscriptionProvider
- UsageContextSubscriptionProviderTest
But probably there are more.

kahowell:
c6bc65f to b4b2ef0: rebase against main; see `meld <(git show 4acf6f91d8bd97cada3e2243a901f79f9d0b3bfa) <(git show ae2e0022b8bf78b9fe7853175d25a537688b0a81)` if you want to quickly verify the equivalence of the primary commit.

kahowell:
@Sgitario latest rebase was a bit messier; it's worth reviewing: `meld <(git show 4acf6f91d8bd97cada3e2243a901f79f9d0b3bfa) <(git show 945042e609982d52077e3ee97e99ab56dab18dcd)` and/or `MeasurementMetricIdTransformer` again, as well as the brand new commits. I'm hoping to squash before merging which should preserve the history better as you requested. It did make it necessary to actually remove the internal subscription API, but I think this is okay.

Sgitario (commented):
Not sure why, but UsageContextSubscriptionProvider is still maked as removed in the monolith, and added to the swatch-contracts, instead of moved.
For UsageContextSubscriptionProviderTest, it's now marked as moved. 

By the way, I submit a commit to fix the sonar issues.

kahowell:
> UsageContextSubscriptionProvider is still marked as removed in the monolith

I think this happened because the amount of changes I made in refactoring was over a threshold that git uses.

> I submit a commit to fix the sonar issues.

:pray: thanks for this!

Any other feedback before this is dev-approved?

Sgitario (approved):

Sgitario:
Approved. Changes look good to me, though failures seems suspicious. I've just rebased to main just in case.

ntkathole:
Test failing on asserting message on contract creation, when same contract payload is received multiple times. 

```
AssertionError: Same contract payload should have been ignored.
assert 'Existing con...tions updated' == 'Redundant message ignored'
  
  - Redundant message ignored
  + Existing contracts and subscriptions updated
```
Is this message change on api response expected ?

Sgitario:
> Test failing on asserting message on contract creation, when same contract payload is received multiple times.
> 
> ```
> AssertionError: Same contract payload should have been ignored.
> assert 'Existing con...tions updated' == 'Redundant message ignored'
>   
>   - Redundant message ignored
>   + Existing contracts and subscriptions updated
> ```
> 
> Is this message change on api response expected ?

No, this is caused by a new bug. I could reproduce the issue, but I'm working on the fix atm.

Sgitario:
@kahowell finally, found the root cause of the issue: the metric id transformer was updating the existing metric keys which made the map that was already storing this metric be inconsistent. 
I fixed the issue by:
- Refactoring the subscription and metric id transformers to not having to store wrong metrics in the subscription, but directly storing the correct metric IDs
- Remove the setters of the SubscriptionMeasurementKey to avoid the same issue to happen in the future
- Remove the redundant ContractServiceSubscriptionTest test and moved the tests to the existing ContractServiceTest

You can find these changes as part of this commit: https://github.com/RedHatInsights/rhsm-subscriptions/pull/3680/commits/2259bb6069c4426139f04f2549cbb688a4e85613

kahowell:
@Sgitario :100: excellent fix, thanks for tracking that down and not only fixing the issue, but also cleaning up the implementation as well!

ntkathole (approved):

kahowell:
1e54fca to 38e7849: squashed and rebased

To see diff of diffs: `meld <(git diff 9852ff4..1e54fca) <(git diff 297f952..38e7849)`

kahowell:
Merging, because failing test is wholly unrelated: `tests.integration.exports.test_export_instances.test_export_instances_payg_product_filters[json-OpenShift-dedicated-metrics]`

PR #3681, opened 2024-08-26
Drop conduit init container and fix param names

kahowell:


kahowell:
Self-merging because this causes confusion in konflux setup

PR #3699, opened 2024-08-30
SWATCH-2855: Truncate start/end dates when syncing contracts from IT Partner gateway

Sgitario:
Jira issue: SWATCH-2855

## Description
The partner gateway uses nanosecond precision timestamps. Postgres/HSQL only support microsecond precision, so there is a loss of precision when the timestamps are persisted. To see a couple of examples of postgres having less precision, you can try in the database:
```
select '2024-08-24T12:00:00.000000501Z'::timestamptz
select '2024-08-24T12:00:00.000000500Z'::timestamptz
```
Notice that they produce different values.
This is causing a number of side-effects, namely, the equals methods on contracts and subscriptions are treating records as unequal when they shouldn't and, when we use the startDate as a key as seen [here](https://github.com/RedHatInsights/rhsm-subscriptions/blob/0da69bb0a921fedb8e7aadea2e0b5cd0b1027729/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java#L290), it causes a miss in the mapping to existing records.

With these changes, we are truncating the start and end dates to use the same precision than stored in database.

## Testing
Added test to reproduce the scenario.

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractEntityMapper.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractEntityMapper.java
@@ -31,6 +31,7 @@ import com.redhat.swatch.clients.rh.partner.gateway.api.model.SaasContractV1;
 import com.redhat.swatch.contract.repository.ContractEntity;
 import com.redhat.swatch.contract.repository.ContractMetricEntity;
 import java.time.OffsetDateTime;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


awood (approved):

awood:
@Sgitario Should we apply these changes to SubscriptionEntityMapper too?  Or does that service not provide dates with nanosecond precision?  It looks like we ignore the startDate but I'm wondering if we need to worry about the end date.  Or maybe it just doesn't even matter since the endDate isn't part of a key anywhere.

Sgitario:
> @Sgitario Should we apply these changes to SubscriptionEntityMapper too? Or does that service not provide dates with nanosecond precision? It looks like we ignore the startDate but I'm wondering if we need to worry about the end date. Or maybe it just doesn't even matter since the endDate isn't part of a key anywhere.

I checked the subscription start date and when getting the payload from IT, we're getting an unix timestamp, so it is not affected by this issue.

PR #3700, opened 2024-08-30
SWATCH-2862: Purge tally measurements in batches

Sgitario:
Jira issue: SWATCH-2862

## Description
After SWATCH-2791, we did limit the number of snapshots to delete using the "limit" clause. The problem is that one snapshot might have too many measurements to be deleted in the same transaction, so we need to first purge the measurements in batches, and then proceed with the snapshots in batches as well.

## Testing
0. podman-compose up
2. insert the small dataset:

```

```


wottop:
lgtm

PR #3710, opened 2024-09-04
SWATCH-2871: Handle azure contracts measurements in MeasurementMetricIdTransformer

kflahert:
Jira issue: SWATCH-2871

## Description

Azure contracts were not setting the correct metricId value on subscription measurements causing errors. This change also includes removing the dependency on internalSubscriptionApi.

## Testing
Should fix failing IQE tests


Sgitario (changes requested):
Changes look good to me. 
The only missing part is about to remove the dependency "implementation project(':clients:swatch-internal-subscription-client')" since we're no longer using it.

Sgitario (approved):

PR #3725, opened 2024-09-09
SWATCH-2160: Increase SubscriptionDefinition lookup speed

awood:
Jira issue: SWATCH-2160

## Description
This commit adds a caching component to the `SubscriptionDefinition` class to improve the efficiency of looking up `SubscriptionDefinition`'s by role, tag, and engineering ID.  I've made several design choices in this commit that I will explain below.

Why Caffeine?  Guava does offer a Cache implementation and we are already using Guava in this project.  However, even the Guava [JavaDocs](https://guava.dev/releases/snapshot/api/docs/com/google/common/cache/CacheBuilder.html) explicitly state that Caffeine is the superior choice.  The Caffeine API is almost identical to Guava's so switching between the two is mostly just a matter of changing import statements.

Why do we need a scheduler?  Caffeine ordinarily performs its maintenance and evictions during writes and occasionally after read operations if writes are rare.  Our cache has an extremely high ratio of reads to writes.  In that case, the documentation suggests providing the cache with the Java 9 system-wide scheduling thread so cleanUp
operations can be scheduled.

Why are we using a time based eviction strategy instead of one based on reference count?  I did a lot of experimenting on this.  The issue is that while `SubscriptionDefinition` objects are long lived (and thus using a weak reference as the cache value and allowing garbage collection to be the de facto eviction method) in the 
`SubscriptionDefinitionRegistry` class, `Optional<SubscriptionDefinition>` objects are created ad hoc in our search methods.  I added an eviction listener, wrote multiple test cases with `System.gc()` calls, and could see our entries getting evicted once the test finished.

Why don't we use `SubscriptionDefinition` as our cache values?  Our cache loading method returns an `Optional<SubscriptionDefinition>`.  While we can do a `.orElse(null)` on the optional to resolve the `Optional`, doing so
decreases the performance of the method.  In my benchmark, the time required when returning a  `SubscriptionDefinition` by resolving the `Optional` increased the run time from about 1.5 seconds to 5.5 seconds.

Why don't we have the cache loading method return a `SubscriptionDefinition` or null?  That approach just moves the problem to a different location.  We're using the Stream API to move through the `SubscriptionDefinition`'s `Variants`.  Any `Collector` is going to either return an `Optional` or a collection that we would then have to crack open and convert an empty Optional/collection to null. 

## Testing
Unit testing suffices for checking for regressions.  Verifying the improvement in speed requires a performance test which I'm including as a patch since it's not something we really need to check-in.

### Setup
1.  Apply the following patch: `patch -p1 < SWATCH-2160-test.patch`:
    ```patch
    diff --git a/swatch-product-configuration/src/main/java/com/redhat/swatch/configuration/registry/SubscriptionDefinition.java b/swatch-product-configuration/src/main/java/com/redhat/swatch/configuration/registry/SubscriptionDefinition.java
    index c160e5c40..5af80122e 100644
    --- a/swatch-product-configuration/src/main/java/com/redhat/swatch/configuration/registry/SubscriptionDefinition.java
    +++ b/swatch-product-configuration/src/main/java/com/redhat/swatch/configuration/registry/SubscriptionDefinition.java
    @@ -63,18 +63,21 @@ public class SubscriptionDefinition {
       // See https://github.com/ben-manes/caffeine/wiki/Cleanup
       static LoadingCache<String, Optional<SubscriptionDefinition>> tagCache =
           Caffeine.newBuilder()
    +          .recordStats()
               .expireAfterAccess(3, TimeUnit.HOURS)
               .scheduler(Scheduler.systemScheduler())
               .build(SubscriptionDefinition::cacheSubscriptionByTag);
     
       static LoadingCache<String, Optional<SubscriptionDefinition>> roleCache =
           Caffeine.newBuilder()
    +          .recordStats()
               .expireAfterAccess(3, TimeUnit.HOURS)
               .scheduler(Scheduler.systemScheduler())
               .build(SubscriptionDefinition::cacheSubscriptionByRole);
     
       static LoadingCache<String, Set<SubscriptionDefinition>> engIdCache =
           Caffeine.newBuilder()
    +          .recordStats()
               .expireAfterAccess(3, TimeUnit.HOURS)
               .scheduler(Scheduler.systemScheduler())
               .build(SubscriptionDefinition::cacheSubscriptionByEngId);
    diff --git a/swatch-product-configuration/src/test/java/com/redhat/swatch/configuration/registry/SubscriptionDefinitionCacheTest.java b/swatch-product-configuration/src/test/java/com/redhat/swatch/configuration/registry/SubscriptionDefinitionCacheTest.java
    new file mode 100644
    index 000000000..c0c13d9c2
    --- /dev/null
    +++ b/swatch-product-configuration/src/test/java/com/redhat/swatch/configuration/registry/SubscriptionDefinitionCacheTest.java
    @@ -0,0 +1,140 @@
    +/*
    + * Copyright Red Hat, Inc.
    + *
    + * This program is free software: you can redistribute it and/or modify
    + * it under the terms of the GNU General Public License as published by
    + * the Free Software Foundation, either version 3 of the License, or
    + * (at your option) any later version.
    + *
    + * This program is distributed in the hope that it will be useful,
    + * but WITHOUT ANY WARRANTY; without even the implied warranty of
    + * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    + * GNU General Public License for more details.
    + *
    + * You should have received a copy of the GNU General Public License
    + * along with this program.  If not, see <https://www.gnu.org/licenses/>.
    + *
    + */
    +package com.redhat.swatch.configuration.registry;
    +
    +import static org.junit.jupiter.api.Assertions.*;
    +
    +import com.github.benmanes.caffeine.cache.Cache;
    +import com.google.common.base.Stopwatch;
    +import java.util.Optional;
    +import java.util.Set;
    +import java.util.concurrent.TimeUnit;
    +import java.util.function.Function;
    +
    +class SubscriptionDefinitionCacheTest {
    +  @Test
    +  void benchmarkLookupSubscriptionByStrategies() {
    +    Stopwatch stopwatch = Stopwatch.createStarted();
    +    System.out.println("TAG:");
    +    var stats = benchmarkLookupSubscriptionsByTag(SubscriptionDefinition::cacheSubscriptionByTag);
    +    stopwatch.stop();
    +    System.out.format("Without cache: %,10d ms%n", stopwatch.elapsed(TimeUnit.MILLISECONDS));
    +    System.out.println(stats);
    +
    +    stopwatch.reset();
    +    stopwatch.start();
    +    stats = benchmarkLookupSubscriptionsByTag(SubscriptionDefinition::lookupSubscriptionByTag);
    +    System.out.format("With cache: %,10d ms%n", stopwatch.elapsed(TimeUnit.MILLISECONDS));
    +    System.out.println(stats);
    +
    +    System.out.println("ROLE:");
    +    stopwatch.reset();
    +    stopwatch.start();
    +    stats = benchmarkLookupSubscriptionsByRole(SubscriptionDefinition::cacheSubscriptionByRole);
    +    System.out.format("Without cache: %,10d ms%n", stopwatch.elapsed(TimeUnit.MILLISECONDS));
    +    System.out.println(stats);
    +
    +    stopwatch.reset();
    +    stopwatch.start();
    +    stats = benchmarkLookupSubscriptionsByRole(SubscriptionDefinition::lookupSubscriptionByRole);
    +    System.out.format("With cache: %,10d ms%n", stopwatch.elapsed(TimeUnit.MILLISECONDS));
    +    System.out.println(stats);
    +
    +    System.out.println("ENG ID:");
    +    stopwatch.reset();
    +    stopwatch.start();
    +    stats = benchmarkLookupSubscriptionsByEngId(SubscriptionDefinition::cacheSubscriptionByEngId);
    +    System.out.format("Without cache: %,10d ms%n", stopwatch.elapsed(TimeUnit.MILLISECONDS));
    +    System.out.println(stats);
    +
    +    stopwatch.reset();
    +    stopwatch.start();
    +    stats = benchmarkLookupSubscriptionsByEngId(SubscriptionDefinition::lookupSubscriptionByEngId);
    +    System.out.format("With cache: %,10d ms%n", stopwatch.elapsed(TimeUnit.MILLISECONDS));
    +    System.out.println(stats);
    +  }
    +
    +  String benchmarkLookupSubscriptionsByTag(
    +      Function<String, Optional<SubscriptionDefinition>> lookupFunction) {
    +    for (int i = 0; i < 1_000_000; i++) {
    +      assertTrue(lookupFunction.apply("BASILISK").isPresent());
    +      assertTrue(lookupFunction.apply("Satellite Server").isPresent());
    +      assertTrue(lookupFunction.apply("rhel-for-x86-ha").isPresent());
    +      assertTrue(lookupFunction.apply("rosa").isPresent());
    +      assertTrue(lookupFunction.apply("RHEL for x86").isPresent());
    +      assertTrue(lookupFunction.apply("OpenShift-metrics").isPresent());
    +      assertTrue(lookupFunction.apply("rhacs").isPresent());
    +      assertFalse(lookupFunction.apply("nothing").isPresent());
    +      assertFalse(lookupFunction.apply("does not exist").isPresent());
    +      assertFalse(lookupFunction.apply("nada").isPresent());
    +    }
    +    return collateStats(SubscriptionDefinition.tagCache);
    +  }
    +
    +  String benchmarkLookupSubscriptionsByRole(
    +      Function<String, Optional<SubscriptionDefinition>> lookupFunction) {
    +    for (int i = 0; i < 1_000_000; i++) {
    +      assertTrue(lookupFunction.apply("BASILISK").isPresent());
    +      assertTrue(lookupFunction.apply("Red Hat Enterprise Linux Server").isPresent());
    +      assertTrue(lookupFunction.apply("Red Hat Enterprise Linux Workstation").isPresent());
    +      assertTrue(lookupFunction.apply("Red Hat Enterprise Linux Compute Node").isPresent());
    +      assertTrue(lookupFunction.apply("rhacs").isPresent());
    +      assertTrue(lookupFunction.apply("moa-hostedcontrolplane").isPresent());
    +      assertTrue(lookupFunction.apply("ocp").isPresent());
    +      assertTrue(lookupFunction.apply("osd").isPresent());
    +      assertFalse(lookupFunction.apply("nothing").isPresent());
    +      assertFalse(lookupFunction.apply("does not exist").isPresent());
    +      assertFalse(lookupFunction.apply("nada").isPresent());
    +    }
    +    return collateStats(SubscriptionDefinition.roleCache);
    +  }
    +
    +  String benchmarkLookupSubscriptionsByEngId(
    +      Function<String, Set<SubscriptionDefinition>> lookupFunction) {
    +    for (int i = 0; i < 1_000_000; i++) {
    +      assertFalse(lookupFunction.apply("250").isEmpty());
    +      assertFalse(lookupFunction.apply("269").isEmpty());
    +      assertFalse(lookupFunction.apply("90").isEmpty());
    +      assertFalse(lookupFunction.apply("70").isEmpty());
    +      assertFalse(lookupFunction.apply("204").isEmpty());
    +      assertFalse(lookupFunction.apply("290").isEmpty());
    +      assertFalse(lookupFunction.apply("251").isEmpty());
    +      assertTrue(lookupFunction.apply("999999999").isEmpty());
    +      assertTrue(lookupFunction.apply("9999999").isEmpty());
    +      assertTrue(lookupFunction.apply("99999").isEmpty());
    +    }
    +    return collateStats(SubscriptionDefinition.roleCache);
    +  }
    +
    +  String collateStats(Cache c) {
    +    var stats = c.stats();
    +    var sb = new StringBuilder();
    +    sb.append("\tHit Rate: " + stats.hitRate() + "\n");
    +    sb.append("\tHit Count: " + stats.hitCount() + "\n");
    +    sb.append("\tMiss Count: " + stats.missCount() + "\n");
    +    sb.append("\tLoad Time: " + stats.totalLoadTime() + "ms\n");
    +    sb.append("\tLoad Success Count: " + stats.loadSuccessCount() + "\n");
    +    sb.append("\tRequest Count: " + stats.requestCount() + "\n");
    +    sb.append("\tEviction Count: " + stats.evictionCount() + "\n");
    +    return sb.toString();
    +  }
    +}
    ```
    This testing method works because the `cacheSubscriptionBy*` methods are just the original lookup methods with  
    the small modification of combining two `filter` calls into one.

### Steps
1.  Run the test
    ```bash
    ./gradlew :swatch-product-configuration:test --tests='SubscriptionDefinitionCacheTest'
    xdg-open "file:///$(git rev-parse --show-toplevel)/swatch-product-configuration/build/reports/tests/test/classes/com.redhat.swatch.configuration.registry.SubscriptionDefinitionCacheTest.html
    ```
1. Click the `Standard Output` button on the page to view the benchmarks.  I've
   reproduced my results here just for future reference.
    ```
    TAG:
    Without cache:     13,163 ms
    	Hit Rate: 1.0
    	Hit Count: 0
    	Miss Count: 0
    	Load Time: 0ms
    	Load Success Count: 0
    	Request Count: 0
    	Eviction Count: 0
    
    With cache:      1,461 ms
    	Hit Rate: 1.0
    	Hit Count: 0
    	Miss Count: 0
    	Load Time: 0ms
    	Load Success Count: 0
    	Request Count: 0
    	Eviction Count: 0
    
    ROLE:
    Without cache:     14,988 ms
    	Hit Rate: 1.0
    	Hit Count: 0
    	Miss Count: 0
    	Load Time: 0ms
    	Load Success Count: 0
    	Request Count: 0
    	Eviction Count: 0
    
    With cache:      1,503 ms
    	Hit Rate: 1.0
    	Hit Count: 0
    	Miss Count: 0
    	Load Time: 0ms
    	Load Success Count: 0
    	Request Count: 0
    	Eviction Count: 0
    
    ENG ID:
    Without cache:     15,797 ms
    	Hit Rate: 1.0
    	Hit Count: 0
    	Miss Count: 0
    	Load Time: 0ms
    	Load Success Count: 0
    	Request Count: 0
    	Eviction Count: 0
    
    With cache:      1,392 ms
    	Hit Rate: 1.0
    	Hit Count: 0
    	Miss Count: 0
    	Load Time: 0ms
    	Load Success Count: 0
    	Request Count: 0
    	Eviction Count: 0
    ```

### Verification
1. The results should show a 90% reduction in execution time.


ben-manes:
> Schedule periodic cleanup so that cache maintenance occurs outside periods of heavy read usage.

The cache maintenance is very inexpensive and runs regularly to replay activity. However it is not a dedicated thread and defers to FJP.commonPool() when triggered. It is asynchronous because an eviction listener may be attached and foreign code has an unknown expense. Many like Infinispan use the calling thread, executor(Runnable::run).

The scheduler is for prompt expiration when the cache is idle, e.g. to close a websocket after a timeout. Some users need predictable cleanup/notification, whereas most don’t care. The GC is lazy anyway, so eagerly evicting an expired entry usually has no memory performance effects.

If your cache is fairly active then you don’t need the scheduler, but it does no harm. I’d also recommend using jmh for benchmarks to avoid jit warmup surprises.

Sgitario (changes requested):
I don't think we need a cache like this since the loaded data is coming from static files that won't change after 3 hours or other amount of time, so in practice, this is like having loaded this data in memory all the time. 

Perhaps, having the data in a weak reference that is released if and only if we're out of memory is a better solution. After released, we can load it again when we need it.

awood:
> I don't think we need a cache like this since the loaded data is coming from static files that won't change after 3 hours or other amount of time, so in practice, this is like having loaded this data in memory all the time.
> 

No, the entries will be evicted after 3 hours if they are not used.  Since nightly tallies (`BaseSnapshotRoller`) are what are doing the lookups, the cache should be empty for most of the day.

If we use weak references, the cache will be emptied much more frequently (degrading performance) since those values will be getting GCed in between tallies of the different orgs (since the values are only referenced by the cache itself).

awood:
> If your cache is fairly active then you don’t need the scheduler, but it does no harm. I’d also recommend using jmh for benchmarks to avoid jit warmup surprises.

Thank you so much for commenting @ben-manes!  Our cache is going to be pretty inactive.  There will be just a few writes when it's empty and then a ton of reads during a nightly job.  After that it'll be inactive until the next night.  I'm not concerned so much about eviction occurring right away.  My concern was more that I didn't want to slow down cache activity while we are performing our burst of reads.  Does this seem like a good use-case for the scheduler?

Thanks again for stopping by our humble repo!

ben-manes:
> My concern was more that I didn't want to slow down cache activity while we are performing our burst of reads.

That shouldn't be a problem. The reads are lock-free and shouldn't block, even under high load. We split the cache into its data and the policy, with a pub-sub model to keep the policy eventually consistent. This way you can get ConcurrentHashMap style performance and behavior (e.g per-entry computes), without being bottlenecked as we keep the policy up-to-date (global shared state, every read is a write). We use bounded buffers where when full on a read it drops events (as popularity skew will mean those items get added more often, so minimal impact to the hit rate) and the write buffer only get throttled when inserts greatly outpace evictions (no runaway memory leak). 

These are the jmh [benchmarks](https://github.com/ben-manes/caffeine/wiki/Benchmarks) run on my M3 Max laptop:

image

Sgitario:
> No, the entries will be evicted after 3 hours if they are not used. Since nightly tallies (`BaseSnapshotRoller`) are what are doing the lookups, the cache should be empty for most of the day.
> 
> If we use weak references, the cache will be emptied much more frequently (degrading performance) since those values will be getting GCed in between tallies of the different orgs (since the values are only referenced by the cache itself).

My main concern is about having extra data loaded in memory just in the tally process that is when we push the service memory's constraints more. For example: if the tally takes more than 3 hours, the cache will be refreshed several times which is not a problem, but the memory will keep this data longer. 

If this extra data is only a few MB, then that's ok. Otherwise, I would like that the garbage collector can free this data via either weak reference or maybe we can configure caffeine for doing this, even if this degradates the performance.

ben-manes:
In that case using the scheduler is appropriate as it will trigger a cleanUp() when the entry expires. You can specify weak values but it might be collected prematurely, so I prefer that for lifecycle management rather than size control.

awood:
> My main concern is about having extra data loaded in memory just in the tally process that is when we push the service memory's constraints more. For example: if the tally takes more than 3 hours, the cache will be refreshed several times which is not a problem, but the memory will keep this data longer.
> 
> If this extra data is only a few MB, then that's ok. Otherwise, I would like that the garbage collector can free this data via either weak reference or maybe we can configure caffeine for doing this, even if this degradates the performance.

These data structures are not large at all.  Our tallies take so long, that I would rather optimize for CPU usage than memory use.  There are 22 SubscriptionDefinitions and according to the profiler that totals to about 1KB of memory used.  Caffeine classes add about 4KB.  We can use the weak references approach, but it will result in substantially more cache evictions since every tally basically will have to repopulate whatever `Optional<SubscriptionDefinition>` objects got GCed.  Personally, I think that is worth the extra 5KB of extra memory use to improve the tally speed as much as possible.

Like @ben-manes mentioned, I can put the scheduler back in to trigger cleanUp's once the entries expire, but again that's extra CPU use for the sake of 5KB of memory.

PR #3738, opened 2024-09-13
SWATCH-2910: Filter snapshots by supported metric in the product

Sgitario:
Jira issue: SWATCH-2910

## Description
When existing snapshots have metrics that are not supported by the project (for example: the product "rhel-for-x86-els-payg-addon" does not support the metrics "Sockets" or "Cores"),  Then we were creating invalid billable usage remittance entries that were kept indefinetely on pending state.

Also, we are updating the aws and azure producers to properly update these records with status failed.

## Testing
Added junit tests to cover these changes.
This addresses a production issue containing ol data in the tally_measurements table.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
@@ -38,6 +38,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.candlepin.clock.ApplicationClock;
 import org.candlepin.subscriptions.db.AccountServiceInventoryRepository;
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -388,24 +389,29 @@ private void updateUsage(AccountUsageCalculation calc, Event event) {
...
                   measurement -> {
-                    var metricId =
-                        MetricId.fromString(
-                            Optional.ofNullable(measurement.getMetricId())
-                                .orElse(measurement.getUom()));
-                    if (Variant.getMetricsForTag(productId).stream()
-                        .map(Metric::getId)
-                        .anyMatch(definedMetricId -> definedMetricId.equals(metricId.toString()))) {
+                    var metricId = toMetricId(measurement);
+                    if (isMetricSupportedByProduct(metricId, productId)) {
```
kflahert, line 396: I think we should log a warning when we skip a metric.

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -528,11 +534,14 @@ private AccountUsageCalculation loadHourlyAccountCalculation(Event event) {
               .forEach(
                   entry -> {
                     var measurementKey = entry.getKey();
-                    calc.addUsage(
-                        usageKey,
-                        measurementKey.getMeasurementType(),
-                        MetricId.fromString(measurementKey.getMetricId()),
-                        entry.getValue());
+                    var metricId = MetricId.fromString(measurementKey.getMetricId());
+                    if (isMetricSupportedByProduct(metricId, usageKey.getProductId())) {
```
kflahert, line 538: Same as above, think we should log when we skip a metric.

kflahert (approved):

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
Sgitario: Added

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
Sgitario: Added

TrayvonMcKnight:
[IQE MR](https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/869)

PR #3743, opened 2024-09-13
SWATCH-2907: Replace custom retry annotation with standard ones

kahowell:
Jira issue: SWATCH-2907

Description
===========

This replaces the workaround we had previously for doing runtime configuration of retry parameters with idiomatic quarkus/smallrye fault tolerance annotations. (This removes the
swatch-common-smallrye-fault-tolerance module).

Our previous pattern was creating high cardinality metrics because the method name wasn't detected properly. With this change, the method name is populated properly, so we can re-enable fault tolerance metrics.

I also removed the env var & template parameter definitions for some of these settings, as we're not using them currently, and we should instead override them in the application.properties file if necessary.

Testing
=======

Setup
-----

1. Temporarily add to `TallySummaryMessageConsumer.consume`:

```java
if (true) {
  throw new RuntimeException("foobar");
}
```

2. Start swatch-billable-usage with the max retries overridden:

```shell
BILLING_PRODUCER_MAX_ATTEMPTS=4 \
./gradlew swatch-billable-usage:quarkusDev
```

Steps
-----

1. Using a kafka producer (IntelliJ kafka producer works well for this), produce a message to `platform.rhsm-subscriptions.tally` with the contents `{"org_id":"foo","tally_snapshots":[]}`.
2. Wait about 15 seconds to allow all retries to finish.
3. Fetch metrics via `http :9000/metrics | grep ft_retry_retries_total`

Verification
------------

1. Notice that the count (the first number in the returned line) is 4.0, corresponding to the retry setting.
2. Notice that the method is specified appropriately.

If you try these steps prior to this change, you should notice that the `method` value is a random UUID.


kahowell:
Requesting review from @Sgitario specifically, since the workaround I'm removing was his. I don't mind another reviewer as well if anyone else wants to take a look.

Sgitario (changes requested):
The custom annotation was added not only because it allows to configure the retry at runtime and because it eases a lot the configuration binding by avoiding to need adding up to four properties for each usage, but also because the ExponentialBackoff annotation is not taking this runtime configuration. 
See https://sgitario.github.io/using-exponential-backoff-retry-strategy-in-quarkus/, the section "Limitations of the Smallrye Fault Tolerance API to configure the retry settings at runtime".

To be precise, these properties:
```
org.acme.CoffeeResource/coffees/ExponentialBackoff/maxDelay=${COFFEE_BACK_OFF_MAX_INTERVAL_MILLIS}
org.acme.CoffeeResource/coffees/ExponentialBackoff/factor=${COFFEE_BACK_OFF_MULTIPLIER}
```
Won't ever take effect and it will use the defaults instead.

I checked that by populating the expected class and method name in the `withDescription` , then it works fine. See https://github.com/RedHatInsights/rhsm-subscriptions/pull/3746 as an alternative solution that will prevent the break of the configuration binding for the `@ExponentialBackoff` annotation.

PR #3756, opened 2024-09-17
SWATCH-2925: Handle NumberFormatException for cores/sockets

kahowell:
Jira issue: SWATCH-2925

Description
===========

We started receiving messages having "None" as the value for SOCKET_LIMIT, and this caused a crashloop in swatch-contracts.

Instead, log a warning and interpret the value as null if it's not "Unlimited" or an integer.

Testing
=======

```shell
./gradlew -s swatch-contracts:test \
  --tests=UpstreamProductDataTest.testOfferingWithNoneSocketsProcessedSuccessfully
```

See logs like:

> Invalid integer value None for attr CORES
> Invalid integer value None for attr SOCKET_LIMIT

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/product/UpstreamProductData.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/product/UpstreamProductData.java
@@ -452,9 +452,9 @@ public class UpstreamProductData {
         Optional.ofNullable(attrs.get(Attr.IFL))
             .map(ifl -> Integer.parseInt(ifl) * CONVERSION_RATIO_IFL_TO_CORES)
             // ... but if IFL is not defined, then use the CORES attr.
-            .orElseGet(() -> nullOrInteger(attrs, Attr.CORES));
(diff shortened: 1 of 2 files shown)


swatch-contracts/src/main/java/com/redhat/swatch/contract/product/UpstreamProductData.java
```diff
@@ -485,11 +485,17 @@ There are no SKUs today (2021-10-27) that provide both standard capacity and hyp
...
+  private static Integer nullOrInteger(Map<Attr, String> attrs, Attr key) {
+    String capacity = attrs.get(key);
     if (capacity == null || hasUnlimitedUsage(capacity)) {
       return null;
     } else {
-      return Integer.valueOf(capacity);
+      try {
+        return Integer.valueOf(capacity);
+      } catch (NumberFormatException e) {
+        log.warn("Invalid integer value {} for attr {}", capacity, key);
```
Sgitario, line 496: We should also log the sku with the invalid value for better troubleshooting.

Sgitario (changes requested):

Sgitario (approved):

PR #3761, opened 2024-09-18
SWATCH-2145: Merge duplicate hosts from different sources in tally

wottop:
Jira issue: SWATCH-2145

## Description

We have hosts in the rhsm-subscription table 'hosts' that have duplicate instance ID's. This migration will merge those entries into a single host entry. This will be followed by a new constraint put on the table as well as new logic to determine if a host already exists in this table.

### Setup

1. import from production:
```
./import-from-gabi.py --host --org-id='5345257'
```

### Steps

1. Find an instance id that is repeated:
```
select id,instance_id, org_id from hosts where instance_id in (
    select instance_id from hosts group by instance_id having count(instance_id) > 1)
  and org_id = '5345257'
order by instance_id asc
```
2. Look at the host entries [keep track of the host ids]:
```
select * from hosts where instance_id = 'i-0002481e85a633e0c' order by last_seen desc
```
3. Look at the monthly totals:
```
select * from instance_monthly_totals where host_id in ('bff818fb-ab19-47a9-bf30-bc59f175491c','1d2d9476-b757-4b0b-895e-752397a48d8f');
```
4. Look at the instance measurements:
```
select * from instance_measurements where host_id in ('bff818fb-ab19-47a9-bf30-bc59f175491c','1d2d9476-b757-4b0b-895e-752397a48d8f')
```
5. Look at the tally buckets
```
select * from host_tally_buckets where host_id in ('bff818fb-ab19-47a9-bf30-bc59f175491c','1d2d9476-b757-4b0b-895e-752397a48d8f');
```

6. Call the migration endpoint:
```
http  POST :8000/api/rhsm-subscriptions/v1/internal/rpc/tally/hosts/merge?org_id=5345257 x-rh-identity:$(echo '{"identity":{"internal":{"org_id":"5345257"}}}' | base64 -w0)
```

### Verification

1. There should only be one host entry for an instance id. The initial query for hosts should return no values.
2. The instance type should be HBI_HOST if either had that value before
3. the last_seen and last_applied_event_record_date should be the most recent
4. the monthly totals should be summed across the hosts per metric_id
5. the measurements should be a single entry per metric id with the value belonging to the most recent last_seen [they really should be the same anyway]
6. The list of tally buckets should include all associated with both.
7. The initial query for hosts should return no values

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/admin/MergeHostsMigration.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/admin/MergeHostsMigration.java
@@ -1,0 +1,240 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package org.candlepin.subscriptions.tally.admin;
+
+import com.redhat.swatch.configuration.registry.MetricId;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.sql.Types;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.candlepin.subscriptions.db.HostRepository;
+import org.candlepin.subscriptions.db.model.Host;
+import org.candlepin.subscriptions.db.model.HostTallyBucket;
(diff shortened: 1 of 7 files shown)


src/main/java/org/candlepin/subscriptions/tally/admin/MergeHostsMigration.java
```diff
@@ -0,0 +1,240 @@
...
+    return lastInstanceId;
+  }
+
+  /**
+   * Merge the identifier values into the primary (latest seen wins)
+   *
+   * @param primaryHost
+   * @param host
+   */
+  private void mergeDataFields(Host primaryHost, Host host) {
```
lindseyburnett, line 155: It might be helpful to adjust the merge methods so that instead of mutating the incoming Host objects, we return a new Host object. This way, we could have unit tests that assert directly on the output, making it easier to validate the behavior of each method in isolation. What do you think?

src/main/java/org/candlepin/subscriptions/tally/admin/MergeHostsMigration.java
```diff
@@ -0,0 +1,240 @@
...
+      primaryHost.addToMonthlyTotal(
+          monthlyKey.getMonth(),
+          MetricId.fromString(monthlyKey.getMetricId()),
+          host.getMonthlyTotal(
+              monthlyKey.getMonth(), MetricId.fromString(monthlyKey.getMetricId())));
+    }
+  }
+
+  @Override
+  public void recordCompleted() {
```
lindseyburnett, line 237: I think this indicates that there's some future work to be done around the DataMigration class and what it's used for.  This seems to be the second time we've extended it and then opted to intentionally leave the implementation for recordCompleted() blank.  To me, this indicates maybe we're not using DataMigration as it was originally intended (also an indicator is how there are references to liquibase changesets in DataMigration) for.

For the sake of this story/PR, I'm ok with it as it is because it's working and adding value, but I'd be interested in having a conversation with @kahowell at some point about intended usage of that abstract class.

lindseyburnett (changes requested):

lindseyburnett (approved):

PR #3766, opened 2024-09-19
SWATCH-2887: Delete host tally buckets after processing event batches

Sgitario:
Jira issue: SWATCH-2887

## Description
After the refactor of the event ingestion, we are handling events in batches ordered only by record date. 

Then, when we handle the events for the same host and the buckets need to be either updated/added/deleted in different transactions, hibernate throws an exception because the bucket version to update or delete does not match with the expected (the bucket version has been updated in other transaction).

Therefore, we are adding a transient "stale" field in the host tally buckets, so we can delete them after processing the batch of events. This way when an event is handled, a previously deleted tally bucket can be "restored" preserving the version.

## Testing
Added test to reproduce the issue.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
@@ -106,7 +106,7 @@ public class MetricUsageCollector {
             Collectors.toMap(Host::getInstanceId, Function.identity(), this::handleDuplicates));
 
     for (Event event : events) {
-      Host host = hostsByInstanceId.getOrDefault(event.getInstanceId(), new Host());
+      Host host = hostsByInstanceId.computeIfAbsent(event.getInstanceId(), i -> new Host());
 
       // During the hourly tally, hosts and snapshots are updated in separate transactions.
       // If an error occurs somewhere in the overall tally process while processing a batch of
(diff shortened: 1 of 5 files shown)


Sgitario:
@kahowell adding you as reviewer as you have some context about the changes.
To sum up, I think the best quick fix to unblock production is to drop the version column in the entity. Indeed, this will cause a performance penalty with an extra select when updating/inserting the buckets, though I don't think it will be very notorious since these entities are indexed.

For a long term fix, I think the way to go is to logically delete the buckets instead of perform delete statements, we'll always perform update statements that will populate the version accordignly. If you're ok with these changes, I will open a follow-up JIRA ticket for this and also to add back the version column to configure the optimistic locking.

barnabycourt:
This needs review by the performance team before merge

kahowell (changes requested):
> We could refactor the event ingestion by grouping the events by host or perform a logical delete for the tally buckets, though these changes might have side effects in performance and other parts of the backend, so it might need more investigation.

I thought about this a bit and I think there's a solution that's similar to logical deletion, but doesn't have the same risk of negative performance impacts:

1. Add a field named `stale` to the `HostTallyBucket` class (default `false`). Make this field transient (it doesn't need to be persisted).
2. When processing events, mark the bucket as stale rather than removing it from the collection.
3. When adding buckets, if there is an existing stale bucket, change the value of `stale` to `false`.
4. After all events have been processed, remove all `stale=true` buckets from hosts in the batch.

I'd rather implement the above than remove optimistic locking.

Sgitario:
> 4. After all events have been processed, remove all `stale=true` buckets from hosts in the batch.

Unfortunately, this is happening because hosts might have in different batches and hence be updated in different transactions, so deleting the stale buckets after each batch would have the same issue.

Sgitario:
@kahowell solution updated according to your feedback. 
I confirmed that it works by running the implemented test case, and also using the prod data that I initially used to reproduce the issue.

kahowell (approved):

PR #3776, opened 2024-09-23
SWATCH-1233: Drop UOM from the swatch

Sgitario:
Jira issue: SWATCH-1233

## Description
UOM is no longer referenced in swatch openapi definitions Automation tests are no longer referencing UOM
UOM is no longer referenced in SWatch code
UOM is no longer stored as part of events. 
Double check that no queries are using UOM
check bin/ scripts to make sure UOM is no longer referenced

## Testing
Only regression testing.


wottop (changes requested):
swatch-core/src/main/java/org/candlepin/subscriptions/db/model/TallyMeasurementKey.java:77:        + ", uom="
swatch-tally/deploy/clowdapp.yaml:708:        metric_id as uom,

Sgitario:
> swatch-core/src/main/java/org/candlepin/subscriptions/db/model/TallyMeasurementKey.java:77: + ", uom=" 

Updated

> swatch-tally/deploy/clowdapp.yaml:708: metric_id as uom,

This is used by the floorplan queries, so it should stay as it is as noted in the JIRA description.

PR updated.

liwalker-rh (approved):
Changes look good in regression testing. I also added an IQE MR to remove variables named UOM in any test and replaced with metric_id when necessary. This will help to avoid anyone being confused about UOM still being seen in tests in the future. https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/874

wottop (approved):

PR #3800, opened 2024-09-26
SWATCH-2970: Fix the contract API certificate configuration in azure svc

Sgitario:
Jira issue: SWATCH-2970

## Description
- Fix the contract API certificate configuration in azure svc
- Not use the wiremock in swatch producer azure in EE
- Configure aws and azure services the same way to deal with the wiremock

## Testing
The test "test_verify_multi_cloud_provider_usage" should pass in CI.

### DIFF
--- a/swatch-producer-azure/src/main/resources/application.properties
+++ b/swatch-producer-azure/src/main/resources/application.properties
@@ -134,9 +134,9 @@ mp.messaging.outgoing.billable-usage-status-out.value.serializer=org.candlepin.s
 quarkus.rest-client-reactive.provider-autodiscovery=false
 
 quarkus.rest-client."com.redhat.swatch.clients.contracts.api.resources.DefaultApi".url=${SWATCH_CONTRACTS_ENDPOINT}
-quarkus.rest-client."com.redhat.swatch.clients.contracts.api.resources.DefaultApi".trust-store=${clowder.endpoints.swatch-subscription-sync-service.trust-store-path}
-quarkus.rest-client."com.redhat.swatch.clients.contracts.api.resources.DefaultApi".trust-store-password=${clowder.endpoints.swatch-subscription-sync-service.trust-store-password}
-quarkus.rest-client."com.redhat.swatch.clients.contracts.api.resources.DefaultApi".trust-store-type=${clowder.endpoints.swatch-subscription-sync-service.trust-store-type}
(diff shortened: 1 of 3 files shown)


swatch-producer-azure/src/main/resources/application.properties
```diff
@@ -43,8 +43,9 @@ AZURE_MARKETPLACE_BASE_URL=https://marketplaceapi.microsoft.com
 AZURE_MARKETPLACE_API_VERSION=2018-08-31
 AZURE_OIDC_SAAS_MARKETPLACE_RESOURCE=20e940b3-4c77-4b0b-9a53-9e16a1b010a7
 
+# wiremock defaults
 %wiremock.SWATCH_CONTRACTS_ENDPOINT=http://${WIREMOCK_HOSTNAME}:8101/mock/internalSub
-%ephemeral.SWATCH_CONTRACTS_ENDPOINT=${WIREMOCK_ENDPOINT}/mock/contractsApi
+# %ephemeral.SWATCH_CONTRACTS_ENDPOINT=${WIREMOCK_ENDPOINT}/mock/contractsApi
```
kflahert, line 48: We depend on this for some component testing so ideally will keep this in.

There is a mismatch between here and wiremock proxy which would prevent it from getting proxied: https://github.com/RedHatInsights/wiremock-consoledot/blob/5f3b7f481f4fa5e6c8f6e2fa4fda437a6a175674/deploy/clowdapp.yaml#L129

```suggestion
%ephemeral.SWATCH_CONTRACTS_ENDPOINT=${WIREMOCK_ENDPOINT}/mock/contractApi
```

liwalker-rh:
test_verify_multi_cloud_provider_usage was being skipped from testing since it was failing often so I removed the skip label in this MR https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/879. Now it will be tested in the CI to see if the fixes in this PR will now make it pass.

swatch-producer-azure/src/main/resources/application.properties
Sgitario: I've re-added and would like @liwalker-rh double check that the SSL issue is gone. If not, this is definetively caused by wiremock being in the middle and not accepting the SSL certificate.

kflahert (approved):

liwalker-rh (approved):
Looks like test_verify_multi_cloud_provider_usage is now passing in latest PR check and output looks to be what is expected. Other tests that are failing are not related.

PR #3813, opened 2024-09-30
SWATCH-2963: Count Splunk transmission failures in Quarkus apps

awood:
Jira issue: SWATCH-2963

## Description
We have a custom Splunk error callback that increments a counter when failures
occur during message transmission.  Unfortunately, that callback was only being
applied to swatch-producer-aws.  This PR applies the callback to all Quarkus
applications.

### Steps
1.  On `main`:
    ```
    ./gradlew :swatch-contracts:quarkusDev &
    http :9000/metrics | grep splunk
    ```

1.  Repeat on this branch
1.  Deploy this branch again like so:
    ```
    ENABLE_SPLUNK_HEC=true \
    SPLUNK_HEC_URL=https://localhost:8088 \
    SPLUNK_HEC_TOKEN=foobar \
    ./gradlew :swatch-contracts:quarkusDev &
    sleep 20
    http :9000/metrics | grep splunk
    ```
    The non-existent URL will result in a bunch of message failures (although
    you will need to wait a bit for them to register).  The counter will get
    incremented.

### Verification
1. On `main` you will not see the metric listed.  On this branch it'll be there.
1. For the last step, the counter will be non-zero.


Sgitario (approved):
should we also add t

PR #3824, opened 2024-10-03
SWATCH-2958: Set billing producer aws when role is moa

Sgitario:
Jira issue: SWATCH-2958

## Description
If the product label on the telemetry feed is (moa-hostedcontrolplane|moa), then force the billing_marketplace to aws.

## Testing
Added junit test to reproduce the issue.

### DIFF
--- a/swatch-model-events/schemas/event.yaml
+++ b/swatch-model-events/schemas/event.yaml
@@ -186,6 +186,7 @@ properties:
       - addon-open-data-hub
       - BASILISK
       - moa-hostedcontrolplane
+      - moa
     required: false
   sla:
     description: Service level for the subject.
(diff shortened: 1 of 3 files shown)


swatch-model-events/schemas/event.yaml
```diff
@@ -186,6 +186,7 @@ properties:
       - addon-open-data-hub
       - BASILISK
       - moa-hostedcontrolplane
+      - moa
```
Sgitario, line 189: As far as I see, we're using an enum Role to parse from string, and then to parse it back to String. 
I think we should remove this Role enum and directly use a string here. wdyt ? @kahowell

kflahert: Thinking we should go with a similar idea as what we did with ProductId. We created a class that only allows values that come directly from our config: https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/swatch-product-configuration/src/main/java/com/redhat/swatch/configuration/registry/ProductId.java

Not required for this fix so I'm approving and can look again if we make updates.

kflahert (approved):

swatch-model-events/schemas/event.yaml
Sgitario: I've created https://issues.redhat.com/browse/SWATCH-3002 for futher discussion.

Aurobinda55:
Looks good to me, 
Added moa-hostedcontrolplane|moa events with billing_provider="". 
swatch replaced billing_provider=aws

Aurobinda55 (approved):

PR #3834, opened 2024-10-08
SWATCH-2951: Allow to filter by metrics in the Instances Table API

Sgitario:
Jira issue: SWATCH-2951

## Description
The payg and non payg views are aggregating all the metrics into a single row in jsonb format. And the problem is that we can't easily filter and order using these metrics because it is in jsonb.

So, instead of dealing with the metrics jsonb column, we're now adding a left join to the measurements table (for payg is "instance_monthly_totals" and for non payg is "instance_measurements") - note about performance: that these tables are indexed.

Therefore, we're now using this relationship to filter and also to order the result set. These relationships will not be used when mapping the query result set to the DTO.

Changes
- Removed the distinct in the query since it wasn't necessary (the distinct is done in the view).
- Implement the order using the Hibernate Criteria API because the Spring Specification API can only order by existing fields in the entity (and we now want to also order by programmatically added columns)
- Don't use the Spring Pageable API since the sort is not supported anylonger (if provided, then the programmatically order by is ignored)
- 
## Testing
Added a JUnit test to mimic the scenario.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resource/api/v1/InstancesResource.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/api/v1/InstancesResource.java
@@ -22,10 +22,8 @@ package org.candlepin.subscriptions.resource.api.v1;
 
 import static java.util.Optional.ofNullable;
 
-import com.google.common.collect.ImmutableMap;
 import com.redhat.swatch.configuration.registry.MetricId;
 import com.redhat.swatch.configuration.registry.ProductId;
-import com.redhat.swatch.configuration.registry.SubscriptionDefinition;
 import com.redhat.swatch.configuration.registry.Variant;
 import com.redhat.swatch.configuration.util.MetricIdUtils;
 import jakarta.ws.rs.BadRequestException;
@@ -37,7 +35,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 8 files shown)


kflahert (approved):

mmusil (commented):
Tested extensively in EE with these products:
- rhel-for-x86-els-payg-addon
- rhel-for-x86-els-payg
- rosa
- rhods
- OpenShift-metrics
- OpenShift Container Platform
- RHEL for x86

I tried sorting with all possible measurements these products have, both asc and desc direction. PAYGO products and annual OCP work even when passing the metric in a camel case. 

Tested also unpaginated data, with 100 items per page and paginated 10 items per page with a data set of more than 200 records. 

Secondary ascending sort by "id" in case there are multiple entries with the same value of measurement seems to work as expected.

RHEL for x86 sorting by metric is still case-sensitive. It **would be nice** and appreciated if it was updated as discussed over the chat. 
`sockets` does not work, while `Sockets` does.

Sgitario:
> RHEL for x86 sorting by metric is still case-sensitive. It **would be nice** and appreciated if it was updated as discussed over the chat. `sockets` does not work, while `Sockets` does.

@mmusil PR updated by doing the "sort" query parameter case insenstive.

mmusil (approved):
I finally got the image on quay and tested the latest change in EE.
RHEL for x86 case insensitivity verified.

Sgitario:
> I finally got the image on quay and tested the latest change in EE. RHEL for x86 case insensitivity verified.

Great! I will wait for the CI gets fixed before merging this one.

Sgitario:
I see many test failures, but I doubt it's related to these changes, @mmusil can you confirm this to proceed with the merge?

mmusil:
> I see many test failures, but I doubt it's related to these changes, @mmusil can you confirm this to proceed with the merge?

The failures do not seem to be unique to the build for this PR, except for perhaps this one https://ibutsu.insights.corp.redhat.com/results/65c3a090-0b7c-4e2a-9ed3-f61942baea72
But I cannot say for sure if it was just a one-time random failure, or something you changed in your latest force push that caused it. In previous builds for this PR, this test was passing. And since it is a Kafka event test parametrized for both `rhel-for-x86-els-payg` and `rhel-for-x86-els-payg-addon` - for which it passed, I'm more inclined to say it is unrelated and we should be good to go.

Sgitario:
> ... or something you changed in your latest force push that caused it. 

Didn't change anything, but I only rebased the branch. 
Then, I'm going to merge this pull request. Thanks!

PR #3829, opened 2024-10-04
SWATCH-2843: Implement logic for subscription table endpoints to swatch-contracts

wottop:
Jira issue: SWATCH-2843

## Description

Moving this endpoint to swatch-contracts while allowing it to still work in the monolith

## Testing

IQE Test MR: 
https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/900

### Setup
To check locally from this branch with the iqe pr
1. Modify iqe_rhsm_subscriptions/conf/rhsm_subscriptions.default.yaml lines 273, 498, and 502 to match your port of choice for swatch-contracts [8003 for this example]
2. Start swatch-contracts
`SERVER_PORT=8003 QUARKUS_MANAGEMENT_PORT=9003 ./gradlew :swatch-contracts:quarkusDev`
3. Start the monolith (for opt in call)
` DEV_MODE=true SERVER_PORT=8000 SPRING_PROFILES_ACTIVE=worker,kafka-queue,api,capacity-ingress,kafka-queue,rhsm-conduit ./gradlew clean :bootRun`
4. Create an entry in the org_config table for an org
`insert into org_config (org_id, opt_in_type,created, updated) values ('5345257','API','2024-06-21T12:50:39.496954Z','2024-06-21T12:50:39.496954Z')`
5. Run the opt-in
`http :8000/api/rhsm-subscriptions/v1/opt-in x-rh-identity:$(echo '{"identity":{"internal":{"org_id":"5345257"}}}' | base64 -w0)
`

### Steps

1. Run from command line to exercise endpoint [either v1 or v2 will work]
`http GET ':8003/api/rhsm-subscriptions/v2/subscriptions/products/RHEL for x86' x-rh-identity:$(echo '{"identity":{"type":"User","org_id":"5345257"}}' | base64 -w0)`
2. Run test command from iqe
`iqe tests plugin rhsm_subscriptions  -k test_subscription_table_api_subscription_type_contracts`

### Verification

1. for the command line, you will see the expected json that you would get using the same command against port 8000
`http GET ':8000/api/rhsm-subscriptions/v2/subscriptions/products/RHEL for x86' x-rh-identity:$(echo '{"identity":{"internal":{"org_id":"5345257"}}}' | base64 -w0)`
2. The 2 iqe tests will succeed.

PR #3841, opened 2024-10-10
SWATCH-2842: Implement logic for capacity endpoints into swatch-contracts

wottop:
PENDING

Jira issue: SWATCH-2843

## Testing

IQE Test MR: 

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.


wottop:
This task is dependent on  https://github.com/RedHatInsights/rhsm-subscriptions/pull/3829 and will likely need updates to match the ones made there.

liwalker-rh (approved):
Testing looks good, PR check failures are not related. Will merge related IQE MR once this has been merged.

Sgitario (commented):
Only a minor thing about removing a dependency in swatch-contracts/build.gradle.
Question: when the remaining capacity endpoints will be removed from the monolith? Perhaps, as part of https://issues.redhat.com/browse/SWATCH-2296?

wottop:
I only saw ' /v1/capacity/products/{product_id}/{metric_id}'. Did I miss one?

swatch-contracts/src/main/resources/META-INF/openapi.yaml
```diff
@@ -1120,6 +1120,118 @@ paths:
       security:
         - customer: [ ]
 
+  /api/rhsm-subscriptions/v1/capacity/products/{product_id}/{metric_id}:
```
Sgitario, line 1123:
```suggestion
  /api/swatch-contracts/v1/capacity/products/{product_id}/{metric_id}:
```

Other paths that were recently ported needs fixing as well:
- /api/rhsm-subscriptions/v2/subscriptions/products/{product_id}
- /api/rhsm-subscriptions/v1/subscriptions/products/{product_id}

Sgitario (changes requested):

Sgitario:
> I only see ' /v1/capacity/products/{product_id}/{metric_id}'. Did I miss one?

There are now two endpoints that return the same data:
- the existing one, the one you ported the API from: https://github.com/RedHatInsights/rhsm-subscriptions/blob/9db5faf77becc33b43385b3ef1180afe8fba07e6/api/rhsm-subscriptions-api-v1-spec.yaml#L243
- the new one, the one you created: https://github.com/RedHatInsights/rhsm-subscriptions/pull/3841/files#diff-8c90c15390db3f8b0ac0ed4a53ac07007c07fd322115ee5f8c804463f53ce357R1123

Btw, I've just realised that the path for the new one is wrong, it should be `/api/swatch-contracts/v1/capacity/products/{product_id}/{metric_id}` (I've just added a comment for this).

wottop:
As per @kahowell, we are keeping the path rhsm-subscription on purpose to make the transition easier on the systems that call it. If you can talk him out of it, I'll gladly change it.

The cards say to leave the old one, so I assume that the old ones will be removed in a later task.

Sgitario:
> As per @kahowell, we are keeping the path rhsm-subscription on purpose to make the transition easier on the systems that call it. If you can talk him out of it, I'll gladly change it.
> 
> The cards say to leave the old one, so I assume that the old ones will be removed in a later task.

Yes, I know. All I asked if what's the later task that will remove this endpoint...

wottop:
That is my understanding.

Sgitario (approved):

PR #3846, opened 2024-10-14
SWATCH-3025: Create swatch-common-export module

Sgitario:
Jira issue: SWATCH-3025

## Description
There is no porting functionality here, but only moving the export request handler/formatters logic into a common library that then will be reused by Quarkus and Spring Boot services (for now, only Spring Boot).

The new swatch-common-export module is framework agnostic. To achieve this, we use some delegates to operate with the actual export API and rbac API. This delegate must be provided by the Spring Boot/Quarkus services.

Captura de pantalla de 2024-10-14 17-00-56

## Testing
Only regression testing.
The existing logic is already covered by the existing junit tests ++ also via iqe tests.


Sgitario:
@kartikshahc @awood this PR has the changes we have discussed. Any of you is volunteer to take the DEV review?
Note that I renamed the ExportService to ExportRequestHandler and ExportConsumer to ExportRequestConsumer to better clarity.

src/main/java/org/candlepin/subscriptions/tally/export/InstancesDataExporterService.java
```diff
@@ -26,6 +26,11 @@
 
 import com.redhat.swatch.configuration.registry.MetricId;
 import com.redhat.swatch.configuration.registry.ProductId;
+import com.redhat.swatch.export.DataExporterService;
+import com.redhat.swatch.export.DataMapperService;
+import com.redhat.swatch.export.ExportServiceException;
+import com.redhat.swatch.export.ExportServiceRequest;
+// NOTE(khowell): this couples our export implementation to the v1 REST API
```
awood, line 33: This comment needs to move down to line 53 now I believe.  Above `import org.candlepin.subscriptions.utilization.api.v1.model.ReportCategory;`

swatch-common-export/src/main/java/com/redhat/swatch/export/ExportRequestHandler.java
```diff
@@ -18,82 +18,43 @@
...
 @Slf4j
-public class ExportSubscriptionListener extends SeekableKafkaConsumer {
+@AllArgsConstructor
+public class ExportRequestHandler {
 
   public static final String ADMIN_ROLE = ":*:*";
   public static final String SWATCH_APP = "subscriptions";
   public static final String MISSING_PERMISSIONS = "Insufficient permission";
+  public static final int INTERNAL_ERROR = 500;
+  public static final int FORBIDDEN_STATUS = 403;
```
awood, line 41: How do you feel about dropping these and just using `HttpURLConnection.HTTP_FORBIDDEN`, etc?

awood (changes requested):
Looks great.  If you could just correct that one mislaid comment; otherwise, excellent.

src/main/java/org/candlepin/subscriptions/tally/export/InstancesDataExporterService.java
Sgitario: yes, thanks for spotting.

swatch-common-export/src/main/java/com/redhat/swatch/export/ExportRequestHandler.java
Sgitario: +1

Sgitario:
@awood I've already updated the PR after your comments

Sgitario:
Merging after applying the suggested changes. Failures are unrelated.

PR #3847, opened 2024-10-14
SWATCH-3027: Create export-client for Quarkus services

Sgitario:
Jira issue: SWATCH-3027

## Description
There is no functional change here, but necessary for SWATCH-2980.

## Testing
Only regression testing. No functional change here.


Sgitario:
Self-merging this. Failures are unrelated.

PR #3857, opened 2024-10-15
Skip azure marketplace message when azureDimension is null

kahowell:


kahowell:
self-merging since hotfixing

PR #3864, opened 2024-10-17
SWATCH-2980: Port export subscriptions to swatch contracts service

Sgitario:
Jira issue: SWATCH-2980
Depends on https://github.com/RedHatInsights/rhsm-subscriptions/pull/3866

## Description
The export of subscriptions was implemented as part of SWATCH-2021 and SWATCH-2023 (epic SWATCH-86).

Since the idea is to delete the subscription sync service, we need to also port this functionality to the swatch contracts.

Important changes:
- the Quarkus REST-Client does not support having multiple Content-Type, so I changed the export API to use instead of using application/json and application/csv, just use "application/octet-stream". 
- I had to disable/comment some tests because we need a Postgresql instance with all the migrations (the migration scripts from the monolith and the ones from the swatch-contracts). This needs to be addressed in a separated ticket which is already reported in SWATCH-2820

## Testing
Regression testing only.

Few notes about how to easily test the changes using the IQE tooling:

```
app.rhsm_subscriptions.rest_client.opt_in_api.put_opt_in_config()
app.user.auth={"username":"xxx"}
application=app
sku_filters={"RH00004": {}}
export_format="json"
product_id = "RHEL for x86"

## SUBSCRIPTIONS


# Typecasting key from the sku_filters dict param to string
sku_id = list(sku_filters.keys())[0]
application.rhsm_subscriptions.add_sku(
    sku=sku_id, quantity=1, count=1, wait=True, product_id=product_id
)
# Retrieve the sku_record only after adding the SKU, if it was missing
sku_record = application.rhsm_subscriptions.get_sku_from_subscription_table(sku_id)
subscription_numbers = []
# Gather all the subscriptions ID in a list
for record in sku_record["subscriptions"]:
    subscription_numbers.append(record["number"])
# Send a export request
export_filters = sku_filters[sku_id]
export_filters["product_id"] = product_id
export = application.rhsm_subscriptions.create_swatch_export_request(
    filters=export_filters, format=export_format
)
# Download export request and convert to json
data, _ = application.rhsm_subscriptions.download_export_to_json(export.id)

## INSTANCES

host = application.rhsm_subscriptions.create_physical_rhel(sockets=8, cores=2)
application.rhsm_subscriptions.sync_swatch(wait=True, hostname=host["display_name"])
system_table = application.rhsm_subscriptions.search_system_table(
    product_id, display_name_contains=host["display_name"]
)
# Send a export request
export = application.rhsm_subscriptions.create_swatch_export_request(
    filters=export_filters, resource="instances", format=export_format
)
data, meta = application.rhsm_subscriptions.download_export_to_json(export.id)
```


swatch-contracts/src/main/resources/application.properties
```diff
@@ -217,6 +217,12 @@ quarkus.rest-client."com.redhat.swatch.clients.rbac.api.resources.AccessApi".tru
 quarkus.rest-client."com.redhat.swatch.clients.rbac.api.resources.AccessApi".trust-store-password=${clowder.endpoints.rbac-service.trust-store-password}
 quarkus.rest-client."com.redhat.swatch.clients.rbac.api.resources.AccessApi".trust-store-type=${clowder.endpoints.rbac-service.trust-store-type}
 
+quarkus.rest-client."com.redhat.swatch.clients.export.api.resources.ExportApi".url=${clowder.private-endpoints.export-service-service.url:http://localhost:10000}
```
awood, line 220: Do you know why it's `export-service-service` instead of just `export-service`?

swatch-contracts/src/main/resources/db/202410171500-move-subscription-capacity-view_h2_only.xml
```diff
@@ -0,0 +1,71 @@
...
+                        inner join offering o on s.sku=o.sku
+                        inner join sku_product_tag spt on s.sku = spt.sku
+                        left join subscription_measurements sm on s.subscription_id = sm.subscription_id and s.start_date=sm.start_date
+                 where s.start_date <= now() and (s.end_date is null or s.end_date >= now())
+                 group by s.subscription_id, s.subscription_number, s.sku, o.has_unlimited_usage, o.description, o.sla, o.usage , s.org_id, s.billing_provider, s.billing_provider_id, s.billing_account_id, spt.product_tag,s.quantity, sm.metric_id, sm.value, sm.measurement_type, s.start_date, s.end_date
+                 order by s.subscription_id asc
+      ]]>
+    </createView>
+    <rollback>
+      <createView
```
awood, line 41: What actually changes in the view?  I'm kind of doing a visual diff between the `createView` and the `rollback` sections.  Or is this more of a logistical thing to move the view to be under swatch-contract's logical ownership?

awood (approved):

swatch-contracts/src/main/resources/application.properties
Sgitario: This is because the quarkus clowder config source utility composes the key using the "{app}-{service}". The app name is "export-service" and the service is "service" as well.

swatch-contracts/src/main/resources/db/202410171500-move-subscription-capacity-view_h2_only.xml
Sgitario: Yes, I didn't change anything here, but only copy these migrations here, so at least we can use it for testing.

PR #3866, opened 2024-10-18
SWATCH-3034: Fix the Quarkus AccessApi generated client interface

Sgitario:
Jira issue: SWATCH-3034

## Description
Changes:
- Join the PermissionsAPI to the AccessAPI (which is related). This way the AccessAPI gets fixed because the `@Path` annotation is now placed at method level
- Added a documentation item in the Quarkus migration section explaining this. 

This change has no effects because we're not using the PermissionsAPI.

## Testing
Regression testing. 
I applied this change in https://github.com/RedHatInsights/rhsm-subscriptions/pull/3864 and it fixed the CI.


Sgitario:
Self-merging this since it's required for https://github.com/RedHatInsights/rhsm-subscriptions/pull/3864.

PR #3872, opened 2024-10-21
SWATCH-3036: Use separated transactions when processing retry remittance

Sgitario:
Jira issue: SWATCH-3036

## Description
Changes:
- Add an index for billable_usage_remittance.retry_after column
- Query the remittances to be retried in batches and single transactions
- Perform update operations in single transactions (after sending the remittance to kafka)
- If the send operation fails, we need to rollback the remittance to its previous state. 

## Testing
Added unit test to reproduce the issue.


kartikshahc (approved):
I don’t foresee any conflicts with previous remittance records. This is because we merge the detached entity first before generating the message, and the status is updated by the Kafka topic in the next hourly cycle.
The pagination approach is a great idea since that will prevent memory issues by processing records in batches of 1024.
Also as we discussed, the records are updated before being sent to the Kafka topic, and if a Exception occurs, we revert the records to their previous state.
Putting QE label for long running test.

kahowell (approved):

PR #3873, opened 2024-10-21
Explicitly set version of commons-io to secure release

wottop:
## Description

The version of commons-io is not explicitly set in the spring boot platform which causes us to use a version that has a security issue

## Testing
View quay.io for advisories for this branch.

### Verification
The advisory for GHSA-78wr-2p64-hpwj should no longer show


Sgitario (changes requested):
We can't guarantee that other dependencies that need the commons-io will continue working fine if we change the commons-io version.

As far I see, the commons-io dependency is used in two places:

- the spring-kafka-test > zookeeper

This is a test dependency, so if tests are passing, it's ok to explicitly change the commons-io version. I also checked whether there is a new version of zookeeper that uses commons-io > 2.14 but there is none. 

- the spring resteasy-multipart-provider > apache-mime4j-dom

This is used by Spring Boot clients like the Export API and others. 
And I also checked whether there is a new version of apache-mime4j-dom without this vulnerability without luck. 

Note that I also tried to exclude the commons-io dependency from the above two dependencies and all the tests also passed. But this is equally risky than upgrading the version but better in the sense that if we don't use this dependency, we don't need to maintain it. 

Therefore, we should contact [the apache-mime4j-dom security team](https://james.apache.org/) for this vulnerability and/or to ask for the compatibility of the commons-io 2.16 with its latest version or ask if we can fully exclude the commons-io dependency.

wottop:
FTR: https://github.com/apache/james-project/pull/2466

chibenwa:
MIME4J do not use XMLStreamParser, you shall be good, but I'll anyway update the dependency asap so that it get's upgraded in next release.

wottop:
Because the update was made in the james-project, we will not need to keep this specification in ours. This MR will be closed and we will see this security warning until the commons-io update lands in the packages we are using.

PR #3879, opened 2024-10-22
SWATCH-3036: Fix pagination to process remittances retries

Sgitario:
Jira issue: SWATCH-3036

## Description
When asking for next pages, since at the same time, we're also updating the remittances to clear the retry_after, we won't be consuming all the remittances. 

To address this issue, we'll always asking for the first page which eventually reach zero. Moreover, we will first count the number of remittances and won't process more than this count (this is to avoid edge conditions and process new usages that might lead in endless

## Testing
Added a test to reproduce the pagination issue.


TrayvonMcKnight (approved):
One of the CI test failures is unrelated - hosts issues. 

The other two failures, the remittance is in pending state but I don't think those failures are impacted by these changes.

PR #3921, opened 2024-11-06
Fix swatch-contracts test error due to wrong property when stub exports

Sgitario:
No JIRA. 
The ExportServiceWireMockResource was using a wrong property and hence the wiremock was not receiving requests.


Sgitario:
Self-merging to fix main.

PR #3924, opened 2024-11-07
SWATCH-2936: Prevent retries when handling export requests

Sgitario:
Jira issue: SWATCH-2936

## Description
Log an error trace when this happens when the export request thas has failed. Change the log warn trace when the request does not contain the needed data.  

## Testing
Added unit tests to cover these scenarios.


mstead (approved):
Looks good to me.

PR #3928, opened 2024-11-08
SWATCH-3066: Log traces for "com.redhat.swatch" in spring services

Sgitario:
Jira issue: SWATCH-3066

## Description
When investaging the random failure of the test "test_export_instances_payg_product_filters" in SWATCH-3066, I found that we're only seeing the WARN traces when handling the export requests. 

This is because we moved the export handling into a separated module that uses the package "com.redhat.swatch" and the spring boot services are only configured to print the info traces for the package "org.candlepin".

For the Quarkus services, this is already working fine because it's configured as:

```
quarkus.log.level=${LOGGING_LEVEL_ROOT}
quarkus.log.category."com.redhat.swatch".level=${LOGGING_LEVEL_COM_REDHAT_SWATCH}
```

After these changes, the package "com.redhat.swatch" should also be configured for the spring services to print the info traces and see them in splunk.

## Testing
IQE Test MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/939

### DIFF
--- a/swatch-api/deploy/clowdapp.yaml
+++ b/swatch-api/deploy/clowdapp.yaml
@@ -165,6 +165,8 @@ objects:
                 value: ${LOGGING_LEVEL_ROOT}
               - name: LOGGING_LEVEL_ORG_CANDLEPIN
                 value: ${LOGGING_LEVEL}
+              - name: LOGGING_LEVEL_COM_REDHAT_SWATCH
+                value: ${LOGGING_LEVEL}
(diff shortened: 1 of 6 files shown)


Sgitario:
Verified using the splunk query: `index=rh_rhsm namespace="ephemeral-r28mvn" "ExportRequestHandler"`. Before, there were no traces found.

Sgitario:
Self-merging since the MR is already merged.

PR #3930, opened 2024-11-08
SWATCH-3072: Better clowder truststore handling

awood:
Jira issue: SWATCH-3072

## Description
Handle unresolved Clowder truststore references better

Truststore handling for the Clowder configuration is complex.  The
Clowder json looks like:

```
"endpoints": [
  {
    "app": "foo",
    "name": "service",
    "hostname": "localhost",
    "tlsPort": 9443
  }
],
"tlsCAPath": "/tmp/ca.pem"
}
```

However, our code generally doesn't want just a bare PEM file.
Instead, we take that path and construct a temporary Java truststore (a
PKCS12 file) from it.  We pass the reference to that keystore along.

Following the pattern of clowder-quarkus-config-source, we also
configure the truststore information on a per-endpoint basis.  To do
that, *if and only if* the endpoint has a `tlsPort` key, then we
generate three synthetic properties: trust-store-path,
trust-store-password, and trust-store-type.  We populate those synthetic
properties with the information from the generated PKS12.

However, if an endpoint doesn't have a `tlsPort` defined on it or if the
`tlsCAPath` isn't specified, the ClowderJsonPropertyResolver can't
resolve the synthetic `clowder.endpoints.[...].trust-store-path`
property. The Spring property resolver behavior is to just return the
literal string instead.

This commit documents this behavior and modifies the CertInfoContributor
to better reflect what is actually going on with the configuration
handling.

### Setup
1. Create a clowder configuration file in the root of the project
    ```
    cat <<EOF > clowder-3072.json
    {
    "kafka": {
     "brokers": [
       {
         "hostname": "localhost",
         "port": 9092
       }
     ],
     "topics": []
    },
    "endpoints": [
      {
        "app": "rbac",
        "name": "service",
        "hostname": "localhost",
        "port": 9000,
        "tlsPort": 9443
      },
      {
        "app": "swatch-contracts",
        "name": "service",
        "hostname": "localhost",
        "tlsPort": 9443
      }
    ],
    "metricsPath": "/metrics",
    "metricsPort": 9000,
    "privatePort": 10000,
    "publicPort": 8000,
    "webPort": 8000,
    "tlsCAPath": "$(pwd)/swatch-core/src/test/resources/ca-test.pem"
    }
    EOF
    ```

### Steps
1. Start the application with Clowder configuration enabled
    ```
    FOO_SERVICE_HOST='f' FOO_SERVICE_PORT='f' ACG_CONFIG=$(pwd)/clowder-3072.json  ./gradlew :bootRun --args="--spring.profiles.active=kafka-queue,api,worker
    ```
1. Hit the `/info` endpoint
    ```
    http :9000/info
    ```

### Verification
1. For something that doesn't have `tlsPort` configured, you'll see
    ```
    "billableUsageClientProperties.truststore": {
        "Unresolved clowder config value: ${clowder.endpoints.swatch-billable-usage-service.trust-store-path}": {}
    },
    ```
1. And if `tlsPort` is configured, you'll see the info for the test cert.
    ```
    "rbacServiceProperties.truststore": {
      "cert-0": {
          "Distinguished Name": "CN=localhost,O=unknown,L=unknown,ST=unknown,C=XX",
          "Issuer Distinguished Name": "CN=localhost,O=unknown,L=unknown,ST=unknown,C=XX",
          "Not After": "2023-04-05T18:57:04Z",
          "SHA-1 Fingerprint": "960f13f186301c2418265ab937b0af589576c248",
          "Serial Number": "17232293264883780949"
      }
    },
    ```
    Matching
    ```
    % openssl x509 -in swatch-core/src/test/resources/ca-test.pem -noout -fingerprint
    SHA1 Fingerprint=96:0F:13:F1:86:30:1C:24:18:26:5A:B9:37:B0:AF:58:95:76:C2:48
    ```

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPropertySource.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPropertySource.java
@@ -44,16 +44,18 @@ import org.springframework.util.ClassUtils;
 import org.springframework.web.context.support.StandardServletEnvironment;
 
 /**
- * Property source that attempts to resolve properties against the Clowder JSON using JSON Path.
- * This class resolves any property starting with "clowder.". Everything after the "clowder." prefix
- * is evaluated by using the following patterns:
+ * Property source that attempts to resolve properties against the Clowder JSON. This class resolves
+ * any property starting with "clowder.". Everything after the "clowder." prefix is evaluated by
+ * using the following patterns:
  *
  * <ul>
  *   <li>clowder.kafka.brokers
  *   <li>clowder.kafka.brokers.(sasl.securityProtocol|sasl.mechanism|sasl.jaas.config|cacert|cacert.type)
  *   <li>clowder.kafka.topics.<topics[?].requestedName(*)>.name
- *   <li>clowder.endpoints.<endpoints[?].app>-<endpoints[?].name>.(url|trust-store-path|trust-store-password|trust-store-type).
- *       For example: "clowder.endpoints.index-service.url"
+ *   <li>clowder.endpoints.<endpoints[?].app>-<endpoints[?].name>.url For example:
+ *       "clowder.endpoints.index-service.url"
+ *   <li>clowder.endpoints.<endpoints[?].app>-<endpoints[?].name>.(trust-store-path|trust-store-password|trust-store-type)
+ *       are special synthetic properties that are generated based on the clowder.tlsCAPath
  *   <li>clowder.privateEndpoints.<privateEndpoints[?].app(*)>-<privateEndpoints[?].name(*)>.(url|trust-store-path|trust-store-password|trust-store-type).
  *       For example: "clowder.privateEndpoints.index-service.url"
  *   <li>When there is no matching rule from above, we navigate from (*) and extract the string
@@ -66,7 +68,7 @@ import org.springframework.web.context.support.StandardServletEnvironment;
  */
 @EqualsAndHashCode(callSuper = true)
 @Slf4j
-public class ClowderJsonPathPropertySource extends PropertySource<ClowderJson>
+public class ClowderJsonPropertySource extends PropertySource<ClowderJson>
     implements OriginLookup<String> {
 
   public static final String PROPERTY_SOURCE_NAME = "Clowder JSON";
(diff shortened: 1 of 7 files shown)


swatch-contracts/src/main/resources/application.properties
```diff
@@ -212,6 +212,8 @@ RBAC_ENDPOINT=${clowder.endpoints.rbac-service.url}
 rhsm-subscriptions.rbac-service.use-stub=${RHSM_RBAC_USE_STUB:false}
 rhsm-subscriptions.rbac-service.stub-permissions=${RHSM_RBAC_STUB_PERMISSIONS:subscriptions:*:*}
 # Note the keystore and truststore values must be prefixed with either "classpath:" or "file:"
+# The .trust-store-* properties are actually synthetic.  They don't appear in the cdappconfig.json file.  Instead,
+# clowder-quarkus-config-source creates references to the Java truststore file we create based off of tlsCaCert file.
```
Sgitario, line 216: I would not add the same comment in all the application properties since it might confuse more than help. 
The relevant point here is to use the trust-store provided by the clowder config source. What is mentioned in the comment is an implementation detail. 

Moreover, we can find the same information in https://github.com/RedHatInsights/rhsm-subscriptions/pull/3930/commits/bfc30d361244f3e13b1de07637fbf6e41423c085#diff-2e2cbfc13c89a5e8c259890f9ff8977aafcd33398a87740195f9e67e716fba39R59 which is the right place

swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPropertySource.java
```diff
@@ -44,16 +44,18 @@
 import org.springframework.web.context.support.StandardServletEnvironment;
 
 /**
- * Property source that attempts to resolve properties against the Clowder JSON using JSON Path.
- * This class resolves any property starting with "clowder.". Everything after the "clowder." prefix
- * is evaluated by using the following patterns:
+ * Property source that attempts to resolve properties against the Clowder JSON This class resolves
```
Sgitario, line 47:
```suggestion
 * Property source that attempts to resolve properties against the Clowder JSON. This class resolves
```

Sgitario (changes requested):

swatch-contracts/src/main/resources/application.properties
awood: My reasoning for putting that comment in the config files themselves was that `clowder.endpoints...trust-store-path` will never actually appear in the `cdappconfig.json`.  So if someone doesn't know that we generate those properties, when they are integrating a new endpoint they'll find it very confusing.  I wanted to save someone from getting into a "Why isn't my endpoint getting a trust-store-path from Clowder!?".  And on top of that, the Clowder team would have no idea what the developer is talking about since that property is unknown to them.

What do you think?

Sgitario: Yes, I see the intention is good to prevent headaches or misunderstandings in the future. However my main concern is that if for some reason this internally changes in the future (it strongly depends to how the Quarkus Clowder config source works plus the cdappconfig.json is structured), we will need to maintain this comment in too many places. 

What about moving this comment into the CONTRIBUTING.md document instead?

Sgitario (approved):

PR #3955, opened 2024-11-18
SWATCH-3108: Use feeds name when openshift cluster uuid does not exist

Sgitario:
Jira issue: SWATCH-3108

## Description
Before these changes, when the system was not an openshift cluster, we were not setting the display name.  When sending the message without display name, the inventory service in [here](https://github.com/RedHatInsights/insights-host-inventory/blob/b4b3f1d6d6ad79cc5ad93ac09f5bbe4df684e232/app/models.py#L97) uses the instance ID as display name.

To prevent the inventory service to use the instance ID as display name, conduit service needs to populate the display name always regardless if it's an openshift system.

## Testing
Added a test to reproduce the issue.

### DIFF
--- a/swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
+++ b/swatch-system-conduit/src/main/java/org/candlepin/subscriptions/conduit/InventoryController.java
@@ -183,11 +183,7 @@ public class InventoryController {
     log.debug("Received message from consumer: {}", consumer);
     ConduitFacts facts = new ConduitFacts();
     facts.setOrgId(consumer.getOrgId());
-    String clusterUuid = rhsmFacts.get(OPENSHIFT_CLUSTER_UUID);
-    // NOTE future displayName logic could consider more facts here
-    if (clusterUuid != null) {
-      facts.setDisplayName(clusterUuid);
-    }
+    extractDisplayName(consumer, rhsmFacts, facts);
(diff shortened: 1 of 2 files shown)


Sgitario:
@kahowell since we're already discussed these changes internally, would you mind taking a look into these changes?

Aurobinda55:
verification steps - 

- If the OpenShift cluster contains a UUID, the display name == cluster UUID.
- If the OpenShift cluster does not contain a UUID, the display name ==  user-provided name.
- For other types of hosts, we observed that currently, when the name field is sent, the conduit service ignores it and sets displayName to null. Once the PR is implemented, it will detect the name field and set display_name to the provided name. Further validation will be conducted once the PR is merged into the stage.

Sgitario:
> verification steps -
> 
> * If the OpenShift cluster contains a UUID, the display name == cluster UUID.
> * If the OpenShift cluster does not contain a UUID, the display name ==  user-provided name.
> * For other types of hosts, we observed that currently, when the name field is sent, the conduit service ignores it and sets displayName to null. Once the PR is implemented, it will detect the name field and set display_name to the provided name. Further validation will be conducted once the PR is merged into the stage.

@Aurobinda55 just checked that when posting new consumers using `app.rhsm_subscriptions.create_physical_rhel()`, will create a new record in the HBI database with this host with a display name. 

So, when we sync the organization using conduit service and we send the message without display name, this won't be updated and hence will keep the previous display name.

PR #3964, opened 2024-11-19
Configure Jenkins pipelines to use kubedock latest image

Sgitario:
Jira issue: No JIRA

## Description
Use latest tag for kubedock image. 

## Testing
Only regression testing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,8 @@ kind: Pod
 spec:
   containers:
     - name: kubedock
-      image: quay.io/cloudservices/kubedock:f8452a8
+      image: quay.io/cloudservices/kubedock:latest
+      imagePullPolicy: Always
       tty: true
       args:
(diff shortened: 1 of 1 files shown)


kahowell:
I'm fine with this given:

> if you omit the `imagePullPolicy` field, and the tag for the container image is `:latest`, `imagePullPolicy` is automatically set to `Always`

per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

Sgitario:
> I'm fine with this given:
> 
> > if you omit the `imagePullPolicy` field, and the tag for the container image is `:latest`, `imagePullPolicy` is automatically set to `Always`
> 
> per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

let's set the imagePullPolicy to always for better clarity.

PR #3966, opened 2024-11-19
Allow response content to be json formatted

wottop:
## Description

In the Prometheus metering endpoint, we will benefit from the response content being JSON formatted instead of plain text. This will allow built-in JSON parsers to be used.

## Testing

IQE Test MR: 

### Setup

1. Startup one of the Quarkus apps locally
`./gradlew :swatch-producer-aws:quarkusDev`

### Steps

1. Run against the Prometheus metrics endpoint
`http :9000/metrics Accept:application/json`

### Verification

1. The response should be in JSON format
2. You should get the plain text response if the Accept header is left off.


Aurobinda55:
Deployed EE against this branch and checked the above URL with **Accept:application/json** and got a response in json format. Again, checked the same URL without **Accept:application/json** and got a response in plain text format.

PR #3976, opened 2024-11-21
SWATCH-3106: Switch to rosa:cluster:vcpu_hours

kahowell:
`hostedcluster:hypershift_cluster_vcpus:vcpu_hours` is not valid for classic clusters. `rosa:cluster:vcpu_hours` is defined as `hostedcluster:hypershift_cluster_vcpus:vcpu_hours` if it exists, otherwise `cluster:usage:workload:capacity_virtual_cpu_hours`

See the definition at https://github.com/Nikokolas3270/telemeter/blob/086530ffb35ecadbc986c0fca465f362c1bfdcd3/jsonnet/telemeter/rules.libsonnet#L218-L221


kahowell:
Self-merging this config change

PR #4005, opened 2024-12-03
SWATCH-3128: Disabling Kafka client config providers

Sgitario:
Jira issue: SWATCH-3128

## Description
I confirmed that we're already using kafka-clients >=3.8.0, so we need to explicitly set the JVM system property "org.apache.kafka.automatic.config.providers=none" to avoid this security risk. 

## Tests
Only regression testing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,5 +62,9 @@ COPY --from=0 /stage/LICENSE /licenses/
 RUN chmod -R g=u /deployments
 
 USER default
+# Custom JVM properties
+## Fix CVE-2024-31141: Disabling Kafka client config providers
+ENV INTERNAL_OPTS_APPEND="-Dorg.apache.kafka.automatic.config.providers=none"
+ENV JAVA_OPTS_APPEND="$INTERNAL_OPTS_APPEND $USER_OPTS_APPEND"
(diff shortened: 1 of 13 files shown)


Dockerfile
```diff
@@ -62,5 +62,8 @@ COPY --from=0 /stage/LICENSE /licenses/
 RUN chmod -R g=u /deployments
 
 USER default
+# Custom JVM properties
+## Fix CVE-2024-31141: Disabling Kafka client config providers
+ENV JAVA_OPTS_APPEND="-Dorg.apache.kafka.automatic.config.providers=none"
```
awood, line 67: Did this work in an ephemeral deployment scenario?  I ask because we define `JAVA_OPTS_APPEND` in many of our [clowdapp.yaml](https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/swatch-api/deploy/clowdapp.yaml#L152-L153) files and my presumption is that what's in the `clowdapp.yaml` file just going to clobber the value set in the image.

Sgitario: Well spotted. So, we can always put this property in place, I added a new USER_OPTS_APPEND to quarkus and spring boot services where in spring boot services, this is used to configure the splunk agent.

Sgitario:
Merging since it's needed for another task.

PR #4031, opened 2024-12-10
SWATCH-3186: Fix null pointer exception when tallying snapshots

Sgitario:
Jira issue: SWATCH-3186

## Description
From the Peeking Collections javadocs, the `.peek()` method might return null values, so we need to prevent the null pointers to happen.

Digging further into this issue, the root cause is that when peeking elements, we're also consuming the inner iterator without checking the iterator.hasNext() method: https://github.com/google/guava/blob/3644d6155fdeca4c97d51329b66d1272d2905735/guava/src/com/google/common/collect/Iterators.java#L1220

So, it sounds like a bug in the peeking collection.

## Testing
Added JUnit test to reproduce scenario.
To reproduce exactly the same issue than in production, you would need to load the HBI and Swatch data of a given organization (more details in the JIRA ticket) and run the tally snapshots:

```
http PUT ":8000/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots/<ORG>" x-rh-swatch-psk:placeholder
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
@@ -189,7 +189,8 @@ public class InventorySwatchDataCollator {
   private OrgHostsData trackActiveHypervisor(
       String hypervisorUuid, PeekingIterator<String> activeSubmanIds) {
     log.debug("Active hypervisorUuid={}", hypervisorUuid);
-    while (activeSubmanIds.hasNext() && activeSubmanIds.peek().compareTo(hypervisorUuid) < 0) {
+    while (activeSubmanIds.hasNext()
+        && compareStrings(activeSubmanIds.peek(), hypervisorUuid) < 0) {
       // discard any subman IDs less than the specified hypervisor UUID
       activeSubmanIds.next();
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
```diff
@@ -205,6 +206,16 @@ private OrgHostsData trackActiveHypervisor(
...
 
+  private int compareStrings(String a, String b) {
+    if (a == null) {
+      return 1;
+    } else if (b == null) {
+      return -1;
+    }
+
+    return a.compareTo(b);
+  }
```
awood, line 217: Would

```
return Ordering.natural().nullsLast().compare(a, b);
```

work here?  The [Ordering](https://github.com/google/guava/wiki/OrderingExplained) class in Guava has a lot of utility methods that we can use instead of writing our own.

Sgitario: For a small thing like this and taking into account that this logic will eventually be removed, I strongly prefer to keep using the basics instead of dealing about what the Ordering, natural, and nullsLast does. 
One example about using third parties libraries is this PR, we're using the PeekingIterator from Guava and had to spend some amount of time to understand why the hasNext was true, but the peek element was empty...

diegomaranhao:
It looks good to me.
I've added comments on the [card](https://issues.redhat.com/browse/SWATCH-3186).

src/main/java/org/candlepin/subscriptions/tally/InventorySwatchDataCollator.java
awood: Sounds good.

awood (approved):

PR #4072, opened 2025-01-03
Fix build to work in any location and timezone

Sgitario:
## Description
There were only two places where we should set the timezone / locate format to fix the tests. This is only reproducible when building in a different time zone like in Spain.

## Testing
Only regression testing.

### DIFF
--- a/swatch-contracts/src/test/java/com/redhat/swatch/contract/service/SubscriptionSyncServiceTest.java
+++ b/swatch-contracts/src/test/java/com/redhat/swatch/contract/service/SubscriptionSyncServiceTest.java
@@ -53,6 +53,8 @@ import io.quarkus.test.InjectMock;
 import io.quarkus.test.junit.QuarkusTest;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.BadRequestException;
+
+import java.time.Clock;
 import java.time.OffsetDateTime;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


swatch-contracts/src/test/java/com/redhat/swatch/contract/service/SubscriptionSyncServiceTest.java
```diff
@@ -339,7 +341,7 @@ void shouldSyncSubscriptionsSkipSubIfTooExpired() {
   void shouldSyncSubscriptionsSkipSubIfTooFutureDated() {
     // When syncing an org's subs, don't sync subscriptions future-dated more than 2 months from now
     var dto = createDto("456", 10);
-    dto.setEffectiveStartDate(toEpochMillis(NOW.plusMonths(2).plusDays(1)));
```
lindseyburnett, line 342: Was it intentional to add `plusMinutes`?

Sgitario: Yes, when troubleshooting this failure in my local using the timezone from Spain, apart from the hours, I spot a minor difference of a few seconds due to slow executions (my laptop is much slower than the old one ... ) , that's why adding the minutes will make this test more stable.

lindseyburnett (approved):

lindseyburnett (approved):

PR #4089, opened 2025-01-09
CI: Use service account for Kubedock with port-forward permissions

Sgitario:


Sgitario:
These changes didn't work. I reported https://github.com/joyrex2001/kubedock/issues/122 looking for clarification.

PR #4090, opened 2025-01-09
CI: Add isolated job to validate format in GitHub actions

Sgitario:
## Description
This will help to have a quick feedback about the formatting and to better spot format issues. Context: https://redhat-internal.slack.com/archives/C01F7QFNATC/p1736361810374569

## Testing
Only regression testing.

## Verification
Checking the checks in PRs, we should now see two pipelines: the new one about running only the format validation and the previous one to validate the floorplan queries.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,9 +1,23 @@
-name: "FloorPlan Queries"
+name: "Build"
 on:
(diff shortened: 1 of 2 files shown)


awood (approved):
Nice!  Thanks, Jose.

PR #4092, opened 2025-01-09
Use ocp-c1 cloud for running pr checks

ntkathole:
Since new jenkins cluster default service account do not have enough permissions, this PR is to fall back to using old cluster in the meanwhile. 

Old cluster definition added in https://gitlab.cee.redhat.com/smqe/swatch-ci/-/merge_requests/332 

This change can be revered once Kubedock issues/122 or CCITJEN-2066 fixed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,6 +9,7 @@ pipeline {
     }
     agent {
         kubernetes {
+            cloud 'ocp-c1'
             label 'swatch-17-kubedock-2023-12-06' // this value + unique identifier becomes the pod name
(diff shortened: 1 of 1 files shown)


Sgitario:
@ntkathole why using this cloud and where is this value coming from?

ntkathole:
> @ntkathole why using this cloud and where is this value coming from?

This value is coming from https://gitlab.cee.redhat.com/smqe/swatch-ci/-/merge_requests/332
where we specified old cluster

Sgitario (approved):

PR #4093, opened 2025-01-09
Clean up configuration for Quarkus services in tests

Sgitario:
## Description
These changes will ensure that containers are not used when not needed and hence will speed up running tests.

## Testing
Only regression testing.


Sgitario:
Self-merging since this is simply a refactor for speeding up testing which are passing according to the CI.

PR #4094, opened 2025-01-09
Move swatchdog

kahowell:
Moved to https://github.com/RedHatInsights/swatchdog


awood (commented):
How do you feel about leaving/modifying the Poetry stuff so people can easily set up bonfire with just a `poetry install`?

lindseyburnett:
> How do you feel about leaving/modifying the Poetry stuff so people can easily set up bonfire with just a `poetry install`?

Migrating some slack discussion highlights for posterity:

We shouldn't maintain python packaging manifests in swatch itself. (We don't have any runtime python components, just scripts).

Recommendation to give `pipx install crc-bonfire` a try, as it's a more succinct version of

```
VENV_DIR=~/bonfire_venv
mkdir -p $VENV_DIR
python3 -m venv $VENV_DIR
. $VENV_DIR/bin/activate
pip install crc-bonfire
```

lindseyburnett (approved):

awood (approved):

PR #4103, opened 2025-01-13
SWATCH-3151: skip setting credential provider when using moto

kflahert:
Jira issue: SWATCH-3151

## Description

Skip setting aws credentials when moto is being used.


Sgitario (changes requested):
We're opening a dangerous door here that allow users to submit usages to other servers using these credentials. Plus, we're losing the ability to test the credentials matching in the iqe tests. 

Instead, we should configure moto to only allow requests for the expected aws seller account IDs credentials.

kflahert:
@Sgitario good point thanks! I'll fix

PR #4116, opened 2025-01-17
Configuration for SonarCloud rules

wottop:
## Description

This is the link to the Sonar rules used in the CI

## Testing

When using Intellij, you should see real-time Sonar reporting for .java files that match the CI.
It does require the SonarQube for IDE plugin.


lindseyburnett (approved):
Approving even with the IQE failure because it's unrelated to this PR.

PR #4127, opened 2025-01-23
SWATCH-3247: Add missing "Managed-nodes" metric when exporting instances

Sgitario:
Jira issue: SWATCH-3247

## Description
When exporting instances that use the "ansible-aap-managed" product, we were not exporting the configured Managed-nodes metric.

I've added a test, so we don't forget to add the mapping when adding a new metric in the future. Also, I covered the ansible export in our integration tests.

## Testing
MR test: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/1008

Added integration tests as part of the pull request changes.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/export/InstancesCsvDataMapperService.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/export/InstancesCsvDataMapperService.java
@@ -62,7 +62,8 @@ public class InstancesCsvDataMapperService implements DataMapperService<TallyIns
               InstancesExportCsvItem::setMeasurementStorageGibibyteMonths,
           MetricIdUtils.getTransferGibibytes(),
               InstancesExportCsvItem::setMeasurementTransferGibibytes,
-          MetricIdUtils.getVCpus(), InstancesExportCsvItem::setMeasurementVcpus);
+          MetricIdUtils.getVCpus(), InstancesExportCsvItem::setMeasurementVcpus,
+          MetricIdUtils.getManagedNodes(), InstancesExportCsvItem::setMeasurementManagedNodes);
 
   @Override
   public List<Object> mapDataItem(TallyInstanceView item, ExportServiceRequest request) {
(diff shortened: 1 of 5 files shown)


awood (approved):
This thought is outside the scope of this task, but looking at `MetricIdUtils`, it strikes me as being an issue that we have to define a method for each MetricId type (`get...`) when the actual metric IDs are determined dynamically at runtime when we load the configuration.  Plus, every call to one of those `get...` methods is running `MetricId.fromString` which does a lot of redundant work every time.

I'm wondering if we can have MetricId call `getAll` and then store the set, turn it into a map with the metric name as the key and build the `get...` methods with reflection or something (and therefore we never have to remember to add a `get...` to MetricIdUtils).  Regardless, I think it would make sense to keep a `getAll` set and then turn it into a map so we don't have to call `fromString` over and over.

What do you think @Sgitario?

kflahert (approved):

Sgitario:
> This thought is outside the scope of this task, but looking at `MetricIdUtils`, it strikes me as being an issue that we have to define a method for each MetricId type (`get...`) when the actual metric IDs are determined dynamically at runtime when we load the configuration. Plus, every call to one of those `get...` methods is running `MetricId.fromString` which does a lot of redundant work every time.
> 
> I'm wondering if we can have MetricId call `getAll` and then store the set, turn it into a map with the metric name as the key and build the `get...` methods with reflection or something (and therefore we never have to remember to add a `get...` to MetricIdUtils). Regardless, I think it would make sense to keep a `getAll` set and then turn it into a map so we don't have to call `fromString` over and over.
> 
> What do you think @Sgitario?

Keeping these metrics as a constant in memory sounds good to me to avoid loading the whole configuration everytime we get a metric. 
Also, I thought this utility was mainly used in tests, but I see that this is also used in many places in our logic.

liwalker-rh (approved):
Looks good testing in PR checks (failing billable usage tests are not related) and all the tests in the related IQE MR seem to be ready as well.

PR #4139, opened 2025-01-29
Add test helper pods to iqe proxy

lindseyburnett:
## Description

After following the guide for standing up the iqe proxy, I was encountering errors running my IQE tests that were due to not being able to connect to various services.  This PR:

- Updates the name of the wiremock service when it's deployed to match the naming convention of the other pods we deploy
- Updates the iqe proxy script so wiremock-service:8101 and swatch-mock-prometheus-test-service:9090 get their ports forwarded
- Adds a reminder to the docs/iqe-ee-proxy.md guide that you need to make sure all the parts and pieces are 

## Testing

Following the iqe-ee-proxy.md guide and then running this iqe test locally is successful

```
iqe tests plugin rhsm_subscriptions -m ephemeral -k test_verify_metering_operation_for_events_received_from_prometheus_for_openshift_products
```

### DIFF
--- a/docs/iqe-ee-proxy.md
+++ b/docs/iqe-ee-proxy.md
@@ -26,6 +26,15 @@ https://insights-qe.pages.redhat.com/iqe-core-docs/tutorial/part3.html#download-
 `oc login …`
 `bonfire namespace reserve`
 `bonfire deploy rhsm …`
+
+If necessary, don't forget to deploy the additional "helper" containers (e.g. wiremock, kafka-bridge, mock prometheus), since they're EE specific and defined outside of the rhsm clowdapp.  If you're using `--source=appsre`, or if you have them defined in your local `~/.config/bonfire/config.yaml`, they should be automatically deployed.
+
+Otherwise, here's how you can deploy them manually.
+
+`oc process -f ../rhsm-subscriptions/stub/wiremock.yaml | oc apply -f -`
+`oc process -f ../rhsm-subscriptions/kafka-bridge/deploy/template.yaml | oc apply -f -`
+`oc process -f ../rhsm-subscriptions/swatch-metrics/deploy/mock-prometheus-clowdapp.yaml | oc apply -f -`
+
 9. Start the proxy:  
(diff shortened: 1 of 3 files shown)


docs/iqe-ee-proxy.md
```diff
@@ -26,6 +26,12 @@ https://insights-qe.pages.redhat.com/iqe-core-docs/tutorial/part3.html#download-
 `oc login …`
 `bonfire namespace reserve`
 `bonfire deploy rhsm …`
+
+Don't forget to deploy the additional "helper" containers (e.g. wiremock, kafka-bridge, mock prometheus), since they're EE specific and defined outside of the rhsm clowdapp.
+`oc process -f ../rhsm-subscriptions/stub/wiremock.yaml | oc apply -f -`
+`oc process -f ../rhsm-subscriptions/kafka-bridge/deploy/template.yaml | oc apply -f -`
+`oc process -f ../rhsm-subscriptions/swatch-metrics/deploy/mock-prometheus-clowdapp.yaml | oc apply -f -`
```
Sgitario, line 33: I never encountered this issue because when I deploy rhsm, I use the ephemeral profile which already includes wiremock, kafka bridge and so on. 
But I acknowledge that this is really helpful when simply doing `bonfire deploy rhsm --no-remove-resources app:rhsm` (which uses the latest production image and templates)

Sgitario (approved):
Changes look good to me. Maybe I would add a note that the step to install the wiremock, kafka bridge and mock prometheus are only needed when not using the ephemeral profile.

docs/iqe-ee-proxy.md
lindseyburnett: Oh good point, I didn't even think about using the ephemeral profile.  I was just using that simple command.

barnabycourt: Where is the documentation on deploying rhsm to ephemeral using the ephemeral profile? There should probably be a link from the beginning of this doc to wherever that is (or include it in this doc) so that the requirements for the setup are linked together.

PR #4143, opened 2025-01-30
Update iqe ee proxy documentation to enable live logging

Sgitario:
## Testing
Follow the iqe ee proxy steps and when running a test directly from pycharm, you should see the logs as in:

Screenshot From 2025-01-30 11-49-57

Before these changes, when not adding the "--log-cli-level" argument, I only saw Passed or Failed :/


mstead (approved):
Verified this works. Thanks for this tip!

PR #4138, opened 2025-01-29
SWATCH-3147: Configure swatch-producer-aws to use wiremock proxy in ephemeral

Sgitario:
Jira issue: SWATCH-3147

Depends on https://github.com/RedHatInsights/wiremock-consoledot/pull/6

## Description
Configure swatch-producer-aws to use wiremock proxy in ephemeral.
By default, wiremock is configured to keep using the moto instance. 
Only when the usage request contains the text marker "_UseWiremock_" in any field, then the wiremock instance will be used.

## Testing
MR Test: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/1021

PR #4145, opened 2025-01-31
SWATCH-3147: Do not overwrite aws marketplace properties in EE

Sgitario:
Jira issue: SWATCH-3147

## Description
This fixes a regression of https://github.com/RedHatInsights/rhsm-subscriptions/pull/4138.
The problem is that the clowder's defined properties overwrite the properties from the application properties.
Let's configure everything (the prod, ee, stage properties) in a single place: the application.properties.

## Testing
CI pipelines must be green


kflahert (approved):

Sgitario:
I'm testing that with these changes and the ones from https://github.com/RedHatInsights/rhsm-subscriptions/pull/4146, we will have a CI green pass. If so, I will merge both PRs at the same time. Adding the on hold label.

PR #4227, opened 2025-02-28
SWATCH-2820: Independent database migrations

awood:
Jira issue: SWATCH-2820

## Description
SWATCH-2820: Run DB migrations independently

This patch creates a new clowdapp, swatch-database, that is now the
controlling entity for all database migrations.  Migrations from the
main project and swatch-contracts have been consolidated under the
swatch-database project.

The migrations are run on OpenShift via a `public static void main` that
does a bit of argument parsing and then invokes the Liquibase CLI.
The application starts, runs the migrations, then quits.  In Openshift,
we assign this container to a job and then for each ClowdApp that's
dependent on the database, we add an initContainer that uses the
["k8s-wait-for"](https://github.com/groundnuty/k8s-wait-for) script
packaged in the swatch-database container to block until the migration
job completes successfully.

The swatch-database ClowdApp now controls the Clowder provisioned
database so all the other ClowdApps needed to be updated to depend on
swatch-database.  The new database is named "swatch-database-db" rather
than "swatch-tally-db".

The gradle Liquibase plugin functionality has been removed.  Instead users
can invoke the `MigrationService` class directly like so:

```
./gradlew :swatch-database:run --args "status --log-level=WARN"
```

Everything in the `--args` argument is passed directly to Liquibase
**except for the first word (potentially)**.  `MigrationService` looks
at the first word and if it matches any of the migration contexts defined
in `MigrationService.Context`, it will only run the command against that
context.  The context name is stripped and the rest of the arguments are
passed on.  Otherwise, everything is passed on and MigrationService will
run Liquibase separately for every context.

For example:

```
./gradlew :swatch-database:run --args "core update --log-level=WARN"
```

will only run updates against `core` (which is the name I used for the main
schema).  

If the command is run with no arguments at all, the default mode is to run
`update` for each context.

Unfortunately, changing the relative path of the migration files causes
some issues.  The relative path is part of the changeset record.  So
changing the path make Liquibase think the migration has never run.
Potentially, we could modify the changelog table or set MARK_RAN on the
changesets that have already run, but ultimately I decided that involves
some risk.  The status quo is that contracts changesets live
under the "db" directory and core changesets live under "liquibase".
This state of affairs is unattractive but innocuous.

### Setup
1. Provision an ephemeral environment through `bonfire`.
1. Ensure that your local bonfire configuration, `~/.config/bonfire/config.yaml`
  defines `swatch-database`
    ```yaml
    - name: swatch-database
      host: local
      repo: /home/awood/devel/rhsm-subscriptions/swatch-database
      path: /deploy/clowdapp.yaml
      parameters:
        REPLICAS: 1
        swatch-database/IMAGE: quay.io/cloudservices/swatch-database
    ```
1. Build the images with `bin/build-images.sh`

### Steps
1. Deploy
    ```
    VER=$(git rev-parse --short=7 HEAD) ; \
    REPO=$(podman login quay.io --get-login) ; \
    bonfire deploy rhsm \
    --source=appsre \
    --ref-env insights-production \
    --optional-deps-method none \
    --frontends false \
    --no-remove-resources app:rhsm \
    --timeout 1000 \
    -i quay.io/$REPO/swatch-metrics=$VER \
    -i quay.io/$REPO/swatch-metrics-hbi=$VER \
    -i quay.io/$REPO/rhsm-subscriptions=$VER \
    -i quay.io/$REPO/rhsm-subscriptions=$VER \
    -i quay.io/$REPO/swatch-producer-aws=$VER \
    -i quay.io/$REPO/swatch-producer-azure=$VER \
    -i quay.io/$REPO/swatch-system-conduit=$VER \
    -i quay.io/$REPO/swatch-billable-usage=$VER \
    -i quay.io/$REPO/swatch-database=$VER \
    -p swatch-metrics/IMAGE=quay.io/$REPO/swatch-metrics \
    -p swatch-metrics-hbi/IMAGE=quay.io/$REPO/swatch-metrics-hbi \
    -p swatch-tally/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-api/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-database/IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-contracts/IMAGE=quay.io/$REPO/swatch-contracts \
    -p swatch-producer-aws/IMAGE=quay.io/$REPO/swatch-producer-aws \
    -p swatch-producer-azure/IMAGE=quay.io/$REPO/swatch-producer-azure \
    -p swatch-producer-red-hat-marketplace/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-subscription-sync/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-billable-usage/IMAGE=quay.io/$REPO/swatch-billable-usage \
    -p swatch-system-conduit/IMAGE=quay.io/$REPO/swatch-system-conduit
    -p swatch-api/WAIT_IMAGE=quay.io/$REPO/swatch-database \                                                                                                             
    -p swatch-producer-red-hat-marketplace/WAIT_IMAGE=quay.io/$REPO/swatch-database \                                                                                    
    -p swatch-subscription-sync/WAIT_IMAGE=quay.io/$REPO/swatch-database \                                                                                               
    -p swatch-tally/WAIT_IMAGE=quay.io/$REPO/swatch-database \                                                                                                           
    -p swatch-billable-usage/WAIT_IMAGE=quay.io/$REPO/swatch-database \                                                                                                  
    -p swatch-database/WAIT_IMAGE=quay.io/$REPO/swatch-database   
    ```

### Verification
1. Watch the database and make sure it gets populated
    ```
    watch "oc exec pod/$(oc get pods | awk '{print $1}' | grep swatch-database-db)  -- psql swatch-database-db -c '\d'"
    ```
1. Examine the init container logs of swatch-tally, swatch-billable-usage, etc.  E.g.
    ```
    oc logs -c swatch-tally-service-init pod/$(oc get pods | awk '{print $1}' | grep swatch-tally-service)
    ```
1. Note, if you want to redeploy to the same environment cleanly, you'll need to
   delete the `cji/swatch-database-migration` along with all the ClowdApps.


src/main/java/org/candlepin/subscriptions/tally/admin/DataMigration.java
```diff
@@ -29,6 +29,16 @@
...
+ * Implement this interface for long-running database migrations that need to be triggered and run
+ * in the background. Once the migration is complete, remove the implementing class. For code
+ * archeologists, previous implementations were
+ *
+ * <ul>
+ *   <li>HardwareMeasurementMigration
+ *   <li>OrgIdToTallySnapshotMigration
+ * </ul>
+ */
 public abstract class DataMigration {
```
Sgitario, line 42: Shouldn't this class be moved to swatch-database as well?

swatch-api/deploy/clowdapp.yaml
```diff
@@ -96,56 +105,17 @@ objects:
...
-                        key: db.name
-                  - name: OTEL_EXPORTER_OTLP_ENDPOINT
-                    value: ${OTEL_EXPORTER_OTLP_ENDPOINT}
-                  - name: OTEL_EXPORTER_OTLP_PROTOCOL
-                    value: ${OTEL_EXPORTER_OTLP_PROTOCOL}
-                  - name: OTEL_SERVICE_NAME
-                    value: ${OTEL_SERVICE_NAME}
-                  - name: OTEL_JAVAAGENT_ENABLED
-                    value: ${OTEL_JAVAAGENT_ENABLED}
+              - image: ghcr.io/groundnuty/k8s-wait-for:v2.0
```
Sgitario, line 108: We should not use images outside of our organization https://quay.io/organization/cloudservices. 
Also, taking into account that we have `dependencies: swatch-database`, should not this be enough? The deployment of this service should wait for the swatch-database deployment and hence the migrations already.

swatch-contracts/build.gradle
```diff
@@ -28,11 +18,11 @@ dependencies {
     implementation 'io.quarkus:quarkus-security'
     implementation 'io.quarkus:quarkus-smallrye-fault-tolerance'
     implementation 'io.quarkus:quarkus-smallrye-openapi'
-    implementation("io.quarkus:quarkus-liquibase")
-    implementation("io.quarkus:quarkus-jdbc-postgresql")
+    implementation 'io.quarkus:quarkus-jdbc-postgresql'
     implementation 'io.quarkus:quarkus-messaging-amqp'
     implementation 'io.quarkus:quarkus-messaging-kafka'
     implementation 'io.smallrye.reactive:smallrye-reactive-messaging-in-memory'
+    implementation 'org.apache.commons:commons-lang3'
```
Sgitario, line 25: We should avoid adding dependencies with is not super justified. In this case, I think it's only for StringUtils.isEmpty, which we could create our own or use another alternatives within the Quarkus framework.

swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractEntityMapper.java
```diff
@@ -128,7 +127,7 @@ default String extractBillingProviderId(PartnerEntitlementV1 entitlement) {
       // for contracts created before ITPART-1180, the client ID is not provided,
       // so we need to default it to empty.
       var azureClientId =
-          Optional.ofNullable(entitlement.getPartnerIdentities().getClientId()).orElse(EMPTY);
+          Optional.ofNullable(entitlement.getPartnerIdentities().getClientId()).orElse("");
```
Sgitario, line 130: Is this a leftover?

src/test/resources/application-test-containers.yaml
```diff
@@ -0,0 +1,4 @@
+spring:
+  liquibase:
+    change-log: classpath:/liquibase/changelog.xml
+    show-summary: summary
```
Sgitario, line 4: We don't need this profile since it's only and always needed when using the ExtendWithSwatchDatabase. Instead, we could add these properties directly in https://github.com/RedHatInsights/rhsm-subscriptions/blob/8d46fa82d826ffc9b76e6eab6864243a725449f4/src/test/java/org/candlepin/subscriptions/test/ExtendWithSwatchDatabase.java#L40.

Sgitario (changes requested):
Thanks @awood , I think these changes go in the right direction. 
To sum up, my concerns are:
- Usage of the k8s wait for image because it's outside of our organization.
- Maintain both gradle liquibase and Quarkus liquibase plugin for running the migrations. See my comment about my objections. 
- Added some extra changes like the commons collections dependencies, or the test-containers profile

build.gradle
```diff
@@ -56,9 +47,9 @@ dependencies {
     implementation "com.fasterxml.jackson.module:jackson-module-jakarta-xmlbind-annotations"
     // used to connect to umb
     implementation "org.springframework.boot:spring-boot-starter-activemq"
-    implementation "org.liquibase:liquibase-core"
     implementation "org.postgresql:postgresql"
     implementation libraries["guava"]
+    implementation libraries["commons-collections4"]
```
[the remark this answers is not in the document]
awood: That dependency was being pulled in transitively through the dependency on liquibase.  Our code already had usages of the classes in commons-collections, but removing liquibase as a dependency meant that classes weren't found.

src/main/java/org/candlepin/subscriptions/tally/admin/DataMigration.java
awood: I don't think so.  This class is for migrations that are performed as the result of an API call.  Moving this class to swatch-database would mean that swatch-database would have to be REST aware, remain resident instead of just being a job, and would just generally make swatch-database substantially more complex.  I can see doing that in the future, but I believe it's out of scope for this card.

swatch-api/deploy/clowdapp.yaml
awood: `dependencies: swatch-database` doesn't work the way we need it to.  The other pods will wait for the ClowdApp to be successfully reconciled, but won't wait for the actual job to be complete.  You can remove the `initContainer` directives and do a deployment and see.

As for the external container.  Yes, either we will need to build a new container based on the wait-for script or else just mirror that container internally.

swatch-contracts/build.gradle
awood: I added this because it was already in use as a transitive dependency.  If you have an alternative that's already on the classpath, let me know and I'll use it.  Certainly we can write our own "isEmpty" method but do we put that in this project?  There's no common utility project used by the quarkus apps.  What happens when we need "isEmpty" in another Quarkus app?  Now we have two versions of the same function in two projects.  I don't really see the benefit to writing our own implementation unless we have a truly central place to put it.

src/main/java/org/candlepin/subscriptions/tally/admin/DataMigration.java
Sgitario: Ok, thanks for the clarification.

build.gradle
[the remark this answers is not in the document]
Sgitario: I don't see this as a strong motivation to maintain a new dependency (before it was maintained by the liquibase dependency which was already wrong), but now it's worse as far as I see it. 
Looking deeper, we're only using it in ProductNormalizer.java for:
```java
if (hostFacts.getSystemProfileArch() != null
        && CollectionUtils.isEmpty(hostFacts.getSystemProfileProductIds())) {
      switch (hostFacts.getSystemProfileArch()) {
```

Where `CollectionUtils.isEmpty` can be easily replaced to simply `hostFacts.getSystemProfileProductIds() == null || hostFacts.getSystemProfileProductIds().isEmpty()`.

swatch-contracts/build.gradle
Sgitario: We were already using this dependency for some utils it provides:
Screenshot From 2025-03-06 09-18-02
So, let's use it here. We can always refactor this as part of another ticket.

build.gradle
```diff
@@ -69,6 +60,7 @@ dependencies {
     testImplementation 'org.testcontainers:junit-jupiter'
     testImplementation libraries["wiremock"]
     testImplementation libraries["awaitility"]
+    testImplementation project(":swatch-database") // to allow tests to run Liquibase
```
Sgitario, line 63: This is causing that the spring boot and quarkus dependencies mixed up in tests which is indeed causing the test failures:
```
Caused by: java.lang.NoClassDefFoundError: org/apache/kafka/server/config/AbstractKafkaConfig
	at org.springframework.kafka.test.EmbeddedKafkaZKBroker.afterPropertiesSet(EmbeddedKafkaZKBroker.java:318)
	at org.springframework.kafka.test.EmbeddedKafkaBrokerFactory.create(EmbeddedKafkaBrokerFactory.java:119)
	at org.springframework.kafka.test.context.EmbeddedKafkaContextCustomizer.customizeContext(EmbeddedKafkaContextCustomizer.java:60)
	at org.springframework.boot.test.context.SpringBootContextLoader$ContextCustomizerAdapter.initialize(SpringBootContextLoader.java:443)
	at org.springframework.boot.SpringApplication.applyInitializers(SpringApplication.java:612)
	at org.springframework.boot.SpringApplication.prepareContext(SpringApplication.java:383)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:317)
	... 98 common frames omitted
Caused by: java.lang.ClassNotFoundException: org.apache.kafka.server.config.AbstractKafkaConfig
	... 105 common frames omitted
```

swatch-contracts/build.gradle
```diff
@@ -71,6 +61,7 @@ dependencies {
     testImplementation("io.quarkus:quarkus-test-kafka-companion")
     testImplementation 'org.testcontainers:postgresql'
     testImplementation project(':swatch-common-testcontainers')
+    testImplementation project(':swatch-database') // To allow test creation of database tables
```
Sgitario, line 64: Adding this dependency is also causing failures in tests for Quarkus projects because it overwrites the DATASOURCE and forcing to use always a postgres database instead of the h2 in-memory database.

Sgitario (changes requested):

Sgitario:
@awood as per my two comments above, we should not make the swatch-database a Quarkus application because it's causing a lot of issues in the Spring Boot and also the Quarkus services.
There are a few alternatives like:
- Either using a frameworkless Java application
- Or performing the migrations in a command line using an ubi9 image (similar to https://dzone.com/articles/enabling-db-migrations-using-kubernetes-init).

swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractEntityMapper.java
awood: It was actually a constant pulled in from an external library.

swatch-database/build.gradle
```diff
@@ -0,0 +1,73 @@
...
+
+def database = config.getOrDefault('DATABASE_DATABASE', 'rhsm-subscriptions')
+def host = config.getOrDefault('DATABASE_HOST', 'localhost')
+def port = config.getOrDefault('DATABASE_PORT', '5432')
+def jdbcUrl = "jdbc:postgresql://${host}:${port}/${database}"
+
+def dbUsername = config.getOrDefault('DATABASE_USERNAME', 'rhsm-subscriptions')
+def dbPassword = config.getOrDefault('DATABASE_PASSWORD', 'rhsm-subscriptions')
+
+liquibase {
```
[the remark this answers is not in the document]
awood: I removed the gradle plugin since after implementing our own way of invoking Liquibase, the UI is much better.  We went from
```
./gradlew :liquibaseRollbackCountSql -PliquibaseCount=1
```

to 

```
./gradlew :swatch-database:run --args 'core rollbackCountSql --count=1'
```

Given that the latter option more clearly reflects the Liquibase arguments and commands being used, I think what we have now is an improvement and the plugin can be discarded.

src/test/resources/application-test-containers.yaml
awood: Removed

build.gradle
awood: Thanks for finding this.  I reimplemented the Liquibase runner as just a basic Java class that does a little light parsing of the command line and then passes off control to Liquibase.

Sgitario (approved):
These seems really promising to me.
It looks much better using a frameworkless Java application and use our image with a custom script wait_for instead of depending on a third party image, really good idea!
I verified that this works perfectly when there is an existing database that was already migrated using the old approach. I only encountered difficulties because I didn't realise that we were using new properties "LIQUIBASE_COMMAND_*" to connect to the database. About this, I'm proposing to not use the new "LIQUIBASE_COMMAND_*" properties, and keep using our known "DATABASE_*" properties as we did before (see commit: https://github.com/RedHatInsights/rhsm-subscriptions/pull/4227/commits/a46a80fc42e911c7338e7251de3bd4f96d374364). But feel free to drop this commit if you strongly feel to use the new properties.

Another thing to think about is the rename of the database from "swatch-tally-db" to "swatch-database-db". I don't think this was strictly necessary, but I also prefer the name "swatch-database-db", though take into account that you will need more changes in the app-interface repository, and I'm not super sure if we can even do this (check https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/rhsm/namespaces/rhsm-prod.yml#L192).

And finally, the CI is failing because it can't create the repository "https://quay.io/cloudservices/swatch-database" in Quay. We should first verify these changes in PR, so any ideas how to sort this out? Not sure if you have privileges, or if we need a MR in app-interface first.

lindseyburnett:
> And finally, the CI is failing because it can't create the repository "[quay.io/cloudservices/swatch-database](https://quay.io/cloudservices/swatch-database)" in Quay. We should first verify these changes in PR, so any ideas how to sort this out? Not sure if you have privileges, or if we need a MR in app-interface first.

Example of creating a quay repo in app-interface: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/119595

awood:
> I verified that this works perfectly when there is an existing database that was already migrated using the old approach. I only encountered difficulties because I didn't realise that we were using new properties "LIQUIBASE_COMMAND__" to connect to the database. About this, I'm proposing to not use the new "LIQUIBASE_COMMAND__" properties, and keep using our known "DATABASE_*" properties as we did before (see commit: [a46a80f](https://github.com/RedHatInsights/rhsm-subscriptions/commit/a46a80fc42e911c7338e7251de3bd4f96d374364)). But feel free to drop this commit if you strongly feel to use the new properties.

Yeah, I didn't really want to change the environment variables, but Liquibase picks up from these environment variables which allows us to connect up the values to the CLI easily; we don't have to pass the username/password on the command line which I preferred to do so that the MigrationService didn't have to be in charge of taking the value from the environment and inserting it into the CLI invocation.

The commit you wrote does that transfer from environment to CLI invocation, but it could cause an issue during development runs.  You won't be able to actually use `--username`, etc. anymore since those are now being passed in.  Occasionally I set up a "rhsm-subscriptions-temp" database or something and it's nice to be able to have full control over the command line.

So with the patch you wrote, this won't work anymore:

```
./gradlew :swatch-database:run --args 'status --url=jdbc:postgresql://localhost:5432/rhsm-subscriptions-temp'
```

You'd have to control it all through environment variables which is, in my opinion, a little more opaque.  I want to try to keep the interface to as close to Liquibase's as possible so people can just use Liquibase docs to do their work.

We can take your patch and change it so that it pulls from `DATABASE_USERNAME` and then does a `System.setProperty("liquibase.username", System.getEnv("DATABASE_USERNAME"))`.  That would still let CLI arguments given by devs take [priority](https://docs.liquibase.com/parameters/home.html)

```
Configuration hierarchy

Liquibase supports the setting of properties from the following locations, from highest to lowest priority:

    Liquibase flow file stage parameters
    Liquibase flow file global parameters
    Command-line arguments
    ServletConfig initialization parameters
    ServletContext initialization parameters
    Java system properties
    OS environment variables
    Configuration data, such as piped standard input, defaults files, and connection profiles and other properties stored in the Liquibase properties file.
```

What do you think?

awood:
> What do you think?

I updated your commit with what I'm thinking of.

Sgitario:
> Unfortunately, we really don't have a choice here. We can't have the database retain the name "swatch-tally-db" when it's managed by the swatch-database ClowdApp. I realize there's a name directive in the databases section of the ClowdApp, but it doesn't seem to do anything. The database still gets created as "swatch-database-db".
> 

I don't follow why this change was forced. swatch-database does not own the database, but only run the migrations.

Sgitario:
There are a couple of things that must be done before merging these changes:
- Clarifying if we can keep using swatch-tally-db instead of swatch-database-db to avoid outages
- Create the swatch-database quay repository to run the CI in this PR

awood:
> I don't follow why this change was forced. swatch-database does not own the database, but only run the migrations.

To me, it made sense that swatch-database should manage the database entirely rather than leave it under the control of swatch-tally.  I figured we're trying to move to a more distributed architecture so it made more sense for the database to be independent of any one specific application.

I can do an experiment to see how it works with swatch-database only doing the migrations while swatch-tally retains control of the database, but in my opinion that makes the architecture a bit tangled up.

Sgitario:
> > I don't follow why this change was forced. swatch-database does not own the database, but only run the migrations.
> 
> To me, it made sense that swatch-database should manage the database entirely rather than leave it under the control of swatch-tally. I figured we're trying to move to a more distributed architecture so it made more sense for the database to be independent of any one specific application.
> 
> I can do an experiment to see how it works with swatch-database only doing the migrations while swatch-tally retains control of the database, but in my opinion that makes the architecture a bit tangled up.

I'm ok with that, I also prefer the swatch-database-db, but we should avoid outages altogether.

awood:
> I'm ok with that, I also prefer the swatch-database-db, but we should avoid outages altogether.

Awesome.  100% agree.

InsightsDroid:
### IQE Tests Summary Report
 1 failed, 199 passed, 2 skipped, 241 deselected, 257 warnings in 3522.86s (0:58:42) 
Result: Failed
|Failures|
|:-| 
|integration/swatch_tally/test_tally_openshift.py::test_validate_tally_on_openshift_cores[1]|

InsightsDroid:
### IQE Tests Summary Report
 2 failed, 198 passed, 2 skipped, 241 deselected, 257 warnings in 3673.49s (1:01:13) 
Result: Failed
|Failures|
|:-| 
|component/swatch_billable_usage/test_swatch_billable_usage.py::test_multiple_tally_summaries_not_aggregated_for_different_metric_ids| 
|integration/swatch_tally/test_tally_openshift.py::test_validate_tally_on_openshift_cores[2]|

lindseyburnett:
> I figured we're trying to move to a more distributed architecture so it made more sense for the database to be independent of any one specific application.

I see where you're coming from in wanting swatch-database to manage the schema independently, but coupling all our microservices to a centralized database project actually moves us away from a truly distributed architecture rather than toward it.  I don't think that "independent database" needs to mean moving all schema management to a single project — it means allowing each service to manage its own data, within its own boundaries (e.g. bounded contexts).

In a distributed system, each service should be responsible for its own database schema.  If we centralize schema management into a new module swatch-database, we're introducing tight coupling between services and prevents them from evolving independently.  Moving schema management to swatch-database effectively makes swatch-database an implicit monolith, introducing the kind of coupling that microservices aim to avoid.

Today, swatch-contracts is responsible for its own domain logic, and its data schema should be part of that responsibility.  I think we were actually heading in the right direction with swatch-contracts by having its liquibase files in its subdirectory.  The best practice for a distributed architecture is Database per Service.  In an ideal world, we would have different database instances/users to help enforce bounded contexts.  Since we don't have that today, we'd have to be super diligent that we're not introducing @Entity classes to a "wrong" service, violating their bounded contexts.  Also, with a database monolith, every service with a database interest becomes dependent on a shared release cycle.

If swatch-database owns the schema, changes to schemas and services will need to be coordinated - which becomes even more problematic if we every try to carve off a service to hand over to another team to manage.  Then schema changes end up being a request on our board that then another team needs to wait for us to release to unblock them.  I think we should be striving for each service having the ability to update its schema as needed in an isolated way.

InsightsDroid:
### IQE Tests Summary Report
 214 passed, 2 skipped, 241 deselected, 271 warnings in 3480.21s (0:58:00) 
Result: Success

awood:
> EDIT: What I'm rambling about here is kind of what I envision as a goal end state, and I still think the changes in this PR are useful and worth going forward with.

I get your point but I want to highlight that I think having swatch-database will actually further this goal.  Swatch-databases can handle multiple independent migration profiles.  Right now, all profiles have to end up in the same DB instance and schema, but we could revise that.  The bottleneck is that all profiles must be fully migrated before any app pod starts.  Given that the migrations don't take all that long, I think this is reasonable.

The major disadvantage is that all apps have to centralize their migration files under swatch-database.

InsightsDroid:
### IQE Tests Summary Report
 214 passed, 2 skipped, 241 deselected, 271 warnings in 3247.03s (0:54:07) 
Result: Success

lindseyburnett (approved):

InsightsDroid:
### IQE Tests Summary Report
 214 passed, 2 skipped, 241 deselected, 271 warnings in 3250.49s (0:54:10) 
Result: Success

Sgitario:
@awood about renaming the db from "swatch-tally-db" to "swatch-database-db", what is the status about this point? Will this cause an outage when deploying to stage/prod?

awood:
> @awood about renaming the db from "swatch-tally-db" to "swatch-database-db", what is the status about this point? Will this cause an outage when deploying to stage/prod?

I've spoke with some of the Clowder and App-SRE folks and we're going to try the experimental procedure [outlined earlier](https://github.com/RedHatInsights/rhsm-subscriptions/pull/4227#issuecomment-2719064386).  MR 135899 creating the alias has already been merged in for Stage, so once this PR passes CI and merges, I'll create the stage namespace target for the app in `deploy-clowder.yml` and we'll see what will happen.

There will be a hiccup in stage since as soon as this merge lands, a stage deployment will start and all the other clowdapps that have been updated to rely on swatch-database won't find it.  This disconnect should not happen in production since the same MR that we use to update all the git refs used by the apps will also have the production namespace target for swatch-database.  In this case, not doing continuous deployment in production works to our advantage since otherwise we'd have a chicken-and-egg type problem.

InsightsDroid:
### IQE Tests Summary Report
 1 failed, 212 passed, 2 skipped, 242 deselected, 269 warnings in 3546.36s (0:59:06) 
Result: Failed
|Failures|
|:-| 
|component/swatch_billable_usage/test_swatch_billable_usage.py::test_verify_failed_remittances_excluded_remittance_usage|

InsightsDroid:
### IQE Tests Summary Report
 1 failed, 216 passed, 1 skipped, 242 deselected, 272 warnings in 3305.46s (0:55:05) 
Result: Failed
|Failures|
|:-| 
|integration/swatch_tally/test_tally_rhel_payg.py::test_verify_multi_cloud_provider_usage[rhel-for-x86-els-payg]|

InsightsDroid:
### IQE Tests Summary Report
 217 passed, 1 skipped, 242 deselected, 272 warnings in 3273.31s (0:54:33) 
Result: Success

InsightsDroid:
### IQE Tests Summary Report
 1 failed, 216 passed, 1 skipped, 242 deselected, 272 warnings in 3658.91s (1:00:58) 
Result: Failed
|Failures|
|:-| 
|component/swatch_billable_usage/test_swatch_billable_usage.py::test_verify_failed_remittances_excluded_remittance_usage|

InsightsDroid:
### IQE Tests Summary Report

Result: Success

InsightsDroid:
### IQE Tests Summary Report
 6 failed, 217 passed, 2 skipped, 242 deselected, 288 warnings in 1864.67s (0:31:04) 
Result: Failed
|Failures|
|:-| 
|integration/swatch_billing/test_remitance.py::test_verify_remittance_of_metered_product[aws-rhel-for-x86-els-payg-addon]| 
|integration/swatch_billing/test_remitance.py::test_verify_end_to_end_payg_metrics| 
|integration/swatch_billing/test_remitance.py::test_verify_remittance_excludes_hard_failures[OpenShift-metrics-aws]| 
|integration/swatch_billing/test_remitance.py::test_verify_rosa_vs_osd_in_swatch| 
|integration/swatch_billing/test_remitance.py::test_verify_payg_usage_when_billing_account_id_updated[rhods]| 
|integration/swatch_tally/test_tally_dedup_rhel_els.py::test_verify_deduplication_remittance_different_event_source[quantities1]|

InsightsDroid:
### IQE Tests Summary Report
 223 passed, 2 skipped, 242 deselected, 288 warnings in 1833.00s (0:30:32) 
Result: Success

lindseyburnett:
MR for stage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/139795

PR #4250, opened 2025-03-11
SWATCH-3373: Fix flaky test verifyCorrectUpdatesWhenRetryOccurs

Sgitario:
Jira issue: SWATCH-3373

## Description
The snapshots difference could be because when running the test, the day changed (midnight). Therefore, so making the test more consistent, I've mocked the clock to be in a fixed date/time. I also added some logs, so in case this test fails again, we can see why the snapshots were different than expected.

## Testing
CI is green.


mstead (approved):
Looks good to me.

PR #4311, opened 2025-03-28
SWATCH-3418: Ensure a step is marked as failure when it fails

Sgitario:
Jira issue: SWATCH-3418

## Description
Before these changes, when a step failed, another later step was marked as failure, having to navigate and inspect all the previous steps to check what was the cause.

After these changes, a step is properly marked as failure and following steps continue if it makes sense (for ex: the upload of the tests report continues even when the previous step fails because we want to see this report).

## Testing
Regression testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,22 +39,14 @@ jobs:
       - name: Run Spotless & Checkstyle
         id: run-lint
         run: |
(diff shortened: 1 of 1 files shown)


vbusch (approved):
LGTM

For testing I created another branch with a failure.  And then a test PR https://github.com/RedHatInsights/rhsm-subscriptions/pull/4313

PR #4294, opened 2025-03-25
TEST - Do not merge

awood:
Test github actions build system with Swatch database PR

PR #4325, opened 2025-04-02
SWATCH-3417: GitHub actions summary improved

awood:
Jira issue: SWATCH-3417

## Description
This PR improves the experience with Github Actions CI.

* Removed `--no-parallel` to speed up builds.  The runners seem to support it just fine and according to a chat with Kevin H, we originally added `--no-parallel` to deal with memory issues on the Jenkins nodes.
* The comment added with the build results now features:
  * Direct links to the checkstyle and JUnit HTML reports
  * A short report of the JUnit results
  * An accordion folded detailed report that is also a direct link to the annotations on the JUnit Report run so you can get to the stacktrace of the failing tests with just one click.
* Coverage results are in a separate sticky comment since we don't run coverage if the tests are failing.  Running coverage every time is possible, and **if** we moved coverage runs to a separate job it could be done in parallel to mitigate any hit to our build time.  But honestly, I feel like devs can run their own coverage report if they are interested in it while they are debugging their tests.

## Testing
None really.  This PR should show what a passing build looks like and you can check out my test PR (#4294) if you want to see what a failure looks like.


InsightsDroid:
### IQE Tests Summary Report

Result: Success

Sgitario (commented):
The removal of `--no-parallel` should have been done in a different pull request since it's unrelated to the task. 

About the changes, I strongly thing we should remove the Job summary because it gives a little benefit vs the high complexity that it brings to understand the CI and maintain it. However, like other times, maybe the team is happy with this direction, so I will let others to bring into the discussion and review the changes.

awood:
> The removal of `--no-parallel` should have been done in a different pull request since it's unrelated to the task.

I have it isolated to its own commit so it can be easily reverted if necessary.  I don't quite see the advantage in having that as a separate PR, especially given the trivial implementation required for the change.  It's not strictly related to the card, but it was about 60 seconds of work and didn't seem worth the overhead of creating a new card, tracking it, and pulling in another reviewer.  Additionally, that commit is only going to apply cleanly to the one before it.  If I drop the commit from this PR, it's just going to take that much longer for the removal of `--no-parallel` and our builds will continue to take longer until then.

> About the changes, I strongly thing we should remove the Job summary because it gives a little benefit vs the high complexity that it brings to understand the CI and maintain it. However, like other times, maybe the team is happy with this direction, so I will let others to bring into the discussion and review the changes.

I see a great deal of value in being able to navigate directly to the list of failed tests.  Especially given that it's not unheard of for unit tests to pass on a local system but fail in CI.  Without the list of failures, you're left with the console log of the test run.  Gradle does not make it obvious which tests actually fail.  You can also download the JUnit report, but that means you have to download a zip file, unzip it, and go to the index.html in your browser all just to find out which test failed.  That seems extremely tedious to me when we can just add a comment with test failures instead.

The step to find the link directly to the annotations is complicated.  The JUnit report action creates a new check to hold the report annotations and doesn't provide a direct link to the annotations which is where the stacktraces are.  I think it's a nice feature to have a direct link and it works for now.  If GitHub changes their API dramatically or something, it might stop working and we could reevaluate whether it's worth keeping then.  But for now, it has value to me.  Other team members are welcome to cast their votes as well.

lindseyburnett:
**About the `--no-parallel` thing**

I agree with both perspectives here to some extent. While it is generally a good practice to keep unrelated changes in separate PRs for clarity and history, like Al mentioned, this specific change is both trivial and self-contained. It’s also captured in a separate commit, making it easy to revert if needed. Since there's negligible overhead involved, I’m okay with it being part of this PR. 

**About the Job Summary Comment**

This one’s a bit more subjective, but based on team sentiment and pragmatic value, my vote is to keep the Summary job as long as it's working correctly.

There is definitely value in having a quick-glance summary of failed tests directly in the PR with the ability to click into them instead of having to download zips of html reports.

Also, being new to GitHub actions, I didn't find it very intuitive to navigate to the right place to find results.  Having a concise, obvious Summary makes it more accessible in my opinion and will shorten debugging time.

While it does introduce some extra wiring in the CI logic, as long as it's working and stable, I'm inclined to keep it.  It's easier to remove things later than it is to recreate it from nothing.  The concern about GitHub API changes is fair, but not imminent as far as I can tell.  If something breaks, we’ll reassess based on who finds value in the feature.  In an ideal world, CI pipeline definitions aren't something that we should need to maintain regularly once they're in place.

Ultimately, like our PR template file or other team norms or tools that may have initially felt “extra,” I think this adds value without being intrusive. Unless it becomes a burden to maintain or no one uses it, I think it's worth having.

lindseyburnett (approved):

PR #4409, opened 2025-04-28
SWATCH-2820: Independent database migrations (again)

awood:
Jira issue: SWATCH-2820

## Description
A reworked version of #4227 that is configured to run the database migration after every deployment.  The original PR used the same job name every time and once the job completed successfully, Clowder wouldn't run it again.  That made deployments in stage fail since the stage environment is long lived.

### Setup
1. Reserve a bonfire environment.
1. `bin/build-images.sh`
1. Deploy with
    ```
    VER=$(git rev-parse --short=7 HEAD) ; \
    REPO=$(podman login quay.io --get-login) ; \
    bonfire deploy rhsm \
    --source=appsre \
    --ref-env insights-production \
    --optional-deps-method none \
    --frontends false \
    --no-remove-resources app:rhsm \
    --timeout 1000 \
    -i quay.io/$REPO/swatch-metrics=$VER \
    -i quay.io/$REPO/swatch-metrics-hbi=$VER \
    -i quay.io/$REPO/rhsm-subscriptions=$VER \
    -i quay.io/$REPO/swatch-producer-aws=$VER \
    -i quay.io/$REPO/swatch-producer-azure=$VER \
    -i quay.io/$REPO/swatch-system-conduit=$VER \
    -i quay.io/$REPO/swatch-billable-usage=$VER \
    -i quay.io/$REPO/swatch-database=$VER \
    -i quay.io/$REPO/swatch-contracts=$VER \
    -p swatch-metrics/IMAGE=quay.io/$REPO/swatch-metrics \
    -p swatch-metrics-hbi/IMAGE=quay.io/$REPO/swatch-metrics-hbi \
    -p swatch-tally/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-api/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-database/IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-contracts/IMAGE=quay.io/$REPO/swatch-contracts \
    -p swatch-producer-aws/IMAGE=quay.io/$REPO/swatch-producer-aws \
    -p swatch-producer-azure/IMAGE=quay.io/$REPO/swatch-producer-azure \
    -p swatch-producer-red-hat-marketplace/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-subscription-sync/IMAGE=quay.io/$REPO/rhsm-subscriptions \
    -p swatch-billable-usage/IMAGE=quay.io/$REPO/swatch-billable-usage \
    -p swatch-system-conduit/IMAGE=quay.io/$REPO/swatch-system-conduit \
    -p swatch-api/WAIT_IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-producer-red-hat-marketplace/WAIT_IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-subscription-sync/WAIT_IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-tally/WAIT_IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-billable-usage/WAIT_IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-database/WAIT_IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-metrics-hbi/WAIT_IMAGE=quay.io/$REPO/swatch-database \
    -p swatch-contracts/WAIT_IMAGE=quay.io/$REPO/swatch-database
    ```

### Steps
1.  The applications should deploy and looking at the logs on the init containers should show the migrations ran.
1.  Apply the following patch:
    ```patch
    diff --git i/swatch-database-migrations/src/main/resources/liquibase/add-test-table.xml w/swatch-database-migrations/src/main/resources/liquibase/add-test-table.xml
    new file mode 100644
    index 000000000..97be41a6a
    --- /dev/null
    +++ w/swatch-database-migrations/src/main/resources/liquibase/add-test-table.xml
    @@ -0,0 +1,26 @@
    +<?xml version="1.0" encoding="UTF-8"?>
    +
    +<databaseChangeLog
    +  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
    +  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    +  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
    +
    +  <changeSet id="202504251140-1" author="awood">
    +    <comment>Create test table</comment>
    +
    +    <createTable tableName="test_table">
    +      <column name="org_id" type="VARCHAR(32)"/>
    +      <column name="service_type" type="VARCHAR(32)"/>
    +      <column name="latest_event_record_date" type="TIMESTAMP WITH TIME ZONE"/>
    +    </createTable>
    +
    +    <addPrimaryKey constraintName="test_table_pkey"
    +      tableName="test_table"
    +      columnNames="org_id, service_type"/>
    +
    +    <rollback>
    +      <dropTable tableName="test_table" />
    +    </rollback>
    +  </changeSet>
    +</databaseChangeLog>
    diff --git i/swatch-database-migrations/src/main/resources/liquibase/changelog.xml w/swatch-database-migrations/src/main/resources/liquibase/changelog.xml
    index 88a2e47eb..55922bff7 100644
    --- i/swatch-database-migrations/src/main/resources/liquibase/changelog.xml
    +++ w/swatch-database-migrations/src/main/resources/liquibase/changelog.xml
    @@ -178,5 +178,6 @@
         <include file="liquibase/202412041537-primary-keys-on-all-tables.xml"/>
         <include file="liquibase/202501081035-remove-backup-tables.xml"/>
         <include file="liquibase/202412091327-drop-hardware-measurement-type-from-billable-usage-remittance.xml"/>
    +    <include file="liquibase/add-test-table.xml"/>
     </databaseChangeLog>
     <!-- vim: set expandtab sts=4 sw=4 ai: -->
    ``` 
1. Repeat all the setup steps (build images and redeploy).

### Verification
1.  All pods should come up and the init containers should show successful migration runs.  You should see 2 completed swatch-database jobs.  The original one from the setup steps and the new one from the test steps.
2. Connect to the database with `oc port-forward $(oc get pods | awk '{print $1}' | grep migrations-db) 5000:5432` and then `psql -h localhost -p 5000 -U <database user from secret here> swatch-migrations-db` and verify that the table `test_table` is there.


InsightsDroid:
### IQE Tests Summary Report
 227 passed, 2 skipped, 239 deselected, 299 warnings in 1875.77s (0:31:15) 
Result: Success

InsightsDroid:
### IQE Tests Summary Report
 227 passed, 2 skipped, 239 deselected, 299 warnings in 1955.62s (0:32:35) 
Result: Success

InsightsDroid:
### IQE Tests Summary Report
 3 failed, 223 passed, 3 skipped, 239 deselected, 299 warnings in 1897.87s (0:31:37) 
Result: Failed
|Failures|
|:-| 
|component/swatch_contracts/test_contracts_capacity.py::test_verify_contract_capacity[aws-ansible-aap-managed]| 
|integration/swatch_billing/test_remitance.py::test_failed_billable_usage_status_subscription_not_found[aws]| 
|integration/swatch_billing/test_remitance.py::test_failed_billable_usage_status_subscription_not_found[azure]|

Sgitario (commented):
I see the pull request has way more changes that I originally reviewed.
I have a few questions/things that I would like to improve:

- why do we need two modules: swatch-database / swatch-database-migrations?
- I see we're repeating the same configuration in all our clowdapps:
```
- image: ${WAIT_IMAGE}:${WAIT_TAG}
              command: ["/bin/sh","-c"]
              args:
                - >
                  echo "Migrating using image swatch-database-$(MIGRATION_IMAGE_TAG)";
                  /deployments/wait_for.sh job -lclowdjob=swatch-database-$(MIGRATION_JOB_ID)
              env:
                - name: MIGRATION_JOB_ID
                  valueFrom:
                    configMapKeyRef:
                      name: migration-conf
                      key: jobId
                - name: MIGRATION_IMAGE_TAG
                  valueFrom:
                    configMapKeyRef:
                      name: migration-conf
                      key: imageTag
```

Can we somehow reduce the number of duplicated lines?
- Why not running the migration as part of the init-containers vs running the migration in a job? 

Moreover, I think the overall strategy should be introduced and explained to the team in a show and tell session.

InsightsDroid:
### IQE Tests Summary Report
 243 passed, 2 skipped, 240 deselected, 309 warnings in 1931.30s (0:32:11) 
Result: Success

awood:
> * why do we need two modules: swatch-database / swatch-database-migrations?

To prevent classpath pollution.  Keeping the migrations isolated in their own module allows other projects to import the module for test purposes.  The migration runner is very simple, but even adding basic dependencies like a logging library created a lot of problems with classpath pollution between Spring and Quarkus.

> * I see we're repeating the same configuration in all our clowdapps:

That hasn't really changed from the previous iteration.  There are now the two environment variable definitions added, but those are necessary to get the job to trigger on each deployment.  On each deployment, the migration job is given a random suffix.  But the other pods need to know what that random suffix is so they can block on the right job.  In a long running environment like production, it would be likely that numerous migrations jobs will have run and completed.  We need our deployments to block on the most recent migration.

The only way I could figure to communicate the suffix between the migration job and the other application pod init-containers is through a ConfigMap.  Strictly speaking, we could drop the bit with `MIGRATION_IMAGE_TAG` but it seemed like it would be useful for debugging to see the actual migration image that got run instead of having to trace the image down by looking up the job first.

> * Why not running the migration as part of the init-containers vs running the migration in a job?

That's what we do today.  Swatch-tally runs an init container to do migrations and swatch-contracts does too.  The issue is that we don't actually want to deploy swatch-tally in some QE scenarios.  Many of the other modules depend on swatch-tally based tables.

The card's original description read

```
 A new clowder service is created that only performs database maintenance. 

This came up in a discussion of how to extract the contract service so that we could set up an IQE environment that only included the contract service and a minimal set of dependencies and it was uncovered that in order to deploy the export service we also had to deploy tally so that the database would be populated. 

All other services depend on the DB migration service
 * Provide a way for contracts to have the database available that doesn't involve spinning up tally
 * Pulling the database out of the monolith. 
 * Clowder sequence deployments (dependencies)
 * Need to verify that if a migration fails then other services are not rolled out.
 * While swatch-contracts already has it's own liquibase, swatch-api also uses the subscriptions tables. This card gives us a clean way to isolate the db stuff until we can get all the calls proxied to their various microservices.  
 *
```

Sgitario:
> > * why do we need two modules: swatch-database / swatch-database-migrations?
> 
> To prevent classpath pollution. Keeping the migrations isolated in their own module allows other projects to import the module for test purposes. The migration runner is very simple, but even adding basic dependencies like a logging library created a lot of problems with classpath pollution between Spring and Quarkus.

I though the only problem was about the logging library which I don't see why we need one in the swatch-database module (?).
Anyway, what about to rename swatch-database-migrations to swatch-common-database, so it's clear the intent?

> 
> > * I see we're repeating the same configuration in all our clowdapps:
> 
> ...
> 
> > * Why not running the migration as part of the init-containers vs running the migration in a job?
> 
> That's what we do today. Swatch-tally runs an init container to do migrations and swatch-contracts does too. The issue is that we don't actually want to deploy swatch-tally in some QE scenarios. Many of the other modules depend on swatch-tally based tables.
> 
> ... 

Yes, I'm aware of the original idea, but from my point of view, it was based on the assumption that we could depend on swatch-database using clowder which we can't. Then the init-container was added back to all our services. However, now we have to sync and confirm that the job is updated in every releases which is a lot of more complicated as the original purpose of simply adding a new service that other services can depend on. 

So, my question is why do we need the job at all? The init-container can migrate the database (by domains as it is now) and liquibase will lock at table when migrating, so multiple services depending on the same domain will work ok too. 

Comparing your changes (without the Job) and what we had before, the situation is much better:
- you managed to create a new swatch-database module to build our custom image that performs migrations by domain
- and also we thought about integration testing, so now we can perform the tally migrations in swatch-contracts at the tests which is something we couldn't do before.

awood:
> I though the only problem was about the logging library which I don't see why we need one in the swatch-database module (?).

That was the only problem...*so far*.  Any additional dependency (including transitive dependencies that Liquibase could add in the future) could create the same problem.  And then some hapless developer is going to have to either redo the work of pulling the migrations out into their own module or struggle with trying to exclude certain dependencies when the module is used as a test dependency.  I tried the latter approach and was not able to get it working in Gradle.  I imagine it is possible, but it's certainly not easy.

> Anyway, what about to rename swatch-database-migrations to swatch-common-database, so it's clear the intent?

I named it "swatch-database-migrations" because it's the module that contains the actual migrations.  To me, that is more descriptive that "common" which doesn't provide much detail.

> So, my question is why do we need the job at all? The init-container can migrate the database (by domains as it is now) and liquibase will lock at table when migrating, so multiple services depending on the same domain will work ok too.

Are you suggesting that each service just have `swatch-database` as an init-container and run it with the directive to only do the migrations it needs?  E.g. a lot of the pods run with `core`, contracts runs it with `contracts`, etc?

That is an interesting idea and worth considering.  The one issue that comes to mind is that some ClowdApp needs to be the ultimate owner of the database.  We don't want that to be `swatch-tally`.  I reckon we might be able to create a swatch-database ClowdApp that doesn't have any pods at all, but I would need to test that.

Personally, I like the aspect of having the pods block and wait on one actor to handle database setup rather than a mad scrabble for grabbing the changelog lock and then a bunch of extraneous Liquibase runs after the fact, but that's more of an aesthetic preference.  And the blocking aspect certainly adds more complexity to the deployment.

awood:
> Are you suggesting that each service just have `swatch-database` as an init-container and run it with the directive to only do the migrations it needs? E.g. a lot of the pods run with `core`, contracts runs it with `contracts`, etc?
> 
> That is an interesting idea and worth considering. The one issue that comes to mind is that some ClowdApp needs to be the ultimate owner of the database. We don't want that to be `swatch-tally`. I reckon we might be able to create a swatch-database ClowdApp that doesn't have any pods at all, but I would need to test that.

So I did an experiment and tried running the swatch-database application as an init-container for every application that needed it.  It doesn't quiet work out of the box.  The issue is that the `swatch-database` application needs all the database connection information provided via the environment.  It doesn't know how to read the Clowder `cdappconfig.json` file.

A lot of the applications still have some of the database information specified in the environment.  `swatch-api` for example.  But others don't: `swatch-billable-usage` and `swatch-tally-service`.  I put the database information in the environment, and with that change, everything will deploy and the pods will reach an eventual consistency.  *Hooray*!

The other downside is that running the Java migration in every required init container takes up more CPU and memory resources than just running a shell script to poll for successful completion.

Those are both pretty minor disadvantages although I am a little ambivalent about adding the `DATABASE_*` environment variables back into the application pods just because I'm not sure of the resolution order between a value set in the environment versus one set in `cdappconfig.json` and if there's the potential there for conflicts.  Of course, all but two of the pods already had the `DATABASE_*` variables set, so it's not like adding them back into two of the pods is a major change.

Sgitario (approved):

InsightsDroid:
### IQE Tests Summary Report
 245 passed, 2 skipped, 240 deselected, 317 warnings in 2046.03s (0:34:06) 
Result: Success

PR #4519, opened 2025-05-16
SWATCH-3528: Update swatch-contracts junit tests using the io.quarkus:quarkus-test-security extension

wottop:
Jira issue: SWATCH-3528

## Testing

Confirm that the modified unit tests test the scenarios as they did before.

### Verification

The unit tests pass with the same assertions.

### DIFF
--- a/swatch-contracts/build.gradle
+++ b/swatch-contracts/build.gradle
@@ -63,7 +63,6 @@ dependencies {
     annotationProcessor libraries["mapstruct-processor"]
     testImplementation 'io.rest-assured:rest-assured'
     testImplementation 'org.mockito:mockito-junit-jupiter'
-    testImplementation 'io.quarkus:quarkus-test-security'
     testImplementation libraries["junit-jupiter"]
     testImplementation libraries["wiremock"]
     testImplementation libraries["awaitility"]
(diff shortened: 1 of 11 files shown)


Sgitario:
@wottop I've just updated this PR by fixing the testReportByMetricIdShouldRespectOffsetAndLimit test you asked. 
To sum up:
- Originally, the test was not working because the json you built with the identity was malformed. You can debug this kind of issues by debugging RhIdentityHeaderAuthenticationMechanism. 
- To provide a correct identity, I'm using an existing header "CUSTOMER_IDENTITY_HEADER" from RhIdentityUtils which uses the org ID "org123"
- The query params begin and end were wrong, the correct ones are beginning and ending. 
- Also, since this API is for customers only, we need to either disable the RBAC / or mock it (I disabled it, but up to you if you want to mock it instead). 

As a final note, it was challenging to me mocking the subscription repository because of the date format used in tests and the API are different. Generally speaking, I prefer not mocking repositories, so we can also cover the queries the API does. But this is optional for these changes, so I would leave this change up to you.

Sgitario (approved):

InsightsDroid:
### IQE Tests Summary Report
 1 failed, 225 passed, 2 skipped, 234 deselected, 301 warnings in 1979.81s (0:32:59) 
Result: Failed
|Failures|
|:-| 
|integration/swatch_tally/test_tally_rhel_payg.py::test_verify_rhel_payg_kafka_event[rhel-for-x86-els-payg]|

InsightsDroid:
### IQE Tests Summary Report
 226 passed, 2 skipped, 234 deselected, 301 warnings in 1984.42s (0:33:04) 
Result: Success

PR #4569, opened 2025-05-30
SWATCH-3619: Update postgresql image to quay.io/sclorg/postgresql-13-c9s

Sgitario:
Jira issue: SWATCH-3619

## Description
Our podman compose and test containers use unsupported Postgress 13 image based on CentOS7. We should use supported image from https://github.com/sclorg/postgresql-container?tab=readme-ov-file#versions:
- CentOS Stream 9 - quay.io/sclorg/postgresql-13-c9s

Additionally, most of these images (except CentOS 8) are multiarch, which would make it easier for running on MacOS.


mmusil:
@Sgitario I think you have missed one more file https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/deploy/dev-clowdenv.yaml#L47

PR #4580, opened 2025-06-03
SWATCH-3571: Adjust the min interval to view the PAYG metrics

Sgitario:
Jira issue: SWATCH-3571

## Description
These changes set the min interval to view the metrics. Without these changes, the min interval is auto-calculated by grafana regardless of the interval value we set in the filter (which in most of the cases, it uses 10 seconds).

## Testing
Extract the new dashboard json using `oc extract -f .rhcicd/grafana/grafana-dashboard-subscription-watch-payg-metrics.configmap.yaml --confirm`

Once you extract it from the .yaml that's checked into this repo, you can import it into the stage instance of grafana by going to `Dashboards -> +Import` from the left nav.

## Verification

After importing the dashboard, if you change the "Interval" filter, you should see all the plots are updated accordingly.

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-subscription-watch-payg-metrics.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-subscription-watch-payg-metrics.configmap.yaml
@@ -171,6 +171,7 @@ data:
             "y": 1
           },
           "id": 3,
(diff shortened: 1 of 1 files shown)


kartikshahc:
1h aggregating looks good and better rather than every 10 mins

PR #4611, opened 2025-06-10
SWATCH-3554: Use our artemis image for local and ephemeral testing

Sgitario:
Jira issue: SWATCH-3554

## Testing
IQE Test MR: TODO

### Setup
1. Deploy the database using `podman compose -f docker-compose.yml up -d`
2. Build the sources without tests `./mvnw clean install -DskipTests`
3. Run database migrations `./mvnw -f swatch-database/pom.xml exec:java`
4. Start Up artemis `podman compose -f config/artemis/docker-compose.yml up -d`
5. Start Up the swatch contracts service: `UMB_ENABLED=true SERVER_PORT=8001 QUARKUS_MANAGEMENT_PORT=9001 ./mvnw -pl swatch-contracts quarkus:dev`
6. Send messages: `podman exec artemis utils/send_amqp_message.py --address VirtualTopic.services.productservice.Product --content '{ "occurredOn" : "2025-03-29", "productCode" : "RH0180191", "productCategory": "Parent SKU", "eventType" : "Create" }' --debug`
7. You should see the swatch-contracts service consumed this message. Note that it fails due to 401, which is also expected.


kartikshahc (commented):
I don't think we should be removing this test class since it was used to tes new umb topic locally. Is there an equivalent test?

Sgitario:
> I don't think we should be removing this test class since it was used to test new umb topic locally. Is there an equivalent test? Just wanted to bring that up even though its in draft state. Someone else can review it since its in draft state and I will be out this week. :)

We should stop adding production logic like this class for only testing purposes. This is a bad habit that might have side consequencies when running the service in production.

As part of this ticket, I'll be adding a IQE test to cover this functionality without needing this class (that's why it's still in draft)

InsightsDroid:
### IQE Tests Summary Report
 1 failed, 228 passed, 2 skipped, 235 deselected, 304 warnings in 1924.96s (0:32:04) 
Result: Failed
|Failures|
|:-| 
|component/swatch_contracts/test_offering.py::test_umb_product_sync_topics_run_in_parallel_with|

PR #4651, opened 2025-06-18
SWATCH-3424: Configure sonar to exclude generated sources

Sgitario:
Jira issue: SWATCH-3424

## Description
The Jacoco plugin is not excluding the generated sources from the openapi or the jsonschema plugins. Also, we can't configure this using the Jacoco Maven plugin.

However, since we're going to pull the code coverage to sonar, we can exclude the generated sources and tests using the Sonar Maven plugin which involves much lesser changes than renaming the Java packages.

## Testing
Regression testing only.
Check the sonar code coverage is 74% now.


Sgitario:
Closing as sonar was already excluding the generated sources and tests, so this configuration is unneeded.

PR #4664, opened 2025-06-20
CI: run migrations using a Maven profile

Sgitario:
## Description
Before these changes, in order to run the migrations, we had to first install the swatch-database-migrations, and then run the following command `mvn -f swatch-database/pom.xml exec:java`.

After these changes, we don't need to install the swatch-database-migrations or remember the long maven command of exec:java. This is now replaced by simply `./mvnw install -Prun-migrations`.

## Testing
Regression testing only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,13 +78,9 @@ jobs:
       - name: Setup PostgreSQL client
         run: sudo apt-get install -y postgresql-client
 
-      - name: Install Dependencies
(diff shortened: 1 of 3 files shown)


InsightsDroid:
### IQE Tests Summary Report
 233 passed, 2 skipped, 229 deselected, 308 warnings in 1907.24s (0:31:47) 
Result: Success

Sgitario:
Self-merging since this is only a tiny nice to have thing.

PR #4705, opened 2025-06-30
fix: CreateUpdateHostHandlerTest test

Sgitario:
## Description
The problem is that the clock used by the service was the fixed one, instead of the default one, what was causing the host not being considered stale.

## Testing
Regression testing only.


InsightsDroid:
### IQE Tests Summary Report
 300 passed, 2 skipped, 161 deselected, 375 warnings in 2359.50s (0:39:19) 
Result: Success

PR #4733, opened 2025-07-09
Fix swagger webpage for the swatch-tally service

Sgitario:
## Description
The swagger page was broken because:
1. the default base path changed to remove the "api" path a while ago
2. we were not copying the openapi definition from the spec folder

## Testing
Start the swatch-tally service, and go to http://localhost:8000/api/swatch-tally/internal/swagger-ui/index.html


InsightsDroid:
### IQE Tests Summary Report
 309 passed, 2 skipped, 162 deselected, 384 warnings in 1986.09s (0:33:06) 
Result: Success

awood (approved):
It does bring up the UI which is a huge improvement, but I still see a couple errors

```
Resolver error at paths./internal-tally-openapi.json.$ref
Could not resolve reference: undefined undefined
Resolver error at paths./internal-tally-openapi.yaml.$ref
Could not resolve reference: undefined undefined
```

I'm fine with merging this since it does make things better but if you want to explore those errors, go for it.

PR #4744, opened 2025-07-11
Set backofflimit for test job

kflahert:
Test pod is retried when the test results show a failure. Instead we should just try once and report the result.


InsightsDroid:
### IQE Tests Summary Report
 309 passed, 2 skipped, 162 deselected, 384 warnings in 2007.98s (0:33:27) 
Result: Success

PR #4777, opened 2025-07-22
SWATCH-3754: Update default ports on Quarkus and Spring services

wottop:
Jira issue: SWATCH-3754

## Description

By default, the Quarks services use 8000 on a local instance. The Nginx service uses that port when the Podman containers are started, which leads to a conflict. This would start the services on unique ports by default.

### Setup

1. Start the containers
 ```podman -compose up -d```
2. Run the liqubase update
```./mvnw -f swatch-database/pom.xml exec:java```

### Steps

1. Start any of the Quarkus services, e.g.
``` ./mvnw -pl swatch-contracts quarkus:dev```
4. Run against the URL with the expected port
```http GET :8011/api/swatch-contracts/internal/offerings/MCT2342/product_tags```

### Verification

1. You will get the correct response on the port
```
HTTP/1.1 404 Not Found
Content-Type: application/json
content-length: 83
traceresponse: 00-00000000000000000000000000000000-0000000000000000-00

{
    "code": "CONTRACTS3003",
    "status": "404",
    "title": "Sku MCT2342 not found in Offering"
}
```

### DIFF
--- a/src/main/resources/application-wiremock.yaml
+++ b/src/main/resources/application-wiremock.yaml
@@ -1,4 +1,4 @@
 # This profile needs to be activated as long as other profiles in the last order
 # For example: SPRING_PROFILES_ACTIVE=worker,api,kafka-queue,wiremock
-SERVER_PORT: 8005
-METRICS_SERVER_PORT: 9005
+SERVER_PORT: 8010
+METRICS_SERVER_PORT: 9010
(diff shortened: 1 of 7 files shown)


Sgitario (changes requested):
There are some requested changes:
- the default port for swatch-contracts is 8011 and 9011 for the management server, not 8001 / 9001. 
- for swatch-billable-usage is 8012 / 9012, not 8002 / 9002
- for swatch-producer-aws is 8013 / 9013, not 8003 / 9003
- for swatch-producer-azure is 8014 / 9014, not 8004 / 9004
- the spring services need to be updated as well (swatch-tally and conduit service)
- changing SERVER_PORT as initially suggested seems not to be working. You can check either why the property clowder.endpoints.xxx.port is not working, or changing the port only in the dev profile (the wiremock profile will inherit from the dev profile).

Sgitario (approved):

PR #4799, opened 2025-07-29
SWATCH-3678: Create job to run ephemeral tests after merge to main

liwalker-rh:
Jira issue: SWATCH-3678

## Description
In order to have ephemeral tests run after a merge to main similar to stage, we need a job to spin up an ephemeral environment and run all the proper tests.

Changes:
- New script to bring up EE and run the ephemeral tests (similar to pr_check.sh)
- New GitHub action to run the script after every merge to main

## Testing
Make sure `Ephemeral Post Merge Health` job on Ibutsu dashboard has a successful run: https://ibutsu.insights.corp.redhat.com/project/3915c900-85fc-1222-833c-10d51af56f2e/dashboard/72a99440-b965-42c6-86a8-a169275a71e7


Sgitario (changes requested):
I don't think it's technically possible to run the EE tests as a GitHub actions because there are a lot of tools (like ibutsu and report portal) that are not public. 
Moreover, we would need to configure the repository and the GitHub action to login into the Quay account.

From my point of view, why not keep using the Jenkins instance as "ci.ext.devshift.net PR build"? 
There are other teams that have an additional pipeline to be run after merging to main. For example: https://ci.ext.devshift.net/job/RedHatInsights-rhsm-api-proxy-gh-build-main/.

PR #4818, opened 2025-08-04
SWATCH-3235: New Billable Usage API to reset status of remittance entities

wottop:
Jira issue: SWATCH-3235

## Description

Adding an internal endpoint designed to 'unstick' billable usage remittances.
Has a configurable number of days for the remittance to be considered 'stuck'.
Only considers 2 states for modification: 'in-progress' and 'sent'.
Updates counters in the metering and logs the reconciliation for tracing.

## Testing

IQE Test MR:  PENDING

### Setup

1. ```podman-compose up -d```
2. ```./mvnw -f swatch-database/pom.xml exec:java```
3. ```./mvnw -pl swatch-billable-usage quarkus:dev```

### Steps

1. Insert rows into the database. Both have the condition for an update, but you can change 'status' and/or 'updated_at' to see other scenarios.
```
 insert into billable_usage_remittance (account_number, org_id, product_id, metric_id, accumulation_period, sla, usage, billing_provider, billing_account_id, remitted_pending_value, remittance_pending_date, retry_after, tally_id, uuid, billed_on, error_code, status, updated_at) values ('1234567890', '1234567890', 'prod1', 'metric1', '2025-01-01', '1', '1', 'aws', '1234567890', 100, '2025-01-01', '2025-01-02', gen_random_uuid(), gen_random_uuid(), '2025-01-01', null, 'in_progress', '2025-01-01');
```
```
insert into billable_usage_remittance (account_number, org_id, product_id, metric_id, accumulation_period, sla, usage, billing_provider, billing_account_id, remitted_pending_value, remittance_pending_date, retry_after, tally_id, uuid, billed_on, error_code, status, updated_at) values ('1234567890', '1234567890', 'prod1', 'metric2', '2025-01-01', '1', '1', 'aws', '1234567890', 100, '2025-01-01', '2025-01-02', gen_random_uuid(), gen_random_uuid(), '2025-01-01', null, 'sent', '2025-01-01');
```
2. Run the endpoint
```http POST :8012/api/swatch-billable-usage/internal/rpc/remittance/reconcile```

### Verification

1. If the remittance was 'in_progress' and 'updated_at' is more than 1 day ago, the state will be 'failed' and the error_code will be 'error_sending_to_aggregate_topic'
2. If the remittance was 'sent' and 'updated_at' is more than 1 day ago, the state will be 'unknown', but no error code will be recorded.
3. Look at the log in the window where you ran the application. You will see entries that begin with 'Billable Usage Remittance Stuck Status Change:' for the remittances that have been updated.

### CODE REVIEW

File: swatch-billable-usage/src/main/resources/META-INF/openapi.yaml [hunk 1]
```diff
@@ -197,6 +197,23 @@ paths:
           $ref: "../../../../../spec/error-responses.yaml#/$defs/Unauthorized"
         '500':
           $ref: "../../../../../spec/error-responses.yaml#/$defs/InternalServerError"
+
+  /api/swatch-billable-usage/internal/rpc/remittance/reconcile:
+    post:
+      operationId: reconcileRemittances
+      summary: 'Reconcile existing remittances matching the configured stuck policy.'
+      responses:
```
vbusch, line 205: With the securitySchemes in this file, a 403 response type could be added.  It would be consistent with several other defined endpoints.

vbusch (commented):
Looks good.  The 403 response could be added for completeness in the API docs.

InsightsDroid:
### IQE Tests Summary Report
 315 passed, 2 skipped, 166 deselected, 389 warnings in 1976.38s (0:32:56) 
Result: Success

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/admin/api/InternalBillableUsageController.java
```diff
@@ -85,6 +94,38 @@ public int resetBillableUsageRemittance(
         productId, start, end, orgIds, billingAccountIds);
   }
 
+  @Transactional
+  public void reconcileBillableUsageRemittances(long days) {
```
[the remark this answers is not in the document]
vbusch: Updated to use single transaction

swatch-billable-usage/src/test/java/com/redhat/swatch/billable/usage/admin/api/InternalBillableUsageControllerTest.java
```diff
@@ -274,6 +290,96 @@ void testDeleteDataForOrg() {
...
 
+  @Transactional
+  @Test
+  void testReconcileBillableUsageRemittances() {
+    // create mocks
+    BillableUsageRemittanceRepository mockRemittanceRepo =
+        mock(BillableUsageRemittanceRepository.class);
+    MeterRegistry mockMeterRegistry = mock(MeterRegistry.class);
+    RemittanceMapper mockRemittanceMapper = mock(RemittanceMapper.class);
+    InternalBillableUsageController testController =
```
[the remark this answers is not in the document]
vbusch: mocks removed

vbusch:
> One more thing the diagram in Jira card needs an update because it says to change "In progress" status to "pending" instead of "failed".

I have updated the diagram

swatch-billable-usage/src/test/java/com/redhat/swatch/billable/usage/admin/api/InternalBillableUsageControllerTest.java
```diff
@@ -274,6 +277,128 @@ void testDeleteDataForOrg() {
...
+        .setParameter(1, OffsetDateTime.now().minusDays(2))
+        .setParameter(2, RemittanceStatus.IN_PROGRESS.getValue())
+        .executeUpdate();
+
+    remittanceRepo
+        .getEntityManager()
+        .createNativeQuery(updateUpdatedAtQuery)
+        .setParameter(1, OffsetDateTime.now().minusDays(2))
+        .setParameter(2, RemittanceStatus.SENT.getValue())
+        .executeUpdate();
```
Sgitario, line 306: why do we need these statements? the setupRemittances are not setting the updated_at and the status already?

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/data/BillableUsageRemittanceRepository.java
```diff
@@ -140,6 +142,69 @@ public int resetBillableUsageRemittance(
...
+            + "where bu.status = :oldStatus "
+            + "and bu.updatedAt <= :cutoffDate";
+    parameters.put("errorCode", errorCode);
+    parameters.put("newStatus", newStatus);
+    parameters.put("updatedAt", now);
+    parameters.put("oldStatus", oldStatus);
+    parameters.put("cutoffDate", cutoffDate);
+
+    // Perform the update and get the count of updated records
+    int updatedCount = update(query, parameters);
```
Sgitario, line 185: this method should return this long updatedCount only, not the list of entities.

Sgitario (changes requested):
The changes look better now, thanks @vbusch !
About the logging, I agree with your thoughts too. We should only log the number of records that were updated, and this number is already provided by the update statement, so we don't need to load all the entities in memory which it will also simplify the logic.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/data/BillableUsageRemittanceRepository.java
kartikshahc: So we were discussing how do we log these updated rows. Won't it be helpful to get that information in log?

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/data/BillableUsageRemittanceRepository.java
```diff
@@ -234,6 +310,12 @@ private static Specification<BillableUsageRemittanceEntity> afterRemittanceDate(
             root.get(BillableUsageRemittanceEntity_.remittancePendingDate), beginning);
   }
 
+  private static Specification<BillableUsageRemittanceEntity> beforeUpdatedAtDate(
```
kartikshahc, line 313: This method seems to be of no use after the refactor

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/data/BillableUsageRemittanceRepository.java
```diff
@@ -140,6 +142,69 @@ public int resetBillableUsageRemittance(
...
+
+    // Perform the update and get the count of updated records
+    int updatedCount = update(query, parameters);
+    flush();
+
+    if (updatedCount == 0) {
+      return List.of();
+    }
+
+    var searchCriteria =
```
kartikshahc, line 192: Based on discussion with Jose if we are not logging the changed entities then we can remove the below code.

swatch-billable-usage/src/test/java/com/redhat/swatch/billable/usage/admin/api/InternalBillableUsageControllerTest.java
vbusch: I have cleaned that up.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/data/BillableUsageRemittanceRepository.java
vbusch: I reverted the last commit, which brings it back to logging just the count.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/admin/api/InternalBillableUsageResource.java
```diff
@@ -51,6 +51,7 @@ public class InternalBillableUsageResource implements DefaultApi {
   private final InternalBillableUsageController billingController;
   private final EnabledOrgsProducer enabledOrgsProducer;
   private final ApplicationConfiguration configuration;
+  private final InternalBillableUsageController internalBillableUsageController;
```
kartikshahc, line 54: Why not just use existing `private final InternalBillableUsageController billingController;`? This can create circular dependency in future.

swatch-billable-usage/src/main/java/com/redhat/swatch/billable/usage/data/BillableUsageRemittanceRepository.java
kartikshahc: This method isn't needed anymore so can we remove it?

kartikshahc (changes requested):

kartikshahc (approved):
LGTM

swatch-billable-usage/src/test/java/com/redhat/swatch/billable/usage/admin/api/InternalBillableUsageControllerTest.java
lindseyburnett: Thanks for everyone's collaboration on this! I have some concerns about the testing approach that I'd like to discuss after Walk The Board:

**The Issue: "Cheating the PrePersist"**
The tests are using native SQL and manual timestamp overrides to bypass the `@PrePersist` lifecycle methods. While this makes the tests happy and pass, it's not testing the actual behavior we want to verify.

**What's Happening:**
- `@PrePersist` automatically sets `updatedAt` on entity creation
- Tests use `UPDATE billable_usage_remittance SET updated_at = ?` to override this
- Tests manually set `updatedAt` in entity builders

**Why This Matters:**
- We're not testing that `@PrePersist` actually works
- We're testing implementation details instead of business behavior
- If the JPA lifecycle logic changes, tests might break even though behavior is correct

**Suggested Approach:**
Instead of bypassing `@PrePersist`, let's test the actual business logic:
1. Create entities and let them age naturally (or use test time manipulation)
2. Test that stale entities get processed correctly
3. Verify the business behavior, not the timestamp mechanics

**Questions:**
- Can we use test time manipulation instead of native SQL?
- What specific business logic are we trying to test here?
- Could we test the `@PrePersist` behavior separately?

lindseyburnett (changes requested):

swatch-billable-usage/src/test/java/com/redhat/swatch/billable/usage/admin/api/InternalBillableUsageControllerTest.java
Sgitario: I agree with @lindseyburnett : we should not manipulate the system behaviour. 
I've submitted a commit that contains the following changes in this test class:
- to not update the updatedAt since this won't actually update this property
- refactor the test to not manipulate the updatedAt column and instead tweak and cover the configuration, so we can also assert this logic

Sgitario:
@kartikshahc @lindseyburnett @vbusch @wottop I've found a important bug in the query to update the entities. Apart from this, I've also added a new commit with the following changes:
- Fix this bug in BillableUsageRemittanceRepository where we were only updating the entities using this criteria bu.updatedAt = :updatedAt. 
- Fix issue since we were using a duration in the configuration to only take the days. In case of we want to configure this property in seconds or minutes, this won't work. Let's use directly the Duration property to calculate the cutoff date. 
- Fix the test InternalBillableUsageControllerTest to not update the updatedAt since this won't actually update this property. 
- Refactor the test InternalBillableUsageControllerTest to not manipulate the updatedAt column and instead tweak and cover the configuration, so we can also assert this logic.

InsightsDroid:
### IQE Tests Summary Report
 318 passed, 2 skipped, 166 deselected, 392 warnings in 2022.01s (0:33:42) 
Result: Success

PR #4908, opened 2025-09-03
Move RBAC configuration into swatch-common-security

mstead:
Required for Jira issue: SWATCH-3862

## Description
This config belongs at the library level and allows for overridding at the application config level.

Always set the RBAC_ENABLED env var to True since
it is only ever used in testing and is forced to
False only in specific unit test cases. By default,
Quarkus security will disable the role augmenters
in development mode, so enabling RBAC all the
time is Ok.


kartikshahc (approved):

kartikshahc:
Changes LGTM

PR #4920, opened 2025-09-08
fix HbiEventOutboxServiceTest test which is in conflict with HbiEventConsumerTransactionBoundaryTest

Sgitario:
## Description
The test HbiEventConsumerTransactionBoundaryTest is trying to create a record that is expected to fail. However, after a service restart, since the mocks are gone, this record is actually created. This is causing the test HbiEventOutboxServiceTest to fail because it does not expect more than 1 event. 

## Testing
CI is green.


InsightsDroid:
### IQE Tests Summary Report
 310 passed, 2 skipped, 185 deselected, 384 warnings in 2123.65s (0:35:23) 
Result: Success

PR #4933, opened 2025-09-10
created parent classes for the azure tests

ericThompson038:
- add a version for the jackson-datatype-jsr310

Jira issue: SWATCH-XXXX

## Testing

IQE Test MR: 

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.

### DIFF
--- a/swatch-producer-azure/ct/java/api/AzureTestHelper.java
+++ b/swatch-producer-azure/ct/java/api/AzureTestHelper.java
@@ -31,9 +31,11 @@ import org.candlepin.subscriptions.billable.usage.BillableUsage;
 import org.candlepin.subscriptions.billable.usage.BillableUsageAggregate;
 import org.candlepin.subscriptions.billable.usage.BillableUsageAggregateKey;
 
-public class AzureTestHelper {
+public final class AzureTestHelper {
 
-  public BillableUsageAggregate createUsageAggregate(
+  private AzureTestHelper() {}
+
+  public static BillableUsageAggregate createUsageAggregate(
       String productId, String billingAccountId, String metricId, double totalValue, String orgId) {
     OffsetDateTime snapshotDate =
         OffsetDateTime.now().minusHours(1).withOffsetSameInstant(ZoneOffset.UTC);
--- a/swatch-producer-azure/ct/java/api/BaseAzureTest.java
+++ b/swatch-producer-azure/ct/java/api/BaseAzureTest.java
@@ -35,8 +35,6 @@ import org.junit.jupiter.api.Tag;
 @Tag("azure")
 public abstract class BaseAzureTest extends BaseServiceTest {
 
-  protected static final AzureTestHelper azureHelper = new AzureTestHelper();
-
   @Wiremock
   protected static AzureWiremockService wiremock = new AzureWiremockService();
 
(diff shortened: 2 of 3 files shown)


swatch-producer-azure/ct/java/api/AzureTestHelper.java
```diff
@@ -0,0 +1,65 @@
...
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import org.candlepin.subscriptions.billable.usage.BillableUsage;
+import org.candlepin.subscriptions.billable.usage.BillableUsageAggregate;
+import org.candlepin.subscriptions.billable.usage.BillableUsageAggregateKey;
+
+public class AzureTestHelper {
```
Sgitario, line 34: this class should be final, with a private constructor.

swatch-producer-azure/ct/java/api/AzureTestHelper.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public BillableUsageAggregate createUsageAggregate(
```
Sgitario, line 36: this method should be static, so we can directly use it.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/core/BaseServiceTest.java
```diff
@@ -0,0 +1,50 @@
...
+package com.redhat.swatch.component.tests.core;
+
+import com.redhat.swatch.component.tests.api.ComponentTest;
+import com.redhat.swatch.component.tests.api.KafkaBridgeService;
+import java.util.Collections;
+import java.util.Set;
+import org.junit.jupiter.api.AfterEach;
+
+@ComponentTest
+public abstract class BaseServiceTest {
```
Sgitario, line 30: why do you need this extra parent level? All the tests in swatch-producer-azure/ct are to be used for azure only, therefore BaseAzureTest should be enough (and move the cleanupAfterEachTest method there.

swatch-test-framework/pom.xml
```diff
@@ -31,6 +31,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jsr310</artifactId>
+      <version>2.18.4</version>
```
Sgitario, line 34: did you add this change by mistake?

Sgitario (changes requested):

swatch-test-framework/pom.xml
ericThompson038: The build was failing without a version for me and @kartikshahc . If this isn't happening for you lets see if we can figure out what the difference is.

Sgitario: I already fixed this as part of another commit.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/core/BaseServiceTest.java
Sgitario: Other services won't see this BaseServiceTest because it's created in swatch-producer-azure/ct.

PR #4976, opened 2025-09-22
SWATCH-3906: Added no subscription found test in azure producer compo…

Mantzounis:
Jira issue: SWATCH-3906

## Description
Added the no subscription found test for azure producer based on the same test in IQE.

### DIFF
--- a/swatch-producer-azure/ct/java/tests/SimpleAzureComponentTest.java
+++ b/swatch-producer-azure/ct/java/tests/SimpleAzureComponentTest.java
@@ -26,6 +26,8 @@ import static com.redhat.swatch.component.tests.utils.Topics.BILLABLE_USAGE_STAT
 
 import java.util.Map;
 import java.util.UUID;
+
+import org.candlepin.subscriptions.billable.usage.BillableUsage;
 import org.candlepin.subscriptions.billable.usage.BillableUsageAggregate;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


Sgitario (approved):

TommasoLencioni (approved):

swatch-producer-azure/ct/java/tests/SimpleAzureComponentTest.java
```diff
@@ -140,4 +142,39 @@ public void testNullUsageMessage() {
...
+      kafkaBridge.waitForKafkaMessage(
+          BILLABLE_USAGE_STATUS,
+          messages ->
+              (messages.contains(billingAccountId)
+                  && messages.contains("failed")
+                  && messages.contains(BillableUsage.ErrorCode.SUBSCRIPTION_NOT_FOUND.toString())),
+          1);
+
+      // Verify service produces appropriate log
+      service.logs().assertContains("Subscription not found for for aggregate=");
```
lindseyburnett, line 175: typo here?

`for for`

Mantzounis: This is the way it is logged in the [AzureBillableUsageAggregateConsumer](https://github.com/RedHatInsights/rhsm-subscriptions/blob/ac21645da5bd00498bb5ec7103cd96b615d53438/swatch-producer-azure/src/main/java/com/redhat/swatch/azure/service/AzureBillableUsageAggregateConsumer.java#L148), can I open a bug card for this?

Sgitario: I don't think this is super important enough to open a bug card from my point of view spite this is what we should generally do. 
Instead, I would be fine to fix it directly in this branch.

Mantzounis: it must also be changed in the respective iqe test and possibly other tests as well, should I also open an iqe branch to change this there?

Sgitario: iqe tests cannot assert the service logs, so no changes from iqe are needed.

PR #5036, opened 2025-10-02
hotfix to replace swatch-database :latest image tag

lindseyburnett:
Relates to: SWATCH-3735

https://redhat-internal.slack.com/archives/C09BCH0GRT7/p1759382029073089

Using the `:latest` tag for the swatch-database image in production caused unpredictable component rollouts.

This led to a critical incident where a Liquibase lock was held by a terminated pod during a rollout. As a result, all services are currently blocked from applying new migrations because they are waiting on a lock that no longer exists.

This commit replaces the `:latest` tag with a specific, stable version to ensure predictable deployments. This aligns the production environment with the configuration that has been stable in stage for some time.

### DIFF
--- a/swatch-api/deploy/clowdapp.yaml
+++ b/swatch-api/deploy/clowdapp.yaml
@@ -37,8 +37,6 @@ parameters:
     value: 300m
   - name: MIGRATION_IMAGE
     value: quay.io/cloudservices/swatch-database
-  - name: MIGRATION_TAG
-    value: latest
   - name: MIGRATION_MEMORY_REQUEST
     value: 256Mi
   - name: MIGRATION_MEMORY_LIMIT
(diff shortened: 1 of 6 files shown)


awood (commented):
My concern here is if different apps use different tags.  To date, that hasn't been an issue, but if that becomes the case, then we'll have a really nasty race condition where the database migrations that get run are dependent on whichever app gets its init container to the finish line first.

That scenario was one of the reasons that I initially wrote `swatch-database` such that all the pods block on the completion of one migration pod instead of today where it's a scrabble between all the initializing applications.

The scramble is mitigated by making sure that every migration attempt is using the same image tag across all applications.  Since `MIGRATION_TAG` isn't tied to the actual application image tag, this is easily done.  But if we use `IMAGE_TAG` and begin the practice of different apps using different tags, we'll run into a problem.

PR #5038, opened 2025-10-02
SWATCH-3994: Introduce the MIGRATION_IMAGE_TAG parameter

Sgitario:
Jira issue: SWATCH-3994

## Description
This new parameter defaults to IMAGE_TAG when MIGRATION_IMAGE_TAG is not set.

The idea is to only set MIGRATION_IMAGE_TAG when running the component tests because when Konflux executes the integration test, the snapshot contains a different image tag for each component.

## Testing
CI is green.


Sgitario:
Closing as you can't define fallback properties in OpenShift, so this solution is not supported.

PR #5040, opened 2025-10-03
SWATCH-3895: remove swatch-tally dependency from swatch-contracts

Sgitario:
Jira issue: SWATCH-3895

## Description
We can deploy swatch-contracts independently from swatch-tally, so we can remove this dependency from the clowder.yaml file in swatch-contracts.

## Testing
CI is green.

### DIFF
--- a/swatch-contracts/deploy/clowdapp.yaml
+++ b/swatch-contracts/deploy/clowdapp.yaml
@@ -130,7 +130,6 @@ objects:
         sharedDbAppName: ${DB_POD}
       dependencies:
         - ${DB_POD}
-        - swatch-tally
         - rbac
         - export-service
 


InsightsDroid:
### IQE Tests Summary Report
 1 failed, 312 passed, 2 skipped, 183 deselected, 70 warnings in 2209.22s (0:36:49) 
Result: Failed
|Failures|
|:-| 
|integration/swatch_tally/test_tally_openshift.py::test_validate_tally_on_openshift_cores_for_new_candlepin_facts|

kartikshahc (approved):
Lgtm. One question we don’t need tally for checking opt in correct? I think we are doing it with Kafka to check enabled orgs so just double checking.

Sgitario:
> Lgtm. One question we don’t need tally for checking opt in correct? I think we are doing it with Kafka to check enabled orgs so just double checking.

Exactly

barnabycourt:
Is swatch-tally no longer a dependency at all? Should it be fully removed or just moved to an optional-decency?

Sgitario:
> Is swatch-tally no longer a dependency at all? Should it be fully removed or just moved to an optional-decency?

We don't need swatch-tally dependency in swatch-contracts. This was a leftover to when we needed the swatch-tally for database migrations.

PR #5997, opened 2026-04-16
SWATCH-4780: Implement the IT Partner Gateway Kafka Consumer

Sgitario:
Jira issue: SWATCH-4780

## Description
This PR creates the IT Partner Kafka Consumer as part of the umb to kafkamigration, and also configures the consumer to use a different Kafka Instance (the IT Managed Kafka instance). Details in SWATCH-4894. 

The idea is that in ephemeral, it will use the Kafka instance that is started within the environment. But in stage and prod, it will use the IT Managed Kafka instance which secret was already added in app-interface. 

## Testing
./mvnw test -Dtest=PartnerEntitlementKafkaMessageConsumerTest -pl swatch-contracts
All tests should pass.

## Verification

I verified that the connection works fine for stage and prod, so this should work as expected after merged.

### DIFF
--- a/deploy/rhsm-clowdapp.yaml
+++ b/deploy/rhsm-clowdapp.yaml
@@ -113,3 +113,17 @@ objects:
     data:
       # dummy-token for testing purposes only
       token: ZHVtbXktdG9rZW4=
+
+  - apiVersion: v1
+    kind: Secret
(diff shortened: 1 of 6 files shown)


vbusch (approved):
LGTM

PR #5998, opened 2026-04-16
SWATCH-4563: add nightly contracts sync all cron job

TommasoLencioni:
Jira issue: SWATCH-4563

## Description

Adds a scheduled CronJob that calls POST /api/swatch-contracts/internal/rpc/contracts/sync with the service PSK. That endpoint only enqueues work: one ContractSyncTask per distinct org that already has contracts in our DB. A Kafka consumer runs the existing per-org sync so the HTTP path stays light and consistent with other “sync all” RPCs (POST, not PUT). ClowdApp wires contracts-sync, CONTRACTS_SYNC_SCHEDULE, and topic platform.rhsm-subscriptions.contract-sync.

Contract sync Kafka topic is created here https://github.com/RedHatInsights/platform-mq/pull/513

## Testing

Run `./mvnw install -Pcomponent-tests -pl swatch-contracts/ct -am`

### DIFF
--- a/swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
+++ b/swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
@@ -22,8 +22,8 @@ package tests;
 
 import static api.PartnerApiStubs.PartnerSubscriptionsStubRequest.forContractsInOrgId;
 import static java.util.Collections.disjoint;
-import static java.util.stream.Collectors.toSet;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -34,6 +34,7 @@ import static utils.DateUtils.assertDatesAreEqual;
 
 import api.PartnerApiStubs;
 import com.redhat.swatch.component.tests.api.TestPlanName;
+import com.redhat.swatch.component.tests.utils.AwaitilityUtils;
 import com.redhat.swatch.component.tests.utils.RandomUtils;
 import com.redhat.swatch.contract.test.model.SkuCapacitySubscription;
 import domain.BillingProvider;
@@ -42,16 +43,14 @@ import io.restassured.response.Response;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.Test;
 
 public class ContractsSyncComponentTest extends BaseContractComponentTest {
 
-  // Status message constants
   private static final String STATUS_SUCCESS = "SUCCESS";
-  private static final String STATUS_ALL_CONTRACTS_SYNCED = "All Contract are Synced";
+  private static final String STATUS_ENQUEUED_PREFIX = "Enqueued";
   private static final String STATUS_NO_CONTRACTS_FOUND = "No active contract found for the orgIds";
 
   @TestPlanName("contracts-sync-TC001")
@@ -156,7 +155,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
     var initialSubscriptionIds =
         initialSkuCapacity.get().getSubscriptions().stream()
             .map(SkuCapacitySubscription::getId)
-            .collect(toSet());
+            .collect(Collectors.toSet());
 
     // Setup new upstream contracts (different from existing - same SKU, different provider)
     Contract newContract =
@@ -193,7 +192,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
       var remainingSubscriptionIds =
           finalSkuCapacity.get().getSubscriptions().stream()
               .map(SkuCapacitySubscription::getId)
-              .collect(toSet());
+              .collect(Collectors.toSet());
 
       // Verify none of the old AWS subscription IDs remain
       assertTrue(
@@ -238,27 +237,11 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
     givenContractIsCreated(secondOrgContract);
 
     // When: Sync all contracts
-    Response syncResponse = service.syncAllContracts();
+    whenSyncAllContracts();
 
-    // Then: Verify sync succeeded
-    assertThat(
-        "Sync all contracts should succeed", syncResponse.statusCode(), is(HttpStatus.SC_OK));
-    syncResponse.then().body("status", equalTo(STATUS_ALL_CONTRACTS_SYNCED));
-
-    // Verify contracts still exist for both orgs after sync
-    var firstOrgContracts = service.getContractsByOrgId(firstOrg);
-    assertEquals(1, firstOrgContracts.size(), "First org should have one contract after sync");
-    assertEquals(
-        BillingProvider.AWS.toApiModel(),
-        firstOrgContracts.get(0).getBillingProvider(),
-        "First org contract should be AWS");
-
-    var secondOrgContracts = service.getContractsByOrgId(secondOrg);
-    assertEquals(1, secondOrgContracts.size(), "Second org should have one contract after sync");
-    assertEquals(
-        BillingProvider.AZURE.toApiModel(),
-        secondOrgContracts.get(0).getBillingProvider(),
-        "Second org contract should be Azure");
+    // Then: Verify contracts
+    thenOrgHasSingleContractWithProvider(firstOrg, BillingProvider.AWS);
+    thenOrgHasSingleContractWithProvider(secondOrg, BillingProvider.AZURE);
   }
 
   @TestPlanName("contracts-sync-TC004")
@@ -307,35 +290,21 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
     var contractsBefore = service.getContractsByOrgId(orgId);
     assertEquals(1, contractsBefore.size(), "Should have exactly one contract before sync");
 
-    // When: Run syncAllContracts the first time
-    Response firstSync = service.syncAllContracts();
-    assertThat("First sync should succeed", firstSync.statusCode(), is(HttpStatus.SC_OK));
-    firstSync.then().body("status", equalTo(STATUS_ALL_CONTRACTS_SYNCED));
+    // When: Run syncAllContracts the first time (async)
+    whenSyncAllContracts();
 
-    var contractsAfterFirst = service.getContractsByOrgId(orgId);
-    assertEquals(
-        1, contractsAfterFirst.size(), "Should still have exactly one contract after first sync");
-    String uuidAfterFirst = contractsAfterFirst.get(0).getUuid();
+    // Wait for async processing, then capture state
+    String uuidAfterFirst =
+        AwaitilityUtils.until(
+                () -> service.getContractsByOrgId(orgId), contracts -> contracts.size() == 1)
+            .get(0)
+            .getUuid();
 
     // When: Run syncAllContracts a second time (same upstream data — idempotency check)
-    Response secondSync = service.syncAllContracts();
-    assertThat("Second sync should succeed", secondSync.statusCode(), is(HttpStatus.SC_OK));
-    secondSync.then().body("status", equalTo(STATUS_ALL_CONTRACTS_SYNCED));
+    whenSyncAllContracts();
 
-    // Then: No duplicate contracts created; record count is still exactly 1
-    var contractsAfterSecond = service.getContractsByOrgId(orgId);
-    assertEquals(1, contractsAfterSecond.size(), "Repeat sync must not create duplicate contracts");
-
-    // UUID must be stable — upsert finds the existing record and leaves it unchanged
-    String uuidAfterSecond = contractsAfterSecond.get(0).getUuid();
-    assertEquals(
-        uuidAfterFirst,
-        uuidAfterSecond,
-        "Contract UUID must remain stable across repeat syncAllContracts calls");
-    assertEquals(
-        contractsAfterFirst.get(0).getBillingProvider(),
-        contractsAfterSecond.get(0).getBillingProvider(),
-        "Billing provider must not change across repeat runs");
+    // Then: Wait for async processing and verify no duplicates
+    thenSingleContractHasStableUuid(uuidAfterFirst);
   }
 
   @TestPlanName("contracts-sync-TC007")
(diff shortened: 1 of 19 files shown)


swatch-contracts/deploy/clowdapp.yaml
```diff
@@ -166,6 +170,9 @@ objects:
         - replicas: ${{KAFKA_SUBSCRIPTION_SYNC_TASK_REPLICAS}}
           partitions: ${{KAFKA_SUBSCRIPTION_SYNC_TASK_PARTITIONS}}
           topicName: platform.rhsm-subscriptions.subscription-sync-task
+        - replicas: ${{KAFKA_CONTRACT_SYNC_REPLICAS}}
+          partitions: ${{KAFKA_CONTRACT_SYNC_PARTITIONS}}
+          topicName: platform.rhsm-subscriptions.contract-sync
```
Sgitario, line 175: This Kafka topic needs to first be created in stage and prod, otherwise it will start failing when promoting these changes in production. 
The kafka topics are created here: https://github.com/RedHatInsights/platform-mq.

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsResource.java
```diff
@@ -147,43 +179,22 @@ public List<Contract> getContract(
   }
 
   @Override
-  @Transactional
   @RolesAllowed({"test", "support", "service"})
   public StatusResponse syncAllContracts() throws ProcessingException {
     log.info("Starting global contract sync");
 
     var orgIds = service.getOrgIdUsedInContracts();
```
Sgitario: there is a potential memory leak issue here (it was previous to your changes) that when loading the orgIds, it's copying the whole orgIds into memory (in a list). To address this issue now with this approach, we can start using Stream API. The difference is that the orgIds will be loaded in chunkes.

Sgitario (changes requested):

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -307,35 +313,34 @@ void syncAllContractsTwiceShouldNotCreateDuplicateContracts() {
...
+    AwaitilityUtils.untilAsserted(
+        () -> {
+          var contractsAfterSecond = service.getContractsByOrgId(orgId);
+          assertEquals(
+              1, contractsAfterSecond.size(), "Repeat sync must not create duplicate contracts");
+          assertEquals(
+              uuidAfterFirst,
+              contractsAfterSecond.get(0).getUuid(),
+              "Contract UUID must remain stable across repeat syncAllContracts calls");
+        });
```
Sgitario: We can create a thenContractAsserts(Condition) method and remove duplicate lines that are reused in a few places.

swatch-contracts/src/main/resources/META-INF/openapi.yaml
```diff
@@ -175,6 +175,29 @@ paths:
         - support: [ ]
         - test: [ ]
       operationId: syncSubscriptionsForContractsByOrg
+  /api/swatch-contracts/internal/rpc/contracts/sync:
+    description: "Trigger a sync for all orgs that have contracts in the upstream database."
+    put:
```
Sgitario: This is a POST operation, not a PUT. 
I know you copied this from the subscriptions sync, but the correct method is POST. Other API that use PUT are wrong.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -237,28 +238,33 @@ void shouldSyncAllContractsAcrossAllOrganizations() {
     givenContractIsCreated(firstOrgContract);
     givenContractIsCreated(secondOrgContract);
 
-    // When: Sync all contracts
+    // When: Sync all contracts (async — enqueues tasks via Kafka)
```
Sgitario: This is an implementation detail " (async — enqueues tasks via Kafka)". I know this is AI generated, but let's avoid this kind of updates that can be easily out to date.

Sgitario (changes requested):

swatch-contracts/deploy/clowdapp.yaml
TommasoLencioni: Thanks, I opened in draft this https://github.com/RedHatInsights/platform-mq/pull/513
I'll make this ready in parallel.

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsResource.java
TommasoLencioni: Right, despite being a relatively low number, it's not a good practice.
I've implemented the Streaming API, thanks for the suggestion!

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Refactored, thanks!

swatch-contracts/src/main/resources/META-INF/openapi.yaml
TommasoLencioni: Right, done!
I also created https://redhat.atlassian.net/browse/SWATCH-4901 to address all other wrong PUT uses.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -238,27 +239,11 @@ void shouldSyncAllContractsAcrossAllOrganizations() {
...
-
-    var secondOrgContracts = service.getContractsByOrgId(secondOrg);
-    assertEquals(1, secondOrgContracts.size(), "Second org should have one contract after sync");
-    assertEquals(
-        BillingProvider.AZURE.toApiModel(),
-        secondOrgContracts.get(0).getBillingProvider(),
-        "Second org contract should be Azure");
+    // Then: Verify contracts
+    thenSingleContractWithProvider(firstOrg, BillingProvider.AWS);
+    thenSingleContractWithProvider(secondOrg, BillingProvider.AZURE);
```
Sgitario: I miss the verb of this method, what is this method doing?

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -307,35 +292,28 @@ void syncAllContractsTwiceShouldNotCreateDuplicateContracts() {
...
-        uuidAfterSecond,
-        "Contract UUID must remain stable across repeat syncAllContracts calls");
-    assertEquals(
-        contractsAfterFirst.get(0).getBillingProvider(),
-        contractsAfterSecond.get(0).getBillingProvider(),
-        "Billing provider must not change across repeat runs");
+    whenSyncAllContracts();
+
+    // Then: Wait for async processing and verify no duplicates
+    thenContractAsserts(
```
Sgitario: `thenContractAsserts` means to me that we're going to assert one contract, but looking at the logic, it's asserting a list of contracts, right?

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -372,38 +350,37 @@ void syncAllContractsWithMultiProviderOrgShouldPreserveBothContracts() {
...
-            .collect(Collectors.toSet());
-    assertEquals(
-        uuidsBefore,
-        uuidsAfter,
-        "Contract UUIDs must remain stable after syncAllContracts for multi-provider org");
+    // When: syncAllContracts (async — enqueues one task for this org)
+    whenSyncAllContracts();
+
+    // Then: Wait for async processing, both contracts must survive
+    thenContractAsserts(
```
Sgitario: is not easier to read something like:

thenContractIsFound(contractIdForAws, BillingProvider.AWS);
thenContractIsFound(contractIdForAzure, BillingProvider.AZURE);
thenContractsSizeAre(2);

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -426,35 +403,36 @@ void syncAllContractsAfterOrgLevelSyncShouldNotInsertAdditionalRecords() {
...
-        "subscriptions: subscription_id must be stable after global sync");
-    assertEquals(
-        measurementCountAfterOrgSync,
-        subsAfterGlobalSync.get(0).getMetrics().size(),
-        "subscription_measurements: row count must be stable after global sync");
+    // When: Run syncAllContracts on top of the already-synced state (async)
+    whenSyncAllContracts();
+
+    // Then: Wait for async processing, no additional records in any table
+    thenContractAsserts(
```
Sgitario: this still needs to be refactored because the logic is duplicated in many tests.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -141,8 +143,16 @@ public ContractResponse createContract(ContractRequest request) {
   }
 
   @Transactional
-  public List<String> getOrgIdUsedInContracts() {
-    return contractRepository.getDistinctOrgIds();
+  public long forEachContractOrgId(Consumer<String> action) {
+    try (Stream<String> orgIds = contractRepository.getDistinctOrgIds()) {
+      long[] count = {0};
```
Sgitario: this is a hack :), do not use it. Moreover, forEachContractOrgId is not following a good pattern since we should give it a better context. 
Instead follow the same pattern than for ex in: CaptureSnapshotsTaskManager.updateSnapshotsForAllOrg:
- create a new service ContractSyncService with a new method that does all this logic: query + enqueue
- instead of using long[] use AtomicLong.

Sgitario (changes requested):
Apart from the new comments, I prefer avoiding using global contracts sync and keep using contracts sync all since:
- it's the convention we have followed in other APIs
- global could stand for updating all the contracts in the upstream services as well (IT Partner gateway) when we're only going to update the ones we have registered in our database.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Clarified method name, thanks for the heads up!

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Yes, improved method name.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
TommasoLencioni: Done, thanks for the suggestion!

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Refactored, ty!

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
```diff
@@ -277,6 +280,96 @@ protected String getContractUuid(String orgId) {
     return contracts.get(0).getUuid();
   }
 
+  protected String getContractUuidByProvider(String orgId, BillingProvider provider) {
```
Sgitario: If this class is only used in ContractsSyncComponentTest, add it there. Otherwise, BaseContractComponentTest will became too large to maintain.

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
```diff
@@ -277,6 +280,96 @@ protected String getContractUuid(String orgId) {
...
+            .filter(c -> c.getBillingProvider().equals(provider.toApiModel()))
+            .toList();
+    assertEquals(
+        1,
+        contracts.size(),
+        "Expected exactly one contract with provider " + provider + " for org " + orgId);
+    return contracts.get(0).getUuid();
+  }
+
+  protected void verifyContractFields(
```
Sgitario: If this class is only used in ContractsSyncComponentTest, add it there. Otherwise, BaseContractComponentTest will became too large to maintain.

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
```diff
@@ -277,6 +280,96 @@ protected String getContractUuid(String orgId) {
...
+            .map(com.redhat.swatch.contract.test.model.Metric::getMetricId)
+            .distinct()
+            .count();
+    assertEquals(
+        (long) metrics.size(),
+        distinctCount,
+        "contract_metrics must not contain duplicate metric_id values for the same contract");
+  }
+
+  protected void verifySubscriptionFields(
```
Sgitario: If this class is only used in ContractsSyncComponentTest, add it there. Otherwise, BaseContractComponentTest will became too large to maintain.

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
```diff
@@ -277,6 +280,96 @@ protected String getContractUuid(String orgId) {
...
+    assertNotNull(actual.getEndDate(), "contracts.end_date must be set");
+    assertEquals(
+        expected.getSubscriptionNumber(),
+        actual.getSubscriptionNumber(),
+        "contracts.subscription_number must match upstream");
+    assertFalse(actual.getProductTags().isEmpty(), "contracts.product_tags must not be empty");
+    assertNotNull(actual.getMetrics(), "contract_metrics must not be null");
+  }
+
+  protected void verifyContractMetricsAreValid(
```
Sgitario: If this class is only used in ContractsSyncComponentTest, add it there. Otherwise, BaseContractComponentTest will became too large to maintain.

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsResource.java
```diff
@@ -70,12 +72,10 @@
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
-import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 @ApplicationScoped
-@AllArgsConstructor
```
Sgitario, line 78: Now, you don't use EmitterService, we can keep using `@AllArgsConstructor` to avoid having a constructor that injects too many services.

swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/ContractsResourceTest.java
```diff
@@ -83,14 +86,21 @@ class ContractsResourceTest {
   private static final String ORG_ID = "org123";
   private static final String ANOTHER_ORG_ID = "org456";
   private static final String ROSA = "rosa";
-  private static final String SYNC_ALL_CONTRACTS_ENDPOINT = "/internal/rpc/syncAllContracts";
+  private static final String SYNC_ALL_CONTRACTS_ENDPOINT =
+      "/api/swatch-contracts/internal/rpc/contracts/sync";
+  private static final String SWATCH_PSK_HEADER = "x-rh-swatch-psk";
```
Sgitario, line 91: This constant is really defined (and hardcoded in SwatchPskHeaderFilter). I think it's good to reuse the same constant in both places, so we can easily see how it's used.

Sgitario (changes requested):

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
TommasoLencioni: Moved, thanks.

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
TommasoLencioni: Moved, thanks.

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
TommasoLencioni: Moved, thanks.

swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/ContractsResourceTest.java
TommasoLencioni: Done, thanks for the suggestion!

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsResource.java
TommasoLencioni: Done, thanks for the suggestion!

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
TommasoLencioni: Moved, thanks.

mmusil (approved):
LGTM, but @Sgitario's approval has more value here, hahaha

Sgitario (approved):
LGTM. Many thanks @TommasoLencioni for addressing all my comments.

PR #5999, opened 2026-04-16
Liquibase utility updates

awood:
Jira issue: None

## Description
The application portion of the `swatch-database` module has been removed.  That code allowed us to easily execute Liquibase tasks during development.  To replace the now missing functionality, I've added in the Liquibase Maven plugin and removed an entry in the Makefile that is obsolete.

Being able to run Liquibase rollbacks, diffs, and the like is very helpful.  Especially since just running the Liquibase CLI is not easy.  On my machine, to use the CLI, I had to checkout Liquibase, run a maven install, unpack the distribution zip, add 2 missing JARs to the classpath, and set the search path correctly. 

### Steps
1. `./mvwn -pl swatch-tally liquibase:status`
1. `./mvnw -pl swatch-billable-usage liquibase:status`
1. `./mvnw -pl swatch-contracts liquibase:status`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ SHELL=/bin/bash
 	install \
 	test \
 	clean \
-	rollback \
 	status
 
 # Add a profile(s) to use like so:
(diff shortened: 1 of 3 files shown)


swatch-billable-usage/pom.xml
```diff
@@ -164,6 +164,17 @@
...
+      <plugin>
+        <groupId>org.liquibase</groupId>
+        <artifactId>liquibase-maven-plugin</artifactId>
+        <configuration>
+          <changeLogFile>db/changelog.xml</changeLogFile>
+          <url>jdbc:postgresql://localhost:5432/rhsm-subscriptions</url>
+          <username>rhsm-subscriptions</username>
+          <emptyPassword/>
+        </configuration>
+      </plugin>
```
Sgitario: I personally never needed this plugin, and I think none of the lightning team did. 
In order to keep the Maven configuration setup minimal, I would ask to keep this change only in swatch-tally.

Sgitario (changes requested):

swatch-billable-usage/pom.xml
awood: Done

lindseyburnett (approved):

Sgitario (approved):

PR #6012, opened 2026-04-20
SWATCH-4509: Set IMPORTANT severity for over usage notifications

Sgitario:
Jira issue: SWATCH-4509

## Description
Updated existing logic for over usage notifications to set the IMPORTANT severity.

## Testing
Added unit tests and new component test to cover the change.

### DIFF
--- a/swatch-utilization/src/main/java/com/redhat/swatch/utilization/model/Severity.java
+++ b/swatch-utilization/src/main/java/com/redhat/swatch/utilization/model/Severity.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


swatch-utilization/src/main/java/com/redhat/swatch/utilization/model/Severity.java
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package com.redhat.swatch.utilization.model;
+
+public enum Severity {
```
TommasoLencioni, line 23: I would either enumerate only used severity or the correct ones from https://www.patternfly.org/component-groups/status-and-state-indicators/severity/.

TommasoLencioni (changes requested):

swatch-utilization/src/main/java/com/redhat/swatch/utilization/model/Severity.java
Sgitario: These are the correct ones, see the document: https://inscope.corp.redhat.com/docs/default/component/notifications-app/user-guide/send-notification/#description-of-the-message-format-and-samples (the info note: "Including severity levels").

TommasoLencioni (approved):

PR #6014, opened 2026-04-20
Unleash config update

wottop:
## Description

An update to the configuration was needed because the value was not properly assembled for the unleash api url for the clowder values

### Verification

Run swatch tally locally and toggle the unleash flag as per the description in https://github.com/RedHatInsights/rhsm-subscriptions/pull/5969

Note that the final verification will be in stage after this is merged to main.

### DIFF
--- a/swatch-core/src/main/resources/swatch-core/application.yaml
+++ b/swatch-core/src/main/resources/swatch-core/application.yaml
@@ -48,6 +48,12 @@ DATABASE_CONNECTION_TIMEOUT_MS: 30000
 DATABASE_MAX_POOL_SIZE: 10
 POSTGRES_STATEMENT_TIMEOUT: 5min
 
+UNLEASH_HOST: ${clowder.featureFlags.hostname:localhost}
+UNLEASH_PORT: ${clowder.featureFlags.port:4242}
+UNLEASH_SCHEME: ${clowder.featureFlags.scheme:http}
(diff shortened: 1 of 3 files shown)


swatch-core/src/main/resources/swatch-core/application.yaml
```diff
@@ -48,6 +48,12 @@ DATABASE_CONNECTION_TIMEOUT_MS: 30000
 DATABASE_MAX_POOL_SIZE: 10
 POSTGRES_STATEMENT_TIMEOUT: 5min
 
+UNLEASH_HOST: ${clowder.unleash.hostname:localhost}
```
awood: I believe this will need to be `clowder.featureFlags.hostname` unless ClowderJsonPropertySource explicitly changes "featureFlags" to "unleash".

swatch-core/src/main/resources/swatch-core/application.yaml
```diff
@@ -48,6 +48,12 @@ DATABASE_CONNECTION_TIMEOUT_MS: 30000
...
+UNLEASH_PORT: ${clowder.unleash.port:4242}
+UNLEASH_SCHEME: ${clowder.unleash.scheme:http}
+UNLEASH_API_TOKEN: ${clowder.unleash.clientAccessToken:default:development.unleash-insecure-api-token}
```
awood: We'll want to make sure the additional colons don't throw a monkey wrench into things.

awood (changes requested):

awood (approved):

PR #6018, opened 2026-04-21
SWATCH-4840: update send-notifications feature flag to allow denylist event types

Sgitario:
Jira issue: SWATCH-4840

## Description
Changes:
- Centralize notification gating in `NotificationsProducer` so every `Action` is evaluated in one place (not only over-usage).
- Extend the `swatch.swatch-notifications.send-notifications` Unleash toggle with an optional variant `send-notifications-config` whose JSON payload may include `event_types_denylist` to suppress specific notification event types while the toggle remains enabled.
- Preserve existing behavior when the variant name or payload does not match, and keep org-level bypass via `swatch.swatch-notifications.send-notifications-orgs-allowlist` (allowlist OR still overrides per-event suppression for that org).

## Testing
- Unit: `FeatureFlagsTest`, `NotificationsProducerTest`; over-usage tests no longer stub `FeatureFlags` for send/allowlist behavior.
- Component: `UtilizationNotificationsFeatureFlagsComponentTest` replaces the allowlist-only suite and adds denylist / bypass coverage; `UtilizationUnleashService` helpers configure Unleash variants for CT.
- Manual / release: `swatch-utilization/TEST_PLAN.md` updated with TC001–TC006 and the new case IDs.


swatch-utilization/TEST_PLAN.md
```diff
@@ -333,3 +338,55 @@ Kafka messages can be injected for event-driven testing.
...
+- **Action**:
+    - Generate enough usage to exceed the threshold for the metric A of the product B
+    - Trigger utilization calculation process
+- **Verification**:
+    - Wait for notification message on notifications topic
+    - Verify notification payload
+- **Expected Result**:
+    - Notification event contains correct information (org_id, product_id, metric_id and utilization_percentage)
+
+**utilization-notifications-featureflags-TC006 - Single allowlisted org bypasses denylisted event type**
```
TommasoLencioni: Could be useful showing both one org in and one out of the allowlist and how the denylist affects the two?

Sgitario: PR just updated with it.

TommasoLencioni (approved):

PR #6024, opened 2026-04-22
SWATCH-4791: GET API for notifications UI to get the user's preferences.

Mantzounis:
Jira issue: SWATCH-4791

## Description
This PR adds a GET endpoint to retrieve per-organization over-usage threshold preferences that notification UI will use to get custom threshold preferences.

Since the UI does not handle 404, we are going to set a default threshold of 80 and always return 200, this has been discussed with product before implementing.

## Testing
IQE tests and component tests are part of the parent task [SWATCH-4789](https://redhat.atlassian.net/browse/SWATCH-4789) and will be implemented later.

[SWATCH-4789]: https://redhat.atlassian.net/browse/SWATCH-4789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swatch-utilization/src/main/java/com/redhat/swatch/utilization/resources/OrgPreferencesResource.java
+++ b/swatch-utilization/src/main/java/com/redhat/swatch/utilization/resources/OrgPreferencesResource.java
@@ -42,6 +42,13 @@ public class OrgPreferencesResource implements OrgPreferencesApi {
 
   @Override
   @RolesAllowed({"customer", "service"})
+  public OrgPreferencesResponse getOrgPreferences(String xRhIdentity) {
+    String orgId = orgIdResolver.getOrgId(securityContext, httpHeaders);
+    return orgPreferencesService.getOrgPreferences(orgId);
+  }
+
+  @Override
+  @RolesAllowed({"customer", "service"})
   public OrgPreferencesResponse updateOrgPreferences(
       String xRhIdentity, OrgPreferencesRequest orgPreferencesRequest) {
     String orgId = orgIdResolver.getOrgId(securityContext, httpHeaders);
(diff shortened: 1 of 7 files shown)


swatch-utilization/src/main/java/com/redhat/swatch/utilization/resources/OrgPreferencesResource.java
```diff
@@ -40,6 +41,18 @@ public class OrgPreferencesResource implements OrgPreferencesApi {
   @Inject OrgIdResolver orgIdResolver;
   @Inject OrgPreferencesService orgPreferencesService;
 
+  @Override
+  @RolesAllowed({"customer", "service"})
+  public OrgPreferencesResponse getOrgPreferences(String xRhIdentity) {
+    String orgId = orgIdResolver.getOrgId(securityContext, httpHeaders);
+    var response = orgPreferencesService.getOrgPreferences(orgId);
+    if (response == null) {
```
Sgitario: Spite returning 404 is correct, my concern is that this will bring an incompatibility when being used by the user-preferences-frontend (repo https://github.com/RedHatInsights/user-preferences-frontend).
Unlikely in advisor, when the GET API always return 200, if we return 404, the UI will prompt an Error window to users.

Sgitario (changes requested):
The decision of using 404 vs returning a default threshold is incompatible with how this API will be used by https://github.com/RedHatInsights/user-preferences-frontend. 

We need to clarify with the user preferences UI team if returning 404 is ok before proceeding with these changes. 

Alternatively, we could work as the Advisor API and always return 200.

Sgitario (approved):

PR #6026, opened 2026-04-22
SWATCH-4563: Standardize log for alerting

TommasoLencioni:
Jira issue: [SWATCH-4563](https://redhat.atlassian.net/browse/SWATCH-4563)

## Description

This PR standardize contracts sync all logs. SUMO alert [https://gitlab.cee.redhat.com/rhsm/swatch/swatch-sumo-content/-/merge_requests/4](https://gitlab.cee.redhat.com/rhsm/swatch/swatch-sumo-content/-/merge_requests/4) will use these logs.
Custom metric is removed since it's not used.

[SWATCH-4563]: https://redhat.atlassian.net/browse/SWATCH-4563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


swatch-contracts/src/test/java/com/redhat/swatch/contract/service/ContractSyncTaskConsumerTest.java
```diff
@@ -38,54 +37,47 @@
...
   }
 
   @Test
-  void testConsumeCallsSyncContractsByOrgId() {
+  void shouldInvokeSyncContractsForOrgId() {
     when(contractService.syncContractsByOrgId("org123"))
         .thenReturn(new StatusResponse().status(ContractService.SUCCESS_MESSAGE));
 
-    consumer.consumeFromTopic(new ContractSyncTask("org123"));
+    Assertions.assertDoesNotThrow(() -> consumer.consumeFromTopic(new ContractSyncTask("org123")));
```
vbusch: I feel that removing the `Assertions.assertDoesNotThrow` would be clearer.  Simply call `consumer.consumeFromTopic(new ContractSyncTask("org123")));`
If an error were to be thrown, the test would fail anyways.

vbusch (dismissed):
Looks good to me.  One very minor optional suggestion.

swatch-contracts/docs/monitoring/sumo-alert-contract-sync-all-org-failures.md
```diff
@@ -0,0 +1,24 @@
+# Sumo Logic: org failures during contracts sync all
```
Sgitario: do we need this document here? 
adding docs like this need maintainance in the long term and this looks like more like an investigation document rather than something we want to include in the repo.

Sgitario (changes requested):
Apart from the comment, it's not clear to me what is the problem that these changes fix. 
I don't understand terms like Sumo-ready logs and `ContractSyncTaskConsumer prefixes fan-out logs with contract-sync-all: and removes Micrometer outcome counters.` sounds super weird to me: having custom metrics should never remove the micrometer counters -> I'm super this is not what it's happening.

Note that I'm ok with removing the custom metrics since it seems we don't need it.

swatch-contracts/docs/monitoring/sumo-alert-contract-sync-all-org-failures.md
TommasoLencioni: Right, not needed, removed.

swatch-contracts/src/test/java/com/redhat/swatch/contract/service/ContractSyncTaskConsumerTest.java
TommasoLencioni: Done, thanks for the feedback!

Sgitario (approved):

vbusch (approved):

PR #6030, opened 2026-04-23
SWATCH-4732: Add heap dump monitor with S3 upload for swatch-utilization

vbusch:
Jira issue: SWATCH-4732

## Description
Add automated heap dump monitoring for OOM debugging 

Implements proactive heap dump capture when container memory reaches 75% threshold to diagnose production OOM issues before pods are killed.

### How
  - **heap-monitor.sh**: Background monitor that polls container memory every 10s
  - **Periodic capture**: NMT snapshots and JVM metrics every 15 minutes
  - **Threshold capture**: Heap dump when memory usage reaches 75% of container limit
  - **Cooldown**: After dump, waits for memory to stay below 60% for 5 minutes before re-triggering
  - **Compression**: Heap dumps gzipped (~75% size reduction: 60MB → 15-30MB)
  - **Upload**: Compressed dumps uploaded to S3/MinIO, then deleted from EmptyDir
  - **JVM OOM dumps**: Automatically processes dumps created by `-XX:+HeapDumpOnOutOfMemoryError`
  - **EmptyDir persistence**: Files survive container restarts but not pod restarts

### Dual Credential Strategy
 - **Ephemeral**: Uses Clowder objectStore (MinIO auto-provisioned)
 - **Stage/Prod**: Uses mounted `heapdump-s3` secret from app-interface

### JVM Options - Prevents JVM from using container defaults, leaves headroom for metaspace, direct memory, thread stacks, and heap monitor overhead.
- `-Xmx150m` - Explicit heap cap
- `-XX:MaxDirectMemorySize=32m` / `48m` - Cap NIO buffers
- `-XX:+HeapDumpOnOutOfMemoryError` - JVM creates dump on OOM
- `-XX:HeapDumpPath=/heapdumps/` - Save to EmptyDir volume
- `-XX:NativeMemoryTracking=summary` - Enable native memory tracking

### Ephemeral
- The swatch-utilization server logs startup of memory monitor
- When heap dump is transferred use the following command to do an ls and cp
POD=$(oc get pods -l app=swatch-utilization -o jsonpath='{.items[0].metadata.name}')

### List heap dumps in MinIO  (telepresence required)
aws s3 ls s3://swatch-utilization-heapdumps/swatch-utilization/heapdumps/ --recursive --endpoint-url http://env-ephemeral-krtdbl-minio.ephemeral-krtdbl.svc:9000

### Download a heap dump
aws s3 cp s3://swatch-utilization-heapdumps/swatch-utilization/heapdumps/swatch-utilization-service-fffdddb5-hnvp9/jvm-metrics-periodic.txt . --endpoint-url http://env-ephemeral-krtdbl-minio.ephemeral-krtdbl.svc:9000

### DIFF
--- a/swatch-utilization/deploy/clowdapp.yaml
+++ b/swatch-utilization/deploy/clowdapp.yaml
@@ -89,10 +89,14 @@ parameters:
   - name: EGRESS_IMAGE_TAG
     value: latest
   - name: USER_JAVA_OPTS
-    # Heap dumps triggered by heap monitor at 75% threshold
+    # Heap dumps triggered by heap monitor at 75% threshold (leaves headroom for compression + upload)
     value: '-Xmx150m -XX:MaxDirectMemorySize=32m -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/heapdumps/ -XX:NativeMemoryTracking=summary'
   - name: ENABLE_HEAP_MONITOR
     value: 'true'
+  - name: HEAP_MONITOR_THRESHOLD_PERCENT
+    value: '60'
+  - name: HEAP_MONITOR_RESET_PERCENT
+    value: '55'
   - name: HEAP_DUMP_UPLOAD_S3
     value: 'false'
   - name: HEAP_DUMP_S3_PREFIX
@@ -100,6 +104,7 @@ parameters:
   - name: HEAP_DUMP_S3_BUCKET_NAME
     value: swatch-utilization-heapdumps
   - name: HEAP_DUMP_S3_SECRET_NAME
+    # Secret name for S3 credentials (stage/prod: 'heapdump-s3' from app-interface)
     value: heapdump-s3
 
 objects:
@@ -248,6 +253,10 @@ objects:
               value: ${USER_JAVA_OPTS}
             - name: ENABLE_HEAP_MONITOR
               value: ${ENABLE_HEAP_MONITOR}
+            - name: HEAP_MONITOR_THRESHOLD_PERCENT
+              value: ${HEAP_MONITOR_THRESHOLD_PERCENT}
+            - name: HEAP_MONITOR_RESET_PERCENT
+              value: ${HEAP_MONITOR_RESET_PERCENT}
             - name: HEAP_DUMP_UPLOAD_S3
               value: ${HEAP_DUMP_UPLOAD_S3}
             - name: HEAP_DUMP_S3_PREFIX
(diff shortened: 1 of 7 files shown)


Sgitario (changes requested):
Modifying the container images to bundle tools like AWS CLI and jq is a security risk and also widens the CVE footprint the team must own through release and compliance.

vbusch:
> Modifying the container images to bundle tools like AWS CLI and jq is a security risk and also widens the CVE footprint the team must own through release and compliance.

> Modifying the container images to bundle tools like AWS CLI and jq is a security risk and also widens the CVE footprint the team must own through release and compliance.

Replaced the AWS CLI with a Java S3Uploader.  The project already uses the dependencies that it requires.

I removed the dependency on `jq`.  It was used for ephemeral testing with the wiremock.  But I added a container test instead.

Sgitario (changes requested):
## Concerns:
- Always adding gzip, two shell scripts, and overriding ENTRYPOINT increases attack surface, and maintenance. 
- Heap dumps and JVM tuning are split between image (wrapper + scripts), deploy (env/volumes), and platform (secrets in app-interface). The story for “how another service turns this on” should be one short checklist, not three places.
- Hard to enable/port these settings to other services. 

## Suggestions:
- Gate everything on ENABLE_HEAP_MONITOR (and keep defaults off)

When false: no wrapper behavior, no extra volumes, no heap-specific env (or only a single feature flag). 

- Compress in Java (e.g. GZIPOutputStream while streaming to S3) and drop gzip from the image entirely.

- Avoid a permanent ENTRYPOINT override if possible: prefer mechanisms that don’t fork from the stock JVM image entrypoint.

- Optional ConfigMap for “all the extra properties”.

## Next actions:
- Before doing any changes, share a live demo with the team and decide the approach to follow: either modifying the classpath+yaml+scripts+image, or using cryostat with agent and smart triggers. 

I think with the feedback from this session, we could leverage the next actions and close this proof of concept.

vbusch:
Thank you @Sgitario! I'll demo this at today's sprint review for both teams to provide feedback.

I've already evaluated some of these alternatives:
- Gating - Everything is already gated on ENABLE_HEAP_MONITOR (default: false).  Toggle requires pod restart currently, But configMap watching is possible to avoid the restart.      
- Absolutely, java could do the zip.
- A ConfigMap would keep the properties together, and make it possible to toggle without a restart.

The ENTRYPOINT is interesting:
- Cryostat might be best here.
- The sidecar implementation didn't require the ENTRYPOINT.  I have one that uses python (which brings in more security issues).  But had more complexity.   Could be switched to java.
- The side car used more resources.
- Being in a different container outside of the JMV the following would be required:
-- jcmd with shareProcessNamespace: true  OR
-- expose an endpoint to give the memory dump.  This could be a good idea

Sgitario (approved):

vbusch:
Merging to test it on stage.  Will get the results and then discuss.

PR #6035, opened 2026-04-23
SWATCH-4854: Add instance report component tests (PAYG)

liwalker-rh:
Jira issue: SWATCH-4854

## Description
These changes add new test coverage for instances report API with PAYG products.

## Testing

IQE Test MR (for non-PAYG): https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/1502

### Setup
1. Clean before running tests:
```
./mvnw clean install -DskipTests
```
2. Bring up services:
```
podman compose up -d
```

### Steps
1. Run all the instances report tests:
```
./mvnw clean install -Pcomponent-tests -pl swatch-tally/ct -am -Dtest=tests.TallyInstancesBillingAccountIdsTest -Dsurefire.failIfNoSpecifiedTests=false
```
```
./mvnw clean install -Pcomponent-tests -pl swatch-tally/ct -am -Dtest=tests.TallyInstancesReportFiltersPaygTest -Dsurefire.failIfNoSpecifiedTests=false
```
```
./mvnw clean install -Pcomponent-tests -pl swatch-tally/ct -am -Dtest=tests.TallyInstancesReportPaginationAndSortTest -Dsurefire.failIfNoSpecifiedTests=false
```

### Verification
1. All 23 instances report API tests should pass

- TallyInstancesBillingAccountIdsTest: 5 tests
- TallyInstancesReportFiltersPaygTest: 12 tests
- TallyInstancesReportPaginationAndSortTest: 6 tests


wottop (dismissed):

swatch-tally/ct/java/tests/TallyInstancesReportFiltersPaygTest.java
TrayvonMcKnight, line 1: Since this files contains the filtering tests, we should be able to cut down the number of events to like 3 or 4 and we can filter on those.

TrayvonMcKnight:
I think you should also look into maybe using the Records? I think it might help with the TCxxx naming of the variables. It's what was done for the TallyReport Tests.

swatch-tally/ct/java/tests/TallyInstancesReportFiltersPaygTest.java
liwalker-rh: So I did try this for the instances tests and it did not work as well I think due to differences in how the apis and queries work for instances vs. tally reports. It was fine to have a few events for tally reports since its totals are rolled up but I think for instances the tests need separate rows and distinct instances so they do not step on each other. Maybe there is another way we can do this in the future but for now this is probably the best way to do it (and at least there is only one tally sync per test file). I will definitely implement the records though for better naming of variables.

liwalker-rh: Added the use of records in all the new test files.

TrayvonMcKnight (approved):
Tests pass for me. LGTM

PR #6060, opened 2026-04-29
SWATCH-4805: Add custom usage threshold evaluation service

TommasoLencioni:
## Description

This PR proposes to introduce `CustomUsageThresholdService`, that evaluates utilization summaries against org-specific thresholds for proactive notification. Extract shared logic into `BaseThresholdService` base class and rename `CustomerOverUsageService` to `OverUsageThresholdService`.

The custom threshold service:
- Looks up org preferences from `org_utilization_preference` table
- Sends proactive-utilization-notification with `MODERATE` severity
- Includes SHA-256 hashed `last_updated` for deduplication
- Skips orgs without explicit threshold preferences
- Only runs when usage percentage is >100%

Assertion object changed from `Action` to `Event type` of that action to make sure the right event is produced (over usage vs proactive).

Jira issue: [SWATCH-4805](https://redhat.atlassian.net/browse/SWATCH-4805)

### Verification

Run unit tests:
`mvn test -pl swatch-utilization`

[SWATCH-4805]: https://redhat.atlassian.net/browse/SWATCH-4805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vbusch (dismissed):
The changes look good to me.  Our pipelines are being difficult, but I think the following 4 iqe tests need to be looked at:
- swatch_utilization.test_notification_for_hourly_and_daily_tallies.test_contract_hourly_tally_emits_notification_message
- swatch_utilization.test_notification_for_hourly_and_daily_tallies.test_nonpayg_subscriptions_tally_emits_notification_message
- swatch_utilization.test_notification_for_hourly_and_daily_tallies.test_over_usage_notification_requires_payg_sla_aligned_with_offering
- swatch_utilization.test_notification_for_hourly_and_daily_tallies.test_over_usage_notification_requires_payg_usage_aligned_with_offering

vbusch (commented):
Waiting for the tests.

swatch-utilization/ct/java/tests/OverUsageThresholdComponentTest.java
```diff
@@ -323,7 +320,7 @@ void thenNotificationShouldBeSent() {
     // Calculate expected utilization percentage
     Measurement measurement = utilizationSummary.getMeasurements().get(0);
     double expectedUtilizationPercent =
-        (measurement.getCurrentTotal() / measurement.getCapacity()) * 100.0;
+        (measurement.getValue() / measurement.getCapacity()) * 100.0;
```
Sgitario: The right field here is `currentTotal`, not value. 
At the moment, we only use `value` for gauge metrics because this logic is wrong in swatch-tally and will change soon.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/AbstractThresholdService.java
```diff
@@ -0,0 +1,205 @@
...
+import io.micrometer.core.instrument.Meter.MeterProvider;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.OptionalDouble;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public abstract class AbstractThresholdService {
```
Sgitario: As a naming convention, we don't use `Abstract` in this project, but `Base`.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/BaseThresholdService.java
```diff
@@ -0,0 +1,205 @@
...
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.OptionalDouble;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public abstract class AbstractThresholdService {
+
+  public static final String DIMENSION_SENTINEL = "_ANY";
```
Sgitario: I'm not sure what do you mean with `SENTINEL`, but to reduce the comprehension hardened, simply use `ANY` please.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/BaseThresholdService.java
```diff
@@ -0,0 +1,205 @@
...
+  private static final String APPLICATION = "subscriptions";
+  private static final boolean NOTIFY_ONLY_ADMINS = false;
+  private static final boolean IGNORE_USER_PREFERENCES = false;
+
+  private final NotificationsProducer notificationsProducer;
+  private final MeterProvider<Counter> counter;
+
+  protected AbstractThresholdService() {
+    this.counter = null;
+    this.notificationsProducer = null;
```
Sgitario: why do we have an empty constructor here? please remove this one since we should not build instances that leave the instance in a wrong state.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomUsageThresholdService.java
```diff
@@ -0,0 +1,131 @@
...
+  public CustomUsageThresholdService(
+      MeterRegistry meterRegistry,
+      NotificationsProducer notificationsProducer,
+      OrgUtilizationPreferenceRepository preferenceRepository) {
+    super(meterRegistry, notificationsProducer, CUSTOM_THRESHOLD_METRIC);
+    this.preferenceRepository = preferenceRepository;
+  }
+
+  @Transactional
+  public void check(UtilizationSummary payload, Measurement measurement) {
```
Sgitario: The abstract pattern was not implemented correctly here, it's more like a base classes with shared methods. 
To correctly implement this pattern, the `check` method should be implemented in the abstract and have placeholder (abstract methods) that are implemented in the different strategies.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomUsageThresholdService.java
```diff
@@ -0,0 +1,131 @@
...
+    var preferenceOpt = preferenceRepository.findByIdOptional(payload.getOrgId());
+    if (preferenceOpt.isEmpty()) {
+      log.debug(
+          "No org preference found for orgId={}, skipping custom threshold check",
+          payload.getOrgId());
+      return;
+    }
+
+    var preference = preferenceOpt.get();
+    var utilizationOpt = computeUtilizationPercent(payload, measurement);
```
Sgitario: I think it will be cheaper to reverse the order of (1) getting the preference organizations and (2) check the utilization percentage. Since (1) requires a SQL query (expensive) when many times won't be necessary.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomUsageThresholdService.java
```diff
@@ -0,0 +1,131 @@
...
+      String lastUpdatedHash = hashLastUpdated(preference.getLastUpdated());
+      sendNotification(
+          payload,
+          metricId,
+          EVENT_TYPE,
+          Severity.MODERATE,
+          buildEvent(utilizationPercent, lastUpdatedHash));
+    } else {
+      log.debug(
+          "Usage within custom threshold: orgId={} productId={} metricId={} utilizationPercent={}% threshold={}%",
```
Sgitario: I don't think we need this log statement ... if you feel strong about keeping it, please update the log to state that the condition for the custom threshold was not met or something like this, otherwise it might be confusing.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
```diff
@@ -0,0 +1,131 @@
...
+          "Usage within custom threshold: orgId={} productId={} metricId={} utilizationPercent={}% threshold={}%",
+          payload.getOrgId(),
+          payload.getProductId(),
+          measurement.getMetricId(),
+          String.format(PERCENT_FORMAT, utilizationPercent),
+          threshold);
+    }
+  }
+
+  static String hashLastUpdated(OffsetDateTime lastUpdated) {
```
Sgitario, line 94: we don't need to implement this, please use `DigestUtils.sha256Hex` instead (I confirmed that commons-codec is already available in swatch-utilization).

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomUsageThresholdService.java
```diff
@@ -0,0 +1,131 @@
...
+    try {
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      byte[] hash = digest.digest(lastUpdated.toString().getBytes(StandardCharsets.UTF_8));
+      return HexFormat.of().formatHex(hash);
+    } catch (NoSuchAlgorithmException e) {
+      throw new IllegalStateException("SHA-256 not available", e);
+    }
+  }
+
+  private Event buildEvent(double utilizationPercent, String lastUpdatedHash) {
```
Sgitario: This method is one candidate to be added in the base class, so we ensure all the events are built the same way (with custom modifications).

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/OverUsageThresholdService.java
```diff
@@ -0,0 +1,123 @@
...
+import com.redhat.swatch.utilization.model.UtilizationSummary;
+import io.micrometer.core.instrument.MeterRegistry;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@Slf4j
+@ApplicationScoped
+public class OverUsageThresholdService extends AbstractThresholdService {
```
Sgitario: We're losing all the git history because you deleted the old class `CustomerOverUsageService` and created this one. Please, use `git mv` instead of removing one file and creating a new one.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/UtilizationSummaryConsumer.java
```diff
@@ -50,7 +51,10 @@ public void process(UtilizationSummary payload) {
 
     for (var measurement : payload.getMeasurements()) {
       if (measurementValidator.isMeasurementValid(payload, measurement)) {
-        customerOverUsageService.check(payload, measurement);
+        boolean overUsageDetected = overUsageThreshold.check(payload, measurement);
+        if (!overUsageDetected) {
+          customUsageThreshold.check(payload, measurement);
```
Sgitario: I think both services should be independent from each other: we iterate over all the services (there are two today), the custom threshold should include a condition to not send anything if the usage is over 100%. 

This is because this class is agnostic about threshold usages and in the future, there will be other kind of notifications (like subscription not found).

Sgitario (changes requested):
I did a first round of review here.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/UtilizationSummaryConsumer.java
TommasoLencioni: That's a very good point, implemented, thanks!

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/UtilizationSummaryConsumer.java
```diff
@@ -34,7 +35,7 @@ public class UtilizationSummaryConsumer {
 
   @Inject UtilizationSummaryPayloadValidator payloadValidator;
   @Inject UtilizationSummaryMeasurementValidator measurementValidator;
-  @Inject CustomerOverUsageService customerOverUsageService;
+  @Inject Instance<BaseThresholdService> thresholdServices;
```
Sgitario: In the future, we will need a more generic service (interface), something like: UtilizationHandlerService. I think that establish a good convention now, it will avoid more renames in the future. 
Also, I would rename the `check` method to `handle` since `check` means to me like won't do anything, but it will only check something.

Sgitario (changes requested):

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/UtilizationSummaryConsumer.java
TommasoLencioni: Makes sense, done, thanks!

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/UtilizationSummaryConsumer.java
```diff
@@ -34,7 +35,7 @@ public class UtilizationSummaryConsumer {
 
   @Inject UtilizationSummaryPayloadValidator payloadValidator;
   @Inject UtilizationSummaryMeasurementValidator measurementValidator;
-  @Inject CustomerOverUsageService customerOverUsageService;
+  @Inject Instance<UtilizationHandlerService> thresholdServices;
```
Sgitario:
```suggestion
  @Inject Instance<UtilizationHandlerService> handlers;
```

Sgitario (changes requested):
Only a minor thing.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomUsageThresholdService.java
```diff
@@ -0,0 +1,114 @@
...
+
+  static final String EVENT_TYPE = "exceeded-custom-utilization-threshold";
+
+  @Inject OrgUtilizationPreferenceRepository preferenceRepository;
+
+  @Transactional
+  @Override
+  public boolean handle(UtilizationSummary payload, Measurement measurement) {
+    return super.handle(payload, measurement);
+  }
```
Sgitario: I didn't see this change in my previous reviews. 
Here, you are forcing a dependency on the parent class's structure. If the signature of super.handle changes in the future, this "empty" override becomes a maintenance liability. Moreover, by annotating the method, the whole process is wrapped in a transaction and if it performs heavy computations, external API calls, or non-DB logic, you are holding onto a database connection much longer than necessary.

My suggestion is to not override the method, and move the annotation into the repository class level (which is closer to the data) or create a new method with the annotation in the repository:

```
@Transactional
public Optional<OrgPreference> getPreferences(String orgId) {
    return preferenceRepository.findByIdOptional(orgId);
}
```

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/BaseThresholdService.java
```diff
@@ -0,0 +1,250 @@
...
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@Slf4j
+public abstract class BaseThresholdService implements UtilizationHandlerService {
```
Sgitario:
```suggestion
public abstract class BaseThresholdUtilizationHandlerService implements UtilizationHandlerService {
```

Normally, the implementions of an interface should also inherit the name (or part of it), so it's clear the hierarchy only by reading the class names.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomUsageThresholdService.java
```diff
@@ -0,0 +1,114 @@
...
+import jakarta.inject.Inject;
+import jakarta.transaction.Transactional;
+import java.time.OffsetDateTime;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.codec.digest.DigestUtils;
+
+@Slf4j
+@ApplicationScoped
+public class CustomUsageThresholdService extends BaseThresholdService {
```
Sgitario:
```suggestion
public class CustomThresholdUtilizationHandlerService extends BaseThresholdService {
```

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/OverUsageThresholdService.java
```diff
@@ -0,0 +1,97 @@
...
+import com.redhat.swatch.utilization.model.Measurement;
+import com.redhat.swatch.utilization.model.Severity;
+import com.redhat.swatch.utilization.model.UtilizationSummary;
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@ApplicationScoped
+public class OverUsageThresholdService extends BaseThresholdService {
```
Sgitario:
```suggestion
public class OverThresholdUtilizationHandlerService extends BaseThresholdService {
```

Sgitario (changes requested):
Changes look good to me, thanks for your hard work on this!

I have a new comment to address about the transaction, and I suggested to rename the classes. Finally, we also need to remove the dependency we have between the custom and the over usage services according to Gagan's comment where he wants both handlers act independently (even if this implies we send two notifications).

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomUsageThresholdService.java
TommasoLencioni: That's a good point that could lead to inefficiency, fixed, thanks for spotting that! :pray:

Sgitario (approved):

vbusch (approved):
Looks good!

PR #6062, opened 2026-04-30
SWATCH-4782: Add panels in Grafana to display message consumption from IT Partner contracts channels

Sgitario:
Jira issue: SWATCH-4782

## Description
Added two dashboards:
- to see the number of messages received from both UMB and Kafka (so we can easily compare it) grouped by day (this is configurable). 
- to see the number of failures from both UMB and Kafka similarly to the above one. 

### Verification
1. Log in with oc
2. Run `oc extract -f .rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml --confirm`
3. Create a new empty dashboard on your folder in Gragana
4. Import generated `subscription-watch.json`
5. Check new dashboards with title "IT Partner - Contracts Messages Consumed" and "IT Partner - Contracts Message Failures"

Screenshoot:
Screenshot 2026-04-30 at 07 05 33

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
@@ -27,7 +27,7 @@ data:
       "editable": true,
       "fiscalYearStartMonth": 0,
       "graphTooltip": 1,
-      "id": 1102302,
+      "id": 1102446,
       "links": [],
       "panels": [
         {
@@ -300,7 +300,7 @@ data:
                 "h": 7,
                 "w": 6,
                 "x": 0,
-                "y": 190
+                "y": 245
               },
               "id": 113,
               "options": {
@@ -400,7 +400,7 @@ data:
                 "h": 7,
                 "w": 6,
                 "x": 6,
-                "y": 190
+                "y": 245
               },
               "id": 115,
               "options": {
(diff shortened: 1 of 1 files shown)


.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
```diff
@@ -4092,1056 +4092,1306 @@ data:
...
+                    "value": 80
+                  }
+                ]
+              }
+            },
+            "overrides": [
+              {
+                "matcher": {
+                  "id": "byName",
+                  "options": "contracts"
```
TommasoLencioni, line 4773: Both UMB and Kafka channels originate from the IT Partner Gateway, so the naming is asymmetric. `contracts` doesn't communicate its source while `contracts-from-gateway` does. Consider a task to rename both consistently (something like `contracts-from-gateway-umb` and `contracts-from-gateway-kafka`).

TommasoLencioni (approved):

.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
Sgitario: we're going to remove the UMB one, the right one is `contracts-from-gateway` which is message broker agnostic.

PR #6064, opened 2026-04-30
SWATCH-4784: Add feature flag to the IT Partner Gateway Kafka/UMB consumers

Sgitario:
Jira issue: SWATCH-4784

## Description
Update the IT Partner Gateway Kafka and UMB contract consumers behind Unleash (feature flags), with optional structured config variant payload to turn each consumer on or off independently.

Changes:
- FeatureFlags: New helper that reads `swatch.swatch-contracts.enable-partner-gateway-contracts` and, when the variant is config with JSON payload, applies kafka_consumer_enabled / umb_consumer_enabled.
- Consumers: ContractsPartnerEntitlementMessageConsumer and ContractUMBMessageConsumer return early when the corresponding flag path is disabled (with debug logging).
- Schema / build: partner_gateway_contracts_feature_flag_variant_payload.yaml plus pom.xml updates so the payload model is generated for FeatureFlags.
- Runtime: application.properties — Quarkus Unleash client settings (URL, token, name-prefix, profiles for test/dev/ephemeral). deploy/clowdapp.yaml — featureFlags: true on the deployment so Unleash is available in managed environments.

### Follow-up (separate MR in app-interface)

After this PR is approved, a separate MR in app-interface will add the swatch.swatch-contracts.enable-partner-gateway-contracts feature flag (and environment rollout as appropriate), including a config variant whose string payload is JSON, for example:

```json
{"kafka_consumer_enabled": true, "umb_consumer_enabled": true}
```

## Testing
- Unit tests: FeatureFlagsTest and updates to both consumer tests for enabled/disabled behaviour.
- Component tests / plan: ContractsUnleashService + @Unleash / @BeforeEach in ContractCreationComponentTest to enable the toggle before UMB-driven contract creation tests; TEST_PLAN.md setup updated; ct/README.md, ct/pom.xml, and ct/resources/test.properties adjusted as needed.

### DIFF
--- a/swatch-contracts/src/test/java/com/redhat/swatch/contract/config/FeatureFlagsTest.java
+++ b/swatch-contracts/src/test/java/com/redhat/swatch/contract/config/FeatureFlagsTest.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package com.redhat.swatch.contract.config;
+
+import static com.redhat.swatch.contract.config.FeatureFlags.CONFIG_VARIANT;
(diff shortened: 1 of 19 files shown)


swatch-contracts/src/test/java/com/redhat/swatch/contract/config/FeatureFlagsTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+    assertTrue(featureFlags.isPartnerGatewayContractsKafkaConsumerEnabled());
+  }
+
+  @Test
+  void shouldReturnFalse_whenConfigVariantEnabled_withInvalidJsonPayload_forUmb() {
+    givenPartnerGatewayContractsEnabledWithConfigPayload("{invalid");
+
+    assertTrue(featureFlags.isPartnerGatewayContractsUmbConsumerEnabled());
+  }
```
vbusch: The test names all start with `shouldReturnFalse....`
Then they `assertTrue`

vbusch (changes requested):
PR looks good.
- There are 3 test names that cause confusion with the True/False.
- Both flags are on by default.  It wasn't specified in the issue, but is a good choice.

swatch-contracts/src/test/java/com/redhat/swatch/contract/config/FeatureFlagsTest.java
Sgitario: fixed

vbusch (approved):

PR #6065, opened 2026-04-30
SWATCH-4547: Support for prod floorist queries

liwalker-rh:
Jira issue: SWATCH-4547

## Description
Changes to get data from rhsm-prod to rhprod snowflake environment.

### DIFF
--- a/swatch-tally/deploy/clowdapp.yaml
+++ b/swatch-tally/deploy/clowdapp.yaml
@@ -243,6 +243,10 @@ parameters:
     name: FLOORIST_BUCKET_SECRET_NAME_PREPROD
     required: true
     value: egress-s3
+  - description: bucket secret name for prod
+    name: FLOORIST_BUCKET_SECRET_NAME_PROD
+    required: true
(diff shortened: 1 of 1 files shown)


swatch-tally/deploy/clowdapp.yaml
```diff
@@ -243,6 +243,10 @@ parameters:
     name: FLOORIST_BUCKET_SECRET_NAME_PREPROD
     required: true
     value: egress-s3
+  - description: bucket secret name for prod
+    name: FLOORIST_BUCKET_SECRET_NAME_PROD
```
kartikshahc, line 247: Like we discussed offline I think we should continue using FLOORIST_BUCKET_SECRET_NAME and the metadata since that will make sure it doesn't create a new job. Also the FLOORIST_BUCKET_SECRET_NAME is unique per stage and prod so it will automatically pick things up from app-interface.

kartikshahc (changes requested):
Maybe we don't need to have a PR as we discussed earlier

PR #6066, opened 2026-04-30
SWATCH-4855 Add is_primary column to host_tally_buckets

mstead:
Jira issue: SWATCH-4855

## Description
Adds is_primary column (boolean, NOT NULL, default false) to host_tally_buckets table and updates the JPA entity to include an isPrimary field with a supporting getter/setter.

## Testing
1. Deploy swatch-tally from **main** -- verify the column does not exist.
2. Deploy from this PRs branch -- verify the column was added and the correct constraints are set on the column (NOT NULL, default false)

### DIFF
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
@@ -52,6 +52,9 @@ public class HostTallyBucket implements Serializable {
 
   @EmbeddedId private HostBucketKey key;
 
+  @Column(name = "is_primary")
+  private boolean isPrimary = false;
+
   private Integer cores;
   private Integer sockets;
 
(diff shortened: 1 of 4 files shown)


kartikshahc (dismissed):
LGTM

mstead:
I need to fix a test issue. The way claude code wrote it doesn't test  it well.

swatch-core/src/main/java/org/candlepin/subscriptions/db/model/HostTallyBucket.java
```diff
@@ -52,6 +52,9 @@ public class HostTallyBucket implements Serializable {
 
   @EmbeddedId private HostBucketKey key;
 
+  @Column(name = "is_primary")
+  private boolean isPrimary = false;
```
kartikshahc, line 56: Changes look good but, quick question: since host tally buckets fetches some data by key, and it is already included in equals() and hashCode(), that already enforces uniqueness, right? So we wouldn’t need to add is_primary to equals() and hashCode()correct?

mstead: Correct, the key enforces the unique constraint that we require. IMO isPrimary is intended to be a marker.

kartikshahc (approved):
LGTM

PR #6072, opened 2026-05-02
swatch-4955: Fix capacity unavailable at contract start boundary

vbusch:
Jira issue: SWATCH-4955

## Description
Fix the capacity queries at the exact contract start or end timestamp, which were incorrectly excluded contracts due to strict `isBefore()` and `isAfter()` boundary checks in `CapacityResourceV1.java`.

This was demonstrated by IQE test `test_verify_contract_capacity` was failing on the first day of each month (May 1st UTC = April 30 8pm EDT to May 1st 8pm EDT).

**Fix:** Changed boundary checks to be inclusive:
- `begin.isBefore(date)`  -> `!begin.isAfter(date)`
- `end.isAfter(date)` -> `!date.isAfter(end)` 

## Testing
iqe tests plugin rhsm_subscriptions -k test_verify_contract_capacity

### Verification
The test_contract_capacity#test_verify_contract_capacity should now pass every day of the month, including May 1st UTC.  

##Unimplemented Alternative Solution - Test Modification
- Modify the test to query at `May 1, 00:01` instead of `May 1, 00:00`
- [swatch_contracts/test_contracts_capacity.py#L64](https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/blob/master/iqe_rhsm_subscriptions/tests/component/swatch_contracts/test_contracts_capacity.py#L64)
- Did continue with that because it eaves a window where production queries at exact boundaries return incorrect results.

### DIFF
--- a/swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/api/v1/CapacityResourceV1Test.java
+++ b/swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/api/v1/CapacityResourceV1Test.java
@@ -26,6 +26,7 @@ import static io.restassured.RestAssured.given;
 import static java.util.Objects.isNull;
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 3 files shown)


vbusch:
In draft until unit tests are added.

Sgitario (approved):

swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/api/v1/CapacityResourceV1Test.java
```diff
@@ -760,4 +761,71 @@ private static Stream<Arguments> generateFinestGranularityCases() {
...
+    OffsetDateTime queryDate = min.truncatedTo(ChronoUnit.DAYS);
+    var subscription = datedSubscription(queryDate, max);
+    subscription.setSubscriptionMeasurements(basicMeasurement());
+
+    when(subscriptionRepository.findByCriteria(
+            argThat(argument -> argument.getOrgId().equals("org123")), any(Sort.class)))
+        .thenReturn(List.of(subscription));
+
+    // When: Querying capacity at exactly the subscription start time
+    CapacityReportByMetricId report =
```
Sgitario, line 777: why not using the contracts service API instead? ContractsSwatchService.getCapacityReportByMetricId.

Sgitario (changes requested):

swatch-contracts/src/test/java/com/redhat/swatch/contract/resource/api/v1/CapacityResourceV1Test.java
vbusch: It was written to be consistent with the existing pattern in CapacityResourceV1Test.java (all other tests use the REST API).                                                     

However, I can change these two tests to call capacityResource.createCapacitySnapshotWithMetricId() directly for a more focused unit test of the boundary logic.  This would be less consistent with the file but a better test.

Note: CapacityReportTemporalBoundariesComponentTest already provides component test coverage with ContractsSwatchService.getCapacityReportByMetric.  In the unit test only the boundary logic needs covering.

vbusch: I reverted the change to the updated test, to keep it consistent with testing the api.

Sgitario (approved):

PR #6082, opened 2026-05-05
Use hermeto to prefetch maven dependencies

Sgitario:
## Description
TBA


Sgitario:
hermeto has the same network issue, so closing.

PR #6084, opened 2026-05-05
chore: use mvn instead of maven wrapper to build images

Sgitario:
TBA


Sgitario:
it also fails when downloading the maven dependencies, so this solution does not fix the problem. Closing.

PR #6086, opened 2026-05-05
SWATCH-4941: Add FloorPlan for swatch-contracts

awood:
Jira issue: SWATCH-4941

## Description
Remove the `sku_product_tag` entry from swatch-tally as that's now duplicated
in the swatch-contracts FloorPlan.

### DIFF
--- a/swatch-contracts/deploy/clowdapp.yaml
+++ b/swatch-contracts/deploy/clowdapp.yaml
@@ -146,6 +146,77 @@ parameters:
     value: signalfx
   - name: OTEL_SIGNALFX_TOKEN_SECRET_KEY
     value: token
+  - name: FLOORIST_DB_SECRET_NAME
+    description: database secret name
+    value: swatch-db-ro
+    # FloorPlan swatch-contracts: Stage=dev, Prod=prod
+  - name: FLOORIST_BUCKET_SECRET_NAME
+    description: bucket secret name
+    required: true
+    value: egress-s3
+    # FloorPlan swatch-contracts-test: Stage=sandbox, Prod=preprod
+  - name: FLOORIST_BUCKET_SECRET_NAME_TEST
+    description: bucket secret name for preprod
+    required: true
+    value: egress-s3
+  - name: FLOORIST_LOGLEVEL
+    description: Floorist loglevel config
+    value: 'DEBUG'
+  - name: FLOORIST_SUSPEND
+    description: Disable Floorist cronjob execution
+    required: true
+    value: 'false'
+  - name: FLOORIST_S3_SNOWPIPE_PATH
+    description: Path for dataproduct in snowpipe s3 bucket
+    value: 'swatch'
+
+# Shared FloorPlan specification for all environments (sandbox, dev, preprod, prod)
+# Only objectStore differs per environment
+.floorplan_spec: &floorplan_spec_base
+  database:
+    secretName: ${FLOORIST_DB_SECRET_NAME}
+  logLevel: ${FLOORIST_LOGLEVEL}
+  suspend: ${{FLOORIST_SUSPEND}}
+  resources:
+    limits:
+      cpu: 400m
+      memory: 500Mi
+    requests:
+      cpu: 200m
+      memory: 250Mi
+  queries:
+  - prefix: ${FLOORIST_S3_SNOWPIPE_PATH}/subscription
+    chunksize: 50000
+    query: >-
+      select * from subscription;
+  - prefix: ${FLOORIST_S3_SNOWPIPE_PATH}/subscription_measurements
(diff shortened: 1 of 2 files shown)


kartikshahc:
@awood changes look good but, two things:
1. We recently decided to add a `last_updated` filter to these queries since the tables can become quite large and may lead to timeouts. Since the changes are without the filter do you think we can skip adding the filter criteria for these tables, or should it be applied consistently across all of them?
2. The branch needs a rebase.

awood:
> 1. We recently decided to add a `last_updated` filter to these queries since the tables can become quite large and may lead to timeouts. Since the changes are without the filter do you think we can skip adding the filter criteria for these tables, or should it be applied consistently across all of them?

For the initial import, we need to *not* have a `last_update` filter.  When that column is created, the value for all pre-existing rows will be `null`.  After the initial import, then we'll add the filter.

kartikshahc:
> > 1. We recently decided to add a `last_updated` filter to these queries since the tables can become quite large and may lead to timeouts. Since the changes are without the filter do you think we can skip adding the filter criteria for these tables, or should it be applied consistently across all of them?
> 
> For the initial import, we need to _not_ have a `last_update` filter. When that column is created, the value for all pre-existing rows will be `null`. After the initial import, then we'll add the filter.

Yeah adding the last_updated filter was recently added as a requirement. I am going to approve this hoping that the full table export doesn't timeout.

kartikshahc (approved):

Sgitario (changes requested):
why do we need two floorplan objects? the test one seems to me like a temporal object we need for stage tests only? 

I see that having two floorplans means that we run the same queries and export the same data, but into two different destinations. And if I didn't see this wrongly, we're doing this in production too which seems to me like a security issue: we're exporting prod data into a preprod environment.

kartikshahc:
> why do we need two floorplan objects? the test one seems to me like a temporal object we need for stage tests only?
> 
> I see that having two floorplans means that we run the same queries and export the same data, but into two different destinations. And if I didn't see this wrongly, we're doing this in production too which seems to me like a security issue: we're exporting prod data into a preprod environment.

@Sgitario This is how the Dataverse team communicated it to Kevin and our team. We need to configure Dataverse `sandbox` and `dev` pipelines from `rhsm-stage`, and `preprod` and `prod` from `rhsm-prod`. That is why we need two floorplans since they require different buckets for individual envs.
@awood if you want to add more to it.

awood:
> why do we need two floorplan objects? the test one seems to me like a temporal object we need for stage tests only?
> 
> I see that having two floorplans means that we run the same queries and export the same data, but into two different destinations. And if I didn't see this wrongly, we're doing this in production too which seems to me like a security issue: we're exporting prod data into a preprod environment.

Both stage and production each have 2 destination locations.  And the names are inconsistent with the standard nomenclature we use.  `preprod` is actually a production location but `prod` is the ultimate final location.  I can't say I fully understand why it's set up like this, but that's what we've got.

Sgitario (approved):

Sgitario:
> > why do we need two floorplan objects? the test one seems to me like a temporal object we need for stage tests only?
> > I see that having two floorplans means that we run the same queries and export the same data, but into two different destinations. And if I didn't see this wrongly, we're doing this in production too which seems to me like a security issue: we're exporting prod data into a preprod environment.
> 
> Both stage and production each have 2 destination locations. And the names are inconsistent with the standard nomenclature we use. `preprod` is actually a production location but `prod` is the ultimate final location. I can't say I fully understand why it's set up like this, but that's what we've got.

Do you know whether this is a final setup we need to use in stage and prod both? Asking because we're running twice of the processes (that involves SQL queries). If the preprod queries were asked for testing purposes, we could create a JIRA tracker for removal and a comment in the code to let others know that this is a temporal change.
Otherwise, if this is a final configuration, maybe we could improve the floorplan object to support two destinations instead of only one. 
Thoughts?

awood:
> Do you know whether this is a final setup we need to use in stage and prod both? Asking because we're running twice of the processes (that involves SQL queries). If the preprod queries were asked for testing purposes, we could create a JIRA tracker for removal and a comment in the code to let others know that this is a temporal change.
> Otherwise, if this is a final configuration, maybe we could improve the floorplan object to support two destinations instead of only one.
> Thoughts?

I am not sure about if this is going to be the final setup.  While, it is possible to alter the FloorPlan object to write to multiple destinations, that would involve making changes to the Floorist operator (to support multiple object stores) and changes to the Floorist Python script to write to multiple destinations.  Ultimately these queries are run on a read-clone so I'm not overly worried about performance here.  Especially for `swatch-contracts` whose largest table is only around 10 million rows.

PR #6144, opened 2026-05-15
SWATCH-4800: Org preference POST and Grafana panel

TommasoLencioni:
## Summary

Adds a Subscription Watch stat panel for how often and the status of customers POST org utilization preferences (custom threshold saves).

## Merge order

**Merge this PR first** (1 of 2). The notification-rate and handler-duration PRs touch the same dashboard ConfigMap; merging in this order avoids unnecessary conflicts.

1. **This PR** — `SWATCH-4800-grafana-customer-threshold`
2. `SWATCH-4800-grafana-notification-event-rate`

PR #6145, opened 2026-05-15
SWATCH-4800: Add Grafana panel for custom threshold notification event sent by swatch-utilization

TommasoLencioni:
## Summary

Add Grafana panel for custom threshold notification event sent by swatch-utilization, same as the one already in place for default threshold events.

## Merge order

**Merge after #6144**.

1. #6144 — `SWATCH-4800-grafana-customer-threshold`
2. **This PR** — `SWATCH-4800-grafana-notification-event-rate`

Expect a ConfigMap merge conflict if #6144 is not merged first; resolve by keeping both panels.

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
@@ -6835,6 +6835,208 @@ data:
                 "x": 0,
                 "y": 82
               },
+              "id": 137,
+              "options": {
+                "cellHeight": "sm",
+                "footer": {
+                  "countRows": false,
+                  "fields": "",
+                  "reducer": [
+                    "sum"
+                  ],
+                  "show": false
+                },
+                "showHeader": true
+              },
+              "pluginVersion": "11.6.3",
+              "targets": [
+                {
+                  "datasource": {
+                    "type": "prometheus",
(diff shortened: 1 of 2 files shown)


Sgitario (changes requested):
Adding event_type in incrementCounter() increases cardinality on every series without clear benefit: each handler already has its own metric (swatch_utilization_over_usage vs swatch_utilization_custom_threshold) and a fixed event type, so the label is redundant with the metric name, and the new Grafana panel queries those metrics directly without using event_type. I'd drop the tag and keep the two metric names. Also, a new label creates new Prometheus series on deploy, so counters without event_type stop updating and sum(rate(...)) may dip until old series age out

TommasoLencioni:
> Adding event_type in incrementCounter() increases cardinality on every series without clear benefit: each handler already has its own metric (swatch_utilization_over_usage vs swatch_utilization_custom_threshold) and a fixed event type, so the label is redundant with the metric name, and the new Grafana panel queries those metrics directly without using event_type. I'd drop the tag and keep the two metric names. Also, a new label creates new Prometheus series on deploy, so counters without event_type stop updating and sum(rate(...)) may dip until old series age out

@Sgitario thanks for the suggestion, I just replicated the panel showing over usage events fields also for custom threshold events.

Sgitario (approved):

PR #6146, opened 2026-05-15
SWATCH-4800: Threshold handler timer and Grafana duration panel

TommasoLencioni:
## Summary

Instruments threshold handler work with a Micrometer **timer** and adds a Subscription Watch panel for mean handler duration by kind.

- **`swatch_utilization_handler`**: timer around `handle()` in `BaseThresholdUtilizationHandlerService`, tagged by **`handler_kind`** (`over_usage`, `custom_threshold` via `handlerKind()` on concrete handlers).
- **Grafana**: panel **Threshold handler duration (avg)** — mean `rate(sum/count)` by `handler_kind`.

## Merge order

**Merge last** (3 of 3).

1. #6144 — `SWATCH-4800-grafana-customer-threshold`
2. #6145 — `SWATCH-4800-grafana-notification-event-rate`
3. **This PR** — `SWATCH-4800-grafana-handler-duration`

Expect a ConfigMap merge conflict with #6144 / #6145; resolve by keeping all utilization panels.

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-subscription-watch.configmap.yaml
@@ -6628,6 +6628,101 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
+              "description": "Mean wall time spent evaluating each utilization measurement in a threshold handler (Kafka pipeline).",
+              "fieldConfig": {
+                "defaults": {
+                  "color": {
+                    "mode": "palette-classic"
+                  },
+                  "custom": {
+                    "axisBorderShow": false,
+                    "axisCenteredZero": false,
+                    "axisColorMode": "text",
+                    "axisLabel": "",
+                    "axisPlacement": "auto",
+                    "barAlignment": 0,
+                    "barWidthFactor": 0.6,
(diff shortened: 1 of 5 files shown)


Sgitario (changes requested):
I'm very concerned about adding metrics or also adding new parameters into existing metrics because it exploits the native memory of the service.

In order to avoid adding a new metric to monitor the processing time of each threshold handler, I think that simply monitoring the consumer lag is enough (we're already doing that). 

Why?
the over usage threshold logic is very simple and will be done almost at the instant.
the custom usage theshold logic requires a (simply) query to database, so if the customer lag starts growing, this would be the root cause of the issue for sure.

TommasoLencioni:
> I'm very concerned about adding metrics or also adding new parameters into existing metrics because it exploits the native memory of the service.
> 
> In order to avoid adding a new metric to monitor the processing time of each threshold handler, I think that simply monitoring the consumer lag is enough (we're already doing that).
> 
> Why? the over usage threshold logic is very simple and will be done almost at the instant. the custom usage theshold logic requires a (simply) query to database, so if the customer lag starts growing, this would be the root cause of the issue for sure.

@Sgitario that's a great point, I'll close this one, thanks for the contribution!

PR #6149, opened 2026-05-15
SWATCH-4954: fix NPE in contract sync and clear orphaned subscription

Mantzounis:
Jira issue: SWATCH-4954

## Description
When partner Gateway returns entitlements with a null purchase field, it causes an NPE in extractBillingProviderId() that crashed the entire org sync. Also, orphaned contract termination didn't clear the linked subscription's billing fields, leaving SubscriptionSyncService treating it as contract-provided and skipping it indefinitely.

## Testing
Unit and component tests included

### Verification

1. podman compose up -d
2. make build component test swatch-contracts

### DIFF
--- a/swatch-contracts/TEST_PLAN.md
+++ b/swatch-contracts/TEST_PLAN.md
@@ -747,6 +747,53 @@ This section verifies the automatic contract termination behavior when contracts
   - Contract `end_date` is unchanged (still in the future, not terminated)
   - Contract count remains 1
 
+**contracts-sync-TC016 - Sync does not crash when upstream returns entitlement with null purchase**
+- **Description**: Verify that when upstream returns a mix of valid entitlements and entitlements with null `purchase` field, the sync skips the ones with null purchase and processes the valid contracts. Reproduces SWATCH-4954 NPE.
+- **Setup**:
+  - Stub upstream Partner API to return one valid AWS contract and one entitlement with `purchase: null`
+  - Stub offering and search API for the valid contract
+- **Action**: POST `/api/swatch-contracts/internal/rpc/sync/contracts/{orgId}`
+- **Verification**:
+  - Sync returns HTTP 200 with status "SUCCESS"
+  - Only the valid contract is persisted
+- **Expected Result**:
+  - HTTP 200 with StatusResponse
+  - StatusResponse status: "SUCCESS"
+  - Exactly 1 contract persisted (the valid one)
+  - The entitlement with null purchase is skipped without crashing the sync
+
+**contracts-sync-TC017 - Orphaned contract termination clears subscription's contract-provided state**
+- **Description**: Reproduces SWATCH-4954 subscription blocking. When a contract is orphaned (not found in upstream), the termination should clear the subscription's billing provider and measurements so subscription sync can update it.
+- **Setup**:
+  - Create a contract-enabled subscription for the org (AWS ROSA contract)
+  - Verify the subscription has billing provider and measurements set
+  - Re-stub upstream Partner API to return no valid contracts for the org
+- **Action**: POST `/api/swatch-contracts/internal/rpc/sync/contracts/{orgId}`
+- **Verification**:
+  - Contract is terminated (end_date set)
+  - Subscription's billing provider is cleared
+  - Subscription's measurements are cleared
+- **Expected Result**:
+  - HTTP 200 with StatusResponse
+  - Contract still exists but is terminated
+  - Subscription billing provider cleared and measurements cleared — subscription sync will no longer skip it
+
+**contracts-sync-TC018 - Legitimate contract termination preserves subscription's contract-provided state**
+- **Description**: When a contract is legitimately terminated (still present in upstream but with end_date in the past), the subscription should retain its billing provider, billing_account_id, and measurements — unlike orphaned contracts (TC017) which clear them.
+- **Setup**:
+  - Create a contract-enabled subscription (AWS ROSA contract with future end_date)
+  - Verify the subscription has billing provider and measurements set
+  - Re-stub upstream Partner API to return the same contract but with end_date in the past
+- **Action**: POST `/api/swatch-contracts/internal/rpc/sync/contracts/{orgId}`
+- **Verification**:
+  - Contract's end_date is updated to the past date from upstream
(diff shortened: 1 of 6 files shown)


swatch-contracts/TEST_PLAN.md
```diff
@@ -747,6 +747,37 @@ This section verifies the automatic contract termination behavior when contracts
   - Contract `end_date` is unchanged (still in the future, not terminated)
   - Contract count remains 1
 
+**contracts-sync-TC016 - Sync does not crash when upstream returns entitlement with null purchase**
```
Sgitario, line 750: I would not mention that entitlements without purchase are malformed messages.

swatch-contracts/TEST_PLAN.md
```diff
@@ -747,6 +747,37 @@ This section verifies the automatic contract termination behavior when contracts
...
+- **Verification**:
+  - Sync returns HTTP 200 with status "SUCCESS"
+  - Only the valid contract is persisted
+- **Expected Result**:
+  - HTTP 200 with StatusResponse
+  - StatusResponse status: "SUCCESS"
+  - Exactly 1 contract persisted (the valid one)
+  - The malformed entitlement is skipped without crashing the sync
+
+**contracts-sync-TC017 - Orphaned contract termination clears subscription's contract-provided state**
```
Sgitario, line 765: Can you update an existing component test (or add a new one) to validate that legitim contracts that were terminated do not clear the subscription's contract-provided state?

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -578,9 +582,11 @@ private PartnerEntitlements syncContractsByOrgId(
         if (entitlement != null
             && ContractSourcePartnerEnum.isSupported(entitlement.getSourcePartner())) {
           String bpId = contractEntityMapper.extractBillingProviderId(entitlement);
-          if (bpId != null) {
-            upstreamBillingProviderIds.add(bpId);
+          if (bpId == null) {
+            log.warn("Skipping entitlement with null purchase for org {}: {}", orgId, entitlement);
+            continue;
```
Sgitario: please, avoid using `continue` because it makes harder to follow the sequence logical order of this logic.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -528,6 +528,10 @@ private void terminateActiveSubscriptionsForContract(
                   "Terminating subscription that no longer exists in upstream partner API: {}",
                   sub);
               sub.setEndDate(terminationDate);
+              sub.setBillingProvider(null);
+              sub.setBillingProviderId(null);
+              sub.setBillingAccountId(null);
+              sub.getSubscriptionMeasurements().clear();
```
Sgitario, line 534: question: why do we need to also clear the subscription measurements?

swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractEntityMapper.java
```diff
@@ -116,11 +119,11 @@ default String extractBillingProviderId(PartnerEntitlementV1 entitlement) {
...
     } else if (isAzureMarketplace(partner)) {
+      var contracts = entitlement.getPurchase().getContracts();
       var azurePlanId =
-          entitlement.getPurchase().getContracts().stream()
-              .map(SaasContractV1::getPlanId)
-              .findFirst()
-              .orElse("");
+          contracts != null
+              ? contracts.stream().map(SaasContractV1::getPlanId).findFirst().orElse("")
+              : "";
```
Sgitario, line 134: is this change necessary ?

swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractEntityMapper.java
```diff
@@ -108,6 +108,9 @@ default String extractBillingAccountId(PartnerIdentityV1 accountId) {
   /** Extract billingProviderId from Partner Gateway API response */
   @Named("billingProviderId")
   default String extractBillingProviderId(PartnerEntitlementV1 entitlement) {
+    if (entitlement.getPurchase() == null) {
+      return null;
```
Sgitario, line 120: can you add a log warning statement saying that this contract was skipped? it's important to also mention the org ID for troubleshooting.

Sgitario (changes requested):

swatch-contracts/src/main/java/com/redhat/swatch/contract/model/ContractEntityMapper.java
Mantzounis: It's a defensive fix against nullable fields in the partner gateway model, contracts != null is a separate npe from getPurchase() == null.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
Mantzounis: in subscription sync,   if (hasBillingProvider && !existingSubscription.getSubscriptionMeasurements().isEmpty()) {
      log.info(
          "Skipping sync of subscriptionId={} because it has contract-provided capacity",
          existingSubscription.getSubscriptionId());
      return;
  }   
  
If either billing fields or measurements remain after contract orphaning, subscription sync keeps skipping the subscription. That's why I think clearing both is necessary since clearing only billing provider but leaving measurements wouldn't be enough because the check is && and we also keep a consistent state this way, so clearing both makes the subscription fully "non-contract-provided" again.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
Mantzounis: thank you I will fix that

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
Sgitario: Thanks for the clarification!

Sgitario (approved):

PR #6162, opened 2026-05-19
SWATCH-4897: Add Playwright container to frontend stage post deploy job

mmusil:
Merge only after related iqe curiosity plugin MR !610 is merged.

### DIFF
--- a/deploy/frontend-stage-internal-test-job.yaml
+++ b/deploy/frontend-stage-internal-test-job.yaml
@@ -8,7 +8,7 @@ objects:
     kind: Job
     metadata:
       name: curiosity-frontend-stage-test-job-${IMAGE_TAG}-${UID}
-      annotatons:
+      annotations:
         "ignore-check.kube-linter.io/no-liveness-probe": "probes not required on Job pods"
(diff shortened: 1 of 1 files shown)


deploy/frontend-stage-internal-test-job.yaml
```diff
@@ -96,6 +86,31 @@ objects:
...
+                - name: pw-shm
+                  mountPath: /dev/shm
+              terminationMessagePath: /dev/termination-log
+              terminationMessagePolicy: File
+            - name: playwright-chromium-${IMAGE_TAG}-${UID}
+              image: ${PLAYWRIGHT_IMAGE}
+              imagePullPolicy: Always
+              env:
+                - name: PW_HEADLESS
+                  value: "false"
```
mmusil, line 99: `False` because the IQE team is running it the same https://github.com/RedHatInsights/insights-pipeline-lib/blob/master/vars/openShiftUtils.groovy#L278-L290
It could help with debugging. The PW container also includes VNC server, so we could look in real time at the session with port forwarding to the running pod.

Sgitario (approved):

TommasoLencioni (approved):

PR #6163, opened 2026-05-19
SWATCH-4953: Reduce currentTotal Calls for Gauge Metrics for performance improvement

wottop:
Jira issue: SWATCH-4953

## Description

To reduce calls, we use an available event value instead of a SQL query

### Setup
1. Start pods
```podman-compose up -d```
2. Start swatch_tally with debugging on for tally
```LOGGING_LEVEL_ORG_CANDLEPIN_SUBSCRIPTIONS_TALLY=DEBUG make build swatch-tally```

### Steps
1. Create hosts in HBI database
```
bin/insert-mock-hosts --hbi --org 13259775 --db-host localhost --db-user insights --db-name insights --num-physical 20
```
2. Run nightly tally 
```
http PUT ":8010/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots/13259775" x-rh-swatch-psk:placeholder
```

### Validation
1. You will see the following logging:
```
2026-05-20 14:10:24,076 [thread=rhsm-subscriptions-task-processor-0-C-1] [DEBUG] [org.candlepin.subscriptions.tally.TallySummaryMapper] - Using event value for GAUGE metric: TallyMeasurementKey{measurementType=PHYSICAL, metricId=SOCKETS}
2026-05-20 14:10:24,095 [thread=rhsm-subscriptions-task-processor-0-C-1] [DEBUG] [org.candlepin.subscriptions.tally.SnapshotSummaryProducer] - Produced 1 DAILY TallySummary messages
```

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/tally/TallySummaryMapper.java
+++ b/src/main/java/org/candlepin/subscriptions/tally/TallySummaryMapper.java
@@ -21,9 +21,7 @@
 package org.candlepin.subscriptions.tally;
 
 import com.redhat.swatch.common.model.ServiceLevel;
-import com.redhat.swatch.configuration.registry.Metric;
 import com.redhat.swatch.configuration.registry.MetricId;
-import com.redhat.swatch.configuration.registry.MetricType;
 import com.redhat.swatch.configuration.registry.ProductId;
 import com.redhat.swatch.configuration.registry.SubscriptionDefinition;
 import com.redhat.swatch.configuration.util.MetricIdUtils;
@@ -126,18 +124,15 @@ public class TallySummaryMapper {
       TallySnapshot snapshot, TallyMeasurementKey measurementKey, Double eventValue) {
 
     String productId = snapshot.getProductId();
-    boolean isPayg = ProductId.fromString(productId).isPayg();
 
-    // Non-PAYG, non-Ansible GAUGE metrics: use event value
-    if (!isPayg
-        && !isAnsibleProduct(productId)
-        && isGaugeMetric(productId, measurementKey.getMetricId())) {
-      // For non-PAYG, non-Ansible GAUGE metrics: use event value directly - no SQL calls
+    // Non-Ansible GAUGE metrics: use event value
+    if (!"ansible-aap-managed".equals(productId)
+        && SubscriptionDefinition.isGaugeMetric(productId, measurementKey.getMetricId())) {
       log.debug("Using event value for GAUGE metric: {}", measurementKey);
       return eventValue;
     }
 
-    // For all other cases (PAYG, Ansible, or COUNTER metrics): use existing SQL logic
+    // For all other cases (Ansible, or COUNTER metrics): use existing SQL logic
     if (featureFlags.isPrimaryRowSearchesEnabled()) {
       gateOnAnyValues(snapshot);
       return snapshotRepository.sumMeasurementValueForPeriodWithPrimary(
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by kartikshahc** [comment]
Tagging @Sgitario since you were interested to see the changes for this

File: src/main/java/org/candlepin/subscriptions/tally/TallySummaryMapper.java [hunk 1]
```diff
@@ -110,13 +117,27 @@ private List<TallyMeasurement> mapMeasurements(TallySnapshot snapshot) {
...
 
   private Double getCurrentlyMeasuredTotal(
-      TallySnapshot snapshot, TallyMeasurementKey measurementKey) {
+      TallySnapshot snapshot, TallyMeasurementKey measurementKey, Double eventValue) {
+
+    String productId = snapshot.getProductId();
+    boolean isPayg = ProductId.fromString(productId).isPayg();
+
+    // Non-PAYG, non-Ansible GAUGE metrics: use event value
+    if (!isPayg
```
kartikshahc: @Sgitario I am sure you will have question on this and we can discuss more here. But the team decision was to have this because we don't want any behavior change for payg.

Sgitario: All the payg products already use counter metrics, except the Ansible product (which is excluded from these changes as I expected). 
If in the future, one payg product is updated to use a gauge metric, we'll immediatelly over bill customers because this logic is wrong (the calculation depends on gauge vs counter, not payg / payg products).

kartikshahc: Sure. @wottop can you remove the payg check and update the acceptance criteria stating the above.

src/main/java/org/candlepin/subscriptions/tally/TallySummaryMapper.java
```diff
@@ -145,6 +163,34 @@ private Double getCurrentlyMeasuredTotal(
...
+        .orElse(false);
+  }
+
+  /**
+   * Check if a product is an Ansible product.
+   *
+   * @param productId the product ID (tag)
+   * @return true if the product platform is "Ansible", false otherwise
+   */
+  private boolean isAnsibleProduct(String productId) {
```
kartikshahc: we don't need to do this since productId gives us `ansible-aap-managed` so you can just say if productId.equals('ansible-aap-managed') unless I am missing something and you are doing it for a reason.

src/main/java/org/candlepin/subscriptions/tally/TallySummaryMapper.java
```diff
@@ -145,6 +163,34 @@ private Double getCurrentlyMeasuredTotal(
...
 
+  /**
+   * Check if a metric is of type GAUGE.
+   *
+   * @param productId the product ID (tag)
+   * @param metricId the metric ID
+   * @return true if the metric is a GAUGE, false otherwise
+   */
+  private boolean isGaugeMetric(String productId, String metricId) {
+    return SubscriptionDefinition.lookupSubscriptionByTag(productId)
```
kartikshahc: This method should be part of SubscriptionDefinition. You can refer to implementation stated in Jira card or https://github.com/RedHatInsights/rhsm-subscriptions/commit/08ed7fb46607e51e7c06f1e34de5b465e28cea02

kartikshahc (changes requested):

src/test/java/org/candlepin/subscriptions/tally/TallySummaryMapperTest.java
```diff
@@ -192,44 +191,56 @@ void assertMappedSnapshot(
...
-                    eq(expected.getBillingProvider()),
-                    eq(expected.getBillingAccountId()),
-                    any(),
-                    eq(expected.getSnapshotDate()),
-                    eq(key));
-          }
-        });
-    verify(clock, times(mappedMeasurements.size())).startOfMonth(expected.getSnapshotDate());
+    int sqlCallCount = 0;
+    for (var m : mappedMeasurements) {
```
kartikshahc, line 195: I think these tests should really be captured as part of component test because it will be more useful to send the event and then do tally and then see the summary value. 

In unit test we can check whether the sql call is done or not but, this can be checked much simpler way.

kartikshahc (changes requested):

src/main/java/org/candlepin/subscriptions/tally/TallySummaryMapper.java
kartikshahc: Nop product tags are 1:1 mapping

kartikshahc (approved):
Changes look good to me. @TrayvonMcKnight can we cover the IQE/component tests with the three scenarios Paygo/non-paygo/ansible

PR #6189, opened 2026-05-27
SWATCH-4858: Update swatch-tally to use is_primary views

awood:
Jira issue: SWATCH-4858

## Description
Creates the `tally_instance` family of views but filtering on `is_primary=TRUE`
and then configures swatch-tally to use the views based on the value of the
feature flag.

### Setup
1. Start with SQL logging turned on
    ```shell
    SPRING_JPA_SHOW_SQL=true SPRING_JPA_PROPERTIES_HIBERNATE_FORMAT_SQL=true \
    LOGGING_LEVEL_ORG_HIBERNATE_ORM_JDBC_BIND=trace make build swatch-tally
    ```

### Steps
1. Do a GET on `/instances/products/` for an org to access the `tally_instance`
   views
    ```shell
    http localhost:8010/api/rhsm-subscriptions/v1/instances/products/RHEL%20for%20x86 \
    sla==Premium \
    usage==Production \
    beginning==2025-05-26T00:00:00.000Z \
    ending==2026-05-26T00:00:00.000Z \
    x-rh-identity:"$(echo -n '{"identity":{"account_number":"","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w 0)"
    ```
1. You'll see the select statement reads from `tally_instance_non_payg_view`
1. Turn on the `swatch.swatch-tally.enable-primary-row-searches` feature flag.
    ```shell
    http post :4242/api/admin/projects/default/features/swatch.swatch-tally.enable-primary-row-searches/environments/development/on Authorization:"*:*.unleash-insecure-admin-api-token"
    ```
1. Re-run the same GET to `/instances/products`.

### Verification
1.  This time you'll see that `tally_instance_non_payg_primary_view` is used.

### DIFF
--- a/src/main/resources/liquibase/202506191100-add-host-service-type-table.xml
+++ b/src/main/resources/liquibase/202506191100-add-host-service-type-table.xml
@@ -244,7 +244,7 @@
                       h.hypervisor_uuid
       from hosts h
              join host_tally_buckets b on h.id = b.host_id
-             inner join MaxHostServiceTypeDate mhstd ON h.id = mhstd.host_id
+             left join MaxHostServiceTypeDate mhstd ON h.id = mhstd.host_id
     </createView>
     <createView
       replaceIfExists="true"
--- a/swatch-core/src/main/java/org/candlepin/subscriptions/db/TallyInstanceViewRepository.java
+++ b/swatch-core/src/main/java/org/candlepin/subscriptions/db/TallyInstanceViewRepository.java
@@ -148,7 +148,9 @@ public class TallyInstanceViewRepository {
 
   private JpaSpecificationExecutor<? extends TallyInstanceView> selectRepository(
       ProductId productId) {
-    boolean usePrimary = featureFlags.isEnabled(FeatureFlags.ENABLE_PRIMARY_ROW_SEARCHES, false);
+    // TODO Enable with featureFlags.isEnabled(FeatureFlags.ENABLE_PRIMARY_ROW_SEARCHES, false);
+    //  for SWATCH-4862
+    boolean usePrimary = false;
     if (productId.isPayg()) {
       return usePrimary ? paygPrimaryViewRepository : paygViewRepository;
     }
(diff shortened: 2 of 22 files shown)


swatch-core/src/main/java/org/candlepin/subscriptions/db/TallyInstanceViewRepository.java
```diff
@@ -136,6 +145,15 @@ public Page<TallyInstanceView> findAllBy(
             ResourceUtils.getPageable(offset, limit));
   }
 
+  private JpaSpecificationExecutor<? extends TallyInstanceView> selectRepository(
+      ProductId productId) {
+    boolean usePrimary = featureFlags.isEnabled(FeatureFlags.ENABLE_PRIMARY_ROW_SEARCHES, false);
+    if (productId.isPayg()) {
+      return usePrimary ? paygPrimaryViewRepository : paygViewRepository;
+    }
+    return usePrimary ? nonPaygPrimaryViewRepository : nonPaygViewRepository;
```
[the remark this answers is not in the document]
awood: Resolved one level up at `TallyInstanceViewRepository.findAllBy`

src/main/resources/liquibase/202506191100-add-host-service-type-table.xml
```diff
@@ -244,7 +244,7 @@
                       h.hypervisor_uuid
       from hosts h
              join host_tally_buckets b on h.id = b.host_id
-             inner join MaxHostServiceTypeDate mhstd ON h.id = mhstd.host_id
+             left join MaxHostServiceTypeDate mhstd ON h.id = mhstd.host_id
```
mstead, line 247: Was this change on purpose? If so, shouldn't it be in a different changeset since it was already run?

mstead (changes requested):
Looks good. Nice way to handle disabling the new view queries, solves the issue we discussed well.

Small question/issue below, but other than that I think this looks good.

Please update [SWATCH-4862](https://redhat.atlassian.net/browse/SWATCH-4862) with details about the feature flag comment.

src/main/resources/liquibase/202506191100-add-host-service-type-table.xml
awood: Yes, this is on purpose.  And it shouldn't cause a problem because that changeset is run only on HSQLDB (i.e. only during unit tests).

mstead: 👍 Missed that it was HSQLDB.

mstead (approved):
👍 All my feedback was addressed.

PR #6198, opened 2026-05-28
SWATCH-4796: Reset deduplication when custom threshold changes

vbusch:
Jira issue: [SWATCH-4796](https://redhat.atlassian.net/browse/SWATCH-4796)

## Description

When an org admin updates their custom utilization threshold, the deduplication
logic in notifications-backend should treat the next notification as a new event,
so the user is re-notified

- Adds `billing_account_id` to the notification event context. This field was already part of the deduplication key in notifications-backend but was never being sent, causing events for different billing accounts to collide in the
  dedup table.
- Adds a `preferences_hash` (SHA-256 of `OrgUtilizationPreference.lastUpdated`)   to the event context for `exceeded-custom-utilization-threshold` events only.   When the threshold changes, `lastUpdated` changes, the hash changes, and the  deduplication key in notifications-backend becomes distinct — resetting
  deduplication for that org.

[SWATCH-4796]: https://redhat.atlassian.net/browse/SWATCH-4796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/integration-test-plans/Notification Service Integration Test Plan.md
+++ b/docs/integration-test-plans/Notification Service Integration Test Plan.md
@@ -77,21 +77,6 @@ Limitation:
    - Deduplication fires for identical `product + metric + SLA + usage` combo.
    - A new notification is produced when SLA changes and when usage changes, confirming both are part of the deduplication key.
 
-## stage-notification-deduplication-TC003
-
-1. **Description:** Exercises the **preferences_hash** element of the deduplication key for custom threshold notifications. Updating the custom utilization threshold changes the `preferences_hash` embedded in the event, causing the deduplication key to differ so the user is re-notified even if the same utilization level was already notified this month.
-2. **Setup:**
-   - The organization is opted-in for custom threshold notifications (`exceeded-custom-utilization-threshold` event type).
-   - The organization has set a custom threshold at a value below current usage (e.g., 80% with usage at 85%).
-   - The organization has already received a custom threshold notification this calendar month.
-3. **Actions and Verifications (in sequence):**
-   1. Trigger the utilization sync job → verify first custom threshold notification is received with a non-null `preferences_hash` in the context.
-   2. Trigger the utilization sync job again (same usage, same threshold) → assert **no new notification** is received (deduplication suppresses the duplicate).
-   3. Update the custom threshold to a different value that is still below current usage (e.g., 79%) → trigger the sync → verify **a new notification is received** and its `preferences_hash` differs from the first notification's hash.
-4. **Expected Result:**
-   - Deduplication suppresses notifications when the custom threshold configuration is unchanged.
-   - A new notification is delivered after the threshold is updated, confirming that the `preferences_hash` change resets the deduplication key.
-
 ## test_over_usage_notification_requires_payg_sla_aligned_with_offering
 
 1. **Description:** Over-usage notifications for PAYG products require that the SLA of the PAYG usage aligns with the contract offering's SLA. Usage at a mismatched SLA tier must not trigger an alert, even if the volume exceeds capacity.
--- a/docs/integration-test-plans/Notification Service Integration Test Plan.md
+++ b/docs/integration-test-plans/Notification Service Integration Test Plan.md
@@ -77,6 +77,21 @@ Limitation:
    - Deduplication fires for identical `product + metric + SLA + usage` combo.
    - A new notification is produced when SLA changes and when usage changes, confirming both are part of the deduplication key.
 
+## stage-notification-deduplication-TC003
+
+1. **Description:** Exercises the **preferences_hash** element of the deduplication key for custom threshold notifications. Updating the custom utilization threshold changes the `preferences_hash` embedded in the event, causing the deduplication key to differ so the user is re-notified even if the same utilization level was already notified this month.
+2. **Setup:**
+   - The organization is opted-in for custom threshold notifications (`exceeded-custom-utilization-threshold` event type).
+   - The organization has set a custom threshold at a value below current usage (e.g., 80% with usage at 85%).
+   - The organization has already received a custom threshold notification this calendar month.
+3. **Actions and Verifications (in sequence):**
+   1. Trigger the utilization sync job → verify first custom threshold notification is received with a non-null `preferences_hash` in the context.
+   2. Trigger the utilization sync job again (same usage, same threshold) → assert **no new notification** is received (deduplication suppresses the duplicate).
+   3. Update the custom threshold to a different value that is still below current usage (e.g., 79%) → trigger the sync → verify **a new notification is received** and its `preferences_hash` differs from the first notification's hash.
+4. **Expected Result:**
+   - Deduplication suppresses notifications when the custom threshold configuration is unchanged.
+   - A new notification is delivered after the threshold is updated, confirming that the `preferences_hash` change resets the deduplication key.
+
 ## test_over_usage_notification_requires_payg_sla_aligned_with_offering
 
 1. **Description:** Over-usage notifications for PAYG products require that the SLA of the PAYG usage aligns with the contract offering's SLA. Usage at a mismatched SLA tier must not trigger an alert, even if the volume exceeds capacity.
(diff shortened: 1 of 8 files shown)


swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/BaseThresholdUtilizationHandlerService.java
```diff
@@ -172,21 +172,32 @@ private Action buildNotificationAction(
...
 
     if (isServiceLevelSet(payload.getSla())) {
       builder.withAdditionalProperty("service_level", payload.getSla().value());
     }
     if (isUsageSet(payload.getUsage())) {
       builder.withAdditionalProperty("usage", payload.getUsage().value());
     }
+    if (payload.getBillingAccountId() != null && !payload.getBillingAccountId().isEmpty()) {
+      builder.withAdditionalProperty("billing_account_id", payload.getBillingAccountId());
+    }
```
vbusch, line 200: billing_account_id needs to be included. It is being used in: https://github.com/vbusch/notifications-backend/blob/master/engine/src/main/java/com/redhat/cloud/notifications/events/deduplication/SubscriptionsDeduplicationConfig.java#L56

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
```diff
@@ -91,7 +89,21 @@ protected String metricName() {
     return CUSTOM_THRESHOLD_METRIC;
   }
 
-  static String hashLastUpdated(OffsetDateTime lastUpdated) {
-    return DigestUtils.sha256Hex(lastUpdated.toString());
+  @Override
+  protected void addAdditionalContextProperties(
+      Context.ContextBuilder builder, UtilizationSummary payload, MetricId metricId) {
+    preferenceRepository
```
Sgitario: this runs an additional SQL query when the value is already loaded in https://github.com/RedHatInsights/rhsm-subscriptions/pull/6198/changes#diff-d35d77d12aa49fc35c392333beb3576a3b4e0efc80135f7b9fef2aceddcc4065R50

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
```diff
@@ -91,7 +89,21 @@ protected String metricName() {
...
+        .getPreferences(payload.getOrgId())
+        .ifPresent(
+            preference -> {
+              String preferencesHash =
+                  computePreferencesHash(
+                      preference.getCustomThreshold(), preference.getLastUpdated());
```
Sgitario: is this change really necessary? when the custom threshold is updated, the last uptaded column is also updated, causing the reset of the dedup logic

Sgitario (changes requested):

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
vbusch: I changed it to pass the parameter in.  much better.  thank you.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
vbusch: This is needed to have the event during the threshold evaluation.  But it wasn't necessary to add the CustomThreshold as part of the hash.  So I have made the update to remove that parameter.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
```diff
@@ -91,7 +97,16 @@ protected String metricName() {
...
+      Context.ContextBuilder builder, UtilizationSummary payload, MetricId metricId, Event event) {
+    String hash = (String) event.getMetadata().getAdditionalProperties().get(PREFERENCES_HASH);
+    if (hash != null) {
+      builder.withAdditionalProperty(PREFERENCES_HASH, hash);
+    }
+  }
+
+  static String computePreferencesHash(OffsetDateTime lastUpdated) {
     return DigestUtils.sha256Hex(lastUpdated.toString());
   }
```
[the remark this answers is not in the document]
vbusch: lastUpdated should never be null.  The flow will prevent this.  And if it is null, the hash won't work.  It is impossible to get the null, so seeing the nullpointer error, would let us know there is a problem.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
```diff
@@ -68,8 +72,10 @@ protected Optional<Event> evaluateThreshold(
           String.format(PERCENT_FORMAT, utilizationPercent),
           threshold);
       var event = buildEvent(utilizationPercent);
-      String lastUpdatedHash = hashLastUpdated(preference.getLastUpdated());
-      event.getPayload().getAdditionalProperties().put("last_updated_hash", lastUpdatedHash);
+      event
+          .getMetadata()
+          .getAdditionalProperties()
+          .put(PREFERENCES_HASH, computePreferencesHash(preference.getLastUpdated()));
```
Sgitario: why not simplying setting the additional property here? the logic seems more complicated now.

swatch-utilization/TEST_PLAN.md
```diff
@@ -552,3 +552,21 @@ Additional cases can be added under the same `utilization-notifications-featuref
     - Both notifications are sent. The two notification types are independent (neither suppresses the other)
     - One custom threshold notification with `severity` `MODERATE`
     - One over-usage notification with `severity` `IMPORTANT`
+
+**custom-threshold-TC006 - Preferences hash changes when custom threshold is updated**
```
Sgitario: since we're not adding logic changes finally, is this new test plan necessary?

Sgitario (changes requested):

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerService.java
vbusch: It is more complicated because of the timing.  I need to wait for buildContext to run before the context is available.

swatch-utilization/TEST_PLAN.md
vbusch: removed it.

docs/integration-test-plans/Notification Service Integration Test Plan.md
```diff
@@ -77,6 +77,21 @@ Limitation:
    - Deduplication fires for identical `product + metric + SLA + usage` combo.
    - A new notification is produced when SLA changes and when usage changes, confirming both are part of the deduplication key.
 
+## stage-notification-deduplication-TC003
```
Sgitario: adding the test plan now makes me think that the plan is to implement this test as part of this ticket.
however, there is another ticket for doing this: https://redhat.atlassian.net/browse/SWATCH-4843

I would remove this new test plan for now (and maybe adding it in the description of SWATCH-4843 instead).

Sgitario (changes requested):
Apart from my comment in the integration test plan.
I've added a new commit: https://github.com/RedHatInsights/rhsm-subscriptions/pull/6198/changes/fca1c8318bcc692989ca2ad48ab0bafc00ce214e that eliminates the need to building the event in two places: one for the payload and another method for the context; and the workaround to pass data in the payload.

The refactor introduces an internal HandlerEvent DTO with three explicit buckets:
- payload → notification Event payload
- additionalContext → notification Action context
- metadata → notification Event metadata

The base class owns the mapping to the Cloud Notifications ingress model; subclasses only populate the DTO.

Let me know what you think about these changes.

docs/integration-test-plans/Notification Service Integration Test Plan.md
vbusch: Removed.

Sgitario (approved):

PR #6210, opened 2026-06-01
SWATCH-4803: Refactor custom threshold handler logic

Sgitario:
Jira issue: SWATCH-4803

## Description
Refactors custom utilization threshold handling to go through `OrgPreferencesService` and treat persisted org preferences as opt-in via `last_updated`. 

These changes centralize the preference logic, align the notification handler with the API contract (`last_updated` = configured org), and hardens against corrupt threshold data in the database.

- Expose `last_updated` on org preferences GET/POST when the org has saved preferences; omit it when returning the system default.
- Validate stored thresholds (0–100); on invalid values, log a warning and evaluate notifications using `ORG_PREFERENCE_DEFAULT_THRESHOLD`.
- Update OpenAPI, component tests, unit tests, and TEST_PLAN for the new behavior.

### DIFF
--- a/swatch-utilization/TEST_PLAN.md
+++ b/swatch-utilization/TEST_PLAN.md
@@ -403,11 +403,11 @@ Additional cases can be added under the same `utilization-notifications-featuref
 - **Verification**:
     - Verify response status code is 200
     - Verify response contains custom_threshold field
-    - Verify response last_updated is absent (field omitted; org has no persisted preferences)
+    - Verify response last_modified is absent (field omitted; org has no persisted preferences)
 - **Expected Result**:
     - HTTP 200 response
     - Response contains custom_threshold matching system default (ORG_PREFERENCE_DEFAULT_THRESHOLD config property)
-    - last_updated is absent (omitted from response body)
+    - last_modified is absent (omitted from response body)
 
 **org-preferences-TC002 - Update organization threshold preferences**
 
@@ -419,11 +419,11 @@ Additional cases can be added under the same `utilization-notifications-featuref
 - **Verification**:
     - Verify response status code is 200
     - Verify response contains the persisted custom_threshold value
-    - Verify response contains last_updated (UTC timestamp of persistence)
+    - Verify response contains last_modified (UTC timestamp of persistence)
 - **Expected Result**:
     - HTTP 200 response
     - Response custom_threshold matches the value from request body
-    - Response last_updated is present and matches the value returned by subsequent GET
+    - Response last_modified is present and matches the value returned by subsequent GET
 
 **org-preferences-TC003 - Retrieve custom threshold after update**
 
@@ -436,11 +436,11 @@ Additional cases can be added under the same `utilization-notifications-featuref
 - **Verification**:
     - Verify GET response status code is 200
     - Verify GET response contains the previously configured custom_threshold
-    - Verify GET response contains last_updated
+    - Verify GET response contains last_modified
 - **Expected Result**:
     - HTTP 200 response
     - Response custom_threshold matches the value set in previous POST request (not the system default)
-    - Response last_updated is present
+    - Response last_modified is present
 
 **org-preferences-TC004 - Update existing custom threshold**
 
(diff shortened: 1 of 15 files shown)


swatch-utilization/src/test/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerServiceTest.java
```diff
@@ -181,9 +180,27 @@ void shouldSendNotification_whenUsageIsOverCapacity() {
...
+    whenCheckSummary(summary);
+
+    verify(notificationsProducer, times(1))
+        .produce(argThat(action -> EVENT_TYPE.equals(action.getEventType())));
+    verify(orgPreferencesService, atLeastOnce()).getDefaultThreshold();
+    assertEquals(EXPECTED_SINGLE_INCREMENT, getCounterValue(PAYG_PRODUCT_ID, CORES_METRIC_ID));
+  }
+
   @Test
   void shouldNotSendNotification_whenNoOrgPreferenceExists() {
```
Mantzounis: I think this is kind of misleading now since the opt-in is now based on lastUpdated being null and not no DB row, maybe whenOrgHasNoPersistedPreferences

swatch-utilization/TEST_PLAN.md
```diff
@@ -477,6 +485,7 @@ Additional cases can be added under the same `utilization-notifications-featuref
 - **Expected Result**:
     - HTTP 200 response for both 0 and 100 threshold values
     - Response custom_threshold matches request value
+    - Response last_updated is present for each successful POST
```
Mantzounis: Based on this TC006 should assert that last_updated is present but that is not currently the case.

Mantzounis (changes requested):
With the changes in this PR this is now dead code.

https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/swatch-utilization/src/main/java/com/redhat/swatch/utilization/data/OrgUtilizationPreferenceRepository.java#L33

swatch-utilization/src/test/java/com/redhat/swatch/utilization/service/CustomThresholdUtilizationHandlerServiceTest.java
Sgitario: +1 renamed

swatch-utilization/TEST_PLAN.md
Sgitario: readded this check

Sgitario:
@Mantzounis please, can you take another look to these changes? Additionally to your comments, I fixed the conflicts with main and added a new commit to address some missing places where we are still using last updated instead of last modified name.

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/OrgPreferencesService.java
```diff
@@ -68,9 +77,11 @@ public OrgPreferencesResponse updateOrgPreferences(String orgId, OrgPreferencesR
     log.info("Updating utilization preference orgId={}", orgId);
     var entity = getOrCreateOrgPreferenceEntity(orgId);
     entity.setCustomThreshold(request.getCustomThreshold());
-    repository.persist(entity);
+    // using persist and flush, so the current_modified column is populated
```
Mantzounis: typo: this should be last_modified.

swatch-utilization/src/test/java/com/redhat/swatch/utilization/service/OrgPreferencesServiceTest.java
Mantzounis: since we are now explicitly calling repository.persistAndFlush(entity); the testname here is misleading, we should remove withoutpersist and maybe add verification.

Mantzounis, line 48: we could add assertNull(response.getLastModified()); here but feel free to ignore this since CT covers the real path.

Mantzounis (changes requested):

swatch-utilization/src/main/java/com/redhat/swatch/utilization/service/OrgPreferencesService.java
Sgitario: fixed

swatch-utilization/src/test/java/com/redhat/swatch/utilization/service/OrgPreferencesServiceTest.java
Sgitario: `persistAndFlush` is an implementation detail and it should not be mentioned in the test name. For avoid misunderstandings, I renamed the test name to `whenPreferenceExists_updatesThreshold`.

Sgitario: no, because we're mocking the repository and hence we're not exercising the real database interaction and this field will be always null.

Mantzounis (approved):
lgtm!

PR #6212, opened 2026-06-01
SWATCH-4980: Enforce org-admin check on org preferences.

Mantzounis:
Jira issue: SWATCH-4980

## Description
Enforces org-admin authorization on POST /api/rhsm-subscriptions/v1/utilization/org-preferences. Previously any authenticated user could update their org's custom utilization threshold. Now only org admins (is_org_admin: true in x-rh-identity) or service principals can write.

## Testing
podman compose up -d
make-build component-test swatch-utilization
Unit tests included

### DIFF
--- a/swatch-utilization/src/main/resources/META-INF/openapi.yaml
+++ b/swatch-utilization/src/main/resources/META-INF/openapi.yaml
@@ -16,6 +16,9 @@ paths:
       tags:
         - OrgPreferences
       operationId: getOrgPreferences
+      security:
+        - customer: []
+        - service: []
       summary: Get organization utilization preferences
       description: >-
         Retrieves the organization utilization preferences for the organization resolved
@@ -44,6 +47,9 @@ paths:
       tags:
         - OrgPreferences
       operationId: updateOrgPreferences
+      security:
+        - admin: []
+        - service: []
       summary: Update organization utilization preferences
       description: >-
         Accepts organization utilization preferences (for example custom over-usage threshold)
@@ -77,6 +83,22 @@ paths:
         "500":
           $ref: "../../../../../spec/error-responses.yaml#/$defs/InternalServerError"
 components:
+  securitySchemes:
+    customer:
+      type: apiKey
+      description: customer facing API
+      name: x-rh-identity
+      in: header
+    admin:
+      type: apiKey
+      description: API is available for org admin users
+      name: x-rh-identity
+      in: header
+    service:
+      type: apiKey
+      description: API is available for services
+      name: x-rh-swatch-psk
+      in: header
   schemas:
     OrgPreferencesRequest:
       type: object
--- a/swatch-utilization/src/main/resources/META-INF/openapi.yaml
+++ b/swatch-utilization/src/main/resources/META-INF/openapi.yaml
@@ -48,7 +48,7 @@ paths:
         - OrgPreferences
       operationId: updateOrgPreferences
       security:
-        - admin: []
+        - org_admin: []
         - service: []
       summary: Update organization utilization preferences
       description: >-
@@ -89,9 +89,9 @@ components:
       description: customer facing API
       name: x-rh-identity
       in: header
-    admin:
+    org_admin:
       type: apiKey
-      description: API is available for org admin users
+      description: API is available for organization administrators (is_org_admin=true in x-rh-identity)
       name: x-rh-identity
       in: header
     service:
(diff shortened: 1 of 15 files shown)


swatch-utilization/src/test/java/com/redhat/swatch/utilization/resources/OrgPreferencesResourceTest.java
```diff
@@ -106,44 +106,50 @@ void updateOrgPreferences_whenPayloadValid_invokesServiceWithResolvedOrgId() {
...
 
-  private static ValidatableResponse whenUpdateOrgPreferencesTo(Integer customThreshold) {
+  private static ValidatableResponse whenUpdateOrgPreferencesTo(Integer customThreshold)
+      throws Exception {
     OrgPreferencesRequest request = new OrgPreferencesRequest();
     request.setCustomThreshold(customThreshold);
 
     return given()
-        .header(RH_IDENTITY_HEADER, base64UserIdentity(ORG_ID))
+        .header(RH_IDENTITY_HEADER, base64UserIdentity(ORG_ID, true))
```
vbusch: A unit test to confirm that a HttpStatus.SC_FORBIDDEN is returned when attempting to updateOrgPreferences as non-admin would be useful.

There is an integration test, but a unit test could catch it sooner.

vbusch (changes requested):
The implementation looks good.  

I would suggest adding a unit test for the non-happy path.  But do realize it is covered at the integration level.

vbusch (dismissed):
The unhappy path is more difficult to test at the unit test level since %test.RBAC_ENABLED=false is set.  Since it is tested at the component level, there is coverage.

Thank you @Mantzounis for the explanation.

swatch-utilization/src/main/resources/META-INF/openapi.yaml
```diff
@@ -77,6 +83,22 @@ paths:
         "500":
           $ref: "../../../../../spec/error-responses.yaml#/$defs/InternalServerError"
 components:
+  securitySchemes:
+    customer:
+      type: apiKey
+      description: customer facing API
+      name: x-rh-identity
+      in: header
+    admin:
```
Sgitario: can we rename this to `customer_admin`? 
`admin` means to me like a service administrator instead.

swatch-utilization/src/test/java/com/redhat/swatch/utilization/resources/OrgPreferencesResourceTest.java
```diff
@@ -106,44 +106,50 @@ void updateOrgPreferences_whenPayloadValid_invokesServiceWithResolvedOrgId() {
   @ParameterizedTest
   @ValueSource(ints = {-1, 101})
   void updateOrgPreferences_whenThresholdOutOfRange_returnsBadRequestWithoutCallingService(
-      int invalidThreshold) {
+      int invalidThreshold) throws Exception {
```
Sgitario: please, do not add `throws Exception` to all the tests, catch the exception where needed and use assert fails accordingly.

swatch-utilization/src/test/java/com/redhat/swatch/utilization/resources/OrgPreferencesResourceTest.java
Sgitario: instead of true, use `IS_ORG_ADMIN` of boolean type with value `true`

Sgitario (changes requested):

swatch-utilization/src/test/java/com/redhat/swatch/utilization/resources/OrgPreferencesResourceTest.java
Mantzounis: +1 addressed this.

swatch-utilization/src/test/java/com/redhat/swatch/utilization/resources/OrgPreferencesResourceTest.java
Mantzounis: added the try catch and removed throws exception, thank you!

swatch-utilization/src/main/resources/META-INF/openapi.yaml
Mantzounis: I renamed this to org_admin to match the discussions and documents around the role, good callout!

Sgitario (approved):

PR #6216, opened 2026-06-02
SWATCH-5049: Expose public v2 OpenAPI from swatch-api.

lindseyburnett:
The v2 openapi URL was routed to swatch-api, which only serves the v1 spec and has no v2 handler, so unauthenticated requests were rejected and the API catalog could not load the document.

Serve the v2 public spec from swatch-contracts and route those requests through swatch-api nginx to the owning service.

## Summary by CodeRabbit

* **New Features**
  * Released API v2 with new SKU capacity report endpoint (`GET /v2/subscriptions/products/{product_id}`) supporting comprehensive filtering and pagination
  * Made OpenAPI specifications publicly accessible via JSON and YAML endpoints

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -16,6 +16,8 @@
   <properties>
     <api_spec_v1>${project.basedir}/rhsm-subscriptions-api-v1-spec.yaml</api_spec_v1>
     <api_config_v1>${project.basedir}/rhsm-subscriptions-api-v1-config.json</api_config_v1>
+    <api_spec_v2>${project.basedir}/rhsm-subscriptions-api-v2-spec.yaml</api_spec_v2>
+    <api_config_v2>${project.basedir}/rhsm-subscriptions-api-v2-config.json</api_config_v2>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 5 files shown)


Sgitario (dismissed):
The v2 public OpenAPI spec should be served from swatch-api, not swatch-contracts. To be precise, the nginx swatch-api service is our gateway and already publishes the v1 spec; v2 should follow the same pattern for consistency and so API catalog clients have a single entry point.

That swatch-contracts is the only service implementing v2 endpoints today is an implementation detail. Routing and ownership can change later (e.g. v2 APIs on swatch-tally or swatch-utilization).

lindseyburnett:
> The v2 public OpenAPI spec should be served from swatch-api, not swatch-contracts. To be precise, the nginx swatch-api service is our gateway and already publishes the v1 spec; v2 should follow the same pattern for consistency and so API catalog clients have a single entry point.
> 
> That swatch-contracts is the only service implementing v2 endpoints today is an implementation detail. Routing and ownership can change later (e.g. v2 APIs on swatch-tally or swatch-utilization).

Thanks for the input. Just a heads-up, this PR is still marked as work in progress and not ready for review yet, as I'm focusing on the immediate bug fix first.

Regarding the long-term architecture, swatch-api isn't actually just an nginx proxy—it's still running the monolith code under a different Spring Boot profile, which complicates using it as a simple gateway. Because this is a cross-cutting concern, I am currently mapping out the technical options and will make the final determination on how we structure this.

Once the bug fix is stable, I'll bring my proposal to the team so everyone can align on the implementation details. Please hold further comments until the PR no longer has the "work in progress; don't review yet." label.

TommasoLencioni (approved):

Sgitario (approved):

PR #6218, opened 2026-06-02
Update Konflux build-container hash

wottop:
## Description

We are getting a warning in the enterprise-contract tests that says this hash will be deprecated soon. 
Updated to the new hash.


wottop:
changed mid stream kruft in the chain of comments.

PR #6224, opened 2026-06-03
SWATCH-4803: Add exception mapper in swatch-utilization to log client validation exceptions

Sgitario:
Jira issue: SWATCH-4803

## Description
Before these changes, the API worked correctly by validating and rejecting wrong requests, but the warnings were not printed into the service logs.

After these changes, we now capture the validation logs using the standard exception mapper and log a warning. Additionally, we also print the wrong value that caused the error in the validation.

## Testing
Updated `updateOrgPreferences_whenThresholdOutOfRange_returnsBadRequestWithoutCallingService` to validate the log. The log now contains:
```
2026-06-03 11:23:09,338 WARN  [com.redhat.swatch.utilization.resources.ConstraintViolationExceptionMapper] (executor-thread-1) Client request failed validation: updateOrgPreferences.orgPreferencesRequest.customThreshold: rejected value -1 (must be greater than or equal to 0)
```

### DIFF
--- a/swatch-utilization/src/main/java/com/redhat/swatch/utilization/resources/ConstraintViolationExceptionMapper.java
+++ b/swatch-utilization/src/main/java/com/redhat/swatch/utilization/resources/ConstraintViolationExceptionMapper.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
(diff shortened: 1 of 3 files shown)


Mantzounis (approved):
lgtm!

PR #6231, opened 2026-06-04
Fix Mockito warning

awood:
Jira issue: None

## Description
This fixes a warning that has been popping up recently:

Mockito is currently self-attaching to enable the inline-mock-maker.
This will no longer work in future releases of the JDK. Please add
Mockito as an agent to your build as described in Mockito's
documentation:
https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3

We have to add the Java agent ourselves to the test JVM since Mockito
can no longer add the agent itself.

### Steps
1. Run `./mvnw test -pl swatch-tally -Dtest=InstancesDataExporterServiceTest` on
   `main` and this branch.  On `main` you'll see the warning message, but not on
   this branch.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,8 @@
     <sonar.branch.target>main</sonar.branch.target>
     <sonar.coverage.jacoco.xmlReportPaths>${maven.multiModuleProjectDirectory}/coverage-report/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
     <sonar.java.binaries>${project.build.directory}</sonar.java.binaries>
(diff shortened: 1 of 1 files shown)


wottop (approved):
The tests will not pass until this is merged.

awood:
Closing.  It's failing Tekton with an error "The forked VM terminated without properly saying goodbye. VM crash or System.exit called" which is what happens if `@{argLine}` in interpreted literally.  Putting `<argLine/>` under `<properties>` is supposed to solve that issue, but it's not working for some reason.  We could remove the `@{argLine}` but that can cause subtle problems with other plugins that use Java agents.  I'll figure it out later.

PR #6243, opened 2026-06-09
SWATCH-3246: fix subscription configuration properties and add subscription sync component tests

Sgitario:
Jira issue: SWATCH-3246

## Changes
The problem is that in clowdapp.yaml, we use SUBSCRIPTION_IGNORE_EXPIRED_OLDER_THAN=2M and SUBSCRIPTION_IGNORE_STARTING_LATER_THAN=2M.
In Spring, these properties standed for 2 months because it uses a custom duration parser, but in Quarkus, 2M stands for 2 minutes. 
The fix is to represent the duration in days: 60d. 

## Testing
Added four scenarios to extend our coverage for the subscription sync scenario: 
- subscriptions-sync-TC003 - Empty upstream org subscription list
- subscriptions-sync-TC004 - Upstream org subscription search failure
- subscriptions-sync-TC005 - Subscription missing from subscription search list
- subscriptions-sync-TC006 - Expired and replacement both returned from subscription search
- subscriptions-sync-TC007 - Subscription filtered because upstream start date is null
- subscriptions-sync-TC008 - Subscription filtered because upstream start date is too far in the future

## Release Notes

* **New Features**
  * Added date-range filtering capability for SKU capacity queries.

* **Tests**
  * Expanded subscription synchronization test coverage with six new scenarios covering reconciliation, search failures, subscription replacements, and edge cases.

### DIFF
--- a/swatch-contracts/TEST_PLAN.md
+++ b/swatch-contracts/TEST_PLAN.md
@@ -923,6 +923,89 @@ This section verifies the automatic contract termination behavior when contracts
   - Subscription entity created  
   - SubscriptionResponse: "Success"
 
+**subscriptions-sync-TC003 - Empty upstream org subscription list**
+- **Description**: When IT subscription search returns an empty list, all subscription rows previously stored for the org are removed.
+- **Setup**:
+  - One active subscription stored for the org
+  - IT subscription search returns HTTP 200 with an empty list
+- **Action**:
+  - Run subscription sync for the org
+- **Verification**:
+  - List subscriptions for the org via internal API
+- **Expected Results**:
+  - No subscription rows remain for the org
+
+**subscriptions-sync-TC004 - Upstream org subscription search failure**
+- **Description**: When IT subscription search fails, the previously stored subscriptions must remain unchanged.
+- **Setup**:
+  - One active subscription stored for the org
+  - IT subscription search returns an HTTP error code (bad gateway or internal server error)
+- **Action**:
+  - Run subscription sync for the org
+- **Verification**:
+  - List subscriptions for the org via internal API
+- **Expected Results**:
+  - The stored subscription is still present and unchanged
+
+**subscriptions-sync-TC005 - Subscription missing from subscription search list**
+- **Description**: When IT subscription search returns only one subscription for the org, any DB row whose `subscription_id` is not in that response is removed. Covers partial upstream responses and renewal: only the replacement is returned, it has a different `subscription_id`, and its start date is after the expired row — so days in the report window before the replacement start lose the expired row’s capacity in the capacity API.
+- **Setup**:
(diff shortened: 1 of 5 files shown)


swatch-contracts/TEST_PLAN.md
```diff
@@ -904,9 +904,9 @@ This section verifies the automatic contract termination behavior when contracts
 - **Note:** This endpoint **SUPPORTS multiple subscriptions** via JSON array
 
 **subscriptions-sync-TC001 - Sync all subscriptions for enabled orgs**  
-- **Description**: Verify POST /rpc/subscriptions/sync enqueues org subscriptions.  
+- **Description**: Verify PUT /rpc/subscriptions/sync enqueues org subscriptions.  
 - **Setup**: Configure sync-enabled orgs.  
-- **Action:** POST `/api/swatch-contracts/internal/rpc/subscriptions/sync`.  
+- **Action:** PUT `/api/swatch-contracts/internal/rpc/subscriptions/sync`.
```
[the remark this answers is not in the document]
Sgitario: fixed

swatch-contracts/ct/java/tests/SubscriptionsSyncComponentTest.java
```diff
@@ -60,10 +86,261 @@ void shouldSyncUmbSubscriptionMessage() {
...
+                + "–"
+                + inclusiveTo);
+      } else {
+        assertEquals(
+            expectedSockets,
+            actual,
+            0.01,
+            "Capacity on " + day + " should report expected socket capacity");
+      }
+    }
```
Mantzounis, line 329: is the a reason for calling the GET inside the for loop here and in the above methods? If not could we get the response once and then filter inside the for loop?

Mantzounis (changes requested):
question: I noticed the additions in domain are not called by any tests, what is the reason for adding them

swatch-contracts/ct/java/tests/SubscriptionsSyncComponentTest.java
Sgitario: I've refactored this method, let me know what you think.

Sgitario:
> question: I noticed the additions in domain are not called by any tests, what is the reason for adding them

I had thought to keep those methods because it was very useful during the investigation, but I'm fine removing them, we can always add it back when it's really used.

Mantzounis:
@Sgitario I'm fine with keeping them there if they are useful for debugging but we should document the decision in a comment to remember its intentional

Sgitario:
> @Sgitario I'm fine with keeping them there if they are useful for debugging but we should document the decision in a comment to remember its intentional

no problem, just removed it. we can always add it back, it's not super complicated.

Mantzounis (approved):
thanks for the updates! lgtm.

PR #6246, opened 2026-06-12
SWATCH-5076: Update the IT Product Service UMB Consumer with logging

Sgitario:
Jira issue: SWATCH-5076

## Description
Added the following log statement:
```
      log.info(
          "IT Product message consumed: source=umb, productCode={}, eventType={}, "
              + "productCategory={}",
          productEvent.getProductCode(),
          productEvent.getEventType(),
          productEvent.getProductCategory());
```

that will allow us to compare the messages received from umb and kafka (future task). 

Moreover, we don't need to add any specific metric because Quarkus already uses `quarkus_messaging_message_count_total` by channel that is the metric we need for grafana. 

## Testing
As in https://github.com/RedHatInsights/rhsm-subscriptions/pull/5964, we don't need to update any tests. 

## Summary by CodeRabbit

* **Chores**
  * Enhanced logging for product status messages: added an informational log that captures structured fields for productCode, eventType, and productCategory to improve visibility into incoming events.


Mantzounis (approved):
lgtm

PR #6254, opened 2026-06-16
SWATCH-5176: Add Kessel (RBACv2) authorization foundation

lindseyburnett:
Jira epic: [SWATCH-4153](https://redhat.atlassian.net/browse/SWATCH-4153)

## Overview

Adds Kessel (RBACv2) authorization as an alternative to legacy RBAC v1, gated behind an Unleash feature flag. When the flag is on, authorization checks go through Kessel Inventory API (gRPC) -> SpiceDB instead of the legacy rbac-service REST API. The flag can be toggled at runtime without restarting services.

### Architecture

```
swatch service
  -> Unleash flag check (swatch.common-security.use-kessel-rbac)
    -> Flag OFF: RbacRolesAugmentor -> legacy rbac-service (REST)
    -> Flag ON:  KesselRolesAugmentor -> kessel-inventory-api (gRPC :9000) -> SpiceDB
```

Quarkus services (swatch-contracts) and Spring services (swatch-tally, swatch-api, rhsm) are wired up.

### Changes

**Kessel gRPC client (both stacks):**
- `KesselAuthorizationService` (Quarkus, swatch-common-security) and `KesselService` (Spring, swatch-core) — gRPC clients using the Kessel Inventory SDK
- Resilience: configurable timeout (`swatch.kessel.timeout-ms`, default 5s), retry on transient gRPC failures (UNAVAILABLE, DEADLINE_EXCEEDED, RESOURCE_EXHAUSTED, ABORTED), automatic channel recreation on UNAUTHENTICATED or SHUTDOWN state
- Endpoint configured via `clowder.endpoints.kessel-inventory-api.url` (Quarkus) and `rhsm-subscriptions.kessel.endpoint` (Spring)
- Insecure mode only for now (OAuth2/TLS support deferred until Kessel enables it)

**Permission mapping:**
- Uses Kessel v2 relation names from [rbac-config PR #648](https://github.com/project-kessel/rbac-config/pull/648)
- `subscriptions:*:*` -> `subscriptions_report_view`
- `subscriptions:reports:read` -> `subscriptions_report_view`

**Principal ID format:**
- The Kessel inventory API prepends the domain prefix (`redhat/`) to principal IDs internally
- The gRPC subject `resource_id` must be the bare user ID (e.g. `user123`), **not** `redhat/user123`
- Using `Utils.principalSubject(id, domain)` from the SDK causes a double-prefix (`redhat/redhat/user123`); we build the `SubjectReference` directly instead

**Quarkus authorization (swatch-contracts):**
- `KesselRolesAugmentor` — Quarkus `SecurityIdentityAugmentor` that grants `customer` role when Kessel Check returns ALLOWED_TRUE
- `RbacRolesAugmentor` — modified to skip when Kessel flag is ON

**Spring authorization (swatch-tally, swatch-api, rhsm):**
- `IdentityHeaderAuthenticationDetailsService` — modified to call `KesselService` when flag is ON, falls back to `RbacService` when OFF
- `InsightsUserPrincipal` — added fallback to top-level `org_id` when `internal.org_id` is absent (backwards compatible; `internal.org_id` takes precedence when present)

**Unleash wiring:**
- Spring side: wired `io.getunleash` properties to Clowder `featureFlags` config (was previously unset, causing Unleash bean to be null)
- Default environment changed to `development` to match ephemeral Unleash flag configuration

**ClowdApp changes:**
- Added `kessel-inventory` as optional dependency to swatch-contracts, swatch-api, swatch-tally, swatch-utilization
- Removed unnecessary `kessel-inventory` dependency from swatch-system-conduit (PSK-only, no RBAC)
- Added `swatch-kessel-sa` placeholder secret to rhsm-clowdapp.yaml
- Registered Unleash flag in `.unleash/flags.json` for local dev

**Testing:**
- `MockKesselServer` — in-process gRPC mock for unit tests
- Canary integration test proving the full Kessel auth path on the v2 subscriptions endpoint
- Draft RBAC parity test plans for component tests

## Unleash Flag

Toggle name: `swatch.common-security.use-kessel-rbac`

```shell
# Enable
http post :4242/api/admin/projects/default/features/swatch.common-security.use-kessel-rbac/environments/development/on \
  Authorization:"*:*.unleash-insecure-admin-api-token"

# Disable
http post :4242/api/admin/projects/default/features/swatch.common-security.use-kessel-rbac/environments/development/off \
  Authorization:"*:*.unleash-insecure-admin-api-token"
```

## Ephemeral Testing

Verified end-to-end in ephemeral environment:
1. Telepresence connects to ephemeral cluster for service DNS resolution
2. Unleash flag toggle switches between RBAC and Kessel paths at runtime
3. Kessel Check calls go through kessel-inventory-api (gRPC :9000) -> SpiceDB
4. SpiceDB tuples (role, role_binding, workspace) grant `subscriptions_report_view` permission
5. Both swatch-contracts (Quarkus) and swatch-tally (Spring Boot) verified with both RBAC and Kessel paths
6. `SWATCH_TEST_APIS_ENABLED=false` / `DEV_MODE=false` required to actually enforce authorization (otherwise `test`/dev roles bypass it)

## Services with NO RBAC v1 calls (no changes needed)

| Service | Reason |
|---------|--------|
| swatch-metrics | All endpoints PSK-only |
| swatch-metrics-hbi | RBAC explicitly disabled (include-rbac=false) |
| swatch-billable-usage | All endpoints internal |
| swatch-producer-aws | All endpoints PSK-only |
| swatch-producer-azure | All endpoints PSK-only |
| swatch-system-conduit | All endpoints PSK-only |

[SWATCH-5176]: https://redhat.atlassian.net/browse/SWATCH-5176
[SWATCH-4153]: https://redhat.atlassian.net/browse/SWATCH-4153

## Summary by CodeRabbit

* **New Features**
  * Added optional Kessel (RBACv2) authorization for customer and service-account access, controlled by `swatch.common-security.use-kessel-rbac`.
  * Added support for `user_id` in identity payloads and principal resolution.
  * Added Kessel connectivity/auth configuration to deployed services via new template parameters and environment variables.
* **Bug Fixes**
  * Improved troubleshooting with clearer logging for allowlist and opt-in checks.
* **Documentation**
  * Updated RBAC parity and endpoint access-control test plan coverage for Kessel vs RBAC.
* **Tests**
  * Added/extended unit, component, and canary tests for Kessel/RBAC parity, permission mapping, and role augmentation.


swatch-common-security/src/main/java/com/redhat/swatch/common/security/KesselAuthorizationService.java
```diff
@@ -0,0 +1,191 @@
...
+import org.project_kessel.api.inventory.v1beta2.ClientBuilder;
+import org.project_kessel.api.inventory.v1beta2.KesselInventoryServiceGrpc.KesselInventoryServiceBlockingStub;
+import org.project_kessel.api.rbac.v2.Utils;
+
+@Slf4j
+@ApplicationScoped
+public class KesselAuthorizationService {
+
+  private static final String RBAC_APP_NAME = "subscriptions";
+  private static final String RBAC_ADMIN_PERMISSION = RBAC_APP_NAME + ":*:*";
```
josejulio: Hi! I'm not sure if this has been discussed before. 
Ideally we would prefer to use granular permissions rather than using these wildcard permissions to offload the logic of `can the user do x?` to Kessel.

Consider these 2 examples
```java
boolean hasSubscriptionWritePermission() {
    // Logic to know if the current user has access to write a subscription
    if hasRole(Admin) {
        return true;
    }

    if (getIdentity().isOrgAdmin()) {
       return true;
    }

    // Maybe other checks

    return false;
}
```

vs

```
boolean hasSubscriptionWritePermission() {
    return kessel.hasPermission(getUser(), RBAC_APP_NAME, "subscription", "write")
}
```

The first is app specific and other apps requiring to check the same permission would need to duplicate exactly to match the behavior, the second relies on that logic being in kessel already.

What we have there today (checking the willcard permission) will work and will be a good first step for the migration, but we will need to think about moving away from roles/logic in the app and moving that to kessel service.

swatch-common-security/src/main/java/com/redhat/swatch/common/security/KesselAuthorizationService.java
```diff
@@ -0,0 +1,278 @@
...
+    this.channel = result.getRight();
+    log.info(
+        "Kessel authorization client initialized: endpoint={} reason={}",
+        properties.endpoint(),
+        reason);
+
+    if (oldChannel != null) {
+      oldChannel.shutdown();
+    }
+  }
```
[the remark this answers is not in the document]
lindseyburnett: Fixed — added `synchronized` to `initializeChannel()` in both `KesselAuthorizationService` and `KesselService`.

swatch-core/src/main/java/org/candlepin/subscriptions/rbac/KesselService.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  private KesselInventoryServiceBlockingStub getClient() {
+    if (channel != null && channel.getState(false) != ConnectivityState.SHUTDOWN) {
+      return stub;
+    }
+    log.warn("Kessel gRPC channel is unhealthy, recreating");
+    initializeChannel("unhealthy_channel");
+    return stub;
+  }
```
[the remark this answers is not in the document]
lindseyburnett: Fixed — added `synchronized` to `initializeChannel()` in both services.

swatch-core/src/main/java/org/candlepin/subscriptions/security/InsightsUserPrincipal.java
```diff
@@ -56,10 +57,68 @@ public void setOrgId(String orgId) {
...
+    this.user = user;
+  }
+
   public String getOrgId() {
-    return internal.getOrgId();
+    String internalOrgId = internal.getOrgId();
+    if (internalOrgId != null && !internalOrgId.isBlank()) {
+      return internalOrgId;
+    }
+    return orgId;
```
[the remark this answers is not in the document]
lindseyburnett: Fixed — added null check: `String internalOrgId = internal == null ? null : internal.getOrgId();`

swatch-system-conduit/deploy/clowdapp.yaml
```diff
@@ -135,6 +135,16 @@ parameters:
...
   - name: OTEL_TRACES_TOKEN_SECRET_KEY
     value: token
+  - name: KESSEL_INSECURE
+    value: 'true'
+  - name: KESSEL_AUTH_ENABLED
+    value: 'false'
+  - name: KESSEL_AUTH_CLIENT_ID
+    value: ''
+  - name: KESSEL_AUTH_CLIENT_SECRET
+    value: ''
```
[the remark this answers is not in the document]
lindseyburnett: Fixed — replaced plain-text `KESSEL_AUTH_CLIENT_ID`/`KESSEL_AUTH_CLIENT_SECRET` params with `secretKeyRef` from `swatch-kessel-sa`, and added missing `KESSEL_ENABLED`/`KESSEL_URL` params.

swatch-system-conduit/deploy/clowdapp.yaml
```diff
@@ -179,6 +189,16 @@ objects:
...
+            - name: KESSEL_INSECURE
+              value: ${KESSEL_INSECURE}
+            - name: KESSEL_AUTH_ENABLED
+              value: ${KESSEL_AUTH_ENABLED}
+            - name: KESSEL_AUTH_CLIENT_ID
+              value: ${KESSEL_AUTH_CLIENT_ID}
+            - name: KESSEL_AUTH_CLIENT_SECRET
+              value: ${KESSEL_AUTH_CLIENT_SECRET}
+            - name: KESSEL_AUTH_OIDC_ISSUER
+              value: ${KESSEL_AUTH_OIDC_ISSUER}
```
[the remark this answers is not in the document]
lindseyburnett: Fixed — aligned with other deployments: `secretKeyRef` for credentials, added `KESSEL_ENABLED`/`KESSEL_URL`, added `optional: true`.

swatch-tally/deploy/clowdapp.yaml
```diff
@@ -467,6 +478,26 @@ objects:
...
+            - name: KESSEL_AUTH_CLIENT_ID
+              valueFrom:
+                secretKeyRef:
+                  key: client_id
+                  name: swatch-kessel-sa
+            - name: KESSEL_AUTH_CLIENT_SECRET
+              valueFrom:
+                secretKeyRef:
+                  key: client_secret
+                  name: swatch-kessel-sa
```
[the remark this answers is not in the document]
lindseyburnett: Fixed — added `optional: true` to all `swatch-kessel-sa` `secretKeyRef` blocks across swatch-tally, swatch-contracts, swatch-api, swatch-utilization, and swatch-system-conduit.

swatch-common-security/src/main/java/com/redhat/swatch/common/security/KesselRolesAugmentor.java
```diff
@@ -0,0 +1,110 @@
...
+  public Uni<SecurityIdentity> augment(
+      SecurityIdentity identity, AuthenticationRequestContext context) {
+    if (!unleash.isEnabled(KESSEL_FLAG)) {
+      return Uni.createFrom().item(identity);
+    }
+
+    Principal principal = identity.getPrincipal();
+    if (principal instanceof RhIdentityPrincipal rhIdentityPrincipal
+        && shouldCheck(rhIdentityPrincipal)) {
+      return context.runBlocking(() -> lookupKesselRoles(identity));
```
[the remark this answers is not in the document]
lindseyburnett: Acknowledged — the TOCTOU window is real but the risk is low since flag toggles are infrequent and the window is tiny (sub-millisecond). Addressing this would require architectural changes (shared request-scoped flag decision) that are out of scope for this PR. Filed as a follow-up.

swatch-api/deploy/clowdapp.yaml
```diff
@@ -144,6 +155,26 @@ objects:
                 value: ${USER_OPTS_APPEND}
               - name: DEV_MODE
                 value: ${DEV_MODE}
+              - name: KESSEL_ENABLED
+                value: ${KESSEL_ENABLED}
+              - name: KESSEL_URL
+                value: ${KESSEL_URL}
```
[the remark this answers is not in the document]
lindseyburnett: The `KESSEL_URL` parameter is wired through to the application via Clowder endpoint injection (`clowder.endpoints.kessel-inventory-api.url`) for Quarkus services and `rhsm-subscriptions.kessel.endpoint` for Spring services. The template parameter provides an override for non-Clowder environments.

Sgitario (changes requested):
Thanks for the solid foundation work on this. I have a few comments before we take this out of draft:

## 1. Prefer the `swatch-common-export` pattern for shared Kessel logic

Right now we have two parallel implementations (`KesselAuthorizationService` in Quarkus / `swatch-common-security`, and `KesselService` in Spring / `swatch-core`) that **are already diverging** (workspace resolution, RBAC OAuth for workspace fetch, principal ID handling, etc.).

For exports we solved this with `swatch-common-export`: keep the **business logic in a shared module**, expose **interfaces** for framework-specific concerns, and let Spring/Quarkus only do **wiring/adapters**.

I'd like Kessel to follow the same pattern:
- Put workspace lookup, permission→relation mapping, Check calls, retries/timeouts, and principal ID resolution in **one shared place**
- Define small interfaces for anything stack-specific (config binding, lifecycle, Unleash, identity types)
- Spring Boot / Quarkus services only implement those interfaces and register beans

## 2. Admin + reader both map to the same Kessel relation

Today both:
- `subscriptions:*:*`
- `subscriptions:reports:read`

map to **`subscriptions_report_view`**.

`getPermissions()` then loops those two v1 permission strings and calls `checkAccess()` for each. In production that means **two identical gRPC Checks** against the same subject/relation/workspace. If Kessel allows view, **both** permissions are added to the granted list.

On the Spring side, `RoleProvider` then grants **both** `SUBSCRIPTION_WATCH_ADMIN` and `SUBSCRIPTION_WATCH_REPORT_READER` whenever Kessel allows — so we lose the admin vs reader distinction that RBAC v1 still provides.

Practical impact today is limited because several Spring annotations already use `hasAnyRole(ADMIN, REPORT_READER)`.

Short-term: do **one** Check for `subscriptions_report_view`, and map the result deliberately (e.g. only emit `subscriptions:reports:read` until we have distinct relations).  
Longer-term (aligned with josejulio's comment): move toward **granular Kessel permissions** instead of reconstructing v1 wildcards in-app.

## 3. Changing the Unleash default environment from `production` → `development`

In `src/main/resources/application.yaml`, Spring Unleash now defaults to:

- `environment: ${UNLEASH_ENVIRONMENT:development}` (was `production`)
- a development-scoped default API token

This is convenient for ephemeral/local (flags often live under the `development` environment), but it is a **behavioral change for every Unleash flag** on Spring services (tally/api/rhsm), not just the Kessel flag.

I don't see `UNLEASH_ENVIRONMENT` set in the clowdapps or set in app-interface, so this change is dangerous. 

Also note the inconsistency: `swatch-system-conduit` still defaults to `production`.

## 4. Missing test coverage for the manual channel resilience paths

We now own a fair amount of custom gRPC lifecycle logic (`initializeChannel`, recreate on `SHUTDOWN` / `UNAUTHENTICATED`, retries on transient codes, fail-closed when `stub == null`). 

I don’t see tests that exercise the failure modes this manual code claims to handle, in particular:

1. **Kessel goes down / restarts mid-process**
   - Channel becomes unhealthy or RPCs return `UNAVAILABLE` / `DEADLINE_EXCEEDED`
   - Verify retry behavior, that we eventually deny (fail-closed) when Kessel stays down, and that requests succeed again after the server recovers — ideally **without** requiring a process restart

2. **Credential / auth rotation (or `UNAUTHENTICATED`)**
   - The code explicitly recreates the channel on `UNAUTHENTICATED` (token expiry path)
   - Please add a test where the stub first fails with `UNAUTHENTICATED`, channel/stub is recreated, and a subsequent `check` succeeds
   - If OAuth/TLS is still deferred and this path is speculative, either add the test against the current recreate behavior, or document that the recreate branch is untested / unused until auth is enabled — leaving production-only recovery logic untested is the risk

`MockKesselServer` and the happy-path unit/canary tests are useful, but they don’t cover “server dies” or “credentials renewed.” 

This is important to demostrate that this logic is actually necessary and/or ellaborate if we can use the standard grpc clients from Spring Boot/Quarkus frameworks instead.

## 5. Commit history is quite heavy for a draft

This branch has a large number of commits (merges + incremental fixes). Before asking for a full review / leaving draft, please **squash / clean up the history** into a small set of logical commits (e.g. shared client, Quarkus wiring, Spring wiring, deploy/config, tests). That will make review and bisect much easier.

## 6. Minor

- Prefer Lombok `@Slf4j` instead of `LoggerFactory.getLogger(...)` for new/touched classes, consistent with most of the codebase (e.g. the Quarkus Kessel classes already use `@Slf4j`).

src/main/java/org/candlepin/subscriptions/security/KesselConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Bean
+  @ConfigurationProperties(prefix = "rhsm-subscriptions.kessel")
+  public KesselProperties kesselProperties() {
+    return new KesselProperties();
+  }
+
+  @Bean(initMethod = "init", destroyMethod = "shutdown")
+  public KesselAuthorizationClient kesselAuthorizationClient(KesselProperties props) {
+    return new KesselAuthorizationClient(props, orgId -> "default");
```
Sgitario: In the Spring Boot services, it's using the "default" workspace, where in the Quarkus services, it uses "FetchWorkspace". Did you do this on purpose?

swatch-core/pom.xml
```diff
@@ -43,6 +43,32 @@
...
+      <groupId>org.project-kessel</groupId>
+      <artifactId>kessel-sdk</artifactId>
+      <version>1.7.1</version>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
+      <version>1.78.0</version>
+      <scope>runtime</scope>
+    </dependency>
```
Sgitario: these dependencies should not longer be necessary there are transitive dependencies from swatch-common-kessel now.

swatch-common-kessel/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.project-kessel</groupId>
+      <artifactId>kessel-sdk</artifactId>
+      <version>1.7.1</version>
```
Sgitario: declare kessel-sdk.version in the parent pom and using the latest version: 1.9.0

swatch-common-security/pom.xml
```diff
@@ -30,6 +30,33 @@
...
+      <groupId>org.project-kessel</groupId>
+      <artifactId>kessel-sdk</artifactId>
+      <version>1.7.1</version>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
+      <version>1.78.0</version>
+      <scope>runtime</scope>
+    </dependency>
```
Sgitario: these dependencies should not longer be necessary there are transitive dependencies from swatch-common-kessel now.

swatch-core/pom.xml
```diff
@@ -43,6 +43,32 @@
...
+    <dependency>
+      <groupId>com.nimbusds</groupId>
+      <artifactId>oauth2-oidc-sdk</artifactId>
+      <version>11.30.2</version>
```
Sgitario: I think this dependency is not used in the spring boot services, maybe because these services are always using the "default" workspace?

swatch-common-kessel/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.nimbusds</groupId>
+      <artifactId>nimbus-jose-jwt</artifactId>
+      <version>10.9.1</version>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
+      <version>1.78.0</version>
```
Sgitario: I'm concerned of independently versioning up to four different dependencies: kessel, nimbus-jose-jwt, grpc-netty and oauth2-oidc-sdk; that are all related to each other. If tomorrow, we bump one of them, it might be incompatible with the rest. 

I think nimbus-jose-jwt is only used to import the Pair class, so we could avoid this. 

However, the grpc dependency is already in conflict with the grpc dependency that is managed by the Spring Boot / Quarkus frameworks. 
If we bump kessel to the latest version 1.9.0, it mitigates this because it also bumps the grpc dependencies to the closer ones to Spring Boot / Quarkus's.

Sgitario (changes requested):
Thanks for the new changes!

I think there are still three important points to address:
- difference of behaviour between spring boot (it uses "default" workspace) and quarkus (it uses FetchWorkspace).
- risk of conflicts / misalignment of the new dependencies and/or declare new dependencies in the parent pom.xml file
- use the latest kessel-sdk to 1.9.0 (there is an email from the Kessel team about asking for using this version)

There was some failures in the swatch-contracts component tests that I fixed in the latest commit.

Sgitario (dismissed):
@lindseyburnett thanks for addressing some of the requested changes. 
I see that keeping the grpd and the oauth2-oidc-sdk dependencies are necessary for kessel to work, still I would like to have the versions declared in the root parent pom.xml as the rest of the important dependencies. 
Also, the comment about the "default" workspace in spring boot has not been addressed yet as long as the CodeRabbit's feedback. 

Finally, I think the class `KesselService` in swatch-tally is dead code now, this could be a left-over since the logic from the swatch-common-kessel is now used.

src/main/java/org/candlepin/subscriptions/security/KesselConfiguration.java
lindseyburnett: Ah good catch.  It needs to be FetchWorkspace

ericThompson038 (dismissed):
lgtm

kartikshahc:
I have got RBAC v1 and v2 wiremock approval for swatch-tally component test from both DEV and QE so, merging the changes to this feature branch. Closing out 
[Enable wiremock RBAC for IQE ephemeral environment and component tests for swatch-tally](https://github.com/RedHatInsights/rhsm-subscriptions/pull/6404#top)

ericThompson038 (dismissed):
lgtm

swatch-common-kessel/src/main/java/com/redhat/swatch/kessel/KesselAuthorizationClient.java
```diff
@@ -0,0 +1,286 @@
...
+    var workspaceId = workspaceResolver.getDefaultWorkspaceId(orgId);
+
+    var request =
+        CheckRequest.newBuilder()
+            .setSubject(
+                SubjectReference.newBuilder()
+                    .setResource(
+                        ResourceReference.newBuilder()
+                            .setResourceType("principal")
+                            .setResourceId(subjectId)
```
josejulio: I think this is missing the `KESSEL_DOMAIN`, but i'm not sure if it is used anywhere else. I could only see `KESSEL_DOMAIN` being used in logs.

There is a [convenience function in the sdk](https://github.com/project-kessel/kessel-sdk-java/blob/44be11c484c5689ffce0001ec8b0dd57ffe7bc85/kessel-sdk/src/main/java/org/project_kessel/api/rbac/v2/Utils.java#L59C36-L59C52) that takes the user id and the domain to build the resource reference with the type.

wottop (dismissed):
Nothing to add over addressing the CodeRabbit suggestions

ericThompson038 (dismissed):
LGTM

swatch-common-kessel/src/main/java/com/redhat/swatch/kessel/KesselAuthorizationClient.java
```diff
@@ -0,0 +1,286 @@
...
+                            .setReporter(ReporterReference.newBuilder().setType("rbac").build())
+                            .build())
```
josejulio: Should be replaced with [org.project_kessel.api.rbac.v2.Utils.principalSubject](https://github.com/project-kessel/kessel-sdk-java/blob/44be11c484c5689ffce0001ec8b0dd57ffe7bc85/kessel-sdk/src/main/java/org/project_kessel/api/rbac/v2/Utils.java#L59C36-L59C52)
```
Utils.principalSubject(subjectId, KESSEL_DOMAIN))
```

josejulio (changes requested):

swatch-tally/RBAC_PARITY_TEST_PLAN.md
TrayvonMcKnight, line 1: We have these test cases documented here: 
https://github.com/RedHatInsights/rhsm-subscriptions/blob/main/swatch-tally/test-plans/rbac-authorization.md

So we don't need this Test Plan file

kartikshahc: Addressed your feedback can you please check and resolve the comment.

swatch-common-kessel/src/main/java/com/redhat/swatch/kessel/KesselAuthorizationClient.java
kartikshahc: Addressed your feedback can you please check and resolve the comment?

swatch-common-kessel/src/main/java/com/redhat/swatch/kessel/KesselAuthorizationClient.java
kartikshahc: Addressed your feedback and refactored the code can you please check and resolve the comment

josejulio (dismissed):
Did a quick check from Kessel side and this lgtm, thanks!

kartikshahc:
We found a critical bug where the service account changes for Kessel aren't working correctly in Spring. We are investigating so holding off to get this PR merged.

kartikshahc:
Found an issue after running component test against RBAC. The naming convention for KESSEL has mismatch. `KESSEL_URL` is set on the pod but nothing reads it. `KESSEL_ENDPOINT` is what the app reads but no template sets it. They should be unified to `KESSEL_ENDPOINT` so changing that.

vbusch (approved):
Changes look complete.  And match with https://github.com/RedHatInsights/curiosity-frontend/pull/1915

wottop (approved):

TrayvonMcKnight (approved):
LGTM!

IQE tests are good - no regressions.

Will followup on Kessel Integration tests.

josejulio (approved):
Kessel permission checks looks good to me, thanks!

PR #6259, opened 2026-06-18
SWATCH-5052: Add not-null constraint to host tally buckets

wottop:
Jira issue: SWATCH-5052

## Description

The tables tally_snapshots and host_tally_buckets all have the four fields of sla/usage/billing_provider and billing_account_id. These are treated as not-null by the code, so the task is to make the database constrain them as such.

Of the eight columns, 7 have had that constraint added over several changesets. This will take care of the final one: host_tally_buckets.sla. 

### Verification

1. Run the following to confirm that no null values exist in the prod and stage databases
```select * from tally_snapshots where sla is null limit 10```

## Summary by CodeRabbit

* **Chores**
  * Enhanced data integrity through the application of stricter validation constraints on core system data to improve reliability and prevent incomplete records.


mstead (approved):
Looks good. Had concerns about the pre-check on the changeset, but it runs fast enough against stage and prod. ✔️

PR #6268, opened 2026-06-23
Fix BaseSMHBIComponentTest so that it flushes the transactional outbox 

barnabycourt:
Make BaseSMHBIComponentTest flush the outbox during thenSwatchEventsAppear
    
 When the thenSwatchEventsAppear() method was entered, if the work was not yet completed, then the first call to flushOutboxSynchronously would do nothing.  The method would never be called again because waitForKafkaMessage() would block for 30 seconds.
    
 In order to work around the blocking nature of KafkaBridgeService.waitForKafkaMessage(..) this uses AwaitUtilitiSettings.onConditionNotMet to trigger the flushOutboxSynchronously on each loop if the conditions have not yet been met.

### DIFF
--- a/swatch-metrics-hbi/ct/java/com/redhat/swatch/hbi/events/ct/tests/BaseSMHBIComponentTest.java
+++ b/swatch-metrics-hbi/ct/java/com/redhat/swatch/hbi/events/ct/tests/BaseSMHBIComponentTest.java
@@ -28,10 +28,8 @@ import com.redhat.swatch.component.tests.api.Quarkus;
 import com.redhat.swatch.component.tests.api.Unleash;
 import com.redhat.swatch.component.tests.api.UnleashService;
 import com.redhat.swatch.component.tests.utils.AwaitilitySettings;
-import com.redhat.swatch.component.tests.utils.AwaitilityUtils;
 import com.redhat.swatch.component.tests.utils.Topics;
 import com.redhat.swatch.hbi.events.ct.api.SwatchMetricsHbiRestService;
-import java.time.Duration;
 import org.candlepin.subscriptions.json.Event;
 
 @ComponentTest(name = "swatch-metrics-hbi")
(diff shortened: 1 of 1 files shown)


awood (dismissed):

swatch-metrics-hbi/ct/java/com/redhat/swatch/hbi/events/ct/tests/BaseSMHBIComponentTest.java
```diff
@@ -50,20 +48,16 @@ public class BaseSMHBIComponentTest {
...
-
-    AwaitilityUtils.untilAsserted(
-        () -> {
-          service.flushOutboxSynchronously();
-          for (var expectedMessage : expectedMessages) {
-            kafkaBridge.waitForKafkaMessage(
-                Topics.SWATCH_SERVICE_INSTANCE_INGRESS, expectedMessage, 1);
-          }
-        },
-        settings);
```
Sgitario, line 67: have you tried using the `AwaitilitySettings.onConditionNotMet` method ? 
This is invoked when a condition is not met in each iteration (it will keep trying) and we use it for flushing the billable usage topic which I think it's the same use case than here. 

See an example of usage in `TallySummaryConsumerComponentTest.whenHourlyAggregatesAreReceived`.

If the `onConditionNotMet` works, we would avoid adding a new parameter in a method that already has many parameters.

Sgitario (changes requested):

swatch-metrics-hbi/ct/java/com/redhat/swatch/hbi/events/ct/tests/BaseSMHBIComponentTest.java
barnabycourt: That's a much cleaner solution. I'll have an update for the PR shortly.  :+1:

Sgitario (approved):
lgtm!

PR #6276, opened 2026-06-25
Update resteasy.version to v7.0.2.Final

lindseyburnett:


Sgitario:
@lindseyburnett we cannot bump resteasy to 7.0.2.Final because it's incompatible with `resteasy-spring-boot-starter-common`. It was reported by SWATCH-4900

PR #6288, opened 2026-06-29
SWATCH-5072: Add IT Subscription Service Kafka tests to the swatch-contracts test plan

vbusch:
Jira issue: SWATCH-5072

## TEST PLAN UPDATES
- Added UMB feature flag tests (subscriptions-creation-TC008, TC011, TC012): flag entirely disabled; UMB consumer disabled via config variant; UMB enabled with Kafka consumer disabled via variant.  These scenarios were previously untested.
- Added "Subscription Management via Kafka (IT Subscription Service)" section  (subscriptions-kafka-TC001–TC014) covering the same happy-path, error, and   feature-flag scenarios as the UMB section, plus three additional edge cases
  (null/empty optional fields, invalid billing provider, null payload).  UMB coverage is intentionally mirrored because the UMB consumer is planned for removal.
- Kafka tests run against a real Kafka broker via Kafka Bridge (same infrastructure  as all other CT-module tests: Docker Compose locally, Bonfire on OpenShift).  
- XML deserialization is exercised end-to-end in every happy-path TC; channel routing is implicitly verified by the message reaching the consumer and producing a subscription in the database.

## Summary by CodeRabbit

* **Documentation**
  * Updated the subscription management test plan header and expanded IT Subscription coverage for both UMB and Kafka consumers.
  * Added scenarios covering null/empty optional billing fields, invalid/unknown billing providers with warning logs, and consumer enablement gating via service flags and configuration variants.
  * Added Kafka message tests for happy-path creation, malformed XML rejection (warn + no partial persistence), missing required fields validation (warn + continued processing), and independence between UMB and Kafka consumers.

### DIFF
--- a/swatch-contracts/TEST_PLAN.md
+++ b/swatch-contracts/TEST_PLAN.md
@@ -863,7 +863,7 @@ This section verifies the automatic contract termination behavior when contracts
   - Contract synced with upstream end_date
   - Subscription retains all contract-provided state
 
-## Subscription Management via UMB
+## Subscription Management via IT Subscription
 
 **subscriptions-creation-TC001 - Process a valid UMB subscription XML message from UMB**  
 - **Description**: Verify subscription creation via UMB XML message.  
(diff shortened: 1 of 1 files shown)


swatch-contracts/TEST_PLAN.md
```diff
@@ -963,6 +963,196 @@ This section verifies the automatic contract termination behavior when contracts
   - `end_date` set to termination date  
   - Status reflects termination
 
+**subscriptions-creation-TC008 - Ignore UMB message when IT subscription service flag is disabled**
+- **Description**: Verify that when the Unleash flag
+  `swatch.swatch-contracts.enable-it-subscription-service` is **disabled**, the UMB
+  consumer exits early and no subscription is persisted.
```
Sgitario, line 999: what does `the UMB consumer exits early` mean?

swatch-contracts/TEST_PLAN.md
```diff
@@ -963,6 +963,196 @@ This section verifies the automatic contract termination behavior when contracts
...
+- **Setup**:
+  - Ensure `UMB_ENABLED=true`
+  - Unleash toggle `swatch.swatch-contracts.enable-it-subscription-service` is disabled
+- **Action**: Publish a valid `CanonicalMessage` XML to `VirtualTopic.canonical.subscription` channel
+- **Verification**: Wait ≥ 3 s, then query subscriptions
```
Sgitario: `Wait ≥ 3 s` we don't wait a fixed amount of time, we should wait using poll for the specific log message.

swatch-contracts/TEST_PLAN.md
```diff
@@ -963,6 +963,196 @@ This section verifies the automatic contract termination behavior when contracts
...
+**subscriptions-creation-TC011 - Ignore UMB message when UMB consumer disabled via variant**
+- **Description**: Verify that a `config` variant payload of
+  `{"umb_consumer_enabled":false}` blocks the UMB consumer while the flag itself
+  remains enabled.
+- **Setup**:
+  - Ensure `UMB_ENABLED=true`
+  - Unleash flag `swatch.swatch-contracts.enable-it-subscription-service` enabled
+  - `config` variant set with `umb_consumer_enabled=false`
+- **Action**: Publish a valid `CanonicalMessage` XML to `VirtualTopic.canonical.subscription` channel
+- **Verification**: Wait ≥ 3 s, then query subscriptions
```
Sgitario: Same here: https://github.com/RedHatInsights/rhsm-subscriptions/pull/6288/changes#r3496372405

swatch-contracts/TEST_PLAN.md
```diff
@@ -963,6 +963,196 @@ This section verifies the automatic contract termination behavior when contracts
...
+  - Ensure `UMB_ENABLED=true`
+  - Unleash flag `swatch.swatch-contracts.enable-it-subscription-service` enabled
+  - `config` variant set with `{"kafka_consumer_enabled":false,"umb_consumer_enabled":true}`
+- **Action**: Publish a valid `CanonicalMessage` XML to `VirtualTopic.canonical.subscription` channel
+- **Verification**: Poll subscriptions until created
+- **Expected Result**:
+  - Subscription created with correct `quantity` and `sku`
+  - Kafka consumer is independently disabled (not tested in this TC)
+
+## Subscription Management via Kafka (IT Subscription Service)
```
Sgitario: the UMB channel will be eventually removed and we will only support the Kafka consumer, so I would not create a new section `subscriptions-kafka-TCXXX`, but probably I would group these new scenarios into `subscriptions-sync-TCXXX`.

Sgitario (changes requested):
The `subscriptions-kafka-TCXXX` are mostly duplicated with `Subscription Management via UMB`. 
Note that the UMB channel will be deleted very soon, so what if we rename this section to `Subscription Management via IT Subscription` and we reuse the existing scenarios to cover both UMB and Kafka for now?

The specific scenarios only for umb or kafka could be grouped to `subscriptions-umb-TCXXX` and `subscriptions-kafka-TCXXX`.

swatch-contracts/TEST_PLAN.md
vbusch: It short circuits without persisting.  Updated the wording.

swatch-contracts/TEST_PLAN.md
vbusch: I updated it to use the delay polling.   That way, it will delay to give it time to appear before assuming the lack of a message is a pass.

swatch-contracts/TEST_PLAN.md
vbusch: same as above.  now updated

swatch-contracts/TEST_PLAN.md
vbusch: UMB and Kafka are under new title 'Subscription Management via IT Subscription'.
I renamed all the umb tests to have 'umb' which will make them easy to delete.

Sgitario (changes requested):
@vbusch I didn't mean to rename all the test cases in Subscription Management via UMB to `subscriptions-creation-umb-TCXXX`
My point was that some of these test cases are umb or kafka specific (like processing invalid UMB xml payload, etc), but most of the test cases are logic specific where umb or kafka is only relevant to send a message, but the rest is how we use the IT Subscription API which is the important part. 

To sum up:
- `subscriptions-creation-TCXXX` to cover the service logic. We could keep using UMB and create a separate follow up ticket to change it to Kafka which is going to be the correct one. 
- `subscriptions-creation-umb-TCXXX` to cover the only relevant to UMB logic
- `subscriptions-creation-kafka-TCXXX` to cover the only relevant to Kafka logic

vbusch:
> @vbusch I didn't mean to rename all the test cases in Subscription Management via UMB to `subscriptions-creation-umb-TCXXX` My point was that some of these test cases are umb or kafka specific (like processing invalid UMB xml payload, etc), but most of the test cases are logic specific where umb or kafka is only relevant to send a message, but the rest is how we use the IT Subscription API which is the important part.
> 
> To sum up:
> 
> * `subscriptions-creation-TCXXX` to cover the service logic. We could keep using UMB and create a separate follow up ticket to change it to Kafka which is going to be the correct one.
> * `subscriptions-creation-umb-TCXXX` to cover the only relevant to UMB logic
> * `subscriptions-creation-kafka-TCXXX` to cover the only relevant to Kafka logic

After the 16th minute discussion, I have renamed the tests.  Added the operation to all of them.  Left many as generic, and a few as umb/kafka.

Sgitario (approved):

PR #6351, opened 2026-07-10
Add feature-flags to the /info endpoint for swatch-tally 

barnabycourt:
When running component tests there was flakiness & extra sleeps being introduced because we did not have a way definitively validate that a feature flag was enabled for the tally service. This does a few things: 
1) add feature flag information to the /info rest endpoint 
2) Update the component tests to query this endpoint to validate the state of a flag 
3) Shorten the polling interval for swatch-tally to query feature flag information to 1 second when working in dev mode 
4) Shorten the AwaitUtilitySettings default polling interval to 200 millisecods

## Summary by CodeRabbit

* **New Features**
  * Added a “feature-flags” section to the Actuator “info” output, and enabled it in Swatch Core management output.
  * Introduced a configurable feature-toggle fetch interval (default 15s; dev profile uses 1s).
* **Bug Fixes**
  * Improved feature-flag change verification by polling the management “info” payload and confirming expected boolean values.
* **Tests**
  * Added/expanded coverage for feature-flags reporting, including empty/null handling.
  * Updated integration test helpers and switched polling to faster millisecond-based checks.


liwalker-rh (dismissed):
Changes look good and tests are all still passing after improvements in run time.

swatch-core/src/main/java/org/candlepin/subscriptions/actuator/FeatureFlagsInfoContributor.java
```diff
@@ -0,0 +1,66 @@
...
+import org.candlepin.subscriptions.configuration.FeatureFlags;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.autoconfigure.info.ConditionalOnEnabledInfoContributor;
+import org.springframework.boot.actuate.autoconfigure.info.InfoContributorFallback;
+import org.springframework.boot.actuate.info.Info.Builder;
+import org.springframework.boot.actuate.info.InfoContributor;
+import org.springframework.stereotype.Component;
+
+/** InfoContributor to report the status of Unleash feature flags. */
+@Component
```
kartikshahc: Actuator package isn't component scanned, so the Component annotation on the class is never evaluated.

swatch-core/src/main/java/org/candlepin/subscriptions/actuator/FeatureFlagsInfoContributor.java
```diff
@@ -0,0 +1,66 @@
...
+@ConditionalOnEnabledInfoContributor(
+    value = "feature-flags",
+    fallback = InfoContributorFallback.DISABLE)
```
kartikshahc: These lines are not needed since this is already covered while defining bean in ApplicationConfiguration
https://github.com/RedHatInsights/rhsm-subscriptions/pull/6351/changes#diff-4283b445b09f2e2aa6386d3c7ca9b8efddc16fd3c0fbc46467bb328dbd52fc4bR164-R166

swatch-core/src/main/java/org/candlepin/subscriptions/actuator/FeatureFlagsInfoContributor.java
```diff
@@ -0,0 +1,66 @@
...
+public class FeatureFlagsInfoContributor implements InfoContributor {
+
+  @Autowired(required = false)
```
kartikshahc: Let's do constructor injection instead 
 private final FeatureFlags featureFlags;

  public FeatureFlagsInfoContributor(@Nullable FeatureFlags featureFlags) {
    this.featureFlags = featureFlags;
  }

src/main/java/org/candlepin/subscriptions/ApplicationConfiguration.java
```diff
@@ -158,4 +159,12 @@ public static RdsSslBeanPostProcessor rdsSslBeanPostProcessor(Environment env) {
   public CertInfoContributor certInfoContributor(ApplicationContext context) {
     return new CertInfoContributor(context);
   }
+
+  @Bean
+  @ConditionalOnEnabledInfoContributor(
+      value = "feature-flags",
+      fallback = InfoContributorFallback.DISABLE)
+  public FeatureFlagsInfoContributor featureFlagsInfoContributor() {
```
kartikshahc: Once you change the constructor injection you can do 
public FeatureFlagsInfoContributor featureFlagsInfoContributor(
      @Nullable FeatureFlags featureFlags) {
    return new FeatureFlagsInfoContributor(featureFlags);
  }

kartikshahc (changes requested):

kartikshahc:
Left a few comments otherwise lgtm

swatch-core/src/main/java/org/candlepin/subscriptions/actuator/FeatureFlagsInfoContributor.java
```diff
@@ -0,0 +1,56 @@
...
+
+    Map<String, Boolean> flagStatus = new LinkedHashMap<>();
+    flagStatus.put(
+        FeatureFlags.ENABLE_PRIMARY_ROW_SEARCHES,
+        featureFlags.isEnabled(FeatureFlags.ENABLE_PRIMARY_ROW_SEARCHES));
+    flagStatus.put(
+        FeatureFlags.ENABLE_HTB_PRIMARY_ROW_SEARCHES,
+        featureFlags.isEnabled(FeatureFlags.ENABLE_HTB_PRIMARY_ROW_SEARCHES));
+
+    builder.withDetail("feature-flags", flagStatus);
```
mstead, line 51: Would be nice if we didn't have to remember to add the feature flags here when we define a new one. Perhaps source the names from the location they are defined so that they automatically get added here.

mstead (changes requested):
One small suggestion.

swatch-core/src/main/java/org/candlepin/subscriptions/actuator/FeatureFlagsInfoContributor.java
barnabycourt: @mstead Do you have any suggestions? The only place I saw the feature flags defined was the static string on the existing FeatureFlags object.

kartikshahc (dismissed):
LGTM so far unless you are making changes for dynamically loading the flags

mstead (dismissed):
My comment was addressed. Looks good.

kartikshahc (approved):

swatch-core/src/main/java/org/candlepin/subscriptions/configuration/FeatureFlags.java
```diff
@@ -25,6 +25,7 @@ public interface FeatureFlags {
   String ENABLE_PRIMARY_ROW_SEARCHES = "swatch.swatch-tally.enable-primary-row-searches";
   String ENABLE_HTB_PRIMARY_ROW_SEARCHES =
       "swatch.swatch-tally.enable-host-tally-bucket-primary-row-searches";
+  String[] FLAG_LIST = {ENABLE_HTB_PRIMARY_ROW_SEARCHES, ENABLE_HTB_PRIMARY_ROW_SEARCHES};
```
kartikshahc: This is duplicate. Please change to ENABLE_PRIMARY_ROW_SEARCHES, ENABLE_HTB_PRIMARY_ROW_SEARCHES

kartikshahc (changes requested):

swatch-core/src/main/java/org/candlepin/subscriptions/configuration/FeatureFlags.java
barnabycourt: Good catch, thanks!

kartikshahc (approved):

PR #6363, opened 2026-07-14
Fix flaky billable usage aggregation component test

TommasoLencioni:
Kafka Streams windows were closing early when flush ran on every waitForKafkaMessage poll, emitting partial hourly aggregates. Drain stale windows before each run, burst-ingest tallies, gate on remittances, then flush once per untilAsserted retry with a strict complete-aggregate validator.

### Verification

To test:
```bash
/mvnw test -Pcomponent-tests -pl swatch-billable-usage/ct -Dtest=tests.BillableUsageAggregationComponentTest#shouldAggregateMultipleTalliesIntoSingleHourlyMessage
```

## Summary by CodeRabbit

* **Tests**
  * Improved hourly billable usage aggregation test coverage.
  * Added validation for aggregate totals, organization identifiers, and remittance counts.
  * Tests now handle pending remittance statuses and clear stale aggregation data between runs.

### DIFF
--- a/swatch-billable-usage/ct/java/tests/BillableUsageAggregationComponentTest.java
+++ b/swatch-billable-usage/ct/java/tests/BillableUsageAggregationComponentTest.java
@@ -30,16 +30,31 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import api.MessageValidators;
 import com.redhat.swatch.component.tests.api.TestPlanName;
 import com.redhat.swatch.component.tests.utils.AwaitilitySettings;
+import com.redhat.swatch.component.tests.utils.AwaitilityUtils;
 import com.redhat.swatch.component.tests.utils.RandomUtils;
 import domain.BillingProvider;
+import domain.RemittanceStatus;
+import java.time.Duration;
+import java.util.List;
+import org.candlepin.subscriptions.billable.usage.BillableUsage;
 import org.candlepin.subscriptions.billable.usage.BillableUsageAggregate;
 import org.candlepin.subscriptions.billable.usage.BillableUsageAggregateKey;
 import org.candlepin.subscriptions.billable.usage.TallySummary;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class BillableUsageAggregationComponentTest extends BaseBillableUsageComponentTest {
 
   private static final double BILLING_FACTOR = ROSA.getBillingFactor(CORES);
+
+  /**
+   * Drain hourly windows left by prior CT runs on shared compose Kafka. An early flush while only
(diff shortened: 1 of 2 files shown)


swatch-billable-usage/ct/java/tests/BillableUsageAggregationComponentTest.java
```diff
@@ -30,17 +30,32 @@
...
 
   private static final double BILLING_FACTOR = ROSA.getBillingFactor(CORES);
 
+  /**
+   * Drain hourly windows left by prior CT runs on shared compose Kafka. An early flush while only
+   * some billable usages are in the window permanently splits the hourly rollup.
+   */
+  @BeforeEach
+  void drainStaleAggregationWindows() {
+    service.flushBillableUsageAggregationTopic();
```
Sgitario, line 56: this is not working as claude thinks it works.
what this API is doing is to send an async request to all the partitions to the hourly kafka topic, so this does not guarante that when running the test, the time window will be closed prematurally.

swatch-billable-usage/ct/java/tests/BillableUsageAggregationComponentTest.java
```diff
@@ -64,50 +79,57 @@ public void shouldAggregateMultipleTalliesIntoSingleHourlyMessage() {
...
 
-    BillableUsageAggregate aggregate =
-        kafkaBridge
-            .waitForKafkaMessage(
-                BILLABLE_USAGE_HOURLY_AGGREGATE,
-                MessageValidators.billableUsageAggregateMatchesOrg(orgId),
-                1,
-                AwaitilitySettings.defaults()
-                    .onConditionNotMet(service::flushBillableUsageAggregationTopic))
-            .get(0);
```
Sgitario, line 91: this is the preferred way of flushing the topic while waiting for the aggregation.
if the original problem was that several aggregations were sent, what we should do (and what the iqe tests do), is to expect a list of aggregations and sum them for the assertions.

Sgitario (changes requested):
@TommasoLencioni I think the test was written correctly as it was before your changes. 
The only required change is to expect several aggregations (which is the same thing that iqe tests do). 
This is really caused a design gap in the hourly aggregation logic that we want to refactor in [SWATCH-3233](https://redhat.atlassian.net/browse/SWATCH-3233)

TommasoLencioni:
Fix will be addressed in [SWATCH-5256](https://redhat.atlassian.net/browse/SWATCH-5256).

[SWATCH-5256]: https://redhat.atlassian.net/browse/SWATCH-5256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4972, opened 2025-09-19
SWATCH-3927: Access capacity information via swatch-contracts

awood:
Jira issue: SWATCH-3927

## Description
Now that swatch-contracts controls the offering/subscription objects, remove as
much direct access to those tables as possible from swatch-tally.  This PR
removes swatch-tally implementations of capacity and replaces them with an API
call to swatch-contracts.  However, it does not remove the AccountResetService's
call into the subscription table to delete all subscriptions for an org.  No
corresponding API for this deletion exists in swatch-contracts.  A follow-on
card will be necessary to create this API and then remove the last pieces of the
offering/subscription objects from swatch-tally.

### Setup
1. Start `swatch-tally`
    ```shell
    RHSM_SUBSCRIPTIONS_CONTRACTS_CLIENT_URL=http://localhost:8011 DEV_MODE=true make swatch-tally
    ``` 

### Steps
1.  Request a tally
    ```shell
    http :8000/api/rhsm-subscriptions/v1/tally/products/RHEL%20for%20x86/Cores \
    granularity==Monthly \
    billing_category==prepaid \
    beginning==2021-08-01T00:00:00.000Z \
    ending==2023-08-31T23:59:59.999Z \
    use_running_totals_format==true \
    x-rh-identity:$(echo -n '{"identity":{"account_number":"","type":"User","user":{"is_org_admin":true},"internal":{"org_id":"org123"}}}' | base64 -w 0)
    ```
1. The request fails with an error message indicating a connection refused.
1. Now start swatch-contracts
    ```shell
    make swatch-contracts
    ```
1. Rerun the `http` command.

### Verification
1.  The tally call returns successfully.

PR #4824, opened 2025-08-05
SWATCH-2296: Remove the "swatch-subscription-sync" clowadpp"

Sgitario:
Jira issue: SWATCH-2296

Re-apply https://github.com/RedHatInsights/rhsm-subscriptions/pull/4807 that was already approved.

## Description
This service is no longer in use.

## Testing
Regression testing only.

MR: https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/1282

PR #6392, opened 2026-07-22
SWATCH-5312: Republish customer-facing RHSM v1 APIs that still exist at runtime

diegomaranhao:
Jira issue: SWATCH-5312

## Description

- Republish customer-facing RHSM v1 APIs that still exist at runtime (nginx → swatch-contracts / swatch-utilization) but were dropped from the public OpenAPI when ownership moved out of swatch-tally.
- Restore `/v1/capacity/products/{product_id}/{metric_id}`, `/v1/subscriptions/billing_account_ids`, deprecated `/v1/subscriptions/products/{product_id}`, and add `/v1/utilization/org-preferences to api/rhsm-subscriptions-api-v1-spec.yaml`, including the schemas needed for those paths.
- Document that the public v1/v2 specs describe the full customer surface, including routes implemented by other services and proxied by swatch-api.

This does not reintroduce tally ownership of those endpoints. Implementations stay in swatch-contracts / swatch-utilization; only the hosted public catalog (`/api/rhsm-subscriptions/v1/openapi.json`) is corrected so console docs and iqe-bindings can regenerate clients (unblocking goatee → iqe-bindings for IQE).

Related: SWATCH-2296 / #4824, SWATCH-3927 / #4972, SWATCH-3995 / #5807.

## Summary by CodeRabbit

* **New Features**
  * Added capacity reports by product and metric, including filtering and pagination with a configurable time range.
  * Added an endpoint to retrieve subscription billing account IDs.
  * Added organization utilization preference retrieval and update endpoints.
  * Introduced new capacity snapshot and utilization preference data models.
* **Documentation**
  * Clarified that public API specs include customer-facing routes provided by supporting services.
* **Deprecation**
  * Marked the existing SKU capacity report endpoint as deprecated.

### DIFF
--- a/api/rhsm-subscriptions-api-v1-spec.yaml
+++ b/api/rhsm-subscriptions-api-v1-spec.yaml
@@ -431,6 +431,365 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/InstanceResponse'
+  /v1/capacity/products/{product_id}/{metric_id}:
+    description: 'Operations for capacity report for a given account and product'
+    parameters:
+      - name: product_id
+        in: path
+        required: true
+        schema:
+          $ref: "#/components/schemas/ProductId"
+        description: "The ID for the product we wish to query"
+      - name: metric_id
+        in: path
+        required: true
+        schema:
+          $ref: "#/components/schemas/MetricId"
+        description: "The metric ID for the product we wish to query"
+      - name: offset
+        in: query
+        schema:
+          type: integer
+        description: "The number of items to skip before starting to collect the result set"
+      - name: limit
+        in: query
+        schema:
+          type: integer
+          minimum: 1
+        description: "The numbers of items to return"
+    get:
+      summary: "Fetch a capacity report for an account and product."
+      operationId: getCapacityReportByMetricId
+      parameters:
+        - name: billing_account_id
+          in: query
+          required: false
+          schema:
+            $ref: "#/components/schemas/BillingAccountId"
+          description: "The billing account ID for the product we wish to query"
+        - name: granularity
+          in: query
+          required: true
+          schema:
+            $ref: '#/components/schemas/GranularityType'
+          description: "The level of granularity to return."
+        - name: category
+          in: query
+          schema:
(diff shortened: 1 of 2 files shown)


api/rhsm-subscriptions-api-v1-spec.yaml
```diff
@@ -431,6 +431,365 @@ paths:
...
+                    product_tag: "rhods"
+        '400':
+          $ref: "../spec/error-responses.yaml#/$defs/BadRequest"
+        '403':
+          $ref: "../spec/error-responses.yaml#/$defs/Forbidden"
+        '500':
+          $ref: "../spec/error-responses.yaml#/$defs/InternalServerError"
+      tags:
+        - contracts
+  /v1/subscriptions/products/{product_id}:
```
Sgitario, line 591: I'm not following why we're reintroducing this deprecated API. 
IQE should already be using the v2. Actually, this v1 codebase should be eventually deleted.

Sgitario (changes requested):

api/rhsm-subscriptions-api-v1-spec.yaml
diegomaranhao: Agreed IQE should use v2 for SKU capacity. Keeping the v1 path in the public OpenAPI is separate: the endpoint is deprecated, not obsolete. It still runs in swatch-contracts and is still routed. We don't have a deprecation policy established, but I think we should keep deprecated APIs (and their docs) until they're removed. Dropping it from the spec early left console/iqe-bindings out of sync with the runtime. Once the API is deleted, we remove it from the spec.

For the IQE side, tracked in [SWATCH-5315](https://redhat.atlassian.net/browse/SWATCH-5315): finish removing the remaining v1 SKU capacity call sites from iqe-rhsm-subscriptions-plugin (most coverage already uses v2).

Sgitario: this API is deprecated and obsolete because it contains bugs that were addressed as part of the v2.

Sgitario (approved):
@diegomaranhao approving, but please, create a folow up JIRA ticket to remove it as soon as possible

PR #6395, opened 2026-07-23
SWATCH-4605: Align Kafka IT partner gateway logic with UMB.

Mantzounis:
Jira issue: SWATCH-4605

## Description
This PR matches the aligns Kafka consumer logic with the existing UMB consumer.
I also changed the log in contract service to be agnostic.

## Testing
Unit tests included

```
If you want to override the default IQE test image tag (rhsm-subscriptions), add /use-iqe-image-tag=XXX anywhere in this PR description, replacing XXX with your tag. If you leave XXX unchanged, CI keeps the default tag.
```

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.

## Summary by CodeRabbit

- **New Features**
  - Incoming partner entitlement messages now automatically create partner contracts.
- **Bug Fixes**
  - JSON parsing failures are now handled more transparently (warned and treated as errors rather than silently ignored).
  - Clearer logging when partner contracts can’t be processed.
  - Null messages are safely ignored before any processing checks.
- **Tests**
  - Expanded and adjusted consumer tests for successful processing, invalid JSON, null inputs, and scenarios where processing is disabled.

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsPartnerEntitlementMessageConsumer.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsPartnerEntitlementMessageConsumer.java
@@ -22,9 +22,12 @@ package com.redhat.swatch.contract.resource;
 
 import static com.redhat.swatch.contract.config.Channels.CONTRACTS_FROM_GATEWAY;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.redhat.swatch.contract.config.FeatureFlags;
+import com.redhat.swatch.contract.model.PartnerEntitlementsRequest;
 import com.redhat.swatch.contract.openapi.model.PartnerEntitlementContract;
+import com.redhat.swatch.contract.service.ContractService;
 import io.smallrye.common.annotation.Blocking;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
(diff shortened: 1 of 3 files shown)


swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsPartnerEntitlementMessageConsumer.java
```diff
@@ -37,52 +40,53 @@ public class ContractsPartnerEntitlementMessageConsumer {
...
+        contract.getAction(),
+        awsCustomerAccountId,
+        productCode,
+        azureResourceId,
+        contract.getRedHatSubscriptionNumber());
 
-    } catch (Exception e) {
-      log.warn("Unable to read IT Partner Kafka message from JSON.", e);
-    }
+    service.createPartnerContract(PartnerEntitlementsRequest.from(contract));
```
Sgitario: in the UMB consumer, we also add a debug log to print the statusresponse:

```
StatusResponse response = consumeContract(dtoContractString);
log.debug("umb response: {}", response);
```

I think it would be good to also align this.

Sgitario (changes requested):
Changes look good to me, only a minor thing.

swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsPartnerEntitlementMessageConsumer.java
Mantzounis: thanks for the suggestion, I pushed the change.

Sgitario (approved):

PR #6402, opened 2026-07-23
SWATCH-5161: Inflated hypervisor usage for a customer account

wottop:
Jira issue: SWATCH-5161

## Description

When we changed over to the is_primary rows for report queries, we found a problem with the implementation. For other report categories outside of HYPERVISOR, summing the values across the search filter result gave the correct answer.

For the HYPERVISOR category, the summation would lead to largely excessive values. The solution in the near term is to revert the query to the '_ANY' version regardless of the unleash flag setting for is_primary.

### Setup

1. You need to pull the most current swatch-support-scripts
2. ```podman compose up -d```
3. oc login to prod
4. in your swatch-support-scripts directory, run
```./import_hbi_hosts_by_org.sh --org 15327614```

### Steps

1. ```make swatch-tally```
2. Opt in (you may need to add the org to the org_config table first)
```http PUT ":8010/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots/15327614" x-rh-swatch-psk:placeholder```
3. Run the nightly tally
```http PUT ":8010/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots/15327614" x-rh-swatch-psk:placeholder```

### Verification

1. Run the report command adjusting for the current date
```http GET ':8010/api/rhsm-subscriptions/v1/tally/products/RHEL for x86/Sockets' x-rh-identity:$(echo '{"identity":{"internal":{"org_id":"15327614"}}}' | base64 -w0) granularity==DAILY category==hypervisor beginning=='2026-08-04T00:00Z' ending=='2026-08-06T23:59Z'```
2. You should see a valid value for the daily. It cannot exceed the number of hypervisor sockets for the org, Note that this is the data from the org in the bug.

```
{
    "data": [
        {
            "date": "2026-08-04T00:00:00Z",
            "has_data": false,
            "value": 0
        },
        {
            "date": "2026-08-05T00:00:00Z",
            "has_data": false,
            "value": 0
        },
        {
            "date": "2026-08-06T00:00:00Z",
            "has_data": true,
            "value": 44
        }
    ],
    "meta": {
        "count": 3,
        "granularity": "Daily",
        "metric_id": "Sockets",
        "product": "RHEL for x86"
    }
}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Hypervisor reports now use snapshot data when primary row searches are enabled.
  * Hypervisor socket totals correctly exclude guest sockets and handle mapped guests, SLA filters, and wildcard reports.

* **Tests**
  * Added coverage for snapshot-based hypervisor reports and socket tally scenarios.

### DIFF
--- a/src/main/java/org/candlepin/subscriptions/resource/api/v1/TallyResource.java
+++ b/src/main/java/org/candlepin/subscriptions/resource/api/v1/TallyResource.java
@@ -168,8 +168,10 @@ public class TallyResource implements TallyApi {
             : Collections.emptyMap();
 
     // Execute query based on feature flag - each path returns unified result
+    // Note that ReportCategory.HYPERVISOR is not supported by the primary row search
     TallyQueryResult queryResult =
         featureFlags.isEnabled(FeatureFlags.ENABLE_PRIMARY_ROW_SEARCHES)
+                && !ReportCategory.HYPERVISOR.equals(category)
             ? executeAggregateQuery(reportCriteria, metricId, category)
             : executeSnapshotBasedQuery(reportCriteria, metricId, category);
     List<UnroundedTallyReportDataPoint> snaps = queryResult.dataPoints();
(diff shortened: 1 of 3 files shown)


mstead (changes requested):
This looks good and functions as expected. A couple of requests:

1. The component test should be added to the test plan and have an appropriate test id. See other tests for an example.
2. Regarding the code rabbit suggestions, they are simple enough to address so I suggest doing them.
3. If you wouldn't mind, the TallyResource.executeAggregateQuery method has an info.log stating "Using primary row searches for tally report" -- This is my bad from another commit, but it should be a debug statement.

mstead (approved):
LGTM

PR #6403, opened 2026-07-23
SWATCH-5295: Persist Partner Gateway licenseArn as contract licenseId

TommasoLencioni:
Persist AWS `partnerIdentities.licenseArn` from Partner Gateway as nullable `licenseId` on `ContractEntity` (DB column + mapper). No remittance or prepaid math changes. API exposure is out of scope (`SWATCH-5296`).

Also scopes AWS contract matching and orphan termination by `subscriptionNumber` when present so contracts that share `billingProviderId` but have different `subscriptionNumber`s stay distinct (legacy null `subscriptionNumber` still matches on `billingProviderId` only).

Jira: [SWATCH-5295](https://redhat.atlassian.net/browse/SWATCH-5295)

### Verification

Unit:
```bash
./mvnw test -pl swatch-contracts -am -Dtest=ContractEntityMapperTest,ContractServiceTest
```

Component:
```bash
podman compose up -d kafka kafka-bridge kafka-setup unleash amqp wiremock db
./mvnw clean install -Pcomponent-tests -pl swatch-contracts/ct -am -Dtest=ContractsSyncComponentTest
```

[SWATCH-5295]: https://redhat.atlassian.net/browse/SWATCH-5295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/swatch-contracts/TEST_PLAN.md
+++ b/swatch-contracts/TEST_PLAN.md
@@ -863,6 +863,43 @@ This section verifies the automatic contract termination behavior when contracts
   - Contract synced with upstream end_date
   - Subscription retains all contract-provided state
 
+**contracts-sync-TC019 - Two AWS contracts with same billingProviderId both persist**
+- **Description**: Two AWS contracts for the same org/product/AWS account (same `billingProviderId`) but different `subscriptionNumber` and `partnerIdentities.licenseArn` must both survive sync. Syncing must not delete one when processing the other.
+- **Setup**:
+  - Stub Partner Gateway with two AWS ROSA entitlements sharing product code, aws customer id, and seller account id
+  - Different `subscriptionNumber` and `partnerIdentities.licenseArn` on each
+  - Stub Search API for both subscription numbers
+  - Sync offering for the shared SKU
+- **Action**: POST `/api/swatch-contracts/internal/rpc/sync/contracts/{org_id}`
+- **Verification**:
+  - Two contracts for the org
+  - Each `subscriptionNumber` present
+  - Re-sync with the same upstream stubs leaves both contracts (stable UUIDs)
+- **Expected Result**: HTTP 200. Both contracts persisted. Second sync is idempotent
+
+**contracts-sync-TC020 - Re-sync updates only one of two AWS contracts sharing billingProviderId**
+- **Description**: After two AWS contracts with the same `billingProviderId` exist, changing capacity/`endDate` for contract A only must update A and leave B unchanged.
+- **Setup**:
+  - Stub Partner Gateway with two AWS ROSA entitlements that share product code, aws customer id, and seller account id (same `billingProviderId`)
+  - Different `subscriptionNumber` and `partnerIdentities.licenseArn` on each entitlement
+  - Stub Search API for both subscription numbers and sync the shared SKU offering
+  - Sync contracts once so both contracts exist
+  - Re-stub Partner Gateway with A’s metrics/`endDate` changed and B unchanged
+- **Action**: POST `/api/swatch-contracts/internal/rpc/sync/contracts/{org_id}`
+- **Verification**:
+  - Contract for A’s `subscriptionNumber` reflects the new end date / capacity
+  - Contract for B’s `subscriptionNumber` keeps prior UUID and end date
+- **Expected Result**: HTTP 200. Only contract A updated
+
+**contracts-sync-TC021 - Entitlement without Partner Gateway licenseArn syncs with null licenseId**
+- **Description**: Legacy entitlements with no `partnerIdentities.licenseArn` still sync successfully; contract is created with `licenseId` null (no failure when the field is absent).
+- **Setup**:
+  - Stub one AWS ROSA entitlement **without** `partnerIdentities.licenseArn`
+  - Stub Search API and sync offering
+- **Action**: POST `/api/swatch-contracts/internal/rpc/sync/contracts/{org_id}`
+- **Verification**: One contract for the org; sync status SUCCESS
+- **Expected Result**: HTTP 200; sync succeeds with no failure when `partnerIdentities.licenseArn` is absent
+
 ## Subscription Management via IT Subscription
 
 **subscriptions-creation-TC001 - Process a valid UMB subscription XML message from UMB**  
--- a/swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
+++ b/swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
@@ -37,9 +37,11 @@ import api.PartnerApiStubs;
 import com.redhat.swatch.component.tests.api.TestPlanName;
 import com.redhat.swatch.component.tests.utils.AwaitilityUtils;
 import com.redhat.swatch.component.tests.utils.RandomUtils;
+import com.redhat.swatch.configuration.registry.MetricId;
 import com.redhat.swatch.contract.test.model.SkuCapacitySubscription;
 import domain.BillingProvider;
 import domain.Contract;
+import domain.Product;
 import io.restassured.response.Response;
 import java.time.OffsetDateTime;
 import java.util.List;
@@ -64,16 +66,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
     Contract azureContract =
         Contract.buildRosaContract(orgId, BillingProvider.AZURE, Map.of(CORES, 20.0), sku);
 
-    // Stub offering and partner API
-    wiremock.forProductAPI().stubOfferingData(awsContract.getOffering());
-    wiremock
-        .forPartnerAPI()
-        .stubPartnerSubscriptions(forContractsInOrgId(orgId, awsContract, azureContract));
-    wiremock.forSearchApi().stubGetSubscriptionBySubscriptionNumber(awsContract, azureContract);
-
-    // Sync offering needed for contracts to persist with the SKU
-    Response syncOffering = service.syncOffering(sku);
-    assertThat("Sync offering should succeed", syncOffering.statusCode(), is(HttpStatus.SC_OK));
+    stubContractsForOrgSync(awsContract, azureContract);
 
     // When: Sync contracts for the organization
     Response syncResponse = service.syncContractsByOrg(orgId);
@@ -161,8 +154,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
     // Setup new upstream contracts (different from existing - same SKU, different provider)
     Contract newContract =
         Contract.buildRosaContract(orgId, BillingProvider.AZURE, Map.of(CORES, 20.0), sku);
-    wiremock.forPartnerAPI().stubPartnerSubscriptions(forContractsInOrgId(orgId, newContract));
-    wiremock.forSearchApi().stubGetSubscriptionBySubscriptionNumber(newContract);
+    stubPartnerAndSearch(newContract);
 
     // When: Delete existing data and re-sync contracts
     service.deleteDataForOrg(orgId);
@@ -320,10 +312,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
 
     wiremock.forProductAPI().stubOfferingData(awsContract.getOffering());
     // Both contracts are returned by the partner API for this org
-    wiremock
-        .forPartnerAPI()
-        .stubPartnerSubscriptions(forContractsInOrgId(orgId, awsContract, azureContract));
-    wiremock.forSearchApi().stubGetSubscriptionBySubscriptionNumber(awsContract, azureContract);
+    stubPartnerAndSearch(awsContract, azureContract);
 
     assertThat(
         "Sync offering should succeed",
@@ -388,8 +377,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
         Contract.buildRosaContract(orgId, BillingProvider.AWS, Map.of(CORES, coresCapacity), sku);
 
     wiremock.forProductAPI().stubOfferingData(contract.getOffering());
-    wiremock.forPartnerAPI().stubPartnerSubscriptions(forContractsInOrgId(orgId, contract));
-    wiremock.forSearchApi().stubGetSubscriptionBySubscriptionNumber(contract);
+    stubPartnerAndSearch(contract);
 
     assertThat(
         "Sync offering should succeed",
@@ -540,8 +528,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
     // Stub upstream to return a different entitlement (different billing_provider_id)
     Contract newContract =
         Contract.buildRosaContract(orgId, BillingProvider.AWS, Map.of(CORES, 20.0), sku);
-    wiremock.forPartnerAPI().stubPartnerSubscriptions(forContractsInOrgId(orgId, newContract));
-    wiremock.forSearchApi().stubGetSubscriptionBySubscriptionNumber(newContract);
+    stubPartnerAndSearch(newContract);
 
     // Action: Sync contracts
     Response syncResponse = service.syncContractsByOrg(orgId);
@@ -644,8 +631,7 @@ public class ContractsSyncComponentTest extends BaseContractComponentTest {
     assertNotNull(endDateBefore, "Contract should have an end_date");
 
     // Stub upstream to return the same Azure entitlement
-    wiremock.forPartnerAPI().stubPartnerSubscriptions(forContractsInOrgId(orgId, azureContract));
-    wiremock.forSearchApi().stubGetSubscriptionBySubscriptionNumber(azureContract);
+    stubPartnerAndSearch(azureContract);
 
     // When: Sync contracts
     Response syncResponse = service.syncContractsByOrg(orgId);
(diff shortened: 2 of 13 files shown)


swatch-contracts/src/test/java/com/redhat/swatch/contract/service/ContractServiceTest.java
```diff
@@ -412,6 +416,73 @@ void syncContractByOrgTerminatesContractsFromVanishedEntitlement() throws Except
     wireMockServer.removeStub(stub);
   }
 
+  @Test
+  @Transactional
+  void syncContractByOrgTerminatesMissingSiblingAwsAgreementSharingBillingProviderId()
+      throws Exception {
```
Mantzounis, line 422: I think there would be value in adding another test like syncContractByOrgPreservesBothConcurrentAwsAgreementsWhenBothPresentUpstream to cover the happy path as well. I know it is covered in component tests, but I think there is merit to testing the happy path earlier as well.

swatch-contracts/TEST_PLAN.md
```diff
@@ -863,6 +863,34 @@ This section verifies the automatic contract termination behavior when contracts
...
+  - Different `subscriptionNumber` and `licenseArn` on each entitlement
+  - Stub Search API for both subscription numbers and sync the shared SKU offering
+  - Sync contracts once so both agreements exist
+  - Re-stub Partner Gateway with A’s metrics/`endDate` changed and B unchanged
+- **Action**: POST `/api/swatch-contracts/internal/rpc/sync/contracts/{org_id}`
+- **Verification**:
+  - Contract for A’s `subscriptionNumber` reflects the new end date / capacity
+  - Contract for B’s `subscriptionNumber` keeps prior UUID and end date
+- **Expected Result**: HTTP 200. Only agreement A updated
+
```
Mantzounis, line 893: in the card we mention this test path: "Legacy without licenseArn: ingest an entitlement with no licenseArn → contract is created/updated with licenseId = null and sync succeeds (no failure)." which is not covered here right now. I know it is implicitly covered but maybe there is value in adding an explicit tests to have documentation of the behavior in the plan. Though feel free to ignore this suggestion.

Mantzounis (changes requested):
Code looks very good to me! I left 2 suggestions on enhancing the test suites.

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -493,27 +497,46 @@ public StatusResponse syncContractsByOrgId(String contractOrgSync) {
...
+    }
+    if (contract.getSubscriptionNumber() == null) {
+      return upstreamBillingProviderIds.contains(contract.getBillingProviderId());
+    }
+    return upstreamAgreementKeys.contains(
+        agreementKey(contract.getBillingProviderId(), contract.getSubscriptionNumber()));
+  }
+
+  private static String agreementKey(String billingProviderId, String subscriptionNumber) {
+    return billingProviderId + '\0' + subscriptionNumber;
```
Sgitario: I'm a bit concerned about what this key implies for future readers. Encoding identity with '\0' concatenation feels fragile and easy to misunderstand later.

Could we replace the two Sets with a single Set of a small type that holds the fields we need (e.g. billingProviderId + subscriptionNumber), instead of concatenating strings?

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
```diff
@@ -493,27 +497,46 @@ public StatusResponse syncContractsByOrgId(String contractOrgSync) {
...
-
-    var orphanedContracts = contractRepository.findContracts(spec).toList();
+    var orphanedContracts =
+        contractRepository
+            .findContracts(ContractEntity.orgIdEquals(orgId).and(ContractEntity.activeOn(now)))
+            .filter(
+                contract ->
+                    !isPresentInUpstream(
+                        contract, upstreamBillingProviderIds, upstreamAgreementKeys))
+            .toList();
```
Sgitario: These changes might introduce a memory degradation: previously the method was querying the database with the constraints living in the JPA spec (billingProviderIdNullOrNotIn), so the DB only returned candidates to terminate. Now we load every active contract for the org and filter in memory.

For orgs with a large number of contracts that becomes unnecessary work (and memory) on every sync. Can we keep this predicate in the query (extended for subscriptionNumber when needed) instead of moving it to runtime?

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -801,6 +804,115 @@ void shouldPreserveSubscriptionBillingStateWhenContractIsLegitimatelyTerminated(
         "Subscription measurements should be preserved after legitimate termination");
   }
 
+  @TestPlanName("contracts-sync-TC019")
+  @Test
+  void shouldPersistBothConcurrentAwsAgreementsSharingBillingProviderId() {
+    // Given: two AWS agreements, same billingProviderId identity, different sub numbers
+    List<Contract> agreements = givenConcurrentAwsAgreementsAreStubbed();
+    Contract agreementA = agreements.get(0);
+    Contract agreementB = agreements.get(1);
```
Sgitario: do we want to incorporate the business concurrent aws agreement domain concept in swatch-contracts? 
asking because concurrent aws agreements is a very specific aws marketplace thing that is modelled by swatch-contracts as contracts with the same billing provider ID and different subscription numbers.

For example, when reading a test like this, I would need to dig into what givenConcurrentAwsAgreementsAreStubbed does which is really something like:


```suggestion
    String billingProviderId = Random.generateRandom();
    Contract contractA = givenContractWithLicenseFor(orgId, billingProviderId);
    Contract contractB = givenContractWithLicenseFor(orgId, billingProviderId);
```

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -801,6 +804,115 @@ void shouldPreserveSubscriptionBillingStateWhenContractIsLegitimatelyTerminated(
...
+        "Agreement A UUID must be stable across sync");
+    assertEquals(
+        uuidB,
+        getContractUuidBySubscriptionNumber(agreementB),
+        "Agreement B UUID must be stable across sync");
+  }
+
+  @TestPlanName("contracts-sync-TC020")
+  @Test
+  void shouldUpdateOnlyOneConcurrentAwsAgreementOnResync() {
```
Sgitario: the test is a bit hard to follow — a lot of setup/assert logic with some duplications lives inline. Could we slim it down and reuse more of the helpers already added for TC019?

Sgitario (changes requested):

swatch-contracts/src/test/java/com/redhat/swatch/contract/service/ContractServiceTest.java
TommasoLencioni: Added, thanks!

swatch-contracts/TEST_PLAN.md
TommasoLencioni: Nice catch, added here for clarity even if covered elsewhere!

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
TommasoLencioni: Yes I share the same feeling, substituted with an AgreementIdentity object, thanks!

swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ContractService.java
TommasoLencioni: Thanks for raising awareness on this. I changed the logic to produce a predicated with the desired logic.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: I rephrased in each suitable place with concepts already belonging to swatch-contracts., thanks!

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Now the test should be more clear.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -801,6 +804,116 @@ void shouldPreserveSubscriptionBillingStateWhenContractIsLegitimatelyTerminated(
...
 
+  @TestPlanName("contracts-sync-TC019")
+  @Test
+  void shouldPersistBothAwsContractsSharingBillingProviderId() {
+    // Given: two AWS contracts, same billingProviderId, different subscriptionNumbers
+    String billingProviderId = RandomUtils.generateRandom();
+    Contract contractA = givenContractWithLicenseFor(orgId, billingProviderId);
+    Contract contractB =
+        givenContractWithLicenseFor(orgId, billingProviderId).toBuilder()
+            .subscriptionMeasurements(Map.of(CORES, 20.0))
```
Sgitario: why are we modifying the metric? I think we don't use it anywhere, so we could reuse the default one from givenContractWithLicenseFor?

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -801,6 +804,116 @@ void shouldPreserveSubscriptionBillingStateWhenContractIsLegitimatelyTerminated(
...
+
+  @TestPlanName("contracts-sync-TC020")
+  @Test
+  void shouldUpdateOnlyOneAwsContractSharingBillingProviderIdOnResync() {
+    // Given: two AWS contracts sharing billingProviderId, both synced
+    String billingProviderId = RandomUtils.generateRandom();
+    Contract contractA = givenContractWithLicenseFor(orgId, billingProviderId);
+    Contract contractB =
+        givenContractWithLicenseFor(orgId, billingProviderId).toBuilder()
+            .subscriptionMeasurements(Map.of(CORES, 20.0))
```
Sgitario: why are we modifying the metric? I think we don't use it anywhere, so we could reuse the default one from givenContractWithLicenseFor?

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -801,6 +804,116 @@ void shouldPreserveSubscriptionBillingStateWhenContractIsLegitimatelyTerminated(
...
+            .build();
+    stubAwsContractsForOrgSync(contractA, contractB);
+    assertThat(
+        "Initial sync should succeed",
+        service.syncContractsByOrg(orgId).statusCode(),
+        is(HttpStatus.SC_OK));
+    thenAssertContractsPresentBySubscriptionNumber(contractA, contractB);
+
+    String uuidB = getContractUuidBySubscriptionNumber(contractB);
```
Sgitario: the uuidB is already coming in beforeB.uuid, don't think we need this line.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -1000,4 +1113,92 @@ private void givenRosaContractsAreStubbed(int count) {
...
     assertEquals(HttpStatus.SC_OK, sync.statusCode(), "Sync offering should succeed");
   }
+
+  /**
+   * AWS ROSA contract with a Partner Gateway {@code licenseArn}. Passing the same {@code
+   * billingProviderId} seed yields the same product code / AWS customer / seller account (and thus
+   * the same persisted billingProviderId) while each call gets its own subscriptionNumber and
+   * licenseId.
+   */
+  private Contract givenContractWithLicenseFor(String orgId, String billingProviderId) {
```
Sgitario: this method is almost duplicated with Contract.buildRosaContract(...). We can reuse it and setting up the license using `toBuilder()`.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -1000,4 +1113,92 @@ private void givenRosaContractsAreStubbed(int count) {
...
+        .subscriptionId(subscriptionNumber)
+        .subscriptionNumber(subscriptionNumber)
+        .subscriptionMeasurements(Map.of(CORES, 10.0))
+        .licenseId("arn:aws:license-manager:us-east-1:000000000000:license:" + subscriptionNumber)
+        .startDate(OffsetDateTime.now().minusDays(1))
+        .endDate(OffsetDateTime.now().plusDays(1))
+        .build();
+  }
+
+  private void stubAwsContractsForOrgSync(Contract contractA, Contract contractB) {
```
Sgitario: this method is not tied to `aws`, we could reuse it also for azure contracts in other tests like TC001. 
Also, instead of using a fixed number of contracts, use varargs: (Contract ... contracts), so we can reuse it for one, two or more contracts.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -1000,4 +1113,92 @@ private void givenRosaContractsAreStubbed(int count) {
...
+    wiremock.forProductAPI().stubOfferingData(contractA.getOffering());
+    stubPartnerAndSearch(contractA, contractB);
+    assertEquals(
+        HttpStatus.SC_OK,
+        service.syncOffering(contractA.getOffering().getSku()).statusCode(),
+        "Sync offering should succeed");
+  }
+
+  private void stubPartnerAndSearch(Contract... contracts) {
```
Sgitario, line 1110: this method can be reused by other tests in this class.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
```diff
@@ -1000,4 +1113,92 @@ private void givenRosaContractsAreStubbed(int count) {
...
+    return service.getContractsByOrgId(orgId).stream()
+        .filter(c -> expected.getSubscriptionNumber().equals(c.getSubscriptionNumber()))
+        .findFirst()
+        .orElseThrow(
+            () ->
+                new AssertionError(
+                    "No contract for subscriptionNumber=" + expected.getSubscriptionNumber()));
+  }
+
+  private String getContractUuidBySubscriptionNumber(Contract expected) {
```
Sgitario: don't think this method is necessary ... why not using directly `getPersistedContract` and then contract.uuid?

swatch-contracts/src/main/java/com/redhat/swatch/contract/repository/ContractEntity.java
```diff
@@ -244,15 +252,46 @@ public static Specification<ContractEntity> billingProviderIdEquals(String billi
...
   }
 
-  public static Specification<ContractEntity> billingProviderIdNullOrNotIn(
-      Set<String> billingProviderIds) {
+  /**
+   * Candidates to terminate after Partner Gateway sync: not an exact match for any {@code
+   * upstreamAgreements} row ({@code billingProviderId} + {@code subscriptionNumber}, null-safe).
+   * Empty collection matches all contracts.
+   */
+  public static Specification<ContractEntity> orphanedByUpstreamAgreements(
```
Sgitario: I'm very confused about the naming. From component tests, you got rid of the agreement business logic abstraction, but you incorporated it to the implementation. 
From my point of view, the implemenation should be closed to the data model (so, the more we use the naming from the data model, the better to understand the logic).

Sgitario (changes requested):

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: The difference in quantity was emphasizing the contract difference. But since it's not used, as you suggested, I reverted to the default one. Thanks!

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: It made the setup less verbose. I followed the suggestion of `buildRosaContract` + `toBuilder()`.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Right, got rid of boilerplate code, thanks!

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Thanks for the suggestion, implemented.

swatch-contracts/ct/java/tests/ContractsSyncComponentTest.java
TommasoLencioni: Agreed, I missed that in the generated code review, thanks for pointing that out!

swatch-contracts/src/main/java/com/redhat/swatch/contract/repository/ContractEntity.java
TommasoLencioni: Right, i didn't meant to create a new term for data model. Now the implementation refers to contract, so does the CT. Thanks for raising the awareness on that!

Sgitario (approved):
Many thanks for addressing all the comments. Great work!

Mantzounis (approved):
lgtm!

PR #6450, opened 2026-08-06
SWATCH-5297: Add optional licenseId to getAwsUsageContext

Sgitario:
Jira issue: SWATCH-5297

## Description
- Add optional `licenseId` query param to `getAwsUsageContext` and filter subscriptions via matching contract `license_id`.
- Echo request `licenseId` on `AwsUsageContext` (null when omitted).
- Update contracts client callers (`swatch-producer-aws` passes null until follow-up).

## Performance

Optional `licenseId` adds an `EXISTS` against `contracts` only when the query param is present; omitting it leaves the previous plan unchanged (current `producer-aws` still passes null).

On stage/prod we ran `EXPLAIN (ANALYZE, BUFFERS)` for the `getAwsUsageContext` lookup shape (org + AWS billing account + product tag + most-recent subscription). With `licenseId`, Postgres correlates
via `contracts_logicalk` (`subscription_number`, `start_date`) — same keys as `SubscriptionEntity.forContract` — then filters `license_id`. Warm-cache overhead on prod was ~1 ms (sub-millisecond vs the baseline
lookup). `contracts` is small (~3k stage / ~19k prod), so no new `license_id` index is needed at current scale.

## Testing
- Unit: `ContractsResourceTest`, `SubscriptionRepositoryTest` (licenseId filter / null license)
- Unit: `AwsBillableUsageAggregateConsumerTest` (7-arg client)
- Component: `AwsUsageContextComponentTest` (TC005–TC007)

## Summary by CodeRabbit

- **New Features**
  - AWS usage-context lookups now support an optional license ID.
  - Results can include the matched AWS license ARN.
  - Contracts are selected accurately when multiple contracts share billing details but have different license IDs.
  - Unknown or mismatched license IDs return no matching contract.

- **Documentation**
  - Updated API documentation to describe the optional license ID parameter and response field.

- **Tests**
  - Added coverage for license-based matching, missing licenses, and multiple contracts sharing billing identifiers.

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/repository/SubscriptionEntity.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/repository/SubscriptionEntity.java
@@ -314,6 +314,28 @@ public class SubscriptionEntity extends ModificationTrackedEntity {
         builder.like(root.get(SubscriptionEntity_.billingAccountId), billingAccountId + "%");
   }
 
+  private static Specification<SubscriptionEntity> hasLicenseId(String licenseId) {
+    return (root, query, builder) -> {
+      var subquery = query.subquery(Integer.class);
+      var contractRoot = subquery.from(ContractEntity.class);
+      subquery
+          .select(builder.literal(1))
+          .where(
+              builder.and(
+                  builder.equal(contractRoot.get(ContractEntity_.licenseId), licenseId),
+                  builder.equal(
+                      contractRoot.get(ContractEntity_.subscriptionNumber),
+                      root.get(SubscriptionEntity_.subscriptionNumber)),
+                  builder.equal(
+                      contractRoot.get(ContractEntity_.startDate),
+                      root.get(SubscriptionEntity_.startDate)),
+                  builder.equal(
+                      contractRoot.get(ContractEntity_.orgId),
+                      root.get(SubscriptionEntity_.orgId))));
+      return builder.exists(subquery);
+    };
+  }
+
   private static Specification<SubscriptionEntity> metricsCriteria(
       HypervisorReportCategory hypervisorReportCategory, String metricId) {
     return (root, query, builder) -> {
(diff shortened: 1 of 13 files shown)


swatch-contracts/src/main/java/com/redhat/swatch/contract/resource/ContractsResource.java
```diff
@@ -278,7 +290,8 @@ private AwsUsageContext buildAwsUsageContext(SubscriptionEntity subscription) {
         .productCode(productCode)
         .customerId(customerId)
         .awsSellerAccountId(sellerAccount)
-        .customerAwsAccountId(billingAccountId);
+        .customerAwsAccountId(billingAccountId)
+        .licenseId(licenseId);
```
TommasoLencioni, line 294: Response `licenseId` is the one in the request param, so omitting it always returns null even when the matched contract has a `licenseId`. Is this the intended behavior?

Sgitario: yes, because the returned response (SubscriptionEntity) is not a ContractEntity (where the licenseId lives).

To give more context:
- if licenseId is null, we don't check the contracts by licenseId, and we return null: **This is intended because the matching license id logic is in swatch-billable-usage, not here.**
- if licenseId is not null:
- a. if the contract does not exist, we do not return any subscription as expected.
- b. if the contract does exist, we return the subscription with the licenseId from the param (because we already know the contract existed).

TommasoLencioni: Right, thanks!

swatch-contracts/src/main/java/com/redhat/swatch/contract/repository/SubscriptionEntity.java
```diff
@@ -314,6 +314,28 @@ private static Specification<SubscriptionEntity> billingAccountIdLike(String bil
         builder.like(root.get(SubscriptionEntity_.billingAccountId), billingAccountId + "%");
   }
 
+  private static Specification<SubscriptionEntity> hasLicenseId(String licenseId) {
```
TommasoLencioni, line 317: `hasLicenseId` sounds like a column check, but it actually runs an `exists`. Could the method be renamed to reflect this behavior? Such as `existsContractWithLicenseId`.

Sgitario: what we're checking here is there is a contract that has a licenseId, *how* we do it, it's an implementation detail. Don't think method names should reflect this. What if tomorrow, we moved the licenseId to the subscriptions table? We would need to rename the method accordingly as well.

TommasoLencioni (approved):

PR #6481, opened 2026-08-12
SWATCH-5246: Port remaining IQE hourly tally component tests

wottop:
Jira issue: SWATCH-5246

## Linked IQE MR
https://gitlab.cee.redhat.com/insights-qe/iqe-rhsm-subscriptions-plugin/-/merge_requests/1573

## Description

The tests listed in the card were ported from IQE. After analysis, the following updates were made to ensure test determination:

### TallyHandlingConflictsTest

  TC005 — Backfill timestamps fixed (testCounterBackfillAmendsHourlyTotals): The second backfill event was placed at ending.minusHours(1) (beginning+6h), which fell outside the 5-event range (beginning through beginning+4h), creating a new bucket instead of amending an existing one. Changed both backfill timestamps to beginning.plusHours(1) and beginning.plusHours(3) so both genuinely amend existing hourly buckets.

### TallyInstancesReportFiltersPaygTest

  TC013 — Weak OR assertion replaced with exact ID-keyed checks (testAnsibleClusterProducesTwoInstanceRows): The original loop asserted each row had Managed-nodes=1.0 OR Instance-hours=1.0, which couldn't distinguish which instance carried which metric. Replaced with a map keyed by instance ID, then asserted exactly the two expected IDs are present (clusterUuid + "/control-host" and clusterUuid + "/managed-host") and that each has its specific metric value.

### TallyPersistenceTest

  TC003 — Missing initial tally added (testPreviousMonthBackfillLeavesCurrentMonthUnchanged): The setupTest() method creates events via Kafka, but TC003 reads baseline sums without tallying first — unlike TC001 and TC002, which both call performHourlyTallyForOrg before their baselines. Added that call before the baseline loop so the setupTest events are processed before baselines are recorded, preventing the current-month-unchanged assertion from failing when the tally processes all events at once.

### TallyReportFiltersPaygTest

  Yesterday ending bound tightened (shouldMatchDailyTotalToSumOfHourlyValues): Changed yesterdayEnding from plusDays(1) (midnight today) to plusDays(1).minusNanos(1) (23:59:59.999999999 yesterday), preventing the query window from potentially including today's tally bucket. This aligns with the pattern used in TallyPersistenceTest.

  Empty-report false-pass eliminated (same test): Both hourlySum and dailyTotal previously fell back to 0.0 on null/empty data, allowing assertEquals(0.0, 0.0) to pass silently. Added assertNotNull/assertFalse(isEmpty) guards on both reports and assertTrue(hourlySum > 0), so empty reports now fail fast with descriptive messages instead of producing a vacuous pass.

### Verification

The successful execution of each ported test

## Tests

- Added coverage confirming unknown metrics are excluded from instance measurements.
- Added validation for hourly tally backfills, including OpenShift Dedicated and Ansible events.
- Added PAYG reporting coverage for Ansible clusters and multiple instance rows.
- Added checks ensuring daily totals match eligible hourly data.
- Added coverage for preserving monthly tally boundaries during backfills.
- Added support for testing Ansible managed-node metrics and reporting.

### DIFF
--- a/swatch-tally/ct/java/tests/TallyHandlingConflictsTest.java
+++ b/swatch-tally/ct/java/tests/TallyHandlingConflictsTest.java
@@ -22,6 +22,8 @@ package tests;
 
 import static com.redhat.swatch.component.tests.utils.Topics.SWATCH_SERVICE_INSTANCE_INGRESS;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static utils.TallyTestProducts.ANSIBLE_AAP_MANAGED;
+import static utils.TallyTestProducts.OPENSHIFT_DEDICATED;
 import static utils.TallyTestProducts.RHACM;
 import static utils.TallyTestProducts.RHEL_FOR_X86_ELS_PAYG;
 import static utils.TallyTestProducts.ROSA;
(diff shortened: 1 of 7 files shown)


swatch-tally/ct/java/tests/TallyReportTests/TallyReportFiltersPaygTest.java
```diff
@@ -765,4 +767,123 @@ void shouldReturnBadRequestForInvalidGranularity() {
...
+              Event.Sla.PREMIUM,
+              Event.Usage.PRODUCTION,
+              Event.BillingProvider.AWS,
+              billingAccountId,
+              Event.HardwareType.CLOUD,
+              rosaProductId,
+              rosaProductTag);
+      kafkaBridge.produceKafkaMessage(
+          com.redhat.swatch.component.tests.utils.Topics.SWATCH_SERVICE_INSTANCE_INGRESS, event);
+    }
```
liwalker-rh: There is a BeforeAll in this file to set up all the payg events for the tests and hourly sync once, so you should be able to define this rosa event there instead. You also then do not need to produce the kafka message for this event because instead the event will be added in BeforeAll along with one hourly sync. The actual test should then just be able to query what is needed from the tally report and assert.

swatch-tally/ct/java/tests/TallyHandlingConflictsTest.java
```diff
@@ -223,6 +318,54 @@ private void createEventForProduct(
...
+            metricId,
+            value,
+            Event.Sla.PREMIUM,
+            Event.Usage.PRODUCTION,
+            Event.BillingProvider.AWS,
+            billingAccountId,
+            Event.HardwareType.CLOUD,
+            product.productId(),
+            product.productTag());
+    event.setServiceType("OpenShift Cluster");
```
liwalker-rh: Since you added serviceType support to TallyTestProducts, you should be able to do something like `event.setServiceType(product.serviceType());` instead.

swatch-tally/ct/java/tests/TallyHandlingConflictsTest.java
```diff
@@ -223,6 +318,54 @@ private void createEventForProduct(
...
+            metricId,
+            value,
+            Event.Sla.PREMIUM,
+            Event.Usage.PRODUCTION,
+            Event.BillingProvider.AWS,
+            billingAccountId,
+            Event.HardwareType.CLOUD,
+            ANSIBLE_AAP_MANAGED.productId(),
+            ANSIBLE_AAP_MANAGED.productTag());
+    event.setServiceType("Ansible Managed Node");
```
liwalker-rh: Same as above, could be `event.setServiceType(ANSIBLE_AAP_MANAGED.serviceType());` instead.

swatch-tally/ct/java/tests/TallyInstancesReportFiltersPaygTest.java
```diff
@@ -864,6 +869,107 @@ public void shouldReturnMeteredInstancesForFullCalendarMonthWindow() {
...
+            Event.BillingProvider.AWS,
+            billingAccountId,
+            Event.HardwareType.CLOUD,
+            ANSIBLE_AAP_MANAGED.productId(),
+            ANSIBLE_AAP_MANAGED.productTag());
+    managedNodeEvent.setServiceType("Ansible Managed Node");
+    managedNodeEvent.setRole(null);
+
+    kafkaBridge.produceKafkaMessage(SWATCH_SERVICE_INSTANCE_INGRESS, controlNodeEvent);
+    kafkaBridge.produceKafkaMessage(SWATCH_SERVICE_INSTANCE_INGRESS, managedNodeEvent);
```
liwalker-rh: Same thing here that I mentioned with the tally report where there is a BeforeAll that creates and tallies all the events, so you can figure out moving these events to there.

swatch-tally/ct/java/tests/TallyInstancesReportFiltersPaygTest.java
```diff
@@ -864,6 +869,107 @@ public void shouldReturnMeteredInstancesForFullCalendarMonthWindow() {
...
+          assertEquals(
+              1.0,
+              byId.get(controlId).get("Managed-nodes"),
+              0.0001,
+              "Control host should have Managed-nodes=1.0");
+          assertEquals(
+              1.0,
+              byId.get(managedId).get("Instance-hours"),
+              0.0001,
+              "Managed host should have Instance-hours=1.0");
```
liwalker-rh: Feel like there should also be an assertion here that the other metric is absent/equals zero for each row. Meaning that controlId should not have Instance-hours populated and managedId should not have Managed-nodes populated. Example is something like `assertEquals(0.0, byId.get(controlId).get("Instance-hours"), 0.0001);` and same for the other row.

liwalker-rh (changes requested):
@wottop Couple suggestions that should help the tests currently failing to pass (both related to new RBAC v2 changes):

- TallyReportFiltersPaygTest.shouldMatchDailyTotalToSumOfHourlyValues: Add `stubRbacAccessForOrg(rosaOrgId);` to BeforeEach function as well, should allow access for the rosa org.
- TallyInstancesReportFiltersPaygTest.testAnsibleClusterProducesTwoInstanceRows: Move the test into the nested FilterTests class that all the other tests are in. Also do the same thing as the tally report test in BeforeEach but with `stubRbacAccessForOrg(ansibleOrgId);`.

liwalker-rh:
@wottop it could be a timing issue so maybe have the  setupAnsibleEvents call happen earlier in the BeforeAll method like before the creation of the other rhel events? If that does not work I would say just have the test go back to how it was before outside the nested class and no data created earlier on just so we can get the test running correctly and passing.

liwalker-rh (approved):
LGTM

PR #6484, opened 2026-08-13
SWATCH-5077: Create the IT Product Service Kafka Consumer

Sgitario:
Jira issue: SWATCH-5077

## Description
This PR creates a new Kafka consumer for the IT Product (offerings) Service, consuming from the IT Managed Kafka instance. Topics:
- stage: stage.product-service.operationalproduct.protected
- prod: product-service.operationalproduct.protected

## Testing
Unit tests included

## Summary by CodeRabbit

* **New Features**
  * Added Kafka-based product offering synchronization through the IT offering event stream.
  * Added support for product-service event topics across deployment environments.
  * Enhanced product consumption logs with product category and child SKU details.
  * Added handling for malformed or empty messages without disrupting processing.

* **Tests**
  * Added coverage for Kafka message processing, deserialization, logging, validation, and synchronization behavior.

### DIFF
--- a/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ProductStatusKafkaMessageConsumer.java
+++ b/swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ProductStatusKafkaMessageConsumer.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Red Hat trademarks are not licensed under GPLv3. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package com.redhat.swatch.contract.service;
+
+import static com.redhat.swatch.contract.config.Channels.IT_OFFERING_SYNC;
+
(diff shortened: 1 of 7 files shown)


swatch-contracts/src/main/java/com/redhat/swatch/contract/service/ProductStatusKafkaMessageConsumer.java
```diff
@@ -0,0 +1,71 @@
...
+      throw e;
+    }
+
+    log.info(
+        "IT Product message consumed: source=kafka, productCode={}, eventType={}, "
+            + "productCategory={}, childSku={}",
+        productEvent.getProductCode(),
+        productEvent.getEventType(),
+        productEvent.getProductCategory(),
+        isChildSku(productEvent.getProductCode()));
```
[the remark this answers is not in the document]
Sgitario: This is intentional. We'll add this logic in a later task.

swatch-contracts/src/main/resources/application.properties
```diff
@@ -212,6 +212,9 @@ IT_PARTNER_KAFKA_TOPIC=partner-integration.entitlement-gateway.partner-entitleme
 %stage.IT_PARTNER_KAFKA_TOPIC=stage.partner-integration.entitlement-gateway.partner-entitlement.protected
 %prod.IT_PARTNER_KAFKA_TOPIC=partner-integration.entitlement-gateway.partner-entitlement.protected
 
+IT_PRODUCT_KAFKA_TOPIC=product-service.operationalproduct.protected
+%stage.IT_PRODUCT_KAFKA_TOPIC=stage.product-service.operationalproduct.protected
+
```
Mantzounis, line 216: we should add the prod option here to match the pattern of the other 2 consumers. Even though the default works right now.

Mantzounis (changes requested):
code looks good to me, I left 1 comment to add the prod option in application properties.

swatch-contracts/src/main/resources/application.properties
Sgitario: I removed the prod option intentionally since the previous pattern made the configuration harder to follow:  (1) we use the prod config by default, (2) then, we replace the prod config by the same duplicated config. 

Instead, we should define the prod kafka topic only once, and overwrite it when needed (for stage).

Mantzounis (approved):
lgtm!

TommasoLencioni (approved):

PR #6512, opened 2026-08-19
Fix Kessel url for tally and contracts

TrayvonMcKnight:
Jira issue: SWATCH-XXXX

Testing RBAC Kessel in IQE

## Testing

IQE Test MR: 

```
If you want to override the default IQE test image tag (rhsm-subscriptions), add /use-iqe-image-tag=XXX anywhere in this PR description, replacing XXX with your tag. If you leave XXX unchanged, CI keeps the default tag.
```

### Setup

1.
1.

### Steps

1.
1.

### Verification

1.
1.

## Summary by CodeRabbit

* **Enhancements**
  * Endpoint configuration now supports resolving hostnames in addition to URLs.
  * Kessel endpoint settings accept complete inventory API URLs.
  * Endpoint handling works consistently with or without an HTTP or HTTPS scheme.

* **Bug Fixes**
  * Improved default endpoint behavior for local development environments.
  * Kessel connections now use the resolved endpoint and configured gRPC port.

### DIFF
--- a/swatch-common-security/src/main/java/com/redhat/swatch/common/security/KesselProperties.java
+++ b/swatch-common-security/src/main/java/com/redhat/swatch/common/security/KesselProperties.java
@@ -22,10 +22,13 @@ package com.redhat.swatch.common.security;
 
 import io.smallrye.config.ConfigMapping;
 import io.smallrye.config.WithDefault;
+import java.net.URI;
 import java.util.Optional;
 
 @ConfigMapping(prefix = "swatch.kessel")
 public interface KesselProperties {
+
+  int GRPC_PORT = 9000;
 
   @WithDefault("localhost:9000")
   String endpoint();
@@ -44,4 +47,16 @@ public interface KesselProperties {
   Optional<String> authClientId();
 
   Optional<String> authClientSecret();
+
+  default String resolvedEndpoint() {
+    var ep = endpoint();
+    if (!ep.contains("://")) {
+      return ep;
+    }
+    var host = URI.create(ep).getHost();
+    if (host == null) {
+      throw new IllegalArgumentException("Cannot resolve hostname from Kessel endpoint: " + ep);
+    }
+    return host + ":" + GRPC_PORT;
+  }
 }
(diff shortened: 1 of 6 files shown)


swatch-common-security/src/main/java/com/redhat/swatch/common/security/KesselProperties.java
```diff
@@ -44,4 +47,16 @@ public interface KesselProperties {
...
+    var ep = endpoint();
+    if (!ep.contains("://")) {
+      return ep;
+    }
+    var host = URI.create(ep).getHost();
+    if (host == null) {
+      throw new IllegalArgumentException("Cannot resolve hostname from Kessel endpoint: " + ep);
+    }
+    return host + ":" + GRPC_PORT;
+  }
```
wottop, line 61: Could this result in double ':9000' as it is the default value below?
Please add testing for this method

ericThompson038 (dismissed):
lgtm

swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPropertySource.java
```diff
@@ -114,6 +115,7 @@ public class ClowderJsonPropertySource extends PropertySource<ClowderJson>
   private static final Map<String, EndpointConfigMapper> ENDPOINTS_PROPERTIES =
       Map.of(
           ".url", ClowderJsonPropertySource::endpointToUrl,
+          ".hostname", ClowderJsonPropertySource::endpointToHostname,
```
wottop, line 118: Do we need both url and hostname?

wottop (changes requested):

swatch-core/src/main/java/org/candlepin/subscriptions/clowder/ClowderJsonPropertySource.java
kartikshahc: Unfortunately the hostname is not supported in quarkus. We need to open a feature request with them so this is an alternative.

swatch-common-security/src/main/java/com/redhat/swatch/common/security/KesselProperties.java
kartikshahc: All the tests has to use port otherwise Kessel is not going to work. This is configuration that we need to make Kessel work. I am not sure what test coverage you are looking for?

wottop: A unit test to show that the conditions in this method operate as expected.

kartikshahc: Done.

wottop (approved):

vbusch (approved):
LGTM
Was able to test it with the UI.  And say kessel being logged in the contract service.

PR #6521, opened 2026-08-21
fix(ct): disable Quarkus dev console to fix service readiness detection

Mantzounis:
## Description
Problem:

After upgrading to Quarkus 3.38, local component tests on Linux with Java 22+ fail with a 5-minute timeout waiting for service readiness. The test framework detects readiness by checking the out.log file for "Installed features", but the Quarkus application output no longer appears in that file, only Maven build output does.

Fix:

Set quarkus.console.enabled=false on the quarkus:dev command. This disables the interactive dev console which is unused in automated tests. With it disabled, application output goes to stdout as expected and the readiness check succeeds.

## Testing
component tests run and pass. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled the Quarkus console during managed test resource startup to prevent unwanted console output and improve test execution.


vbusch (approved):
Thank you.  Fixes the service startup nicely.

PR #6522, opened 2026-08-21
Refactoring the HBI Seeder erthomps/SWATCH 5437

ericThompson038:
Jira issue: SWATCH-5437

## Description
This is an improvement and redesign of the HBI host data seeding process. These changes will allow a clean test creation by providing specific classes for three host types: RhsmHost, SatelliteHost and QpcHost.  It also decouples the HbiDbconnector from the Host life cycle manager (renamed to the HostStateManager ) so we can easily replace it with a kafka based seeding in the future. The core parts that could be leveraged by other services have all been moved from the tally services to the testing framework. The HostBuilder class can now be used to set up host using the builder pattern along side preset templates ( an example can be found in testCanInsertMultipleHosts ). 

### Notes
- A single Host — there's only one kind of HBI host; what varies is which reporters contributed facts to it.
- Typed, discoverable fact classes — RhsmFacts, SatelliteFacts, QpcFacts, SystemProfileFacts each expose only the fields real HBI normalization actually reads, via a builder (RhsmFacts.builder().defaultFacts().products(...).build()) instead of raw Map<String,Object> puts.
- HostBuilder.update() alongside insert(), so a test can seed a host, tally, mutate a fact, and tally again against the same row.
- HostTemplates for common host shapes (physicalRhel(sockets, cores), awsRhel(sockets, cores)), applied via HostBuilder.apply(...) — kept separate from HostBuilder's own orthogonal setters so a template can't be silently chained with a conflicting one.

### Build and insert a host
```
HostStateManager hostManager = new HostStateManager(new HbiDbConnector(hbiDatabase));

SeededHost seeded = hostManager.createHost(orgId)
    .displayName("My RHEL Host")
    .rhsmFacts(RhsmFacts.builder()
        .defaultFacts()
        .products(List.of("69"))
        .build())
    .systemProfileFacts(SystemProfileFacts.builder()
        .infrastructureType("physical")
        .arch("x86_64")
        .numberOfSockets(2)
        .numberOfCpus(4)
        .build())
    .insert();
```
```
hostManager.createHost(orgId)
    .satelliteFacts(SatelliteFacts.builder().sla("Standard").build())
    .rhsmFacts(RhsmFacts.builder().defaultFacts().sla("Premium").build())
    .insert();
```

### Create a host from a template:
Templates are just `Function<HostBuilder, HostBuilder>` factories in HostTemplates.java and
are used to define a specific reusable shape of a host.

```
SeededHost physical = hostManager.createHost(orgId)
    .apply(HostTemplates.physicalRhel(2, 4)) // cores, sockets
    .insert();

SeededHost cloud = hostManager.createHost(orgId)
    .apply(HostTemplates.awsRhel(1, 1)) // cores, sockets
    .insert();
```

### Adding a new template
```
public static Function<HostBuilder, HostBuilder> satelliteRhel(int sockets, int cores) {
  return builder ->
      builder
          .satelliteFacts(SatelliteFacts.builder().defaultFacts().build())
          .systemProfileFacts(
              SystemProfileFacts.builder()
                  .infrastructureType("physical")
                  .arch("x86_64")
                  .numberOfSockets(sockets)
                  .numberOfCpus(cores)
                  .build());
}
```  

### Update a previously-seeded host

Keep the HostBuilder reference and call update() instead of insert(). Fact groups you don't re-set are preserved as-is, since they're still held on the same builder's Host.

```
HostBuilder hostBuilder = hostManager.createHost(orgId)
    .rhsmFacts(RhsmFacts.builder().defaultFacts().build())
    .systemProfileFacts(SystemProfileFacts.builder().numberOfSockets(2).numberOfCpus(2).build());

SeededHost seeded = hostBuilder.insert();
service.tallyOrg(orgId);
// ...assert initial tally...

SeededHost updated = hostBuilder
    .systemProfileFacts(SystemProfileFacts.builder().numberOfSockets(4).numberOfCpus(4).build())
    .update();
service.tallyOrg(orgId);
// ...assert tally reflects the update, not a second host...
```
  
## Testing 
 Run the new tests:                                                                                                     
  ```bash                                                                                                                
  ./mvnw test -Pcomponent-tests -pl swatch-tally/ct -Dtest=TallyNightlyTest                                              
  ./mvnw test -Pcomponent-tests -pl swatch-tally/ct -Dtest=TallyNightlyHbiTest 
  ```

## Summary by CodeRabbit

- **Tests**
  - Expanded nightly tally test coverage for host creation, cleanup, socket totals, API results, and SLA-based reporting.
  - Added coverage for multiple host configurations, reporter precedence, and fact-handling scenarios.
  - Improved test reliability by using consistent host lifecycle management and validation.

- **Chores**
  - Added reusable test utilities for modeling, creating, updating, tracking, and cleaning up HBI hosts.
  - Added templates for common physical and cloud-based RHEL host configurations.
  - Clarified existing test-support documentation.


swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/Host.java
```diff
@@ -0,0 +1,416 @@
...
+    this.virtualHostUuid =
+        source.getVirtualHostUuid(); // Hypervisor UUID (from system_profiles_static)
+    this.hostType = source.getHostType(); // e.g., "edge" (filtered by query line 153)
+    this.account = source.getAccount(); // Deprecated, but included for completeness
+
+    // ===== Facts (stored as JSONB in HBI) =====
+    this.rhsmFacts = source.getRhsmFacts();
+    this.qpcFacts = source.getQpcFacts();
+    this.satelliteFacts = source.getSatelliteFacts();
+
```
wottop: These should be new maps. Cuts down on alterations in unexpected places
i.e. this.rhsmFacts = new HashMap<>(source.getRhsmFacts());

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/Host.java
```diff
@@ -0,0 +1,416 @@
...
+    // ===== Reporter Info =====
+    // This made need to be appended to the existing reporters array, not replaced.
+    this.reporter = source.getReporter();
+    this.reporters = source.getReporters();
+
```
wottop: Same here with cloning the array
source.getReporters().clone()

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/Host.java
```diff
@@ -0,0 +1,416 @@
...
+    this.billingModel = source.getBillingModel();
+  }
+
+  // ===== Core Identity Setters =====
+  // Do not use this setter unless you have to
+  public Host inventoryId(UUID inventoryId) {
+    this.inventoryId = inventoryId;
+    return this;
+  }
+
```
wottop, line 88: insertHost uses gen_random_uuid() for the id so anything put here is thrown away.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HostStateManager.java
```diff
@@ -0,0 +1,303 @@
...
+ * </pre>
+ */
+public class HostStateManager {
+
+  private final HostConnector connector;
+  private final List<UUID> trackedHostIds = new ArrayList<>();
+
+  // RHEL product ID for templates
+  private static final String RHEL_PRODUCT_ID = "69";
+
```
wottop: Never used. It comes from HostBuilder in practice

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HbiDbConnector.java
```diff
@@ -0,0 +1,334 @@
...
+
+      // Calculate cores_per_socket if not explicitly set
+      Integer coresPerSocket = host.getCoresPerSocket();
+      if (coresPerSocket == null
+          && host.getCores() != null
+          && host.getSockets() != null
+          && host.getSockets() > 0) {
+        coresPerSocket = host.getCores() / host.getSockets();
+      }
+
```
wottop, line 320: This calculation is in 2 places. Is that needed?

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/SatelliteHost.java
```diff
@@ -0,0 +1,108 @@
...
+  //    // Sets this hosts values to the host passed in
+  public SatelliteHost(Host host) {
+    super(host);
+    if (!(host instanceof SatelliteHost)) {
+      applySatelliteDefaults();
+    }
+  }
+
+  // Neet to convert this over to the constructor
+  /**
```
wottop: Best to have comments like this prepended with "TODO:". It makes them searchable in the code

wottop (changes requested):

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/SatelliteHost.java
```diff
@@ -0,0 +1,107 @@
...
+   * @param host the host to apply defaults to
+   */
+  private void applySatelliteDefaults() {
+    // Apply reporter defaults (only if not set)
+
+    if (getReporter() == null || "component-test".equals(getReporter())) {
+      reporter("satellite"); // Since you can have multiple reporters, which is set to this value?
+    }
+
+    // Need to determine how we should handle multiple reporters via data in Stage
```
wottop: Same comment about TODO:
Please check your comments across the files for this type of comment

wottop (changes requested):

swatch-tally/ct/java/tests/TallyNightlyTest.java
```diff
@@ -110,19 +114,20 @@ void test_validate_tally_on_physical_rhel_sockets(
...
-            .subscriptionManagerId("subman-" + actualSockets)
+        hostManager
+            .createRhsmHost(orgId)
             .displayName(displayName)
+            .rhsmFact("RH_PROD", List.of("69"))
+            .rhsmFact("ARCHITECTURE", "x86_64")
             .cores(cores)
-            .sockets(actualSockets)
+            .sockets(startingSockets)
             .insert();
```
mstead, line 132: I'd prefer that when setting facts, we didn't use raw key/value pairs. Instead, each type of fact set should define the facts that are applicable and the builder pattern is used to define them.

A HostBuilder should always define a Host and I'm not sure that we need the RhsmHost/QpcHost implementations of these classes since they are really 'reporters' and we are building an HBI host.

Here the HostManager creates a HostBuilder (Host) that allows setting the applicable RHSM facts (rhsm-conduit reporter) using an RhsmFacts builder pattern which allows facts to be defined in a single place and allows the facts to be discoverable via the class.
```java
    SeededHost host = hostManager
            // Initializes a HostBuilder for the new Host.
            .createHost(orgId) 
            // Allow setting common HBI properties for the host 
            .displayName("Multi-Reporter-Host")
            // Defines the RHSM facts that were reported for this HBI host.
            // The setter should also ensure that 'rhsm-conduit' is added as a reporter
            // which is stored via the Host held by the HostBuilder. 
            .rhsmFacts(
                RhsmFacts.builder()
                    // Set the default facts
                    .defaultFacts()
                    // Override any defaults
                    .products(List.of("69"))
                    .architecture("x86_64")
                    // builds and returns the RhsmFacts object. This object could define a toMap() that
                    // encapsulates the key/value usage for a connector.
                    .build()
            )
            // Cores/sockets do not exist at the root of the host and are facts that get reported
            // as part of the system profile (normalized into cores/sockets on ingestion).
            // We could follow suit here and use a builder patter those as well.
            .setSystemProfileFacts(
                 SystemProfileFacts.builder()
                    .coresPerSocket(2)
                    .numberOfSockets(2)
                    .build()
             )
            .insert();
```

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HostBuilder.java
```diff
@@ -0,0 +1,380 @@
...
+
+  /** Apply physical RHEL defaults (infrastructure=physical, IS_VIRTUAL=false, RHEL product). */
+  public HostBuilder physicalRhelDefaults() {
+    host.infrastructureType("physical");
+    host.arch("x86_64");
+    host.rhsmFact("IS_VIRTUAL", "false");
+    host.rhsmFact("RH_PROD", List.of(RHEL_PRODUCT_ID));
+    host.rhsmFact("ARCHITECTURE", "x86_64");
+    return this;
+  }
```
mstead: These templates can be built using the pattern/example above.

swatch-tally/ct/java/tests/TallyNightlyHbiTest.java
```diff
@@ -81,19 +79,175 @@ void testHbiSeederCanInsert() {
...
+            .rhsmFact("Memory", 63)
+            // Set to envoke the nomalization logic that will overwrite the normalize from satellite
+            // facts with the
+            .rhsmFact("SYSPURPOSE_SLA", "Premium")
+            .rhsmFact("SYSPURPOSE_ROLE", "Red Hat Enterprise Linux Server")
+            .rhsmFact("SYSPURPOSE_USAGE", "Component Testing")
+            .sockets(1)
+            .cores(8)
+            .arch("x86_64")
+            .insert();
```
mstead, line 189: Here is another example following the pattern shown in TallyNightlyTest.java. Whenever we set facts, we add the appropriate reporter.

```java
eededHost host2 = hostManager
            .createHost(orgId) // HostBuilder
            .displayname("Multi-Reporter-Host")
            // Add satellite as a reporter as part of this function.
            .satelliteFacts(
                    SatelliteFacts.builder()
                            .defaultFacts()
                            .version("6.17.6.2")
                            .virtualHostName("virt-name-for-testing.com")
                            .virtualHostUUID(UUID.randomUUID().toString())
                            .systemPurposeSla("Standard")
                            .systemPurposeUsage("Component Testing") // Must map to one of our Usage classes like "Production"
                            .systemPurposeRole("Red Hat Enterprise Linux Server")
                            .build())
            .rhsmFacts(
                    RhsmFacts.builder()
                            .syncTimestamp(OffsetDateTime.now())
                            .virtual(false)
                            .products(List.of("69"))
                            .architecture("x86_64")
                            .memory(63)
                            .systemPurposeSla("Premium")
                            .systemPurposeRole("Red Hat Enterprise Linux Server")
                            .systemPurposeUsage("Component Testing")
                            .build())
            .qpcFacts(QpcFacts.builder().defaultFacts().build())
            .systemProfileFacts(
                    SystemProfileFacts.builder()
                            .infrastructureType("physical")
                            .cloudProvider("aws")
                            .arch("x86_64")
                            .providerId(UUID.randomUUID().toString())
                            .build()
            )
            .insert();
```

mstead (dismissed):
This is definitely moving in the right direction. I added a couple of comments with some concerns/suggestions that I had discussed with you already, but I wanted it to be more visible to you and others.

**My primary concerns:**
1) Using raw key/value pairs to allow tests to set arbitrary facts; these should be encapsulated and discoverable when writing tests.
2) Top level Host properties exist that are actually system profile facts should be set as such.

My suggestions should be vetted by others as well to make sure that we agree given the goal of this is to be able to write tests quickly and easily.

swatch-tally/ct/java/tests/TallyNightlyHbiTest.java
kartikshahc: Agreed good suggestion

swatch-tally/ct/java/tests/TallyNightlyTest.java
kartikshahc: Agreed good suggestion

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/Host.java
mstead: Addressed. Copy constructor has been removed since it is no longer needed.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/Host.java
mstead: Addressed. Copy constructor has been removed since it is no longer needed.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/Host.java
mstead: This field was replaced with `id`. The setter is still required to support the `update()` case.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HostStateManager.java
mstead: Removed.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HbiDbConnector.java
mstead: I would like to chat with @ericThompson038 about this calculation in general. I'm unclear if the intent here is simply a test helper, but in general the connector shouldn't be doing this. Facts should be set raw and helpers should be at a different level.

We can address this later IMO.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/SatelliteHost.java
mstead: No longer relevant.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/SatelliteHost.java
mstead: No longer relevant.

liwalker-rh (dismissed):
Changes look good and tests work great. Up to you if you want to address the remaining coderabbit comments or not. 

There is definitely some more cleanup needed to address the old TallyHbiDbSeeder and tests that are still using that, but we can work on that while porting over IQE tests.

mstead:
> There is definitely some more cleanup needed to address the old TallyHbiDbSeeder and tests that are still using that, but we can work on that while porting over IQE tests.

I didn't realize that the old seeder wasn't removed. Based on the existing tests, this may be a bit of a lift. Some of them create hosts/buckets directly in the SWatch DB. Will need to properly seed them in HBI and run a tally beforehand. It seems this is a bit of a work-around to get around the nightly tally deleting hosts that no longer exist in the DB (adding paygo bucket so the host stays around).

Let's all chat tomorrow about what we want to do here because having two seeders isn't ideal, though, we can't hold off getting this in IMO.

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HostTemplates.java
```diff
@@ -0,0 +1,72 @@
...
+ *     .displayName("Custom")
+ *     .insert();
+ * </pre>
+ */
+public final class HostTemplates {
+
+  private HostTemplates() {}
+
+  /** Physical RHEL host with the given socket/core counts. */
+  public static Function<HostBuilder, HostBuilder> physicalRhel(int sockets, int cores) {
```
kartikshahc: I like the HostTemplate idea. Should we add the reporter in the method name to determine when we write test?

swatch-tally/ct/java/tests/TallyNightlyHbiTest.java
```diff
@@ -174,40 +373,225 @@ void testNightlyTallyRhelProduct() {
...
+                "granularity",
+                "Daily",
+                "beginning",
+                beginning.toString(),
+                "ending",
+                ending.toString(),
+                "sla",
+                "Premium"));
+
+    var standardData =
```
kartikshahc: These two seems to be separate tests. One for premiumData and other for standard because I dont like if else in the test.

kartikshahc (changes requested):

swatch-tally/ct/java/tests/TallyNightlyHbiTest.java
mstead: I agree. Good catch. I missed this when I took the card over. This is definitely a pattern we should stay away from.

I'll fix this up.

swatch-tally/ct/java/utils/TallyHbiDbSeeder.java
```diff
@@ -60,6 +60,19 @@
  */
 public final class TallyHbiDbSeeder {
 
+  // separate functionality of the Seeding, DB connection, and HBI  into different classes
+  // Need to be specific about what kind of RHEL host we are creating, so we can use the correct one
+  // Need to specify what facts per RHEL host ( factory for the RHEL host )
+  /*example for Kartik
+  *  public HostBuilder rhelHost(String orgId) {
+      return new HostBuilder(orgId, true);
+    }
```
kartikshahc: I think this comment is no longer valid and if we can remove it.

swatch-tally/ct/java/tests/TallyNightlyHbiTest.java
mstead: DONE

swatch-tally/ct/java/utils/TallyHbiDbSeeder.java
mstead: DONE

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HostTemplates.java
mstead: DONE

kartikshahc (approved):

liwalker-rh (approved):

swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/hbi/HbiDbConnector.java
mstead: I'll address this in a follow up PR that removes the old seeder class. This is more of a test calculation and doesn't belong in the DBConnector so I'm going to remove it.

wottop (approved):
lgtm

PR #6565, opened 2026-08-31
SWATCH-5082:Write component tests for the IT Product Service Kafka consumer

diegomaranhao:
Jira issue: SWATCH-5082

## Description

- Adds component tests for the IT Product Service Kafka and UMB consumers: parent/child SKU happy paths, malformed JSON, Unleash flag and variant gating, duplicate-SKU processing on both channels, and edge payloads (missing fields, empty JSON, extra fields).
- Covers product-umb-TC004 (UMB_ENABLED=false) as a unit test. That property is fixed at startup in the shared quarkus:dev component-test process, so it cannot be flipped per test.
- Updates COMPONENT_TEST_PLAN.md to match production behavior: topic product-service.operationalproduct.protected, UMB address VirtualTopic.services.productservice.Product, consume/sync log assertions, and Unleash flag swatch.swatch-contracts.enable-product-service-consumer.

## Testing
**Unit tests**
```
./mvnw test -pl swatch-contracts \
  -Dtest=ProductStatusKafkaMessageConsumerTest,ProductStatusUMBMessageConsumerTest,ProductStatusUMBMessageConsumerWhenUmbDisabledTest \
  -Dsurefire.failIfNoSpecifiedTests=false \
  -Dquarkus.analytics.disabled=true
```

**Component tests**
```
./mvnw test -pl swatch-contracts/ct -Pcomponent-tests \
  -Dtest=ProductKafkaComponentTest,ProductUmbComponentTest,ProductDuplicateComponentTest,ProductEdgeComponentTest \
  -Dsurefire.failIfNoSpecifiedTests=false \
  -Dquarkus.analytics.disabled=true
```

### DIFF
--- a/swatch-contracts/ct/java/tests/BaseProductConsumerComponentTest.java
+++ b/swatch-contracts/ct/java/tests/BaseProductConsumerComponentTest.java
@@ -20,15 +20,38 @@
  */
 package tests;
 
+import static com.redhat.swatch.component.tests.utils.Topics.IT_PRODUCT_SYNC;
+
 import com.redhat.swatch.contract.test.model.OperationalProductEvent;
 import com.redhat.swatch.contract.test.model.OperationalProductEvent.EventTypeEnum;
 import com.redhat.swatch.contract.test.model.OperationalProductEvent.ProductCategoryEnum;
 import java.time.OffsetDateTime;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
 
 public abstract class BaseProductConsumerComponentTest extends BaseContractComponentTest {
 
+  @BeforeAll
+  static void enableProductServiceConsumer() {
+    unleash.enableProductServiceConsumerBothConsumers();
+  }
+
+  private static boolean isChildSku(String productCode) {
+    return productCode != null && productCode.startsWith("SVC");
+  }
+
+  @AfterEach
+  void restoreProductServiceConsumerFlag() {
+    unleash.enableProductServiceConsumerBothConsumers();
+  }
+
   protected OperationalProductEvent givenParentSkuEvent(String productCode) {
-    return givenProductEvent(productCode, ProductCategoryEnum.PARENT_SKU);
+    return givenParentSkuEvent(productCode, EventTypeEnum.UPDATE);
+  }
+
+  protected OperationalProductEvent givenParentSkuEvent(
+      String productCode, EventTypeEnum eventType) {
+    return givenProductEvent(productCode, ProductCategoryEnum.PARENT_SKU, eventType);
   }
 
   protected OperationalProductEvent givenChildSkuEvent(String productCode) {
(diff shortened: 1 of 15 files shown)


swatch-contracts/src/test/java/com/redhat/swatch/contract/service/ProductStatusUMBMessageConsumerWhenUmbDisabledTest.java
```diff
@@ -0,0 +1,62 @@
...
+    LoggerCaptor.registerHandler(ProductStatusUMBMessageConsumer.class);
+  }
+
+  @BeforeEach
+  void setUp() {
+    LoggerCaptor.clearRecords();
+  }
+
+  @Test
+  void shouldIgnoreMessagesWhenUmbIsNotEnabled() {
```
TommasoLencioni, line 56: Can this be marked in some way, referring to "product-umb-TC004"?

swatch-contracts/ct/java/tests/BaseProductConsumerComponentTest.java
```diff
@@ -0,0 +1,120 @@
...
+    event.setEventType(eventType);
+    event.setOccurredOn(OffsetDateTime.now().toString());
+    return event;
+  }
+
+  protected void whenKafkaMessageIsPublished(OperationalProductEvent event) {
+    kafkaBridge.produceKafkaMessage(IT_PRODUCT_SYNC, event);
+  }
+
+  protected void thenMessageIsNotConsumed(OperationalProductEvent event, String source) {
```
TommasoLencioni, line 80: Sync assertion on logs not containing an entry could be flaky.
Do we have any other mean of checking?

swatch-contracts/src/test/java/com/redhat/swatch/contract/service/ProductStatusUMBMessageConsumerWhenUmbDisabledTest.java
diegomaranhao: I would not put @TestPlanName("product-umb-TC004") on that unit test.

@TestPlanName is a component-test annotation. It is meta-annotated with @ExtendWith(ComponentTestExtension.class), so it pulls in the CT lifecycle (compose services, quarkus:dev, etc.).

TommasoLencioni: A simple comment would be fine, just to keep track of that.

swatch-contracts/ct/java/tests/BaseProductConsumerComponentTest.java
diegomaranhao: Good point. assertDoesNotContain is synchronous, so a bare negative check could race. In the disabled-consumer tests, we first wait for the "consumer is disabled by feature flag" log, which means the message was processed and returned before sync. We also use random SKUs to avoid log collisions in the shared dev process. If we want to harden this further, we could add a polling negative helper in LogsVerifier, or assert no WireMock product-tree request for the SKU.

PR #6569, opened 2026-08-31
swatch-5269: Implement RBAC component tests for swatch-contracts

vbusch:
Jira issue: SWATCH-5269

## Description
This card implements the RBAC v1 and v2 component tests for swatch-contracts.

To run the tests against ephemeral without running out of memory and threads.  The following changes were made to reduces thread count, HTTP connections, and memory footprint in Konflux CT runner:
- Stop background log polling that created unbounded OkHttp dispatcher threads
- Changed `OpenShiftLoggingHandler` to fetch logs on-demand instead of continuous watching
- Removed startWatching() - logs now pulled only when tests read them

## Testing
podman compose up -d
./mvnw clean install -Pcomponent-tests -pl swatch-contracts/ct -am

## Summary by CodeRabbit

* **New Features**
  * Added billing account ID retrieval for organizations.
  * Added access to V1 and V2 capacity reports, including metric-specific daily reports.
  * Improved export request processing and callback handling.

* **Security**
  * Restricted billing account lookups to customer and support roles.

* **Testing**
  * Expanded authorization coverage across subscription, capacity, billing, and export workflows.
  * Improved validation across legacy and modern authorization models.

PR #6576, opened 2026-09-02
fix(ct): stop recreating OpenShift port-forwards on sticky websocket errors

Sgitario:
## Description

Konflux `bonfire-component-tests` for **swatch-contracts** hangs in `run-tests` with `OutOfMemoryError: unable to create native thread` (~1.7Gi RSS, not a 4Gi heap OOM). Same failure on main (including Konflux references PRs), not only on later feature branches.

[#6569](https://github.com/RedHatInsights/rhsm-subscriptions/pull/6569) stopped listing pods on every RestAssured call (good) and started recreating the Fabric8 port-forward when `errorOccurred()` is true. That flag is **sticky**: after the first websocket glitch it stays true, and Kafka keep-alive / RestAssured call `port()` on every request. Each recreation leaks a `newSingleThreadExecutor` in `PortForwarderWebsocketListener` until nproc is exhausted (`pool-2800+`).

## Testing
Regression testing only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Port forwarding now recreates tunnels only when the connection process has stopped, improving stability when recoverable process errors occur.

### DIFF
--- a/swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/clients/BaseKubernetesClient.java
+++ b/swatch-test-framework/src/main/java/com/redhat/swatch/component/tests/api/clients/BaseKubernetesClient.java
@@ -274,7 +274,7 @@ public abstract class BaseKubernetesClient<
      * Konflux nproc via Fabric8's unbounded OkHttp dispatcher.
      */
     boolean needsToRecreate() {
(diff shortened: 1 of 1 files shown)


vbusch (approved):
LGTM

Mantzounis (approved):

PR #6592, opened 2026-09-03
SWATCH-5518: Default null SLA/Usage values

wottop:
- Use the defaults for the product if they exist
- Premium/Production instead of EMPTY during tally

Jira issue: SWATCH-5518

## Description

We want to stop using empty string as the placeholder when no value exists for SLA and Usage. The default of Premium and Production are already used for the hourly tallies for payg instances.

### Setup

1. Start the pods locally
2. oc login to the production OpenShift
3. Import data for HBI hosts
```./import_hbi_hosts_by_org.sh --org-id=4592029 --batch-mode```

### Steps

1. Start tally
```make swatch-tally  PROFILES="dev api worker"```
2. Run the nightly tally (you may need to opt in if this is a new org in your db)
```http PUT ":8010/api/rhsm-subscriptions/v1/internal/rpc/tally/snapshots/4592029" x-rh-swatch-psk:placeholder```

### Verification

1. Find a host in hbi.hosts that does not have either a "SYSPURPOSE_USAGE" or "SYSPURPOSE_SLA" in the 'facts' json column. Use one of the most current 'modified_on' timestamp ones. Copy the 'id' value.
2. Use this query to see the result:
```select host_id, product_id, usage, sla from hosts a join host_tally_buckets b on a.id = b.host_id where a.inventory_id = '[hbi.host.id]';```

## Summary by CodeRabbit

- **New Features**
  - Product-specific SLA and usage defaults are now applied when event data does not provide values.
  - Hosts and usage records without explicit SLA or usage information now default to **Premium** and **Production**.
  - Empty SLA or usage values are normalized to Premium/Production instead of creating incomplete classifications.

- **Bug Fixes**
  - Tally reports now correctly reflect explicit Standard and Development/Test settings while preserving Premium/Production defaults when values are missing.

### CODE REVIEW

File: src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorCreateHostUsageKeysTest.java [hunk 1]
```diff
@@ -0,0 +1,320 @@
...
+    normalizedFacts.setProducts(Set.of(RHEL_PRODUCT));
+    normalizedFacts.setSla(ServiceLevel.EMPTY);
+    normalizedFacts.setUsage(Usage.PRODUCTION);
+
+    Set<Key> keys = collector.createHostUsageKeys(products, normalizedFacts);
+
+    assertNotNull(keys);
+    assertFalse(keys.isEmpty());
+    assertTrue(keys.stream().allMatch(k -> RHEL_PRODUCT.equals(k.getProductId())));
+    assertTrue(keys.stream().anyMatch(k -> k.getSla() == ServiceLevel._ANY));
```
liwalker-rh: I am a little confused with some of the assertions in this test class. Since the host has an empty SLA value wouldn't it be more valuable to assert that getSla == ServiceLevel.PREMIUM? And same for the usage test below this and possibly some other tests in this file as well?

swatch-tally/ct/java/tests/TallyNightlyHbiTest.java
```diff
@@ -609,4 +609,265 @@ private void assertHasNoData(
...
+  /**
+   * - **Description**: Verify that hosts without SLA/Usage facts default to Premium/Production -
+   * **Setup**: Component test environment with swatch-tally is running - **Action**: Insert a host
+   * without SLA/Usage facts and run nightly tally - **Verification**: - Tally report with
+   * sla=Premium and usage=Production filters returns the host's data - Tally report with
+   * sla=Standard or usage=Development/Test filters returns no data - **Expected Result**: Host
+   * without SLA/Usage is tallied as Premium/Production
+   */
+  @Test
+  void testNightlyTallyDefaultsToPremiuProductionWhenSlaUsageNotSet() {
```
liwalker-rh: I feel like both of the component tests here are important for SLA/usage default values in tally reports and maybe should be in a different file. I think these both could be good candidates for tests in a `TallyReportFiltersNonPaygTest.java` file or another name involving defaults instead. They also should have their description, setup, etc. added to the `tally-reports-api.md` test plan.

swatch-tally/ct/java/tests/TallyNightlySlaUsageDefaults.java
```diff
@@ -0,0 +1,327 @@
...
+    }
+  }
+
+  /**
+   * - **Description**: Verify that hosts without SLA/Usage facts default to Premium/Production -
+   * **Setup**: Component test environment with swatch-tally is running - **Action**: Insert a host
+   * without SLA/Usage facts and run nightly tally - **Verification**: - Tally report with
+   * sla=Premium and usage=Production filters returns the host's data - Tally report with
+   * sla=Standard or usage=Development/Test filters returns no data - **Expected Result**: Host
+   * without SLA/Usage is tallied as Premium/Production
```
liwalker-rh: You should be able to remove these from both the tests now that they are in the test plan doc. Besides this and possibly addressing other remaining coderabbit comments though, the testing steps work as expected and everything else looks good.

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -351,13 +352,40 @@ private Host createSwatchSystem(
...
+      Set<Usage> usageSet = new HashSet<>();
+      if (!Usage.EMPTY.equals(facts.getUsage())) {
+        usageSet.add(facts.getUsage());
+      } else {
+        // Use product's default if available then global default
+        Usage productDefault =
+            Usage.fromRegistryUsage(SubscriptionDefinition.getProductDefaultUsage(product));
+        usageSet.add(productDefault != null ? productDefault : Usage.PRODUCTION);
+      }
+      usageSet.add(Usage._ANY);
```
mstead: To make `createHostUsageKeys` is a little easier to read, I'd suggest that we create individual methods to get the sla/usage sets for a product.

```
Set<ServiceLevel> slaSet = determineSlaSet(....).
Set<Usage> usageSet = determineUsageSet(....).

usageKeySet.addAll(
    createKeyCombinations(
        Set.of(product), slaSet, usageSet, Set.of(BillingProvider._ANY), Set.of("_ANY")));
```

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -433,16 +433,20 @@ private Set<List<Object>> buildBucketTuples(Event event) {
...
 
+    // Default to product-specific defaults first, then global PREMIUM/PRODUCTION
+    // Treat blank/empty strings (including __EMPTY__) as absent
     ServiceLevel effectiveSla =
         Optional.ofNullable(event.getSla())
             .map(Event.Sla::toString)
+            .filter(s -> !s.isBlank())
             .map(ServiceLevel::fromString)
-            .orElse(sla.map(ServiceLevel::fromString).orElse(ServiceLevel.EMPTY));
+            .orElse(sla.map(ServiceLevel::fromString).orElse(ServiceLevel.PREMIUM));
```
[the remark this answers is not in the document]
mstead: This is a legit concern. Is there a test around this?

src/main/java/org/candlepin/subscriptions/tally/MetricUsageCollector.java
```diff
@@ -437,12 +437,12 @@ private Set<List<Object>> buildBucketTuples(Event event) {
...
             .map(Event.Sla::toString)
             .map(ServiceLevel::fromString)
-            .orElse(sla.map(ServiceLevel::fromString).orElse(ServiceLevel.EMPTY));
+            .orElse(sla.map(ServiceLevel::fromString).orElse(ServiceLevel.PREMIUM));
     Usage effectiveUsage =
         Optional.ofNullable(event.getUsage())
             .map(Event.Usage::toString)
             .map(Usage::fromString)
-            .orElse(usage.map(Usage::fromString).orElse(Usage.EMPTY));
+            .orElse(usage.map(Usage::fromString).orElse(Usage.PRODUCTION));
```
mstead, line 452: This feels like it could suffer from the same issue that code rabbit pointed out about SLA.

src/test/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollectorCreateHostUsageKeysTest.java
```diff
@@ -0,0 +1,318 @@
...
+import org.candlepin.subscriptions.db.model.Usage;
+import org.candlepin.subscriptions.tally.UsageCalculation.Key;
+import org.candlepin.subscriptions.tally.facts.NormalizedFacts;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+
+@SpringBootTest
```
mstead: Because this test doesn't need to be a heavy weight spring boot test since it is only testing usage key generation. I'd use mocks and initialize the collector directly in the overall setup of the test. This way we can cut the startup time caused by `@SpringBootTest` and make it more of a unit test.

src/test/java/org/candlepin/subscriptions/tally/MetricUsageCollectorTest.java
```diff
@@ -1236,6 +1236,60 @@ void testFilterSupportedMetricsByProduct() {
             .getMeasurement(MetricIdUtils.getStorageGibibyteMonths()));
   }
 
+  @Test
+  void testEmptySlaNdUsageDefaultsToPremiumProduction() {
```
mstead: Typo in the test name?

src/main/java/org/candlepin/subscriptions/tally/InventoryAccountUsageCollector.java
```diff
@@ -351,13 +352,40 @@ private Host createSwatchSystem(
...
+    for (String product : applicableProducts) {
+      // Build SLA set: for each product, apply its default if host SLA is EMPTY
+      Set<ServiceLevel> slaSet = new HashSet<>();
+      if (!ServiceLevel.EMPTY.equals(facts.getSla())) {
+        slaSet.add(facts.getSla());
+      } else {
+        // Use product's default if available then global default
+        ServiceLevel productDefault =
+            ServiceLevel.fromRegistrySla(SubscriptionDefinition.getProductDefaultSla(product));
+        slaSet.add(productDefault != null ? productDefault : ServiceLevel.PREMIUM);
```
mstead: I'm not sure I like binding the model SLA to the registry SLA for a single use via the ServiceLevel class. It feels like this logic should remain in the calling class.

**Untested**, but something like this could be used here instead to accomplish the same thing, which would eliminate the static utility method. The same approach can be applied to Usage as well.
```
        slaSet.add(
           Optional.ofNullable(SubscriptionDefinition.getProductDefaultSla(product))
             // If a default is found in the product config, get the model instance
             // based on the config's sla value.
             .map(sla -> ServiceLevel.fromString(sla.getValue()))
             // If not found in config, default to PREMIUM.
             .orElse(ServiceLevel.PREMIUM)
        );
//        ServiceLevel productDefault =
//            ServiceLevel.fromRegistrySla(SubscriptionDefinition.getProductDefaultSla(product));
//        slaSet.add(productDefault != null ? productDefault : ServiceLevel.PREMIUM);

```

mstead (changes requested):
Change requests outlined below. Apart from those, I'd like to see a component test covering the Hourly tally as well (if it doesn't already exist).

mstead (approved):
LGTM

PR #6608, opened 2026-09-07
SWATCH-5332: Added contracts-update-TC010: create contract after offering sync

Sgitario:
Jira issue: SWATCH-5332

## Description
Added new component test to fill the gap to cover the scenario where after an offering is created via message broker, then the new contract can be created using the new offering.

Note that the document linked in the JIRA also mentions the following gaps:
- "Partner-entitlement UMB happy path + update + malformed payload" which is not really a gap since it's covered by component tests (contracts-update-TC0XX)
- "Product UMB happy and negative path in integration scope" which is also not a gap since it's covered by component tests (offering-update-TC0XX)

## Testing
New component test added.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end coverage for creating a contract after an offering is synchronized through an event.
  - Verifies the offering becomes available before contract processing.
  - Confirms the created contract includes the expected organization, SKU, subscription number, and metrics.
- **Documentation**
  - Documented the new contract update test scenario and its expected results.

### DIFF
--- a/swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
+++ b/swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
@@ -23,6 +23,10 @@ package tests;
 import static api.PartnerApiStubs.PartnerSubscriptionsStubRequest.forContract;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import api.ContractsArtemisService;
 import com.redhat.swatch.component.tests.api.Artemis;
@@ -38,7 +42,6 @@ import io.restassured.response.Response;
 import java.time.OffsetDateTime;
 import java.util.Map;
 import org.apache.http.HttpStatus;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 public class ContractsUpdateComponentTest extends BaseContractComponentTest {
@@ -65,9 +68,9 @@ public class ContractsUpdateComponentTest extends BaseContractComponentTest {
     // then: The existing contract should be updated with the new end date
     service.logs().assertContains("Existing contracts and subscriptions updated");
     var contracts = service.getContracts(initialContract);
-    Assertions.assertEquals(1, contracts.size());
-    Assertions.assertNotNull(contracts.get(0).getEndDate());
-    Assertions.assertTrue(contracts.get(0).getEndDate().isAfter(initialContract.getEndDate()));
+    assertEquals(1, contracts.size());
+    assertNotNull(contracts.getFirst().getEndDate());
+    assertTrue(contracts.getFirst().getEndDate().isAfter(initialContract.getEndDate()));
   }
 
   @TestPlanName("contracts-update-TC002")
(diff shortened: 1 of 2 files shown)


swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
```diff
@@ -23,6 +23,8 @@
 import static api.PartnerApiStubs.PartnerSubscriptionsStubRequest.forContract;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
```
Mantzounis, line 28: we should using Assertions.assertEquals instead of static imports to match current patterns, is there a reason we are using these for this PR?

swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
```diff
@@ -333,6 +335,35 @@ void shouldRemoveMetricFromExistingContract() {
...
+    thenOfferingShouldExist(contract.getOffering().getSku());
+
+    // And: Partner entitlement message creates the contract
+    wiremock.forPartnerAPI().stubPartnerSubscriptions(forContract(contract));
+    wiremock.forSearchApi().stubGetSubscriptionBySubscriptionNumber(contract);
+    artemis.forContracts().sendAsText(contract);
+
+    // Then: Contract is created end-to-end
+    AwaitilityUtils.until(() -> service.getContracts(contract).size(), is(1));
+    var actual = service.getContracts(contract).get(0);
```
Mantzounis: nit: we could use the same pattern we use in TC001 here for consistency.
AwaitilityUtils.until(() -> service.getContracts(contract).size(), is(1));
var contracts = service.getContracts(contract);
Assertions.assertEquals(1, contracts.size());
var actual = contracts.get(0);

Mantzounis (changes requested):
I also think the comment coderabbit left as part of the review is worth addressing.

swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
Sgitario: I don't understand this comment, what are the current patterns? I'm already using the static imports from Assertions API.

Mantzounis: That's my bad, in this specific file all other tests are using Assertions.assertEquals instead of just assertEquals (see lines 70-72 as an example) which stuck out to me from a stylistic perspective, looking at other files I see that we almost always use them without the Assertions. part beforehands. 

So could you remove that from the rest of the tests in this file and use the one you used for this PR?

swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
```diff
@@ -333,6 +335,35 @@ void shouldRemoveMetricFromExistingContract() {
...
+    assertEquals(orgId, actual.getOrgId());
+    assertEquals(contract.getSubscriptionNumber(), actual.getSubscriptionNumber());
+    assertEquals(contract.getOffering().getSku(), actual.getSku());
+    assertNotNull(actual.getMetrics());
+    assertEquals(1, actual.getMetrics().size());
```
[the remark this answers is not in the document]
Sgitario: added

swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
Sgitario: changed

swatch-contracts/ct/java/tests/ContractsUpdateComponentTest.java
Sgitario: I see your point. Just changed it to use the static import. Thanks for spotting!

Mantzounis (approved):
lgtm! thank you for the updates.

PR #6615, opened 2026-09-08
SWATCH-5437 Follow up PR to remove old seeder code

mstead:
Jira issue: SWATCH-5437

## Description
Follow up PR for HBI Host Seeder refactor that removes the old seeder code and updates existing tests to use the new version.

### HBI Seeder Change
To properly port old tests, I needed a way to update fact objects
set by the builder.  When modifying the underlying Host *Fact objects,
the fact object must be recreated and the builder's setter object
must overwrite the fact object. To facilitate this, I enabled  lombok's
`toBuilder` function to handle cloning when going from the underlying
facts back to a builder.

To do so, store the HostBuilder in a variable before calling insert().
Then, overwrite immutable fact properties as follows:
```
builder.rhsmFacts(builder.rhsmFacts().toBuilder().sla("Standard"));
```

### Hypervisor Test Plan Update
There were some tests that marked a non-existent entry in the test plan
 so I updated the plan to include them.

## Summary by CodeRabbit

- **Tests**
  - Expanded tally validation for hypervisor socket totals, guest handling, and overlapping SLA scenarios.
  - Improved test host setup, updates, cleanup, and data consistency across tally component tests.
  - Added coverage for multiple hosts with distinct display names and product information.
  - Added reusable support for creating and managing test hosts in database-backed scenarios.

- **Documentation**
  - Updated the tally processing test plan with three new hypervisor test cases.


swatch-test-hbi-support/src/main/java/com/redhat/swatch/component/tests/api/hbi/HostTemplates.java
```diff
@@ -38,17 +38,21 @@
  */
 public final class HostTemplates {
```
Sgitario, line 39: @mstead why the `hbi` package is in the test-framework dependency? this seems to me like it should be in the swatch-tally component tests since it's only used in here (at least, today).

mstead: @Sgitario It will be needed in swatch-metrics-hbi when we start working on event-driven tally (next quarter).

wottop (changes requested):
After dealing with the coderabbit suggestions, lgtm

swatch-test-hbi-support/src/main/java/com/redhat/swatch/component/tests/api/hbi/HostTemplates.java
Sgitario: I see that it's very convenient for you to keep the model classes in the testing framework dependency, however the idea of this module is to have the foundation for component tests infrastructure, no business/model classes only for a few swatch services. 

This is important to keep the isolation of dependencies: if tomorrow, some of these models fail: all the component tests will be broken (it has already happened in the past that the lombok bump dependency broke the YupanaFacts builder). 

Another question I have is why these models are necessary only in component tests? How the services use the same kind of classes? I see the `swatch-model-metrics-hbi` dependency looks like to be very related. The right pattern is that services and component tests use exactly the same classes.

mstead: > I see that it's very convenient for you to keep the model classes in the testing framework dependency, however the idea of this module is to have the foundation for component tests infrastructure, no business/model classes only for a few swatch services.

Fair point. I'll make sure that this gets moved and we can put them in another module that can be shared across component tests 

> This is important to keep the isolation of dependencies: if tomorrow, some of these models fail: all the component tests will be broken (it has already happened in the past that the lombok bump dependency broke the YupanaFacts builder).

This is true for any dependency and not just this Lombok example. If lombok was updated and a breaking change made it in, then that's a miss on the dependency update IMO and should have been caught/fixed as part of that update.

> Another question I have is why these models are necessary only in component tests? How the services use the same kind of classes? I see the swatch-model-metrics-hbi dependency looks like to be very related. The right pattern is that services and component tests use exactly the same classes.

Yes, this is already on our radar to address and has been talked about within our team. Definitely something we will do.

swatch-tally/ct/java/tests/TallyNightlyHbiTest.java
```diff
@@ -119,21 +118,14 @@ void testCanInsertMultipleHosts() {
...
-                RhsmFacts.builder().isVirtual(false).products(List.of(RHEL_PRODUCT_ID)).build())
-            .systemProfileFacts(
-                SystemProfileFacts.builder()
-                    .infrastructureType("physical")
-                    .arch("x86_64")
-                    .numberOfSockets(1)
-                    .numberOfCpus(0)
-                    .build())
+            .apply(HostTemplates.conduitReportedPhysicalRhel(1, 0))
+            .displayName("Test Host - 2")
```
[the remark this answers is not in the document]
mstead: This was addressed.

wottop (approved):

PR #6646, opened 2026-09-15
SWATCH-5591: Stabilize TC015 per-license remittance component test

TommasoLencioni:
Jira issue: SWATCH-5591

- Fix flaky shouldExposeSeparateMonthlyAccountRemittanceRowsPerLicenseId in ContractCoverageComponentTest
- First tally runs with only licenseA contracted, with an intermediate assertion on that row
- licenseB is added before the second tally so the final row count is deterministic
## Testing

IQE Test MR: 

```
./mvnw clean install -Pcomponent-tests -pl swatch-billable-usage/ct -am
```

## Summary by CodeRabbit

* **Tests**
  * Updated contract coverage validation for remittance splitting across two license IDs and sequential tallies.


swatch-billable-usage/ct/java/tests/ContractCoverageComponentTest.java
```diff
@@ -361,18 +361,18 @@ void shouldExposeSeparateMonthlyAccountRemittanceRowsPerLicenseId() {
...
 
     givenContracts(
         ROSA.getName(),
         List.of(
-            new ContractStub(firstSnapshot.minusDays(5), end, Map.of(awsDimension, 2.0), licenseA),
             new ContractStub(
-                firstSnapshot.minusMonths(1), end, Map.of(awsDimension, 2.0), licenseB)));
+                firstSnapshot.minusMonths(1), end, Map.of(awsDimension, 2.0), licenseA),
+            new ContractStub(
+                firstSnapshot.minusDays(5), end, Map.of(awsDimension, 2.0), licenseB)));
```
Sgitario, line 376: these changes do not run what is described in the test plan when it should have two contracts with a different licenseId and the tally should count total 5. (now, it's only one contract).

also, I'm curious about what is the real fix here... maybe what makes the test more determistic is:
```
thenAccountRemittanceEquals(ROSA.getName(), INSTANCE_HOURS.toString(), 1.0, licenseA);
```

which is compatible with having two contracts as stated in the component test?

Sgitario (changes requested):

swatch-billable-usage/ct/java/tests/ContractCoverageComponentTest.java
TommasoLencioni: > these changes do not run what is described in the test plan when it should have two contracts with a different licenseId and the tally should count total 5. (now, it's only one contract).

I don't like the test case in first instance (despite me being the author). Having the same contracts re-stubbed is not a real behavior.
Starting with one contract, remitting, then adding a new one it's more concrete scenario.
I'm changing the test plan without changing the final assertion.

> I'm curious about what is the real fix here... maybe what makes the test more determistic is:
> `thenAccountRemittanceEquals(ROSA.getName(), INSTANCE_HOURS.toString(), 1.0, licenseA);`

Yes, the fix is checking the remittance after the first tally snapshot is published, so we avoid the condition of remitting both the first and second usage on only the second contract.

> which is compatible with having two contracts as stated in the component test?

Yes.

Sgitario (approved):

PR #6656, opened 2026-09-17
SWATCH-5192: Add contracts-update-TC011 for subscription sync date replacement

diegomaranhao:
Jira issue: SWATCH-5192
## Description
Adds `contracts-update-TC011` to `ContractsUpdateComponentTest`. An AWS ROSA contract is created with original dates. Partner Gateway and Subscription Search then return the same entitlement with a new start and end date. Org subscription reconciliation is triggered by publishing an `EnabledOrgsResponse` to `platform.rhsm-subscriptions.subscription-sync-task`. After sync finishes, the old contract UUID is gone, one replacement contract remains, and the remaining subscription keeps the same subscription ID with the updated dates.
Shared helpers: `whenSubscriptionSyncRunsForOrg()` now lives on `BaseContractComponentTest`, and `givenOfferingIsSynced` can skip the Partner stub so the initial offering sync does not lock in the old entitlement dates.
IQE `test_verify_subscription_sync_overwrites_invalid_records` should be removed in a follow-up after this merges.
## Testing
./mvnw clean install -Pcomponent-tests -pl swatch-contracts/ct -am -Dtest=ContractsUpdateComponentTest#shouldReplaceContractAndSubscriptionDatesAfterSubscriptionSync -Dsurefire.failIfNoSpecifiedTests=false

## Summary by CodeRabbit

- **New Features**
  - Contract records now include billing provider information and a default quantity of one.
  - Subscription synchronization replaces outdated contract records when entitlement dates change while preserving subscription identity and key details.

- **Bug Fixes**
  - API responses now omit quantity when subscription quantity data is unavailable instead of returning a null value.

- **Tests**
  - Added coverage for subscription date reconciliation and preservation of contract and subscription details.


swatch-contracts/ct/java/api/SearchApiStubs.java
```diff
@@ -181,7 +181,8 @@ private Map<String, Object> mapToApiModel(Subscription subscription) {
     var apiSubscription = new HashMap<String, Object>();
     apiSubscription.put("id", subscription.getSubscriptionId());
     apiSubscription.put("subscriptionNumber", subscription.getSubscriptionNumber());
-    apiSubscription.put("quantity", subscription.getQuantity());
+    apiSubscription.put(
+        "quantity", subscription.getQuantity() != null ? subscription.getQuantity() : 1);
```
Sgitario: I would prefer avoiding magic logic here: if we don't pass any quantity, then we don't populate the quantity property. wdyt?

swatch-contracts/ct/java/domain/Contract.java
```diff
@@ -93,23 +93,33 @@ public static Contract buildRosaContract(
       String orgId, BillingProvider billingProvider, Map<MetricId, Double> capacity, String sku) {
     Objects.requireNonNull(billingProvider, "billingProvider cannot be null");
     String seed = RandomUtils.generateRandom();
+    String productCode = "product" + seed;
+    String customerId = "customer" + seed;
     String sellerAccountId = billingProvider == BillingProvider.AWS ? "seller" + seed : null;
+    // Search API AWS refs are productCode;customerId;sellerAccountId. Without this, org sync
+    // queries Partner Gateway with vendorProductCode "null" and never applies new dates.
+    String billingProviderId =
+        billingProvider == BillingProvider.AWS
```
Sgitario: I know previously, it did nothing with azure, but for completeness, can you add also the logic for azure? 
Also, use switch(billingProvider):
case BillingProvider.AWS ... 
case BillingProvider.AZURE...

so, if in the future, we add a new billing provider, the compiler will ask to add the case for the new billing provider.

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
```diff
@@ -129,8 +131,14 @@ void givenContractIsCreated(Contract contract) {
   }
 
   protected void givenOfferingIsSynced(Contract contract) {
+    givenOfferingIsSynced(contract, true);
+  }
+
+  protected void givenOfferingIsSynced(Contract contract, boolean stubPartner) {
     wiremock.forProductAPI().stubOfferingData(contract.getOffering());
-    wiremock.forPartnerAPI().stubPartnerSubscriptions(forContract(contract));
+    if (stubPartner) {
```
Sgitario: I would prefer avoiding the pattern of adding boolean parameters that handles special logic like adding mocks.
Also, it's super weird to have a `givenOfferingIsSynced` when the parameter is a Contract :S

Sgitario (changes requested):

swatch-contracts/ct/java/api/SearchApiStubs.java
diegomaranhao: Agreed. The mapper should not invent a quantity of 1. Dates already skip missing fields; quantity should do the same. I'll fix it.

swatch-contracts/ct/java/domain/Contract.java
diegomaranhao: Azure has been added.
thanks.

swatch-contracts/ct/java/tests/BaseContractComponentTest.java
diegomaranhao: Agreed on both points. Fixed.

Sgitario (approved):

